PR #3, opened 2019-06-28
Error state handling

helenvholmes:
This is a fake PR in that the changes below are literally just committing whitespace—I just wanted to pull out one full example of a data flow that's built out two levels. It only includes one button (although the 'organism' at the end of the chain has three). This is because the code for all of the buttons is basically the same, just with different CSS classes.

@gkallenberg, how should I handle errors? Currently everything silently fails.


helenvholmes:
@gkallenberg Yeah, the file names are awful ☹️ I'm trying to use what the design team has in Figma but I ended up forcing pushing up a bunch of changes because I originally committed the wrong files after getting confused. I kinda want to take some time sometime soon to think about naming and organization in this repo

helenvholmes:
Going to close this out. Thanks Greg.

PR #29, opened 2019-10-23
[WIP] Twig Tests

hokei:
[ResearchNow Jira Ticket](http://jira.nypl.org/browse/SFR-671)

- Implements Unit Tests for Twig

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -10,5 +10,5 @@
     "src/*",
     "storybook"
   ],
-  "version": "1.0.1"
+  "version": "1.0.2"
 }
--- a/patternlab/package-lock.json
+++ b/patternlab/package-lock.json
@@ -6840,9 +6840,9 @@
       }
     },
     "design-system-styles": {
-      "version": "0.0.2",
-      "resolved": "https://registry.npmjs.org/design-system-styles/-/design-system-styles-0.0.2.tgz",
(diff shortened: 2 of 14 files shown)


hokei:
Mostly putting this out as a WIP to see if the approach looks ok?  This approach doesn't have any semantic understanding of HTML, so we're just doing string queries.  

I may try to work on a more PHP based approach, which may allow some more useful tests.

helenvholmes:
I started a review of this, but I actually think it would be good to pull @gkallenberg or @jberube-nypl into the discussion as I'm out of my depth here.

Crystal is working on a test suite for the Twig templates that have been created thus far. At the moment we don't have any tests for our components. What's she's created here is a test suite using mocha and chai, both of which are JavaScript unit testing frameworks.

Twig's documentation seems to suggest it provides [testing out of the box](https://twig.symfony.com/doc/2.x/tests/index.html). I am at a disadvantage here to provide any real guidance as I a) don't have any prior experience writing tests nor b) in choosing a test suite. However, it feels right that the tests for our Twig templates should derive from a PHP world over a JavaScript one.

Do either of you have any experience in this area and can point us in a particular direction?

hokei:
We've decided to go with researching a more pure PHP approach so that we can test functionality similar to the way the React tests work.  Closing.

PR #30, opened 2019-10-24
SFR-686: Fixing Lerna Setup

hokei:

[ResearchNow Jira Ticket](http://jira.nypl.org/browse/SFR-686)

- Hoists packages so that building and linking happen automatically
- Creates scripts in root package.json so that there isn't a need to work from storybook/patternlab directories
- Updates documentation


hokei:
Making this a PR into the tests branch for now, we can also merge this first in which case I'll do a rebase against `development` 

I /hope/ this gets rid of the package-lock churn

For review:  Please pull this and run the instructions in the readme (and let me know if I can improve on them in any way).  I think there might be some `npm link` work that needs to be done as well, and I'd love to be able to script that in.  Obviously if it doesn't work at all, I gotta fix that!

lerna.json
```diff
@@ -3,6 +3,9 @@
   "command": {
     "init": {
       "exact": true
+    },
+    "patternlab": {
```
hokei, line 7: nit: unneccesary

patternlab/gulpfile.js
```diff
@@ -4,7 +4,7 @@
 'use strict';
 
 // General
-var gulp = require('gulp-help')(require('gulp'));
+var gulp = require('gulp');
```
hokei, line 7: revert this

PR #31, opened 2019-10-25
SFR-686 Fixing Lerna Setup

hokei:
[ReesearchNow Jira Ticket](http://jira.nypl.org/browse/SFR-686)

- Hoists packages so that building and linking happen automatically
- Creates scripts in root package.json so that there isn't a need to work from storybook/patternlab directories
- Updates documentation


hokei:
Change to Independent mode

README.md
```diff
@@ -5,14 +5,17 @@ Information about the Design System, philosophy, and timeline is available on [C
 
 ## Getting Started
 This repo is versioned with [Lerna](https://github.com/lerna/lerna).  
-Then, cd into the docs for your framework: `storybook` for React, or `patternlab` for Twig, and then run 
+To use Patternlab, you must have installed PHP7+ and [https://getcomposer.org/](composer)
 
-``` npm install && npm start ```
+After pulling the repo, run `npm run install-all`
```
helenvholmes: I got `sh: lerna: command not found`. Think we might want to call out globally installing lerna before running this command. (Might not be necessary but whoever if working on this will need it anyway.)

README.md
```diff
@@ -5,14 +5,17 @@ Information about the Design System, philosophy, and timeline is available on [C
...
+Then, run `npm run patternlab` or `npm run storybook` depending on the product you are working on.
```
helenvholmes: Both of these commands are failing for me. Here's my [failure log](https://gist.github.com/helenvholmes/26417831127e8ee7cf244b22ea15f0b0).

It seems like it can't find my `start-storybook` script, which might be a result of an `npm link` not happening correctly. Should users be running `npm run link-all` before this?

UPDATE: Running `npm run link-all` doesn't seem to be doing the trick either. We might have to debug this in person to figure out what your local set-up has that mine needs...

storybook/package.json
```diff
@@ -15,7 +15,7 @@
   "scripts": {
     "build-storybook": "build-storybook",
     "start": "start-storybook -p 9009",
-    "clean": "rimraf node-modules package-lock.json"
+    "clean": "rimraf node-modules"
```
helenvholmes, line 18: What's the difference between `rimraf` and `rm -rf`, do we know?

helenvholmes (changes requested):
I'm having trouble getting this up and running locally. We might need to debug this in-person to figure out what information needs to be added to the README.

storybook/package.json
hokei: I had always been using it out of habit, so thanks for finally making me look it up!  So it's cross-platform (eg: works on windows) which might not be something we need, and it's faster.  I had always thought that it was safer (than rm -rf on an environment I didn't know anything about) but it seems like that's not the case.  
I'm happy keeping it or not

README.md
hokei: sounds good.  You shouldn't need to run link at all, and I haven't seen this particular error before - this is good, let's find some time to debug this week

README.md
hokei: Good point, yes we need to globally install lerna and I think it's okay to ask people to do that.

storybook/package.json
helenvholmes: From cursory googling it seemed like `rm -rf` and I didn't understand the difference, but those all sound like great reasons to keep it.

Going to resolve this.

mwbenowitz (commented):
I was able to run both `npm run patternlab` and `npm run storybook` successfully (able to see both sites).

I do get the `breakpoint-sass` error, which seems to be coming from an internal test in that package? I'm unclear on what the proposed fix for that is, but that isn't super important.

I also get a lot of linting errors from `node_modules` it looks like you've got `tslint` in your `package.json` but `stylelint` in your `package-lock.json` not sure which of those is intended for use?

I would just recommend adding explicit steps to the README for starting this up: 

1) Clone repo
2) Install lerna
3) run install-all
4) Edit files in the `src` directory

helenvholmes:
Going to try re-cloning the repo and see if that solves my local errors.

helenvholmes:
So that _didn't_ solve all my problems (unfortunately), but I don't feel comfortable blocking this PR on my ability to get it up and running locally if Mike is able to. Let's investigate my local setup later.

If you can make Mike and I's changes to the README I feel fine merging this in.

hokei:
I realized that independent versioning still needs to happen on this PR, and that's relatively important to get in.  Happy to do that and do that on a different PR, or leave this one open - whichever is easier!

helenvholmes:
@hokei Up to you! ¯\_(ツ)_/¯

helenvholmes (approved):
👍🏻

PR #32, opened 2019-10-28
SFR-671 create tests

hokei:

[ResearchNow Jira Ticket](http://jira.nypl.org/browse/SFR-671)

- Creates unit tests for twig templates
- Runs them using a [puppeteer](https://github.com/GoogleChrome/puppeteer) 
- Reconfigures the `bem` function such that it is not reliant on drupal

### UAT Review:
- [ ] Navigate to [the page in Drupal](https://reno-XXX-nypl.pantheonsite.io/)

### Front End Review:
- [ ] View [the example in Pattern Lab](https://reno-XXX-nypl.pantheonsite.io/themes/custom/nypl_emulsify/pattern-lab/public)


src/twig/_twig-components/functions/bem.function.php
```diff
@@ -4,73 +4,44 @@
...
-    $attributes = new Attribute();
-
-    // Iterate the attributes available in context.
-    foreach($context['attributes'] as $key => $value) {
-      // If there are classes, add them to the classes array.
-      if ($key === 'class') {
-        foreach ($value as $class) {
-          $classes[] = $class;
-        }
+function bem(\Twig_Environment &$env, $config) {
```
hokei, line 7: This looks to be re-generated from patternlab and I should find a way to make sure this used going forward

src/twig/package.json
```diff
@@ -1,12 +1,18 @@
...
   "description": "",
   "main": "index.js",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "mocha __tests__/*.js __tests__/**/*.js"
   },
   "keywords": [],
   "author": "",
-  "license": "Apache"
+  "license": "Apache",
```
helenvholmes: Is it `Apache` or `Apache 2.0`? I feel like we've been swapping back and forth.

src/twig/_twig-components/functions/bem.function.php
helenvholmes: This entire function file, or just certain lines of the file?

helenvholmes (approved):
Looks good to me! 👍🏻

src/twig/package.json
hokei: It throws me a warn when I use "Apache" and so I think it should be "Apache 2.0"

helenvholmes:
Just to be clear, are we blocking this on the `package-lock.json` issue?

hokei:
This is blocked on preventing the automatic replacement of `bem.function.php`
Unblocking it hasn't been my highest priority, but it would entail going in to the emulsify scripts and disabling the scripts that generate the function

It might also be unblocked by the twig->storybook work.

patternlab/scripts/twig_functions.sh
```diff
@@ -25,7 +25,6 @@ then
   # Array of twig functions to copy, starting from the vendor directory
   twig_functions=(
     "drupal-pattern-lab/add-attributes-twig-extension/add_attributes.function.php"
-    "drupal-pattern-lab/bem-twig-extension/bem.function.php"
```
hokei, line 28: This is the main thing that changed

helenvholmes: Ah, so just pulling this up out of the library code and into our code?

hokei: Pretty much!  We're manually keeping it in the `src/twig` npm module so it's getting checked in, there's no need to replace it.

helenvholmes: I think we'll still potentially need it with the storybook/twig changes to interpolate the bem function in all of those files, but we should either include it as a dependency or not have some install script moving it around 😅

helenvholmes (approved):
👍🏻

PR #38, opened 2019-11-01
SFR-679 Removing Emulsify, by removing Patternlab

hokei:
[ResearchNow Jira Ticket](http://jira.nypl.org/browse/SFR-679)

- Removes Patternlab and configures Storybook to run both Twig and React components. 
- Updates READMEs to reflect this
- Modifies twig tests to remove the necessity of PHP


hokei:
So this CR is mostly configuration and pretty dense.  Might be easier to do in person!  
There are files in twig/_annotations, twig/_data, etc that we might also be able to take out, too!  Will defer to you for that!

storybook/.storybook-react/webpack.config.js
```diff
@@ -0,0 +1,70 @@
...
+        loader: 'sass-loader',
+        options: {
+          importer: globImporter(), 
+        }
+      }
+    ],
+    include: [path.resolve(__dirname, './styles/**/*')]
+  });
+
+  // Add twig support
```
helenvholmes: This is no longer needed for the storybook-react since we have storybook-html, correct?

helenvholmes (approved):
So damn pleased. 🥳

storybook/.storybook-react/webpack.config.js
hokei: Yeah, I could take this out.  I'll do that before merging

PR #40, opened 2019-11-05
Added Breadcrumbs

hokei:
[ResearchNow Jira Ticket](http://jira.nypl.org/browse/SFR-694)

- Adds the React version of the breadcrumb

### Front End Review:
- [ ] Check against the [Metronome documentation](http://themetronome.co/components/mn-11-breadcrumb-navigation/?fresh=true)


hokei:
In addition to wanting to know your thoughts about the approach, I'm also wondering where the breadcrumb background comes from?  Both the Twig and React versions of this have white text, and in patternlab/storybook this shows up as white text on a white background, which is not helpful.  

There's also an icon at the end, which doesn't show up in the Metronome to me 

This is not finished, but I think before working further I'd love to clarify what we want from the storybook re:knobs, variations, previews, etc!

helenvholmes:
We chatted about where the background-color comes from in person yesterday, but I'm going to capture that info here.

The background-color is determined by the section that contains the component. For example, in Research, all of the components that have section-based colors—breadcrumbs, article card-no image, etc.—will update to be the Research teal. In the abstract, this means the breadcrumbs _have_ no background-color. In Patternlab we handle this by having Patternlab-specific styles. I think the in the future it means that we'll probably mock up data that includes the section it's in.

Since none of that work has been completed, it means for the time being we should probably have some Storybook-specific CSS as well that's only added and used in the stories. You can see an example of that [here](https://github.com/NYPL/nypl-d8/blob/e752f642775d381aa58010b1aceebf175fa6cf7b/web/themes/custom/nypl_emulsify/components/_patterns/02-molecules/menus/breadcrumbs/_breadcrumbs.scss#L68).

src/react-components/stories/1-Button.stories.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from "react";
 
-import Button from "../components/Button/Button";
+import Button from "../src/components/Button/Button";
```
helenvholmes: Not a line comment but a filename comment. Why are we prefixing the story file with a "1"? I was under the assumption that this was how you ordered stories in the file viewer. If we're fine with letting them fall alphabetically, we can let them be unprefixed.

If we'd like them to begin matching the structure in the Twig storybook, we could probably begin mimicking the file structure in the other places of `atoms/`, `molecules`, and `organisms`, and within those folders not numbering them.

src/react-components/stories/1-Breadcrumb.stories.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from "react";
+
+import Breadcrumb from "../src/components/Breadcrumb/Breadcrumb";
+import { action } from "@storybook/addon-actions";
+
+export default {
+  title: "Breadcrumb",
+};
+
+export const regular = () => <Breadcrumb items={[{ url:'#', text:"Home"}, { url:'#', text:"Parent"}]}></Breadcrumb>;
```
helenvholmes: Can you added two more levels to show off what happens when there are more than two items? "Grandparent" and "Great-Grandparent"?

Might be more readable to have the key/value pairings as a blog with a variable name outside of the function so adding on to it is easier in the future if we want to generate some stories for this that _do_ include a ton of links.

src/react-components/src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,42 @@
...
+    static defaultProps = {};
+
+    constructor(props: BreadcrumbProps) {
+        super(props);
+    }
+
+render(): JSX.Element {
+    const { items, className } = this.props;
+
+    let bcrumbClassName = `breadcrumb${className ? `${className}` : ""}`
```
helenvholmes: If we need to pass multiple classes to a breadcrumb in a story (say, we need to pass in a series of modifiers like `--section--books-and-more`, how does that work? Relevant to [the background discussion](https://github.com/NYPL/nypl-design-system/pull/40#issuecomment-550347049).

helenvholmes (changes requested):
Let me know if you have any questions. I think this is blocked potentially on the `bem` function in your other PR so I'll get cracking on that as well.

src/react-components/stories/1-Button.stories.tsx
hokei: I think it would be better to mimic the Twig structure since it's the same structure as in Metronome, right?  Also could take this time to move stories to be in the same folder as the tsx.

helenvholmes: I think it's best to stay consistent across our three packages. It's a lot of folders, but at least you always know where you're going instead of `find . --name`ing for files.

src/react-components/src/components/Breadcrumb/Breadcrumb.tsx
hokei: AFAICT you can't do that with the Twig right now either, and I think for now it's ok to leave it at parity.

helenvholmes: So you can do it with Twig right now—it's what the `bem` function does. I was curious how that worked here

hokei:
Putting this here: https://github.com/NYPL/nypl-design-system/pull/40/files/0cac9c1b7a98bbf639ec7d012d242d243c7c85a5#diff-b5f9152b3289350c22a6edcf5514f9f8
to make sure we're continuing this conversation.  But also lmk if there's a more sane way to do this via github

src/react-components/src/components/Breadcrumb/Breadcrumb.tsx
hokei: So I think I might be misunderstanding how to do this 
In Button, you can pass a new class into `button_modifiers` (https://github.com/NYPL/nypl-design-system/blob/SFR-694_breadcrumb_react/src/twig/_patterns/01-atoms/buttons/_button.stories.js#L20) that will add in a new class (https://github.com/NYPL/nypl-design-system/blob/SFR-694_breadcrumb_react/src/twig/_patterns/01-atoms/buttons/_button.twig#L20)
However, I don't see anything in breadcrumbs.twig that lets you do anything similar?

src/react-components/CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
...
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
+When it is released, this project will adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ========
-### v1.0.1
-> Add description
+
+## [1.0.2] - 2019-11-13
```
helenvholmes: So a question for our changelogs and Lerna... do we think we'll reset all of the major numbers on these to "0" when we decide we are ready for our first major release?

hokei: I don't know.  NPM won't let us remove and re-publish the package, so if we wanted to do something like this we'd have to pick a different package name.

hokei: Note to self to bump the package.json version numbers

helenvholmes: Renaming the design system is thus a _necessity_ 

`reservoir` here we come

src/react-components/src/components/Breadcrumb/Breadcrumb.tsx
helenvholmes: Ohhh, okay, I think I understand better now.

In breadcrumbs, you're able to use that same function whenever there's a class, basically—being able to pass it each of those three parameters. I was kinda lost in the JS here, could but that that works here as well.

src/react-components/CHANGELOG.md
hokei: Maybe we keep calling it a 0 version until we move to the permanent name... ?

hokei: Or maybe we append -prerelease?

hokei: Added "Prerelease" to the changelogs.  Hopefully this is ... ok for what we need for the time being?

helenvholmes: I'm fine with it. I would prefer to have a `0.x.x` but we got what we got!

src/react-components/src/components/02-molecules/menus/Breadcrumb/Breadcrumb.stories.tsx
```diff
@@ -0,0 +1,16 @@
...
+import Breadcrumb from "./Breadcrumb";
+import { action } from "@storybook/addon-actions";
+
+export default {
+  title: "Breadcrumb",
+  component: Breadcrumb,
+  decorators: [storyFn => <div style={{ backgroundColor: "grey" }}>{storyFn()}</div>],
+};
+
+const shortItems = [{ url: "#", text: "Parent" }, { url: "#", text: "Home" }];
```
helenvholmes: Creates longer files, but can we split these up onto multiple lines? I think it makes them a little more legible, especially for `longItems`.

src/react-components/src/components/02-molecules/menus/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,49 @@
...
+      </ li >);
+    });
+
+    return (
+      <nav
+        className={bem(breadcrumbs__base_class)}
+        role="navigation"
+        aria-label="Breadcrumbs"
+      >
+        <ul className={bem("list", [], breadcrumbs__base_class)}>
```
helenvholmes, line 43: 🌟

src/styles/00-base/02-typography/01-typefaces/_typefaces.scss
```diff
@@ -1,4 +1,4 @@
 // Typography Variables
-$font-body: 'HelveticaNeue-Light', 'Helvetica Neue', Helvetica, Arial,
-  'Lucida Grande', sans-serif;
-$font-heading: 'HelveticaNeue-Light', 'Arial', sans-serif;
+@import '../../node_modules/system-font-css/system-font.scss';
+
+$font-body: system-ui, "Segoe UI", Tahoma;
```
helenvholmes: You shouldn't need to define Segoe and Tahoma here, they're [already built into the font-stack](https://github.com/jonathantneal/system-font-css/blob/gh-pages/system-font.css#L7) that `system-font` declares. More appropriate would just be a `sans-serif` at the end of this imo.

helenvholmes: Man, I'm seeing that it's defined in Reno too... well, let's rip it out here at least.

src/styles/00-base/02-typography/02-type-scale/_type-scale.scss
```diff
@@ -1 +1,57 @@
-// Type Scale info goes here.
```
helenvholmes, line 1: Removing this comment is fine (it's no longer descriptive) but I'd rather add some information about how this file _should_ be used. "Type scale information to keep our font-sizes consistent across the system", or something.

src/react-components/CHANGELOG.md
hokei: If we released under a new package, we could use `0.x.x` - the reason why we weren't doing it was because Lerna treats every pre-1 release as a major release, which ... actually works for us now haha

src/styles/CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
...
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
+
+This repo is in Prerelease. When it is released, this project will adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ========
-### v1.0.1
-> Add description
+## Prerelease
+
+## [1.0.2] - 2019-11-13
```
helenvholmes: Based on our conversation elsewhere, I do think we should be pre-1.0 right now, even though everything is a major release—I mean, everything is a major release! C'est la vie.

src/twig/CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
-### CHANGE LOG
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
+This repo is in Prerelease. When it is released, this project will adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
```
helenvholmes, line 6: Can you add a break to match the other Changelog? Also, 👏🏻, good info here

src/twig/CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
...
+## Prerelease
+
+## [1.0.2] - 2019-11-13
+### Added
+- Started a change log
+- Added Breadcrumb Template Tests
+
+### Changed
+- Breadcrumb uses spaceless to better enforce rendering consistency
+- Breadcrumb uses inline-flex so that word breaks wrap properly
```
helenvholmes: I think this line might belong in the `styles` changelog over here. Also we _removed_ the inline flex in favor of just `inline`, I know that happened very quickly during our working session

hokei:
There is a changelog in root (https://github.com/NYPL/nypl-design-system/blob/development/CHANGELOG.md) I think it makes sense to keep the changelog in the packages so that their consumers have access, but what are your thoughts?

src/react-components/src/utils/bem.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  // If extra non-BEM classes are added.
+  if (extra) {
+    extra.forEach((extraClass) => {
+      classes.push(extraClass);
+    });
+  }
+
+  let attributes = classes.join(" ");
```
helenvholmes, line 36: Not for this pull request, but I have some thoughts about how we can combine this and `bem.js`, I think.

src/twig/_patterns/02-molecules/menus/breadcrumbs/breadcrumbs.stories.js
```diff
@@ -0,0 +1,21 @@
...
+import { action } from "@storybook/addon-actions";
+import { storiesOf } from "@storybook/html";
+
+export default {
+  title: "Breadcrumbs",
+  component: Breadcrumbs,
+  // decorators: [storyFn => <div style={{ backgroundColor: "grey" }}>{storyFn()}</div>],
+};
+
+const shortItems = [{ url: "#", text: "Parent" }, { url: "#", text: "Home" }];
```
helenvholmes: Same comment as one in the `tsx` stories file.

storybook/.storybook-html/config.js
```diff
@@ -1,5 +1,5 @@
 import '!style-loader!css-loader!sass-loader!import-glob-loader!design-system-styles/style.scss';
-import { configure } from '@storybook/html';
+import { configure, load, addDecorator } from '@storybook/html';
```
helenvholmes, line 2: Fingers crossed

helenvholmes (changes requested):
> There is a changelog in root (https://github.com/NYPL/nypl-design-system/blob/development/CHANGELOG.md) I think it makes sense to keep the changelog in the packages so that their consumers have access, but what are your thoughts?

I think that makes sense. I put it in the root without thinking too deeply clearly!

helenvholmes:
Great work, we're almost there. 💫

src/styles/CHANGELOG.md
hokei: https://semver.org/#spec-item-4 
That's intended (and a thoughtful part of semver!)

helenvholmes: We live, we learn, we laugh, we cry, we sob over our new understanding

hokei: yup, that's how I feel
Thankfully, NPM does let me publish previous versions so we have this lovely https://www.npmjs.com/package/@nypl/design-system-twig

helenvholmes: Does that mean no one will accidentally install our old stuff at the other numbers? This might not be something you know tbf

hokei:
The package will be bumped when the release is cut!

src/react-components/src/components/02-molecules/menus/Breadcrumb/Breadcrumb.stories.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from "react";
+
+import Breadcrumb from "./Breadcrumb";
+import { action } from "@storybook/addon-actions";
+
+export default {
+  title: "Breadcrumb",
+  component: Breadcrumb,
+  decorators: [storyFn => <div style={{ backgroundColor: "grey" }}>{storyFn()}</div>],
```
helenvholmes, line 9: Wait, so these are building now????

hokei: it's always built in tsx/React
It's never built in js/Twig (and is still commented out)

src/styles/00-base/02-typography/01-typefaces/_typefaces.scss
```diff
@@ -1,4 +1,2 @@
 // Typography Variables
-$font-body: 'HelveticaNeue-Light', 'Helvetica Neue', Helvetica, Arial,
-  'Lucida Grande', sans-serif;
-$font-heading: 'HelveticaNeue-Light', 'Arial', sans-serif;
+@import '../../node_modules/system-font-css/system-font.scss';
```
helenvholmes, line 2: I pushed up a fix that adds those two variables back in—unfortunately we can't delete them even though they reference the same font-stack as they're used somewhere 😬

helenvholmes (approved):
_I_ pushed up a few small commits so I think you should take a look at those. If you're good with them, I'm good with all the work you did, this is rad 👍🏻

src/styles/00-base/02-typography/01-typefaces/_typefaces.scss
hokei: It all looks good, yeah, I only realized after I pushed that the font thing wasn't working o.o. 
Anyway I'll push and cut a release tomorrow morning!

PR #49, opened 2019-11-21
SFR-716 Added Icons to the React Design system

hokei:
[ResearchNow Jira Ticket](http://jira.nypl.org/browse/SFR-716)

- Added Icons to the React version of the design system
- The Twig version doesn't work yet, but I think it has to do with file-path resolutions that might be a quick remedy if I can figure out how it works.


src/icons/src/scripts/build-icon.js
```diff
@@ -0,0 +1,41 @@
...
+      namespaceIDs: true, // Add namespace token to all IDs in SVG shapes
+      namespaceClassnames: true, // Add namespace token to all CSS class names in SVG shapes
+      dimensionAttributes: true // Width and height attributes on the sprite
+  }
+}
+
+var spriter = new SVGSpriter(config);
+
+// Add SVG source files
+let dir = path.join(__dirname, "../svg/")
```
EdwinGuzman, line 27: Not a big deal but if you can use `let`, use it instead of `var`, for consistency and updated style.

src/icons/src/scripts/build-icon.js
```diff
@@ -0,0 +1,41 @@
...
+fs.readdirSync(dir).forEach(file => {
+  spriter.add(dir + file, null, fs.readFileSync(dir + file, {encoding: 'utf-8'}));
+});
+
+// Compile the sprite
+spriter.compile(function(error, result) {
```
EdwinGuzman: Same for functions, if you can write arrow functions then using that to be consistent would be nice.

src/react-components/src/__tests__/setupEnzyme.ts
```diff
@@ -1,5 +1,34 @@
-
 import { configure } from "enzyme";
 const Adapter = require("enzyme-adapter-react-16");
+const { JSDOM } = require("jsdom");
+
+const jsdom = new JSDOM("<!doctype html><html><body></body></html>");
+const { window } = jsdom;
+
+function copyProps(src, target) {
```
EdwinGuzman: For deep copy?

src/react-components/src/components/01-atoms/Images/Icons/Icon.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+      return (
+        <svg {...iconProps} >
+            {title &&
+              <title id={`title-${name}`}>{ title }</title>
+            }
+            {desc &&
+              <desc id={`title-${name}`}>{ desc }</desc>
+            }
+            <use xlinkHref={`${Icons}#${name}`} />
```
EdwinGuzman: I think the `xlinkHref` is deprecated and not recommend, no? The idea for this is that you'd declare an SVG icon on the page and reuse it whenever this is instantiated, right? If so, how will that first SVG be loaded?

hokei: Yes, it is.  The way that this was done in Twig/Drupal was that all the svgs get compiled in one file and referenced by their ID (https://una.im/svg-icons/) and that's what SVGSpriter is doing. I'm not sure if it's tied to xlinkhref - I'll investigate that, I'm guessing I could just replace with href and it would be ok.

I don't actually know how they're being loaded, though.  I'm not ... actually sure the svg gets reused?

src/icons/.npmignore
```diff
@@ -0,0 +1 @@
+src
```
helenvholmes: I wonder if this is truly what we want to do... I don't know if there are any debugging capabilities that not including the raw icons would truly hinder. For now I assume this is fine

src/icons/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "@nypl/design-system-icons",
+  "version": "0.0.0",
+  "description": "Icons for the Design System",
+  "main": "dist/img/sprite/sprite.svg",
+  "scripts": {
+    "prepare": "npm run build",
+    "build": "node src/scripts/build-icon.js",
+    "test": "echo \"No Icon tests\""
```
helenvholmes, line 9: Some simplistic tests we might consider for icons could be:
- Validating the SVG markup
- E2E testing that the individual icons in `src/` produce some visible black paths in the window when loaded in (video icon's "play" button being the sole exception here—it's the only icon we release with a white fill instead)

src/icons/src/scripts/build-icon.js
```diff
@@ -0,0 +1,41 @@
+var SVGSpriter =  require('svg-sprite');
+var fs = require('fs');
+var path = require('path');
+var mkdirp = require('mkdirp');
+
+var config = {
+  mode: {
+    symbol: {
+      // symbol mode to build the SVG
+      dest: 'dist/img/sprite', // destination foldeer
```
helenvholmes: tiny typo

src/react-components/CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ Currently, this repo is in Prerelease.  When it is released, this project will a
 
 ========
 ## Prerelease
+
+## [ 0.0.4 ]
+### Changed 
+- `Breadcrumb` now shows icon in mobile view
+
+### Added
+- `Icon` component that uses Icons.
```
helenvholmes: Maybe "`Icon` component that uses icons package"?

src/react-components/src/__tests__/components/Breadcrumbs-test.tsx
```diff
@@ -27,4 +27,15 @@ describe("Breadcrumbs", () => {
...
+
+  it("Renders icon", () => {
+    wrapper = Enzyme.shallow(<Breadcrumbs breadcrumbs={breadcrumbComponent} />);
+    let links = wrapper.render();
+
+    expect(links.find("svg")).to.have.lengthOf(1);
+  });
+  it("Throws error when nothing is passed into Breadcrumb", () => {
+    expect(() => Enzyme.mount(<Breadcrumbs breadcrumbs={[]} />))
+    .to.throw("Breadcrumbs cannot be empty");
```
helenvholmes: I'm not really a wordsmith for good error messages but maybe something like "Breadcrumbs must contain a set of links. Breadcrumbs currently empty" might help users understand what they're supposed to be sending it?

Should we be throwing an error above if there is no svg? "Breadcrumbs must contain an icon for mobile. Breadcrumbs icon currently empty"?

src/react-components/src/components/01-atoms/Images/Icons/Icon.tsx
```diff
@@ -0,0 +1,46 @@
+
+import * as React from "react";
+import Icons from "@nypl/design-system-icons/dist/img/sprite/sprite.svg";
+import bem from "../../../../utils/bem";
```
helenvholmes, line 4: lol. We should probably pull these out into a package too

hokei:
@helenvholmes I don't want this to block merging (because there's a lot of other branches in line to be PRed) but it'd be really nice to get this working in twig as well.  I /think/ we just need to figure out what this line: https://github.com/NYPL/nypl-design-system/blob/development/src/twig/_patterns/01-atoms/images/icons/_icon.twig#L17 should be set to.  
Ideally I would like to make `design-system-icons` a dependency of `design-system-twig` so that it can do something like `icon-url = require.resolve(node_modules/design-system-icons)` (and also make it so that Drupal doesn't have to build icons and twig-components separately) but I'm having trouble finding that syntax in Twig.  Do you know what it is/where I can find it?

src/react-components/src/components/01-atoms/Images/Icons/Icon.tsx
helenvholmes: I've come to like the solution proposed above, but it looks like the library that Twig is using hasn't been updated since 2016, which could be the reason it's producing `xlinkHref`s instead of `Hrefs` which is what MDN suggests you update to.

Messing with the icons in Patternlab, it looks like changing the `xlinkHref` to `href` doesn't have any adverse affects that I'm seeing in Chrome.

hokei: So just taking changing `xlinkHref` to `href` seems to work out here.  It looks like it's being cached and re-used, too, I wonder if the import is doing that work for me.

helenvholmes:
I think you want `{{ attach_library('.../...') }}`.

src/icons/.npmignore
hokei: I think the best reason to exclude this is if we don't want consumers of `design-system-icons` using the raw icons vs using `sprite.svg` -- Excluding it makes it impossible to use the raw icons

helenvholmes: I guess we're including them then 😂 we can investigate how to lock this (arguably singular) component down more in the future

src/twig/_patterns/02-molecules/menus/breadcrumbs/breadcrumbs.twig
```diff
@@ -15,7 +15,7 @@
                 icon_name: 'arrow-xsmall',
                 icon_blockname: breadcrumbs__base_class,
                 icon_modifiers: ['small'],
-              } %}
+              } %}
```
helenvholmes: Seems like you might have just added some whitespace here? Are you showing whitespace in your editor? It will help with committing sort of random gunk like this

storybook/.storybook-html/webpack.config.js
```diff
@@ -11,6 +11,21 @@ module.exports = ({ config, mode }) => {
   // You can change the configuration based on that.
   // 'PRODUCTION' is used when building the static version of storybook.
 
+  config.resolve.extensions.push('.svg');
+
+  config.module.rules = config.module.rules.map( data => {
+      if (/svg\|/.test( String( data.test ) ))
+          data.test = /\.(ico|jpg|jpeg|png|gif|eot|otf|webp|ttf|woff|woff2|cur|ani)(\?.*)?$/;
```
helenvholmes: I don't understand what this is testing for. Is the module rule for svgs not just that the filename extension is `.svg`?

helenvholmes (changes requested):

hokei:
> I think you want `{{ attach_library('.../...') }}`.

Spent some more time on this and ... think that maybe we circle back to this when we integrate with D8.  `attach_library` is a drupal specific method, and we'd need to figure out how to replace it.  Right now there's a `pl.attach_library` that does it for patternlab, but and I've been messing around with what that might look like for storybook, but I don't think I'm getting anywhere.  In addition, I'm not sure how it knows whether or not it's in drupal at all.  I think this is going to require some digging into emulsify/drupal build code, and unless there's a good reason why that has to happen _now_, I'd rather defer it to later.

src/react-components/src/__tests__/setupEnzyme.ts
hokei: Every time I think I need JSDOM, it turns out I don't.  I'm gonna take all of it out for now, enzyme should be able to do all that I need ...

storybook/.storybook-html/webpack.config.js
hokei: you're right.  As with all webpack dealings, this is a result of me copying over a bunch of webpack configurations to see if they work - this is one that doesn't work, but isn't visible because storybook-twig doesn't show icons anyway.  Better to take it out.

src/twig/_patterns/02-molecules/menus/breadcrumbs/breadcrumbs.twig
hokei: Ugh I had a bunch of format-on-save stuff from before, which didn't play well with my editor not being able to understand `.twig` files

src/react-components/src/components/01-atoms/Images/Icons/Icon.tsx
hokei: a task for another ticket, I think

hokei: (It's https://jira.nypl.org/browse/SFR-718 and I remember you mentioning you had a solution in mind for the slight bem differences?  You should grab it! -- I probably won't get to it for a while because of the template stuff on my plate)

src/react-components/src/__tests__/components/Breadcrumbs-test.tsx
hokei: oh gosh, you and me both.  This is better though.  
Breaadcrumbs should always have svg, it's just whether or not the hidden styling is applied in mobile.  You're not passing in an icon - the breadcrumbs generates one, so there's no need to tell the user if breadcrumbs doesn't have an icon.  If breadcrumbs fails to generate an icon, the "renders icon" test should fail.

src/react-components/src/components/01-atoms/Images/Icons/Icon.tsx
helenvholmes: Definitely—consider this a note-to-self. Hmmm, okeydoke, sure

src/react-components/src/__tests__/setupEnzyme.ts
hokei: I don't think it actually needs this, just the re-assignment of Windows and stuff

src/icons/package.json
hokei: 1st test:  The validation would mostly be a test for `spriter.compile`, which isn't our code and probably isn't necessary to test.  Could be good to have the test serve as documentation on what happens when you pass in a bad svg, though.

2nd test:  Is this to make sure that designers can't drop in bad (for us) svgs?  I think this could be super useful, and I also think it's a big enough task that we should either extract into a ticket or into a backlog task so as not to overload this one.

helenvholmes: 1st test: I agree, I was hoping we would run that on our `src/` directory, since that's the one we control. This could also be where we would catch uncompressed icons?
2nd test: I'm mostly thinking of testing bad svgs. It's not uncommon when you preprocess them for something funky to happen and nothing shows up.

Let's definitely dump this into the backlog—I don't want to make this ticket larger!

hokei: re: first test - I started writing these and it feels like testing `svg-sprite` so we shouldn't do that.

src/react-components/src/components/01-atoms/Images/Icons/Icon.tsx
```diff
@@ -0,0 +1,46 @@
...
+    name: string;
+    blockName: string;
+    modifiers?: string[];
+    decorative: boolean;
+    role?: string;
+    title?: string;
+    desc?: string;
+  }
+
+  export default class Icon extends React.Component<IconProps, {}> {
```
EdwinGuzman, line 17: Not necessary for this PR, but look into functional components in React. This component is just a presentational component that doesn't have any logic, so writing it as a function can reduce some code (and provide more efficiency later on according to React).

src/react-components/src/components/02-molecules/menus/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -19,32 +20,36 @@ export default class Breadcrumb extends React.Component<BreadcrumbProps, {}> {
...
-        // TODO: SFR-716 Incorporate Icon
-      }
       let linkComponent: string | JSX.Element;
+
       if (isTextBreadcrumb(item)) {
         linkComponent = item.url ? <a href={item.url} className={bem("link", [], breadcrumbs__base_class)}>{item.text}</a> : item.text;
       } else {
         let props = {
-          className: bem("link", [], breadcrumbs__base_class),
+          className: bem("link", [], breadcrumbs__base_class)
```
EdwinGuzman, line 40: This is the second time `bem("link", [], breadcrumbs__base_class)` appears. Either make that a variable or maybe make a curry function (or HOC) function since many of the variables are constant. Kinda like
```
const withBreadcrumbsBaseClass = (type: string) => (
  bem(type, [], breadcrumbs__base_class);
);

className={withBreadcrumbsBaseClass("link")}
```
but maybe a shorter name?

hokei: bem already is a helper class that takes a bit of thinking to wrap the head around.  I'm hesitant to give it other helpers because it'll make it harder to associate the compiled classname with where it's used and where to make CSS changes (Unless we come up with a better source-mapping solution, which ... 🤔 ) 

I wonder if the frustration with this is more with how long `breadcrumbs__base_class` is as a variable name, making everything look ugly?  I don't mind assigning it to a variable, but on the other hand it doesn't seem that bad to me that this is being calculated twice - it's string concatenation without renders in between, which is cheap, right?

src/react-components/src/components/01-atoms/Images/Icons/Icon.tsx
hokei: 👍

src/react-components/src/components/02-molecules/menus/Breadcrumbs/Breadcrumbs.tsx
helenvholmes: I would agree with Crystal here—I think the code duplication is fine as it maps to the way our Twig templates are working.

EdwinGuzman: Yeah, that makes sense, but this would be more of a helper just for this component since the modifiers and block name params don't change. It seems only the base class changes. It's definitely not necessary but I find HOCs to be easier to work with, personally. Those are cheap so it's fine.

helenvholmes: I guess the question then is if the helper should be abstracted, as this isn't the only component with this use case. If it's all right I think this is a really good conversation that would be aided with more data. I've added a ticket to the backlog [here](https://github.com/NYPL/nypl-design-system/projects/2#card-29725568) to look and some more use cases and to figure out how we should move forward, but let's not block this ticket on it.

src/icons/package.json
helenvholmes: I've added a [ticket to the backlog](https://github.com/NYPL/nypl-design-system/projects/2#card-29721175) to think through how we want to test icons. Let's not block on it here.

src/twig/_patterns/01-atoms/images/icons/_icon.twig
```diff
@@ -40,5 +41,5 @@
   {% if icon_desc %}
     <desc id="desc-{{ icon_name }}">{{ icon_desc }}</desc>
   {% endif %}
-  <use xlink:href="{{ icon_url }}sprite.svg#{{ icon_name }}"></use>
+  <use href="{{ blah }}#{{ icon_name }}"></use>
```
helenvholmes: Can you check this out? @hokei

helenvholmes (changes requested):
I think a random string popped in, but otherwise this looks good to me.

helenvholmes (approved):
💫

PR #51, opened 2019-11-26
SFR-713 body text

hokei:
[ResearchNow Jira Ticket](http://jira.nypl.org/browse/SFR-713)

- Added (the beginnings of) a Body Text component
- Only really handles a situation where text is passed - which is what is needed for RN as of right now.  Still need to come up with a plan for links, bolds, italics, lists, etc.


helenvholmes (approved):
This looks good to me. We will probably need to sync up and strategize around how we handle body text across our two templating languages—I reviewed what we're doing in Reno and think that one of the two will be what you want 99% of the time. In any event: I don't think it's worth blocking on at this time, just something to sort out pre-V1.

hokei:
Added https://github.com/NYPL/nypl-design-system/projects/1#card-29822889 to track this

PR #56, opened 2019-12-13
SFR-711: Search Bar

helenvholmes:
[ResearchNow Jira Ticket](http://jira.nypl.org/browse/SFR-711)

- Adds component + subcomponents, respective stories + styles, for `search-bar`

### Front End Review:
- [ ] Check against the [Metronome documentation](http://themetronome.co/components/mf-46-search-bar-with-search-by-parameter/?fresh=true)


helenvholmes:
Blocked by #56.

src/react-components/src/components/01-atoms/Forms/FormDropdown.stories.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from "react";
+
+import FormDropdown from "./FormDropdown";
+import { action } from "@storybook/addon-actions";
+import { withKnobs, text } from "@storybook/addon-knobs";
+
+
+export default {
```
EdwinGuzman, line 7: Not a big deal but I think it's a convention to put the `default` exports at the very bottom of a file.

src/react-components/src/components/01-atoms/Forms/FormDropdown.tsx
```diff
@@ -0,0 +1,47 @@
...
+  options: string[];
+  selectedOption?: string;
+  onSelectBlur: (event: React.FormEvent) => void;
+  onSelectChange: (event: React.FormEvent) => void;
+}
+
+export default function FormDropdown(props: FormDropdownProps) {
+  const { dropdownId, blockName, labelText, options, ariaLabel, disabled, selectedOption, onSelectBlur, onSelectChange } = props;
+  if (!labelText) {
+    if (!ariaLabel) {
```
EdwinGuzman: You can do `if (!labelText || !ariaLabel)`

src/react-components/src/components/01-atoms/Forms/FormDropdown.tsx
```diff
@@ -0,0 +1,47 @@
...
+    onBlur={onSelectBlur}
+    aria-label={ariaLabel}
+    disabled={disabled ? disabled : false}>
+        { options.map((child, key) => {
+          return <option key={key.toString()} aria-selected={child === selectedOption} value={child}>{ child }</option>;
+        }) }
+    </select>
+
+    <Icon decorative={true} name={'arrow'} modifiers={['small']} />
+  </div>
```
EdwinGuzman: I'm going to be unnecessarily nitpicky:
I think it'll be more readable if it was formatted like below but definitely not needed! Just a preference and if you want to introduce `[prettier](https://prettier.io/)`, it'll format it similarily. You are missing a comma at the end of the `return` (the linter didn't pick it up?).
```
return (
    <div className={bem("dropdown", [], formItemBlockName)}>
      {labelText && (<label htmlFor={dropdownId} className={labelClassName}>{labelText}</label>)}

      <select id={dropdownId}
        className={bem("select", [], formItemBlockName)}
        value={selectedOption}
        onChange={onSelectChange}
        onBlur={onSelectBlur}
        aria-label={ariaLabel}
        disabled={disabled ? disabled : false}
      >
          { options.map((child, key) => {
            return <option key={key.toString()} aria-selected={child === selectedOption} value={child}>{ child }</option>;
          }) }
      </select>

      <Icon decorative={true} name={'arrow'} modifiers={['small']} />
    </div>
);
```

src/react-components/src/components/01-atoms/Forms/FormDropdown.tsx
EdwinGuzman: Actually, here it says if both are not defined but the error says "Either". So that's a bit confusing.

src/react-components/src/components/02-molecules/SearchBar/SearchBar.tsx
```diff
@@ -0,0 +1,82 @@
...
+    callback: searchSubmitHandler,
+    blockName: searchbar__base_class,
+    content: "Search",
+    type: "filled",
+    iconPosition: "left",
+    iconName: "search-small",
+    iconDecorative: true,
+  };
+
+  return <form className={bem(searchbar__base_class, modifiers)} id="searchBarId" aria-label="Search for keyword, author, title, or subject">
```
EdwinGuzman: This trips me up but having a static `id` can be troublesome if the component gets rendered more than once, the accessibility error is having duplicate ids on the page. For the `SearchBar` I imagine there will only be one, a main one, but if the design of any page reuses this, then the `id` should be dynamic.

src/react-components/src/components/02-molecules/SearchBar/SearchBar.tsx
```diff
@@ -0,0 +1,82 @@
...
+    placeholderText: placeholderText,
+    modifiers: modifiers,
+  };
+
+  let buttonProps = {
+    id: searchBarId,
+    callback: searchSubmitHandler,
+    blockName: searchbar__base_class,
+    content: "Search",
+    type: "filled",
```
EdwinGuzman, line 62: Reading this, I thought that `type` was the HTML attribute for the button, but it's not. Maybe reconsider naming this slightly different to not confuse with the `type` attribute for a `button` (i.e. the type in `<button type="submit" />`). This is NOT for this PR but just a comment.

src/react-components/src/__tests__/components/FormDropdown-test.tsx
```diff
@@ -0,0 +1,54 @@
...
+  let wrapper: Enzyme.ShallowWrapper<{}, {}>;
+  let blurCallback;
+  let changeCallback;
+
+  beforeEach(() => {
+    changeCallback = stub();
+    blurCallback = stub();
+  });
+
+  it("Form Dropdown", () => {
```
EdwinGuzman: `"should render a FormDropdown"` is more descriptive.

src/react-components/src/__tests__/components/FormDropdown-test.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+    expect(blurCallback.callCount).to.equal(1);
+  });
+
+  it("requires either label or aria-label", () => {
+    expect(() => Enzyme.mount(<FormDropdown dropdownId="hi" options={["opt1", "opt2", "opt3"]} onSelectChange={changeCallback} onSelectBlur={blurCallback} />))
+    .to.throw("Either labelText or ariaLabel must be defined");
+  });
+
+  it("renders empty select form when passed no options", () => {
```
EdwinGuzman, line 49: I don't think there's a case where you would want to render an empty `select` element. If it's okay to _not_ render it, then you should not render it if the options are empty.

src/react-components/src/__tests__/components/FormDropdown-test.tsx
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  it("sends callback with no value on blur", () => {
+    wrapper = Enzyme.shallow(<FormDropdown dropdownId="hi" labelText="hi" options={["opt1", "opt2", "opt3"]} onSelectChange={changeCallback} onSelectBlur={blurCallback} />);
+    wrapper.find("select").simulate("blur");
+
+    expect(blurCallback.callCount).to.equal(1);
+  });
+
+  it("requires either label or aria-label", () => {
```
EdwinGuzman: If either of these props are required, then I'd have this as the first test to make it obvious that one of these values is needed for the component to properly work. So a structure for the test can be:
1 - default rendered component with missing props (not required props)
2 - default rendered component with minimum props
3 - more props
4 - behavior

But I do see how the previous ones are using a shallow render. I'd still include one of the needed props so that it's clear to developers that it should be added when using the component.

src/react-components/src/__tests__/components/SearchBar-Test.tsx
```diff
@@ -0,0 +1,41 @@
+import { expect } from "chai";
+import { stub } from "sinon";
+import * as Enzyme from "enzyme";
+import * as React from "react";
+
+import * as Mocha from "mocha";
```
EdwinGuzman: You don't need this.

src/react-components/src/__tests__/components/SearchBar-Test.tsx
```diff
@@ -0,0 +1,41 @@
...
+  it("Search Bar has an input field and button", () => {
+    let wrapper = Enzyme.shallow(<SearchBar searchBarId="id" searchButtonId="searchButtonId" searchSubmitHandler={searchCallback}/>);
+    expect(wrapper.find("TextField")).to.have.lengthOf(1);
+    expect(wrapper.find("Button")).to.have.lengthOf(1);
+  });
+
+  it("Search Bar calls callback on Submit ", () => {
+    wrapper = Enzyme.shallow(<SearchBar searchBarId="id" searchButtonId="searchButtonId" searchSubmitHandler={searchCallback}/>);
+    wrapper.find("Button").dive().find("button").simulate("click");
+    expect(searchCallback.callCount).to.equal(1);
```
EdwinGuzman: If you `mount` the component, you can do `wrapper.find("button").simulate("click");` and it'll work the same.

src/react-components/src/__tests__/components/SearchBar-Test.tsx
```diff
@@ -0,0 +1,41 @@
...
+  });
+
+  it("SearchBar calls optional onChange property", () => {
+    wrapper = Enzyme.shallow(<SearchBar searchBarId="id" searchButtonId="searchButtonId" searchChangeHandler={onChangeCallback} searchSubmitHandler={searchCallback}/>);
+    wrapper.find("TextField").dive().find("input").simulate("change", { target: { value: "Hello" } });
+    expect(onChangeCallback.callCount).to.equal(1);
+  });
+
+  it("Shows 'error' span when error is passed", () => {
```
EdwinGuzman, line 35: There should also be a test when error is passed but there is no error message to catch the thrown error in `SearchBar`.

src/react-components/src/components/02-molecules/SearchBar/SearchBar.stories.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { action } from "@storybook/addon-actions";
+import { withKnobs, text } from "@storybook/addon-knobs";
+
+
+export default {
+  title: "SearchBar",
+  component: SearchBar,
+};
+
+export const searchWithParams = () => <SearchBar searchBarId="searchBarId" searchButtonId="searchButtonId" dropdownId="dropdownId" dropdownOptions={["Keyword", "Other Keyword"]}
```
EdwinGuzman: These lines are long and I'd cut them back a bit.

Also, when you click on "seach" in storybook, the page changes. The `searchSubmitHandler` is being called, but since there's no preventing the default action, the page changes. You could change that in the `Button` component but not sure if that's a big deal, it's more for Storybook to stay on the same page and show the action when search is clicked.

src/react-components/src/components/02-molecules/SearchBar/SearchBar.stories.tsx
```diff
@@ -0,0 +1,16 @@
...
+  searchSubmitHandler={action("searched")} selectChangeHandler={action("selectChanged")} selectBlurHandler={action("selectChanged")} />;
+export const searchWithNoParams = () => <SearchBar searchBarId="searchBarId" searchButtonId="searchButtonId" searchSubmitHandler={action("searched")} />;
+export const searchWithFormError = () => <SearchBar searchBarId="searchBarId" searchButtonId="searchButtonId" searchSubmitHandler={action("searched")} hasError={true} errorMessage={"Error text"}/>;
```
EdwinGuzman: There should be a story with the dropdown and an error.

src/react-components/src/components/03-organisms/SearchPromo/SearchPromo.tsx
```diff
@@ -0,0 +1,51 @@
...
+export default function SearchPromo(props: SearchPromoProps) {
+  const { headingText, modifiers, blockName, titleId, searchBarId, searchButtonId, hasError, errorMessage,
+    dropdownId, searchDropdownOptions, advancedSearchMessage, 
+    selectChangeHandler, selectBlurHandler, searchSubmitHandler, textChangeHandler } = props;
+
+  const searchpromo__base_class = "search-promo";
+
+
+  return <div className={bem(searchpromo__base_class, [], blockName)}>
+    <div className={bem('content', [], searchpromo__base_class)}>
```
EdwinGuzman: Is there a reason why there must be two divs here?

src/react-components/src/components/03-organisms/SearchPromo/SearchPromo.stories.tsx
```diff
@@ -0,0 +1,32 @@
...
+export const searchPromo = () => <SearchPromo headingText="Search the World's Research Collections"
+  titleId="title"
+  advancedSearchMessage={<p>click here for <a href="#apple-pie">stuff</a></p>}
+  searchButtonId="searchButtonId"
+  searchBarId="searchBarId" dropdownId="dropdownId"
+  searchDropdownOptions={["hi there", "bye there"]}
+  searchSubmitHandler={action("searched")}
+  selectChangeHandler={action("changed")}
+  selectBlurHandler={action("selectChanged")} />;
+export const searchPromoWithError = () => <SearchPromo headingText="Search the World's Research Collections"
```
EdwinGuzman, line 22: The dropdown is off when there's an error:
Screen Shot 2019-12-16 at 3 11 17 PM

src/react-components/src/components/03-organisms/SearchPromo/SearchPromo.tsx
helenvholmes: The outer div handles the the `breakout` that extends the `background-color` while the inner div handles keeping the content on the grid.

src/react-components/src/__tests__/components/FormDropdown-test.tsx
hokei: I think it's more about thinking on how it will be used.  Ideally, I want the person consuming this to have tests on their end making sure that their options lists aren't ever empty.  However, I also think it's overly punishing to throw an error / not render at all if the user runs into a data situation where there's nothing in the list ... thoughts?

helenvholmes: I'm with @hokei on this one—I don't think we want the components to not appear in the library itself, I think that should be handled by the apps on integration in their business logic.

src/react-components/src/components/02-molecules/SearchBar/SearchBar.stories.tsx
hokei: I'm gonna add a thing to the backlog to handle that in buttons - I've noticed it too, but I think it's in the real of storybook improvements that we want to come back to!

src/react-components/src/components/02-molecules/SearchBar/SearchBar.tsx
hokei: I made a note here:  https://github.com/NYPL/nypl-design-system/projects/2#card-30647396

src/react-components/src/components/01-atoms/Forms/FormDropdown.stories.tsx
hokei: I think for storybook stories we want them at the top, cuz that's what the docs do:  https://storybook.js.org/docs/basics/writing-stories/

src/react-components/src/components/02-molecules/SearchBar/SearchBar.tsx
helenvholmes: We could potentially use `modifier-type`, it's what we're calling those for the CSS

src/react-components/src/__tests__/components/SearchBar-Test.tsx
hokei: Should `mount` be avoided as much as possible, or is it just speed of running tests?

EdwinGuzman: I think there are times when you want to test things that can't be tested using a shallow render so it's fine to mount. This is convenience more than anything so it's fine if you don't want to switch it. Usually I find that it's the tests with complex behavior and API calls that slow down tests rather than searching in the DOM.

src/react-components/src/__tests__/components/FormDropdown-test.tsx
EdwinGuzman: That sounds fine. In general, I find it hard to trust another developer to use a component as it should be use and could be misused without clear examples or documentation.

src/react-components/src/components/01-atoms/Forms/FormDropdown.stories.tsx
EdwinGuzman: Makes sense.

src/react-components/src/__tests__/components/FormDropdown-test.tsx
helenvholmes: That's also a solid reason. I'm going to vote we move forward with this as it is and I'm going to file this conversation in the backlog to revisit at a later date.

hokei: +1 to having clear documentation as to how this should be used.  Not sure where something like that would live, yet, but we're gonna start piling up best practices like this, that aren't explicitly checked for in the code.

src/react-components/src/__tests__/components/SearchBar-Test.tsx
hokei: Nah I just wanted to know in general, the internet is very divided on when/whether to use mount

src/react-components/src/__tests__/components/FormDropdown-test.tsx
```diff
@@ -17,7 +17,12 @@ describe("FormDropdown", () => {
     blurCallback = stub();
   });
 
-  it("Form Dropdown", () => {
+  it("requires either label or aria-label", () => {
+    expect(() => Enzyme.mount(<FormDropdown dropdownId="hi" options={["opt1", "opt2", "opt3"]} onSelectChange={changeCallback} onSelectBlur={blurCallback} />))
+    .to.throw("Either labelText or ariaLabel must be defined");
+  });
+
+  it("Form Dropdownshould render a FormDropdown", () => {
```
EdwinGuzman: small typo.

src/react-components/src/components/02-molecules/SearchBar/SearchBar.stories.tsx
```diff
@@ -10,7 +10,29 @@ export default {
...
+  hasError={true} errorMessage={"Error text"} />;
+
+export const searchFormWithParamsAndError = () => <SearchBar searchBarId="searchBarId"
+searchButtonId="searchButtonId"
+dropdownId="dropdownId"
+dropdownOptions={["Keyword", "Other Keyword"]}
+searchSubmitHandler={action("searched")}
+selectChangeHandler={action("selectChanged")}
+selectBlurHandler={action("selectChanged")}
+hasError={true} errorMessage={"Error text"} />;
```
EdwinGuzman: The rest of the functions are indented but this one isn't.

EdwinGuzman:
Latest updates looks good.  I think the dropdown with the label seems a bit off for the label and the down arrow.
Screen Shot 2019-12-16 at 2 42 21 PM

hokei:
@EdwinGuzman What are you seeing the label error in?  I think a lot of the styles existed in a different branch and are now merged into here, so let me know it's still here!

src/react-components/src/components/03-organisms/SearchPromo/SearchPromo.tsx
```diff
@@ -0,0 +1,56 @@
...
+          searchBarId={searchBarId}
+          searchButtonId={searchButtonId}
+          dropdownId={dropdownId}
+          dropdownOptions={searchDropdownOptions}
+          hasError={hasError}
+          errorMessage={errorMessage}
+          selectBlurHandler={selectBlurHandler}
+          selectChangeHandler={selectChangeHandler}
+          searchSubmitHandler={searchSubmitHandler}
+          searchChangeHandler={textChangeHandler}></SearchBar>
```
EdwinGuzman: Not a big deal but if you don't have children, you can do `<SearchBar ... />` for any component.

EdwinGuzman:
It seems it's fixed now in the "Search Promo with Error" story:
Screen Shot 2019-12-17 at 10 43 16 AM

EdwinGuzman (approved):

PR #60, opened 2019-12-18
Webpack react test

EdwinGuzman:
[ResearchNow Jira Ticket #772](http://jira.nypl.org/browse/SFR-772)

- Adds a webpack file to compile and minify the react components into a `design-system-react-components.min.js` file along with a source maps file. This compiles the SVG sprite into the build which was the main motivation for this task.

The `lib/` folder is still around but the main file is now what lives in `dist/`. Importing components should work the same way.

### Front End Review:
- Test and make sure it works in SFR.


EdwinGuzman:
@hokei the tests fail because running tests for the `design-system-icons` repo times out. I'm having the same issue and not sure why it's a permission issue. Tomorrow, can you test this by branching off this branch and just pushing that new branch and see if the tests pass?

hokei:
I tried it this morning and got the same problem that you did. Running `npm run prepare` from the repo itself works, so can we comment that out so it's not installing on build, and put a note in the readme that whoever's using this will have to manually build `sprite.svg` until we figure this out?  I think it's much more valuable to get the sprites in than it is to have this block us.

EdwinGuzman:
That's a good idea and I'll try it, re: commenting out the build for the sprite.

PR #61, opened 2019-12-19
Attemping to make a new branch of working build.

EdwinGuzman:
[ResearchNow Jira Ticket #773](http://jira.nypl.org/browse/SFR-772)

- Adds a webpack file to compile and minify the react components into a design-system-react-components.min.js file along with a source maps file. This compiles the SVG sprite into the build which was the main motivation for this task.

The lib/ folder is still around but the main file is now what lives in dist/. Importing components should work the same way.

I don't know why this branch works but the other branch #60 does not.

### Front End Review:
- Test and make sure it works in SFR.


EdwinGuzman:
So I created a new branch with the _same_ changes but somehow this works but the other one does not. I saw the similar permission issue for `icons` in the Travis build in the previous branch. It may have to do with the "Troubleshooting" commands in the readme. In any case, it seems subsequent commits to this branch build fine.

The `dist` folder shouldn't need to be built for travis but should be done before `npm publish` and that is handled by the `prepublishOnly` command. SFR and any other repo (I tried with with SimplyE's Circulation Admin) should be able to import components like before.

src/icons/package.json
```diff
@@ -8,7 +8,6 @@
     "url": "https://github.com/NYPL/nypl-design-system/tree/development/src/icons"
   },
   "scripts": {
-    "prepare": "npm run build",
```
hokei: this still needs to be removed?

EdwinGuzman: Yeah. I can try to see what Travis thinks and then revert if it fails.

EdwinGuzman:
Adding back ` "prepare": "npm run build",` in `icons` works!

src/icons/package.json
hokei: 🎉

hokei:
I've been having trouble getting `npm link` to work here ... I think with the addition of webpack, running `npm link @nypl/design-system-react-components` in ResearchNow is causing issues?  Are you getting similar problems?  

It seems to work if you start the server, and then do the link (Though icons don't show up) 
I'm also ok with this if there's another way to work simultaneously in both DS and the project that consumes it (I'm not that big of a fan of `npm link`, but I do think we need a way to work locally in both projects at the same time)

EdwinGuzman:
`npm link` works for me but that command never actually compiles the webpack bundle. You'd have to run `npm run dist` locally either before or after running `npm link`. I do not have issues seeing the icons or working locally.

There is the issue that if you need to work locally on both, once the repos are linked, you'd have to run `npm run dist` locally every time you make a new change in DS for RN. I can add a command to watch the files and build using webpack. That sounds more manageable (even if you have to remember to run that whenever developing locally.

hokei:
I think I'm still getting the write access permission - I don't have it on install, but when I cd into `react-components` and `npm link` it still tells me `no such file or directory, access '/Users/crystalngai/NYPL/nypl-design-system/src/react-components/node_modules/@nypl/design-system-icons'` which I assume isn't a problem for you anymore?


Edit:  removing `react-componens/node_modules/@nypl` before running `npm link` solves this problem.  why it's adding an empty folder, I have no clue ... 

Edit 2:  Got it running after installing webpack and webpack-cli!

hokei:
OK yay I'm working with this and the icons look great!  I've pushed up a change with the in-project webpack installations.  Source maps don't work for me quite yet, we can go through that in the new year!

hokei (approved):

src/react-components/package.json
```diff
@@ -6,14 +6,16 @@
...
   },
-  "main": "lib/index.js",
+  "main": "dist/design-system-react-components.min.js",
   "types": "lib/index.d.ts",
   "author": "NYPL",
   "license": "Apache-2.0",
   "scripts": {
     "prepare": "npm run build",
-    "prepublishOnly": "npm run test",
+    "prepublishOnly": "npm run lint && npm run test && npm run dist",
```
hokei: nit: can take out `npm run lint` because it is already run in `npm run test` -- dw if it's already ready to merge tho

PR #62, opened 2019-12-19
Updating documentation in the readmes.

### LINKED ISSUE #59: Individual component documentation should have instructions on how to incorporate styles
Right now instructions to consumers re: compiling styles lives in the main `readme.md` but they should live in the readmes under `react-components` and `twig`, since that's what shows up on NPM and also where people will look when they're trying to install stuff.

EdwinGuzman:
Resolves #59.

- Update the readmes in the styles and react component packages. Not very familiar with the twig components so did not include it.

### DIFF
--- a/src/styles/README.md
+++ b/src/styles/README.md
@@ -1,4 +1,36 @@
 # NYPL Design System Styles
+
+For an overview of the entire Design System, check out the Design System [README](https://github.com/NYPL/nypl-design-system/blob/development/README.md).
+
 ## How to Use
+The Design System Styles provides the full and uncompiled scss files. The product is responsible to import the styles into its project. To import the package, run
 
+```
+  npm install --save @nypl/design-system-styles
+```
 
+Configurations need to be updated to support importing files. Check out the [storybook config](https://github.com/NYPL/nypl-design-system/blob/development/storybook/.storybook-react/webpack.config.js) as a full example.
+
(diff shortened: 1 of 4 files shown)


hokei (approved):

src/styles/README.md
```diff
@@ -1,4 +1,36 @@
...
+For an overview of the entire Design System, check out the Design System [README](https://github.com/NYPL/nypl-design-system/blob/development/README.md).
+
 ## How to Use
+The Design System Styles provides the full and uncompiled scss files. The product is responsible to import the styles into its project. To import the package, run
+
+```
+  npm install --save @nypl/design-system-styles
+```
+
+Configurations need to be updated to support importing files. Check out the [storybook config](https://github.com/NYPL/nypl-design-system/blob/development/storybook/.storybook-react/webpack.config.js) as a full example.
```
hokei: change this to [react storybook config](https://github.com/NYPL/nypl-design-system/blob/development/storybook/.storybook-react/webpack.config.js) or [twig storybook config]((https://github.com/NYPL/nypl-design-system/blob/development/storybook/.storybook-html/webpack.config.js)

hokei:
Looks good! The twig stuff... isn't used anywhere right now, and we're not actually sure it can be, so I think it's ok to leave it out for now.

PR #64, opened 2019-12-23
Testing storybook.

EdwinGuzman:
- POC to get two instances of storybook running at the same time and trying to use Github pages for it.

I did this under my own forked repo because I wanted to test on Github Pages. Ignore everything under `storybook/storybook-static` (for now) since I don't have Travis and wanted to deploy right away.

If you think the workflow is fine, I'll close this and make a better PR in the repo itself with a Travis hook similar to [Reusable-Components](https://github.com/NYPL-Simplified/reusable-components/blob/master/.travis.yml). The `build-storybook` script already gets installed by storybook so not much needs to be done. The main entrypoint, however, does need to be created. The `gh-pages` branch also needs to be created ahead of time. In this case, `index.html` gets picked up by the `gh-pages` branch.

Travis will run `npm run storybook:static` and then deploy it to `gh-pages`. Locally, this can be tested locally with `npm run storybook:static` and then opening up `index.html` in a browser (no server needed). Since I wanted to test it on Github Pages, here's a link from my repo - http://www.edwinguzman.com/nypl-design-system/. Of course, the main page can be updated however it needs to be and it'll have a URL that is similar to https://nypl.github.io/@nypl/design-system.


.gitignore
```diff
@@ -2,6 +2,5 @@ node_modules
 dependencyGraph.json
 public
 .DS_Store
-storybook/storybook-static/
```
EdwinGuzman, line 5: This will be added back later.

hokei:
Pinging @helenvholmes for a second opinion -- In general, even though it's not going to be a thing that works with PRs, I think it's still useful for us to have an easy way to share with the org what we've got going on in design system, so this seems like a natural extension of our workflow so far -- I like it!

helenvholmes:
I agree with @hokei—we have nothing right now, and this is infinitely better than that. I say ship it 👍🏻

hokei (approved):

EdwinGuzman:
👍 will make a better PR soon.

PR #65, opened 2019-12-31
SFR-620 header with search

hokei:
[ResearchNow Jira Ticket](http://jira.nypl.org/browse/SFR-620)

## **Things this PR has changed** 

- `Link` missing URL error message
- Consolidated `Header` and `PageTitle`
- Added ResearchNow specific `SearchBar` error state 
- Made `SearchBar` follow the A11Y guidelines in Metronome

### Front End Review:
- [ ] View [the example in Storybook](http://localhost:9001/?path=/story/headerwithsearch--search-promo)
- [ ] Check against the [metronome page](http://themetronome.co/components/oh-31-header-with-search/?fresh=true)


src/styles/02-molecules/search-bar/_search-bar.scss
```diff
@@ -39,6 +39,10 @@ $search-bar--breakpoint: 856px;
 
     flex: 1 1 70%;
     min-height: 40px; // Accommodates the border for the search-bar without parameters
+    
+    &--error::placeholder { //TODO: IE11 support
```
hokei, line 43: @helenvholmes can I get 👀 on this?  (https://caniuse.com/#search=%3A%3Aplaceholder) 
(this is trying to deal with the special ResearchNow error situation - http://themetronome.co/components/oh-31-header-with-search/?fresh=true)

PS if this seems like a bigger thing, let's file a bug.

helenvholmes: Looks like styling it with red isn't possible in IE11, is that what I'm understanding?

I say we follow up with Willa and ask if just having the text without a styled color is a graceful enough degradation for IE11. Unfortunately I'm not a CSS expert when it comes to IE :(

hokei: Styling a placeholder isn't possible with IE11, and I think needs the extra `webkit-` `moz-` whatever tags for the older stuff.  (I'll add those in, it's not a big lift and we'll need them eventually, I bet)

src/styles/01-atoms/text/headings/_headings.scss
```diff
@@ -6,7 +6,7 @@
 @mixin heading-xl($color: $black) {
   color: #{$color};
   font-size: 2.25rem;
-  font-weight: normal;
+  font-weight: lighter;
```
hokei: @helenvholmes Every time I see H1 (even in screenshots from reno) it seems like it's light -- is this an ok change to make?

helenvholmes: You are completely right. I'm going to push up some changes I've made in Reno to this file.

I got there by following the weights set in [this library](https://github.com/jonathantneal/system-font-css/blob/gh-pages/system-font.css)—I assume `lighter` works too, but I felt this was easier for mapping back later... this would probably be good documentation to write down at some point.

hokei:
[ ] note to self: Add a story for AT-264: H2

src/styles/02-molecules/search-bar/_search-bar.scss
helenvholmes: We don't support older than IE11. Since IE11 styling isn't supported, I think it's moot—we probably ship what's here and let the browser style it natively.

hokei: Oh sorry I mean ... older than current, newer than IE11 stuff, like Edge 18 (needs `-ms-`)

helenvholmes: Sounds like something we should make sure our webpack config handles—is this blocking for MVP?

hokei: Nope, let's throw it on the pile 
https://github.com/NYPL/nypl-design-system/projects/2#card-31347325

hokei:
@helenvholmes 

figma: 
Screen Shot 2020-01-08 at 8 29 33 AM
sb: 
Screen Shot 2020-01-08 at 8 30 28 AM

Metronome: http://themetronome.co/components/at-264-h2/?fresh=true
I'm thinking this class looks a lot like AT-251: H3 and if it's easy we can include that under h2 as well, but if not, throw it on the pile too 

EDIT: talked online, decided to punt.

src/react-components/src/components/02-molecules/SearchBar/SearchBar.tsx
```diff
@@ -28,14 +33,27 @@ export interface SearchBarProps {
...
 
-  const { searchBarId, buttonId, dropdownId, dropdownOptions, selectedField, searchValue, placeholderText, textFieldAriaLabel,
-    hasError, errorMessage, selectBlurHandler, searchSubmitHandler, selectChangeHandler, searchChangeHandler } = props;
+  const { blockName,
+    searchBarId, buttonId,
+    searchBarAriaLabel, searchBarAriaLabelledBy,
+    dropdownId, dropdownOptions, dropdownAriaLabel,
+    selectedField, searchValue,
+    placeholderText,
+    textFieldAriaLabel,
```
EdwinGuzman: This is not being used in this component.

src/react-components/src/components/02-molecules/SearchBar/SearchBar.tsx
```diff
@@ -28,14 +33,27 @@ export interface SearchBarProps {
...
+    helperVariant,
+    hasError, errorMessage,
+    selectBlurHandler, searchSubmitHandler, selectChangeHandler, searchChangeHandler } = props;
 
   if (dropdownOptions) {
-    if (!(dropdownId && selectBlurHandler && selectChangeHandler)) {
+    if (!(dropdownId && dropdownAriaLabel && selectBlurHandler && selectChangeHandler)) {
       throw new Error("If dropdownOptions are passed, dropdownId, selectChangeHandler, and selectBlurHandler must also be passed");
```
EdwinGuzman: Missing `dropdownAriaLabel` from the error string.

src/react-components/src/components/02-molecules/SearchBar/SearchBar.tsx
```diff
@@ -67,7 +86,11 @@ export default function SearchBar(props: SearchBarProps) {
     iconDecorative: true,
   };
 
-  return <form className={bem(searchbar__base_class, modifiers)} id={searchBarId} aria-label="Search for keyword, author, title, or subject">
+  return <form className={`${bem(searchbar__base_class, modifiers, blockName)} ${bem(searchbar__base_class, modifiers)}`}
+    id={searchBarId}
+    role="search"
+    aria-label={searchBarAriaLabel}
+    aria-labelledby={searchBarAriaLabelledBy}>
```
EdwinGuzman, line 96: Is this a case where you need either but not both?

src/react-components/src/components/03-organisms/Headers/HeaderWithSearch.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+};
+
+export const researchNowSearchHeader = () => <HeaderWithSearch
+  searchButtonId="searchButtonId"
+  searchBarAriaLabel="Search research catalog"
+  advancedSearchElem={<UnderlineLink url={"advanced-search-link"}>Advanced Search</UnderlineLink>}
+  searchBarId="searchBarId"
+  dropdownId="dropdownId"
+  textFieldAriaLabel="Research Now"
+  headingContent={<span>Research<span className={"rn-section-title__emphasis"}>Now</span></span>}
```
EdwinGuzman: Adding this prop doesn't do anything. It's declared but not being used in `HeaderWithSearch`.

src/react-components/src/components/03-organisms/Headers/HeaderWithSearch.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+  headingId="researchNow-page-title-id"
+  headingUrl="#research-now-url"
+  headingBaseClass="rn-section-title"
```
EdwinGuzman: This also doesn't do anything (`blockName` is another prop that is declared but not used in `HeaderWithSearch`).

src/react-components/src/components/03-organisms/Headers/HeaderWithSearch.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  React.cloneElement(advancedSearchElem, {modifiers: ["dark-background"]});
+
+  return (
+    <div className={bem(base_class)}>
+      <div className={bem("content", [], base_class)}>
+        <RNSectionTitle
+          blockName={base_class}
+          id={headingId}
+          url={headingUrl} />
```
EdwinGuzman: Because you use a non-generic component, I think `RNHeaderWithSearch` is a more appropriate name for this component. I imagine this will work great for RN, but for other projects there may be confusion if this `HeaderWithSearch` component is pulled in and it renders "ResearchNOW". I do get this should be out rather quickly so if this name will be temporary, that's fine, but it should be reconsidered.

src/react-components/src/components/03-organisms/Headers/HeaderWithSearch.stories.tsx
hokei: I'm okay either way, but eventually we'll want this to accept a generic header, not just the RN specific one.  It's just a variation that hasn't been implemented yet.

src/react-components/src/components/03-organisms/Headers/HeaderWithSearch.tsx
```diff
@@ -0,0 +1,78 @@
+// OH-31 Header with Search
+import * as React from "react";
+import SearchBar from "../../02-molecules/SearchBar/SearchBar";
+import UnderlineLink from "../../01-atoms/Links/UnderlineLink";
+import bem from "../../../utils/bem";
+import BodyText from "../../01-atoms/Text/Text/BodyText";
+import SectionTitle from "../../02-molecules/Headings/SectionTitle";
+import Heading from "../../01-atoms/Text/Headings/Heading";
```
EdwinGuzman: These three, along with `UnderlineLink`, are not being used in this file.

src/react-components/src/components/03-organisms/SearchPromo/SearchPromo.tsx
```diff
@@ -27,7 +27,9 @@ export interface SearchPromoProps {
   textChangeHandler?: (event: React.FormEvent) => void;
 }
 export default function SearchPromo(props: SearchPromoProps) {
-  const { headingText, modifiers, blockName, titleId, searchBarId, searchButtonId, hasError, errorMessage,
+  const { headingText, modifiers, blockName, titleId,
```
EdwinGuzman, line 30: `modifiers` isn't being used in this file.

EdwinGuzman:
I think the description link to the Storybook change is actually http://localhost:9001/?path=/story/headerwithsearch--research-now-search-promo-header.

src/react-components/src/components/02-molecules/SearchBar/SearchBar.tsx
```diff
@@ -18,6 +22,7 @@ export interface SearchBarProps {
   placeholderText?: string;
   textFieldAriaLabel?: string;
   textFieldAttributes?: {};
+  helperVariant?: "ResearchNow";
```
EdwinGuzman, line 27: This says that `helperVariant` can only be of type "ResearchNow" (just this value). This should be `helperVariant?: string;` and the default prop value for this can then be "ResearchNow". Otherwise, any other value won't be accepted. I don't _think_ this is what you meant if this component is meant to be generic with any value for `helperVariant`.
Screen Shot 2020-01-08 at 4 55 04 PM

src/react-components/src/components/03-organisms/Headers/HeaderWithSearch.stories.tsx
EdwinGuzman: Makes sense but I don't think it should be included as part of an example if it's not working and will be implemented later. It will bring confusion for _this_ release, although for a future implementation it won't be a problem.

src/react-components/src/components/02-molecules/SearchBar/SearchBar.tsx
hokei: This variant is kind of a hack for ResearchNow that is a compromise between existing functionality and the "proper" way to do stuff.  You shouldn't use it unless you're ResearchNow and I hope we're not in the habit of making variants in the future.

src/react-components/src/components/03-organisms/Headers/HeaderWithSearch.stories.tsx
hokei: Makes sense, it's not a big deal to put this kind of stuff back in in the future!

src/react-components/src/components/03-organisms/SearchPromo/SearchPromo.tsx
hokei: So unlike the the other stuff I feel like `blockname`, `baseclass` and `modifiers` are stuff that are useful to have even if they're not used - mostly because it's /such a pain/ to get them all the way down the nested component if we ever need them in the future.  Does it hurt to have them in here?  Do unused variables get stripped out when it builds?  (Is there an easy way to copy a component all the way down the component tree?)

src/react-components/src/components/03-organisms/Headers/HeaderWithSearch.tsx
hokei: yeah that's a better approach to thinking about this work.

src/react-components/src/components/02-molecules/SearchBar/SearchBar.tsx
hokei: Yes, but it should take labelledBy rather than reject it outright if both happen to be passed.  I'll put in test cases to reflect this.

src/react-components/src/components/03-organisms/Headers/HeaderWithSearch.tsx
helenvholmes: I agree but we should be matching Metronome, as its intended to be our source of truth. I'll crosspost this to Slack to discuss.

src/react-components/src/components/02-molecules/SearchBar/SearchBar.tsx
EdwinGuzman: I see. If this should only be used in RN, then I don't think the Design System is the right place for it. I meant my comment to be more of a Typescript issue rather than DS philosophy.

src/react-components/src/components/02-molecules/SearchBar/SearchBar.tsx
EdwinGuzman: You can create a new object to destruct the props so only one attribute is added to the element.

src/react-components/src/components/02-molecules/SearchBar/SearchBar.tsx
hokei: Yeah, this ends up kinda dovetailing into DS philosophy - there's a world in which DS (at least, this storybook) contains everything that exists, so that we can test everything that ever changes to DS.  I don't know what the pros/cons of that world are, I'm about to push something that might be even more tricky re: that so uh... stay tuned. 

Re typescript:  knowing this rationale, do you think this is ok as is (Maybe with a comment or a test that makes this clear?) or do you think it should still accept a generic string?

src/react-components/src/components/01-atoms/Text/Headings/RNSectionTitle.tsx
```diff
@@ -0,0 +1,11 @@
+/* AT-79 Page Title - Level 1
+  This is the ResearchNow Variant, which isn't a semantic heading.*/
+
+// This story serves as documentation and example code for ResearchNow's Section Title.
+// Because ResearchNow uses React-Router's Link, the component itself must be passed in for routing to work properly.
+
+import * as React from "react";
+
+export default function RNSectionTitle(props: React.PropsWithChildren<{}>) {
```
hokei, line 9: @EdwinGuzman 
> I see. If this should only be used in RN, then I don't think the Design System is the right place for it.

So this component (and its story) is gonna go against that whole philosophy.  I think there's a large conversation to be had here re: how to do project-specific overrides.  (Does the CSS for this component belong in the RN project?  I think there are obviously good arguments for "yes", and the argument for "no" is that having all the possibilities live somewhere in DS will make it harder for people to do stuff not explicitly blessed by DS, and possible to set up better automated testing).  Bigger conversation, right now project-specific overrides don't exist so ... I'm hoping this is the ... least ugly of uglies in our menu of options?

src/react-components/src/components/02-molecules/SearchBar/SearchBar.tsx
hokei: Can you link me to an example of this?  I'd love to do and don't know how to haha ... it'll be cleaner than this: https://github.com/NYPL/nypl-design-system/pull/65/files#diff-22a3559aa6a3f1be78ca9ee249fb4172R29

src/react-components/src/components/01-atoms/Text/Headings/RNSectionTitle.tsx
EdwinGuzman: Yeah, I'm all for going with this approach short-term. You're right that seeing specific examples might make it harder for others to not try something else.

Ideally, RN should not have to do overrides (or maybe just write minimal CSS) to create a component like this. I think a goal that should be aimed for later is that RN (and any other app) takes in both the styles and the react components and can compose them however it needs to and _not_ have to do overrides. But, yes, larger conversation and this might make more sense once the components are complete.

src/react-components/src/components/02-molecules/SearchBar/SearchBar.tsx
EdwinGuzman: I think it should accept a generic string and make the value `"ResearchNow"` a default prop. But, if you leave it as is and add a comment (and maybe a todo) about why this is the case, that'd be just as helpful.

src/react-components/src/components/02-molecules/SearchBar/SearchBar.tsx
EdwinGuzman: Yeah, that's what I was thinking:
```
let formProps = {};
if (searchBarAriaLabel) {
  formProps['aria-label'] = searchBarAriaLabel;
} else ....

<form {...formProps} />
```

or `"aria-label": !ariaLabelledBy ? ariaLabel : undefined,`.

src/react-components/src/__tests__/components/Heading-test.tsx
```diff
@@ -15,14 +15,34 @@ describe("Section Headings", () => {
...
-    .to.throw("Headings can only be plain text or bold");
+  it("Can pass heading content as child span", () => {
+    wrapper = Enzyme.shallow(<Heading id="h1" level={1}>Text <span>hi</span></Heading>);
+    expect(wrapper.find("h1")).to.have.lengthOf(1);
+  });
+  it("Throws error when neither child nor text is passed", () => {
+    expect(() => Enzyme.mount(<Heading id="h1" level={9} />))
+      .to.throw("Heading only supports levels 1-6");
+  });
+  it("uses child when both child and text are passed", () => {
```
EdwinGuzman, line 25: This doesn't test that the text for the wrapper is "Text".

src/react-components/src/__tests__/components/Heading-test.tsx
```diff
@@ -15,14 +15,34 @@ describe("Section Headings", () => {
...
   });
   it("Can pass heading content as child", () => {
     wrapper = Enzyme.shallow(<Heading id="h1" level={1}>Text</Heading>);
     expect(wrapper.find("h1")).to.have.lengthOf(1);
   });
-  it("Throws error when invalid heading is passed as child", () => {
-    expect(() => Enzyme.mount(<Heading id="h1" level={3}><span>oh no</span></Heading>))
-    .to.throw("Headings can only be plain text or bold");
+  it("Can pass heading content as child span", () => {
+    wrapper = Enzyme.shallow(<Heading id="h1" level={1}>Text <span>hi</span></Heading>);
```
EdwinGuzman, line 21: This could use an `expect` for the content it renders so it includes the span.

src/react-components/src/__tests__/components/Heading-test.tsx
```diff
@@ -15,14 +15,34 @@ describe("Section Headings", () => {
...
+    expect(wrapper.find("h1")).to.have.lengthOf(1);
+  });
+  it("Throws error when neither child nor text is passed", () => {
```
EdwinGuzman: Not necessary for this PR: It'd be nice to organize the tests so that there are two levels for `describe("Section Headings")`. One would be for errors `describe("errors")` and one for expected renders `describe("renders")` (those names could be better but just an example). For larger components, we like to break it down so we also have a level for component behaviors `describe("behaviors")`.

src/react-components/src/components/01-atoms/Forms/TextField.tsx
```diff
@@ -14,22 +14,22 @@ export interface TextFieldProps {
...
+    throw new Error("Either label or ariaLabel must be defined");
+  }
+
   let textFieldProps = {
     className: bem("input", modifiers, textfieldBlockName),
     "type": "text",
     value: value,
     onChange: onChange,
+    "aria-label": !ariaLabelledBy ? ariaLabel : undefined,
+    "aria-labelledby": ariaLabelledBy
```
EdwinGuzman, line 30: You can include `aria-required` here if you don't want the extra `if (isRequired)` check. Maybe do something like `"aria-required": isRequired`, make `isRequired` optional, and assign a default prop value of `false` (?) It makes sense to not make it required unless if it's necessary and it could help reduce props when calling it - `<TextField ariaLabel="text field" onChange={() => {}}/>;` instead of `<TextField ariaLabel="text field" isRequired={false} onChange={() => {}}/>;`
But this is minor.

src/react-components/src/components/01-atoms/Text/Headings/Heading.tsx
```diff
@@ -5,38 +5,36 @@ import bem from "../../../../utils/bem";
...
-    if (!isValid) {
-      throw new Error("Headings can only be plain text or bold");
-    } else {
-      content = props.children;
-    }
+    content = url ? React.createElement("a", { href: url, className: urlClass }, text) : text;
   }
 
-  let heading_base_class = "heading";
+  let heading_base_class = baseClass ? baseClass : "heading";
```
EdwinGuzman: For these type of props, I like using default props. An alternative is to make "heading" a default value when destructuring (line 20):
```
const { level, id, baseClass = "heading", modifiers, ... } = props;
```

Then, `let heading_base_class = baseClass;` will work for "heading" or any passed in value.

src/react-components/src/components/01-atoms/Text/Headings/RNSectionTitle.stories.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export default {
+  title: "Research Now Section Title",
+  component: RNSectionTitle
+};
+
+
+
+export const researchNowSectionTitle = () => <RNSectionTitle>
+{<a className={`${bem("rn-section-title", [])} rn-section-title`} href={"researchNow-home-url"}>
```
EdwinGuzman: The curly braces are not necessary.

src/react-components/src/components/01-atoms/Text/Headings/RNSectionTitle.tsx
helenvholmes: I think a component like `Header-with-search` belongs in the DS. I think the UX team needs to iterate on this component more to get it into a more abstract state so that it's no longer RN-specific.

src/react-components/src/components/03-organisms/Headers/RNHeaderWithSearch.tsx
```diff
@@ -0,0 +1,66 @@
...
+  modifiers?: [];
+  searchBarId: string;
+  searchButtonId: string;
+  searchBarAriaLabel: string;
+  dropdownId: string;
+  sectionTitle: JSX.Element;
+  textFieldAriaLabel: string;
+  hasError?: boolean;
+  // To make into jsx.element (eg: if error has link)
+  // must first handle the placehodlerText=errorMessage in ResearchNow's Search Header
```
EdwinGuzman: Small typo.

src/react-components/src/components/02-molecules/SearchBar/SearchBar.tsx
```diff
@@ -28,14 +33,28 @@ export interface SearchBarProps {
...
-  const { searchBarId, buttonId, dropdownId, dropdownOptions, selectedField, searchValue, placeholderText, textFieldAriaLabel,
-    hasError, errorMessage, selectBlurHandler, searchSubmitHandler, selectChangeHandler, searchChangeHandler } = props;
+  const { blockName,
+    searchBarId, buttonId,
+    searchBarAriaLabel, searchBarAriaLabelledBy,
+    dropdownId, dropdownOptions, dropdownAriaLabel,
+    selectedField, searchValue,
+    placeholderText,
+    helperVariant,
+    hasError, errorMessage,
```
EdwinGuzman, line 45: The following is more of an idea/example rather than something you _have_ to do.

In SimplyE we have a `FetchErrorData` interface defined as:
```
export interface FetchErrorData {
  status: number;
  response: string;
  url: string;
}
```

This is for making API requests and the object conforms to what the server returns. In components we check (as an example):
```
// ...
error: FetchErrorData
// ...

{ error && <ErrorMessage message={error.response} /> }
```

I'm not sure how RN is structured, but it might be easier to define an interface for errors you expect so you don't have to (1) pass in two props to check and render an error message (2) throw an error if one variable is passed but not the other. This interface pattern can also be used for the other components that use this two-prop pattern. But, again, something to think about.

hokei:
> I think the description link to the Storybook change is actually http://localhost:9001/?path=/story/headerwithsearch--research-now-search-promo-header. 

What do you mean here?

(Oh do you mean the github description?  I think that was us being pre-emptively excited for multidevs, I've been kind of ignoring that until PR multidevs get set up)

src/react-components/src/components/02-molecules/SearchBar/SearchBar.tsx
hokei: ah darn, I thought there was some nifty react syntax to tell React "one of these things must be passed"

src/react-components/src/components/01-atoms/Forms/TextField.tsx
hokei: `TextField` should rarely (almost never) be used on its own, so I like the idea of forcing people to pass in isRequired (because it really should be other DS components calling `TextField`), and those DS components should know whether or not it's required.

src/react-components/src/components/01-atoms/Text/Headings/Heading.tsx
hokei: hmm this won't override baseClass if it exists already?

src/react-components/src/components/02-molecules/SearchBar/SearchBar.tsx
hokei: I'm gonna keep this as is and keep this in mind - Right now errors in RN are pretty dumb but if the error component becomes a problem while trying to implement, I'll come back to it.  

The thing I'm quickly learning is that passing a raw `url` is almost always going to cause problems in RN ... but also this is going to be a recurring problem so we do want to come up with a solid solution.

src/react-components/src/components/01-atoms/Forms/TextField.tsx
EdwinGuzman: Maybe there should be a subset of "helper" components (a folder for them) that only DS uses  so anyone reading the source doesn't think that it's a component they can use.

src/react-components/src/components/01-atoms/Text/Headings/Heading.tsx
EdwinGuzman: No, it's a default value but if anything is passed to `Heading`, it'll use that for `baseClass` instead.

src/react-components/src/components/01-atoms/Text/Headings/Heading.tsx
```diff
@@ -17,9 +17,11 @@ export interface HeadingProps {
 }
 
 export default function Heading(props: React.PropsWithChildren<HeadingProps>) {
-  const { level, id, baseClass = "heading", modifiers, url, urlClass, text, blockName, attributes } = props;
+  const { level, id, baseClass, modifiers, url, urlClass, text, blockName, attributes } = props;
+
+  console.log("baseClass", baseClass);
+  let heading_base_class = baseClass ? baseClass : "heading";
```
EdwinGuzman: The default value didn't work?

src/react-components/src/components/01-atoms/Forms/TextField.tsx
hokei: Agreed, though this should probably be reflected in Metronome as well (components that should only be used by other components).  Yet another thing to go on the pile!

src/react-components/src/components/01-atoms/Text/Headings/Heading.tsx
hokei: It was a different styles issue.  It does work!

EdwinGuzman (approved):

PR #68, opened 2020-01-03
SFR-633 Search Results Item

hokei:
[ResearchNow Jira Ticket](http://jira.nypl.org/browse/SFR-633)

## Description
Added Edition Card and Search Results Items
Changed LinkHelper to BasicLink because it can be used as a generalized link within other stuff

### Front End Review:
- [ ] Check against the [Metronome documentation](http://themetronome.co/components/mc-25-search-results-item/?fresh=true)


src/styles/01-atoms/links/link/_link.scss
```diff
@@ -5,6 +5,10 @@ $color-link--hover: $nypl-blue-dark;
 @mixin link {
   color: $color-link;
 
+  &--no-underline {
```
hokei, line 8: @helenvholmes This is to get that blue text link with no underline -- is this ok? (http://themetronome.co/components/mc-25-search-results-item/?fresh=true)

helenvholmes: Ah, on the header? Willa might actually ask us to change that eventually, now that I'm looking at it...

Code looks good to me, though!

helenvholmes (approved):
👍🏻Your very small CSS updates look fine to me!

src/react-components/src/components/01-atoms/Links/Link.tsx
```diff
@@ -0,0 +1,55 @@
+// Base Link Helper Class
```
EdwinGuzman: I don't think you meant to commit this component. It seems to be an older version of `BasicLink`.

src/react-components/src/components/02-molecules/Cards/EditionCard.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  editionHeadingElement: JSX.Element;
+
+  editionInfo: string[];
+
+  readOnlineLink?: string;
+  downloadLink?: string;
+}
+
+export default class EditionCard extends React.Component<EditionCardProps, {}> {
```
EdwinGuzman: This doesn't have state so it could be a functional component.

src/react-components/src/components/02-molecules/Cards/EditionCard.tsx
```diff
@@ -0,0 +1,66 @@
...
+        <Heading id={id} level={3} blockName={blockName ? blockName : baseClass} >{editionHeadingElement}</Heading>
+        <div className={bem("card-content", [], baseClass)}>
+          <div className={bem("card-image", [], baseClass)}>
+            <Image src={coverUrl} isDecorative={true}></Image>
+          </div>
+          <div className={bem("edition-info", [], baseClass)}>
+            {editionInfo.map((value, index) => {
+              return <span className={bem("edition-info-item", [], baseClass)} key={index}>{value}</span>;
+            })}
+          </div>
```
EdwinGuzman, line 60: It's fine if you need a case where not passing in `editionInfo` still renders this component, but I would do a check around the parent div. In the story example, there's an unnecessary div that is being rendered.
```
// something like
{ editionInfo.length > 0 &&
   (<div ...
   </div>)
}
```

src/react-components/src/components/02-molecules/Cards/EditionCard.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+  readOnlineLink="#readOnlineUrl"
+  downloadLink="#downloadUrl">
+</EditionCard>;
+
+export const editionCardWithEmptyEditionInfo = () => <EditionCard
+  id="card#1"
+  blockName=""
+  coverUrl="https://placeimg.com/300/400/arch"
+  editionHeadingElement={<a href="edition-link" className={bem("link", [], "heading")} >2004 Edition</a>}
+  editionInfo={[]}
```
EdwinGuzman, line 55: I think it's annoying, as a developer using a library, to have to pass in empty values if the prop is required. On a more realistic use-case, if you are getting data from a server, it's likely that instead of returning an empty array (or empty object), you can get `undefined` as a value. In that case, this will break. It's possible that NYPL's API will return empty values if there is no data, but that may not be the case for anyone else. I think a default prop of `[]` may work better.

src/react-components/src/components/02-molecules/Cards/EditionCard.tsx
```diff
@@ -0,0 +1,66 @@
...
+  constructor(props: EditionCardProps) {
+    super(props);
+  }
+
+  getButtonsElement(readOnlineLink: string, downloadLink: string, baseClass: string): JSX.Element {
+    if (!readOnlineLink && !downloadLink) {
+      return <div className={(bem("missing-links", [], baseClass))}>Unavailable to read online</div>;
+    }
+
+    return <div className={bem("card-ctas", [], baseClass)}>
```
EdwinGuzman, line 38: What does `ctas` stand for?

src/react-components/src/components/02-molecules/Cards/EditionCard.tsx
```diff
@@ -0,0 +1,66 @@
...
+      {readOnlineLink &&
+        <BasicLink className={bem("card-info-link", [], baseClass)} url={readOnlineLink}>Read Online</BasicLink>
+      }
+      {downloadLink &&
+        <BasicLink className={bem("card-info-link", [], baseClass)} url={downloadLink}>Download</BasicLink>
```
EdwinGuzman, line 43: Should this be a `button` instead since it's performing an action?

src/react-components/src/components/02-molecules/Cards/EditionCard.tsx
helenvholmes: Calls to action

src/react-components/src/components/02-molecules/Cards/EditionCard.tsx
EdwinGuzman: You should start incorporating some documentation to the components in typedoc or esdoc notation. So in this case, it'll be like:
```
/**
 * EditionCard component that renders information for an edition.
 */
export default class EditionCard...
```

I was going to comment that it would be useful if the component was more generic and allowed more than one online or download link, mostly because this is true for items in SimplyE that can have multiple epub files, pdf files, or other links. But it makes sense to have a specific component where you know it will always have at most two links. In cases like that, I think it could use documentation to make that clear.

helenvholmes: I agree, this is a good suggestion. The twig templates have not-quite-identical but a similar idea in the head defining everything that the template accepts: https://github.com/NYPL/nypl-design-system/blob/edad8da424279a9ae7b3b0ec7b722e74e5f8269a/src/twig/_patterns/03-organisms/event-section/_event-section.twig#L1

src/react-components/src/components/02-molecules/Cards/SearchResultItem.tsx
```diff
@@ -0,0 +1,88 @@
...
+    try {
+      React.Children.only(headingContent as React.ReactElement);
+    } catch (e) {
+      const children = React.Children.map(this.props.children, child =>
+        (child as JSX.Element).type);
+      // Catching the error because React's error isn't as helpful.
+      throw new Error(`Please only pass one child into SearchResultItem, got ${children.join(", ")}`);
+    }
+
+    // TODO: Decide whether this needs to be in DS, and write/find utilities for us to count text within child components
```
EdwinGuzman: Didn't know you couldn't count text in children!

src/react-components/src/components/02-molecules/Cards/SearchResultItem.tsx
```diff
@@ -0,0 +1,88 @@
...
+    // }
+
+    return (
+      <div className={bem(baseClass, modifiers, blockName)}>
+        <Heading id={id} level={2} blockName={blockName ? blockName : baseClass} >{headingContent}</Heading>
+        <div className={bem("subtitle", [], baseClass)}>{subtitleText}</div>
+        <div className={bem("author", [], baseClass)}>
+          {authorLinkElement}
+        </div>
+        <EditionCard
```
EdwinGuzman: Is it possible for an item to have more than one edition or will it always have one?

src/react-components/src/components/02-molecules/Cards/SearchResultItem.tsx
```diff
@@ -0,0 +1,88 @@
...
+    // if (headingContent.innerText > 80) {
+    //   throw new Error("Section Title (h2) Text must be fewer than 80 characters");
+    // } else if (headingContent.textContent > 60) {
+    //   console.warn("Section Title (h2) Text should be fewer than 60 characters");
+    // }
+
+    return (
+      <div className={bem(baseClass, modifiers, blockName)}>
+        <Heading id={id} level={2} blockName={blockName ? blockName : baseClass} >{headingContent}</Heading>
+        <div className={bem("subtitle", [], baseClass)}>{subtitleText}</div>
```
EdwinGuzman: From the story example with no subtitle, would it better to have it as a default prop instead of having to pass an empty string? Or perhaps make the empty string a default (`subtitleText = ''` on line 46)? If it's empty, this div is still being rendered.

src/react-components/src/components/02-molecules/Cards/SearchResultItem.tsx
hokei: There's this utility that exists: https://github.com/fernandopasik/react-children-utilities 
But I'll need time to figure out whether we want to re-implement the text counting or include the package from them.  Probably good to put in a DS ticket

src/react-components/src/components/02-molecules/Headings/SectionTitle.tsx
```diff
@@ -30,9 +30,9 @@ export default class SectionTitle extends React.Component<SectionTitleProps, {}>
     const baseClass = "heading-section";
 
     if (headingText.length > 80) {
-      throw new Error("Section Title Text must be fewer than 80 characters");
+      throw new Error("Section Title (h2) Text must be fewer than 80 characters");
     } else if (headingText.length > 60) {
-      console.warn("Section Title Text should be fewer than 60 characters");
+      console.warn("Section Title (h2) Text should be fewer than 60 characters");
```
EdwinGuzman, line 35: I think I better understand the whole DS system a little better now. A developer would be checking multiple documentation in metronome as well as this React library, right? That way they know that `SectionTitle` will only render an h2, right?

src/react-components/src/components/02-molecules/Cards/SearchResultItem.tsx
helenvholmes: https://github.com/NYPL/nypl-design-system/projects/2#card-31474290

src/react-components/src/components/02-molecules/Headings/SectionTitle.tsx
helenvholmes: Yes. Developers will receive Metronome documentation that will (in the future) have Storybook embeds of the relevant stories for both the Twig and React implementation of the component. Those pages have the requirements that the component was built to.

src/react-components/src/components/02-molecules/Cards/SearchResultItem.tsx
```diff
@@ -0,0 +1,88 @@
...
+  render(): JSX.Element {
+    const { id,
+      resultIndex,
+      modifiers, blockName, headingContent,
+      subtitleText,
+      authorLinkElement,
+      editionInfo, editionsLinkElement} = this.props;
+    const baseClass = "search-result-item";
+
+    try {
```
EdwinGuzman: The one-child test is being done here but `this.props.children` isn't being rendered in the final component so it doesn't seem necessary (and there's no example of it being used in the stories).

src/react-components/src/__tests__/components/SearchResultsItem-test.tsx
```diff
@@ -0,0 +1,45 @@
+import { expect } from "chai";
+import * as Enzyme from "enzyme";
+import * as React from "react";
+
+import SearchResultItem from "../../components/02-molecules/Cards/SearchResultItem";
+
+describe("SearchResultsItem", () => {
```
EdwinGuzman: Note that this file has `Result` as plural while the component is singular.

src/react-components/src/components/02-molecules/Cards/SearchResultItem.tsx
EdwinGuzman: Okay! I get this now. This is for `headingContent` to check that it is just one element that is being passed but _not_ for `SearchResultItem`. This should be changed to:
```
    try {
      React.Children.only(headingContent.props.children as React.ReactElement);
    } catch (e) {
      const children = React.Children.map(headingContent.props.children, child =>
        (child as JSX.Element).type);
      // Catching the error because React's error isn't as helpful.
      throw new Error(`Please only pass one child into headingContent, got ${children.join(", ")}`);
    }
```

There should be a test to catch this.

EdwinGuzman: I'd like to say that maybe it shouldn't be the responsibility of `SearchResultItem` to check this but it makes sense to have it here. The parent container could make sure that the prop that gets passed to `headingContent` only has one child, but that delves into the app's responsibility.

src/react-components/src/components/02-molecules/Cards/EditionCard.tsx
EdwinGuzman: This has been resolved: download action falls into the link category.

src/react-components/src/components/02-molecules/Cards/SearchResultItem.tsx
hokei: It seems like `<Heading>` should really test for having only one child.  It isn't like this because `Research<em>Now</em>` is counting as two children, but is there harm in requiring something like that to be wrapped in a span?

src/react-components/src/components/02-molecules/Cards/SearchResultItem.tsx
hokei: It will always only have one, per the metronome spec.

src/react-components/src/components/02-molecules/Cards/EditionCard.stories.tsx
hokei: Agreed!  I'm gonna go in and add defaultProps to stuff in this PR, but I would love to go back and add them to everything - especially all the `baseClass`, `modifiers`, etc.  (They're just new to me (thank you!) and I think they're super nifty)

helenvholmes: Satisfying future work for post-MVP

src/react-components/src/components/02-molecules/Cards/SearchResultItem.tsx
EdwinGuzman: That's true about `Heading` testing it and maybe this part should go there instead of the check happening here. It could be either 1) have every individual component test that it receives one child or 2) check the data at the top level and then pass it down appropriately. If RN needs that heading name then that's fine and enforcing that the child is wrapped in a span is a good idea.

src/react-components/src/__tests__/components/SearchResultItem-test.tsx
```diff
@@ -0,0 +1,51 @@
...
+import * as Enzyme from "enzyme";
+import * as React from "react";
+
+import SearchResultItem from "../../components/02-molecules/Cards/SearchResultItem";
+
+describe("SearchResultItem", () => {
+  const validHeading = <a href="hello">headingText</a>;
+const invalidHeading =  <>< a href = "hello" > headingText</a> <span>badText</span></>;
+
+const EditionInfo = (heading: JSX.Element) => {
```
EdwinGuzman: Check your spacing.

src/react-components/src/__tests__/components/SearchResultItem-test.tsx
```diff
@@ -0,0 +1,51 @@
+import { expect } from "chai";
+import * as Enzyme from "enzyme";
+import * as React from "react";
+
+import SearchResultItem from "../../components/02-molecules/Cards/SearchResultItem";
+
+describe("SearchResultItem", () => {
+  const validHeading = <a href="hello">headingText</a>;
```
EdwinGuzman, line 8: From your earlier comment, it does make sense to have the logic and this test in the `Heading` component.

EdwinGuzman (approved):

src/react-components/src/components/02-molecules/Cards/SearchResultItem.tsx
hokei: The way it's been happening now is that the bottom level component is throwing the error.  However, React isn't tracing its path through (could be the way we set up source maps or something?) so I agree that having a "You passed too many children into SearchPromo" is gonna be a lot more useful than "You passed too many children into Heading".  But I think it's ideal if we could be like, "you passed too many children into Heading, which is called by searchheading, which is called by searchpromo" the way that it would show up if everything was in one project.  
Gonna file another DS card to think about this, but it's yet another "later" thing

PR #69, opened 2020-01-03
Adding missing script and rearranging.

EdwinGuzman:
- I forgot to add `npm run storybook:static` in the Travis file but I think it needs to be merged from a PR in order for Github/Travis to deploy.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,9 @@ script: npm test
 deploy:
   provider: pages
   script: npm run storybook:static
+  on:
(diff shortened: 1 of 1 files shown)


hokei:
wait is it added here then?  Or did it have to be added in Travis?

EdwinGuzman:
It's added in the Travis file but not coming up as a diff. I previously and accidentally committed that change directly to `development`.

hokei (approved):

PR #70, opened 2020-01-03
SFR-603 changed normalize

hokei:
[ResearchNow Jira Ticket](http://jira.nypl.org/browse/SFR-603)

- Changes `normalize.css` for `scss-resets/_normalize` for better importing
- Changes `breakpoint` in `wrapper` mixin so that it reads `max-width` correctly

### Front End Review:


package.json
```diff
@@ -47,6 +47,7 @@
     "mocha": "^6.2.2",
     "prop-types": "^15.7.2",
     "rimraf": "^3.0.0",
+    "scss-resets": "^0.5.2",
```
helenvholmes: I don't think we need this dependency—if we're looking for parity with Storybook, we should be including normalize. If we're looking to the future when we'll be delivering the design system as a minified `style.scss`, we should probably be including nothing.

src/styles/00-base/base/_03-mixins.scss
```diff
@@ -33,7 +33,7 @@
   margin: #{$v-margin} auto;
   padding: #{$v-padding} #{$h-padding};
 
-  @include breakpoint($container-max-width) {
+  @include breakpoint(max-width $container-max-width) {
```
helenvholmes, line 36: What is the addition here doing? I could be wrong but I would have expected this to throw a compiler error.

helenvholmes (changes requested):
As this PR currently stands I have some major concerns about this branch. I don't think we're thinking about this problem in the right way.

In SFR, the contents of `normalize.css` are being injected in a `<style>` tag in the header. It seems that it's messing with the order of operations with our styles. It's unclear why this is happening on SFR and it shouldn't be happening.

In Storybook, webpack compiles our SCSS for us so we're able to see it in the browser. `normalize` is a dependency in our project and is included before any of our written styles, so the order of operations is correct, hence the Storybook styles always looking correct.

It sounds like the current version of SFR has a dependency management issue where `normalize.css` isn't getting pulled in correctly when SFR compiles the `nypl-design-system` SCSS. In the future, we're hoping to also deliver compiled, minified CSS as a `style.css`, so it'll cease to be a problem.

As for how to proceed here, since we need a solution ASAP for SFR, my recommendation is to remove all of the imports and dump `normalize.css`'s contents from [here](https://github.com/necolas/normalize.css/blob/master/normalize.css) into a `lib/` folder and figure out imports later. It preserves the exact code we've been using thus far, and maintains a reference to the library we're using that I'm more comfortable with.

helenvholmes:
Resolving all my PR comments after writing up that huge comment—let's just dump `normalize` into a file and import it first like the other files and fix imports later.

src/styles/00-base/base/_03-mixins.scss
hokei: Per the documentation (http://breakpoint-sass.com/) 

> // assume min-width (by default) if only a number
$high-tide: 500px;

> // if one value is a string, assume a feature/value pair
$surfboard-width: max-width 1000px;

In our storybook: before -- 
Screen Shot 2020-01-03 at 3 30 19 PM

after -- 
Screen Shot 2020-01-03 at 3 31 34 PM

helenvholmes: Super weird, that wasn't broken before... too many mysteries these days 😂

helenvholmes (approved):
👍🏻

PR #71, opened 2020-01-03
Not ignoring the static storybook folder.

EdwinGuzman:
### Front End Review:
- Made this a PR because it's important: the static sites will need the local directory `storybook/storybook-static` that gets generated by Storybook. Unfortunately, this cannot be git-ignored. It just needs to be known that this subfolder cannot/should not be committed to the repo and that the static script is only used by Travis.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@ node_modules
 dependencyGraph.json
 public
 .DS_Store
-storybook/storybook-static/
 src/react-components/lib/
 src/react-components/dist/
 src/icons/dist
--- a/README.md
+++ b/README.md
@@ -51,4 +51,6 @@ Once the Design System Styles package is imported, follow the steps in the React
 
 ### Storybook
 
+There should be no need to run the static Storybook instances. Following the instructions above, the two development Storybook instances can run on a server that tracks changes. If the static Storybook npm script is run, make sure to not commit the `storybook/storybook-static` directory.
+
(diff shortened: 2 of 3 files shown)


EdwinGuzman:
The main landing page works https://nypl.github.io/nypl-design-system/ but the links don't because the folder is currently ignored by git.

hokei:
Hmm is the reason why `reuseable-components` doesn't need this is because it doesn't have that extra landing page? Like, Travis needs to be able to read it from git to deploy?  
I'm a bit wary of committing stuff that we explicitly can't touch - is there a way we deploy from a generated version of something, or a temporary branch, or something like that?

EdwinGuzman:
`reusable-components` does the same pattern, as in, the `storybook-static` directory is not in that repo's gitignore file. Besides the landing page (which I don't think is a big difference), the main difference is that in `reusable-components` I'm telling Travis to deploy a specific directory only. That directory has only the static site files.

Here, however, I cannot do that same subdirectory because of the dependencies from the other subdirectories. The main issue was storybook not being able to find `../../node_modules/src/react-components` which made sense since those files didn't exist. So now it reverted back to pushing everything up to the `gh-pages` but unfortunately, having to be careful about not committing any static storybook files that a user could generate locally. Ideally, a developer wouldn't need to run the static storybook script anyway.

hokei (approved):

hokei:
Is there a built-in mechanism to make and re-make `storybook-static`, then? I'm thinking about the situations where new components are added or old components are removed - for example there's a known issue where `lib` can get stale, but because it doesn't get committed it feels ok to remove and re-generate at any time.

EdwinGuzman:
The `lib` folder should be regenerated in each Travis container when it runs tests and when it runs the static command. So, there shouldn't be a problem with the `lib` folder getting stale. Every time a new PR is merged into `development`, Travis will run the storybook static command and push the new changes.

PR #72, opened 2020-01-06
SFR-637 pagination

hokei:
[ResearchNow Jira Ticket](http://jira.nypl.org/browse/SFR-637)

- Adds Dropdown and Pagination functionality, and its supporting components

### Front End Review:
- [ ] Check against the [Metronome documentation: Dropdown](http://themetronome.co/components/mf-49-drop-down/?fresh=true) [Pagination](http://themetronome.co/templates/t-53-search-results-/?fresh=true)


src/react-components/src/components/01-atoms/Links/Link.tsx
```diff
@@ -41,8 +42,12 @@ export default class Link extends React.Component<LinkProps, {}> {
       }
     });
 
+    elementChildren.map(child => {
+      return React.cloneElement(child, {key: uid(child)});
```
hokei, line 46: @EdwinGuzman I'd love an extra pair of eyes on this - it seems overkill to use a generator but I'm also not sure what other good candidates there are for keys

src/react-components/src/components/02-molecules/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,80 @@
...
+    labelId, labelPosition, labelText, ariaLabel,
+    helperContent, helperContentId,
+    hasError, errorContent, errorContentId,
+    dropdownOptions,
+    onSelectChange, onSelectBlur } = props;
+
+  let modifiers = [];
+
+  const baseClass = "dropdown";
+  if (!["left", "top", "none"].includes(labelPosition)) {
```
EdwinGuzman, line 37: You'd be better off declaring the type of `labelPosition` to be of type `"left" | "top" | "none"` instead of `"string"`. Then you do not have to make this check here. This is great for Typescript but I do wonder how the compiled version of it holds up. It probably does this same error check and throw behinds the scenes for you.

You should also declare the default prop value of it to be `"none"`.

src/react-components/src/components/01-atoms/Links/Link.tsx
EdwinGuzman: Hmm I'm not too sure what this does or how it's working. If you're using `.map` then it's returning an array which you're not assigning to anything. So maybe a `forEach` is better. Are you attempting to add a key to each element in `this.props.children`?

hokei: I'm attempting to add a key to each element in `this.props.children` - creating an element by passing in `children` as array means that React thinks a list is passed, so everything needs a key.  It throws warnings, but I'm not actually sure that the thing it's warning me about is the thing it thinks it's warning me about ...

hokei: (Or is it that using `React.createElement` with an array of children is an ... anti-pattern?)

EdwinGuzman: Short recap of 1-1 talk: I'm not sure if it's an anti-pattern. It may be better to not require the user to pass in an anchor element and instead render it for the user.

src/react-components/src/components/02-molecules/Dropdown/Dropdown.tsx
EdwinGuzman: I'd like to see this change in the Typescript code, but when it is used by others, it actually doesn't matter unless if they use Typescript in the app and use the Typescript source files. Otherwise, the compiled code doesn't really check for this and the dev could potentially use a value other than "left", "top", "none". So maybe you can include both the typescript type check as well as keeping this error check. I'll try to test it out.

src/react-components/src/components/02-molecules/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,80 @@
+/* MN-10: Pagination */
```
EdwinGuzman, line 1: This component, along with `FormDropdown`, are missing from the `index.ts` file to allow it to be exported in the package.

src/react-components/src/components/02-molecules/Dropdown/Dropdown.tsx
EdwinGuzman: Okay, I tested this in a SimplyE typescript app and in SFR and the compiled code doesn't care about the Typescript type. So I would have both checks in place to catch compile time and run time errors in Typescript or non-Typescript projects.

src/react-components/src/components/01-atoms/Links/Link.tsx
hokei: I made a PR for this issue separately, because it's affecting other work it's better to merge it first -- https://github.com/NYPL/nypl-design-system/pull/73

src/react-components/src/__tests__/components/Dropdown-test.tsx
```diff
@@ -0,0 +1,161 @@
...
+    );
+    expect(wrapper.find("FormDropdown").dive().find("select")).to.have.lengthOf(1);
+    expect(wrapper.find("Label")).to.have.lengthOf(0);
+  });
+
+  it("rejects invalid labelPosition", () => {
+    expect(() => Enzyme.mount(<Dropdown
+      dropdownId="dropdown"
+      isRequired={false}
+      labelPosition="invalid"
```
EdwinGuzman: If you add the Typescript type check, this test can be removed.

src/react-components/src/__tests__/components/FormDropdown-test.tsx
```diff
@@ -16,20 +16,36 @@ describe("FormDropdown", () => {
...
+    wrapper = Enzyme.shallow(<FormDropdown labelId="label" isRequired={false} dropdownId="hi" options={["opt1", "opt2", "opt3"]} onSelectChange={changeCallback} onSelectBlur={blurCallback} />);
+    expect(wrapper.find("Icon")).to.have.lengthOf(1);
   });
 
   it("sends callback with value on change", () => {
     wrapper = Enzyme.shallow(<FormDropdown
-      dropdownId="hi" labelText="hi"
+      labelId="label"
+      isRequired={false}
+      dropdownId="hi"
```
EdwinGuzman: Purely stylistic: some of these calls are either on one long line, missing some indentation, or in this "one line per prop". I prefer the latter and would stick with one just for consistency, but again, this is not detrimental to the PR and can be done post-MVP. Maybe introducing [prettier](https://prettier.io/) might be helpful later.

helenvholmes: Seconding prettier. [Added a card](https://github.com/NYPL/nypl-design-system/projects/1#card-31366341) to our backlog.

src/react-components/src/components/02-molecules/Helpers/HelperErrorText.tsx
```diff
@@ -0,0 +1,31 @@
...
+import * as React from "react";
+import bem from "../../../utils/bem";
+
+export interface HelperErrorTextProps {
+  id: string;
+  isError: boolean;
+  modifiers?: string[];
+  blockName?: string;
+  baseClass?: string;
+  attributes?: {};
```
EdwinGuzman: Is this meant to be any other HTML attribute such as `aria-label`, `data-something`, etc?

If so, you can remove this `attributes` prop. On line 15 you can do
```
const { id, isError, ...rest } = props;
```
And on line 30 you can write
```
return <span id={id} className={bem(blockName, modifiers, baseClass)} {...rest}>{props.children}</span>;
```

`...rest` can be called anything so you can call it `...attributes` if you want to. It just destructures `props` so it holds everything that hasn't been explicitly assigned a variable. It may be easier than having to do 
`<HelperErrorText attributes={ 'aria-label': 'label', ...}>`.

src/react-components/src/components/02-molecules/Helpers/HelperErrorText.tsx
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export default function HelperErrorText(props: React.PropsWithChildren<HelperErrorTextProps>) {
+  const { id, isError, attributes } = props;
+  let blockName = props.blockName ? props.blockName : "input-description";
+  let baseClass = props.baseClass ? props.baseClass : "form-item";
```
EdwinGuzman: These should be default props. You can assign them after the function declaration:

```
HelperErrorText.defaultProps = {
  blockName: "input-description",
  baseClass: "form-item",
};
```
and here destructure them: `const { id, isError, attributes, blockName, baseClass } = props;`.

src/react-components/src/components/02-molecules/Helpers/HelperErrorText.tsx
```diff
@@ -0,0 +1,31 @@
...
+  if (isError) {
+    modifiers.push("error");
+  }
+  if (props.modifiers) {
+    modifiers.push(...props.modifiers);
+  }
+
+  if (!props.children) {
+    throw new Error("Helper and Error Text must have content, or else it should not be used");
+  }
```
EdwinGuzman: This is not a big deal here because this is a small component but for larger components with more logic, this check should go higher up so it can throw error right away and not do anything else.

src/react-components/src/__tests__/components/Pagination-test.tsx
```diff
@@ -0,0 +1,48 @@
...
+      previousPageHandler={previousCallback}
+      nextPageHandler={nextCallback}
+      onSelectChange={changeCallback}
+      onSelectBlur={blurCallback} />
+    );
+    expect(wrapper.find("Button")).to.have.lengthOf(2);
+    expect(wrapper.find("Dropdown")).to.have.lengthOf(1);
+  });
+
+  it("Renders two buttons and a Select when there are zero pages", () => {
```
EdwinGuzman, line 38: This doesn't make much sense to me. Why render anything if the total pages is 0? You'll have a non-functional UI element.
Screen Shot 2020-01-08 at 3 25 11 PM

src/react-components/src/components/01-atoms/Button/Button.tsx
```diff
@@ -42,11 +44,14 @@ export default class Button extends React.Component<ButtonProps, {}> {
       buttonModifiers.push(type);
     }
 
-    let btnContent = [content];
+    let btnContent: any[] = [];
```
EdwinGuzman: I'm not a fan of `any`. You can create a type and assign it for the `content` prop:
```
type BtnContent = string | JSX.Element;
...

content: BtnContent; // instead of `content: string | JSX.Element`
...

let btnContent: BtnContent[] = []; // instead of `let btnContent: any[] = [];`
```

src/react-components/src/components/01-atoms/Forms/FormDropdown.tsx
```diff
@@ -16,30 +18,39 @@ export interface FormDropdownProps {
...
-  if (!labelText && !ariaLabel) {
-      // TODO: Assign aria-labeledBy to labelText
-      throw new Error("Either labelText or ariaLabel must be defined");
+  const { dropdownId, blockName, options, labelId, isRequired, ariaLabel, disabled, selectedOption, onSelectBlur, onSelectChange } = props;
+  const modifiers = props.modifiers ? props.modifiers : [];
+  if (!labelId && !ariaLabel) {
+    throw new Error("Must either have labelId or aria-label");
   }
 
   let formItemBlockName = blockName ? blockName : "form-item";
```
EdwinGuzman: This can be a default prop value.

EdwinGuzman: So can `disabled` below.

src/react-components/src/components/01-atoms/Forms/FormDropdown.tsx
```diff
@@ -16,30 +18,39 @@ export interface FormDropdownProps {
...
-        aria-label={ariaLabel}
-        disabled={disabled ? disabled : false}
-      >
-          { options.map((child, key) => {
-            return <option key={key.toString()} aria-selected={child === selectedOption} value={child}>{ child }</option>;
-          }) }
+    <div className={bem("dropdown", modifiers, formItemBlockName)}>
+
+      <select {...selectProps}>
+        {options.map((child, key) => {
```
EdwinGuzman: I would say that there should be a check before the `return` where if `options` is empty, then don't render anything. But, you do have that test where `Pagination` renders everything even if the total pages is 0. I'm not sure of the acceptance criteria for this, but I wouldn't allow components to render anything if they don't have to.

src/react-components/src/components/01-atoms/Forms/Label.stories.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from "react";
+
+import Label from "./Label";
+
+export default {
+  title: "Label",
+  component: Label,
+};
+
+export const label = () => <Label id="label-id" referenceId="should-never-be-used-alone"><div>Lonely</div></Label>;
```
EdwinGuzman, line 10: What a lonely sounding story.

src/react-components/src/components/01-atoms/Forms/Label.tsx
```diff
@@ -0,0 +1,28 @@
+/* AT-50 Label */
+import * as React from "react";
+import bem from "../../../utils/bem";
```
EdwinGuzman, line 3: This isn't being used.

src/react-components/src/components/01-atoms/Forms/Label.tsx
```diff
@@ -0,0 +1,28 @@
...
+  referenceId: string;
+  labelText?: string;
+}
+
+export default function Label(props: React.PropsWithChildren<LabelProps>) {
+  const { id, referenceId, labelText } = props;
+
+  if (!labelText && !props.children) {
+    throw new Error("Label must have content passed as labelText or children.  Got labelText: " + labelText + " and children: " + props.children);
+  } else {
```
EdwinGuzman: You don't need an `else` when you're returning here, but not a big deal.

src/react-components/src/__tests__/components/Label-test.tsx
```diff
@@ -0,0 +1,30 @@
...
+  it("Renders a <label> when passed element as children", () => {
+    wrapper = Enzyme.shallow(<Label id="label" referenceId="so-lonely"><span>Cupcakes</span></Label>);
+    expect(wrapper.find("label")).to.have.lengthOf(1);
+  });
+
+  it("Throws error when neither text nor children are passed", () => {
+    expect(() => Enzyme.mount(<Label id="label" referenceId="so-lonely" />))
+    .to.throw("Label must have content passed as labelText or children.  Got labelText: undefined and children: undefined");
+  });
+});
```
EdwinGuzman, line 36: This might benefit from having a test where `labelText` and text as a child is passed. From the component, the `labelText` takes priority so it would be great to see as a test.

src/react-components/src/components/02-molecules/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  if (hasError && (!errorContent || !errorContentId)) {
+    throw new Error(`prop hasError is ${hasError}, but is missing props errorMessage (got ${errorContent}) and/or errorTextId (got ${errorContent})`);
+  }
+
+  return <div className={bem(baseClass, modifiers)}>
+
+    {labelPosition !== "none" &&
+      <div><Label id={labelId} referenceId={dropdownId}>{labelText}</Label>
+        {isRequired && <span className={bem("required-message", [], baseClass)}>Required</span>}
```
EdwinGuzman: Should the `Required` have some styling?
Screen Shot 2020-01-08 at 3 43 40 PM

src/react-components/src/components/02-molecules/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,80 @@
...
+    <FormDropdown
+      dropdownId={dropdownId}
+      labelId={labelPosition !== "none" ? labelId : undefined}
+      isRequired={isRequired}
+      ariaLabel={labelPosition === "none" ? ariaLabel : undefined}
+      options={dropdownOptions}
+      onSelectChange={onSelectChange} onSelectBlur={onSelectBlur} />
+
+    {helperContent && <HelperErrorText id={helperContentId} isError={false}>{helperContent}</HelperErrorText>}
+    {hasError && <HelperErrorText id={errorContentId} isError={true}>{errorContent}</HelperErrorText>}
```
EdwinGuzman: `HelperErrorText` could use a different name if it's being used for helper content that is not an error, but not necessary for this PR. First time reading it I thought it was _just_ for errors but seeing it used in a different context changes it.

EdwinGuzman (commented):
I _think_ I'm done reviewing this for now.

src/react-components/src/components/01-atoms/Forms/Label.tsx
helenvholmes: This was probably for my benefit—I'll end up using it when I get to this component's styles.

src/react-components/src/__tests__/components/Pagination-test.tsx
hokei: I'm still on a stubborn path where as much as possible I want problems to not break the page (hence not throwing an error).  This test serves as documentation for that case.  I think this is an approach we will want to revisit re: front end strategy.

EdwinGuzman: Hmmm it would be an approach where the DS _always_ renders the component and the app is responsible for appropriately rendering or not rendering it. I'm okay with that approach but it should be noted somewhere for reference.

hokei: that's a good way of putting it - It would eventually go in some introductory documentation that goes through DS philosophy, but I'm gonna leave it here as a reminder to put a bullet point in the readme.

hokei:
@helenvholmes I had to push a bunch of changes to make it work with SFR.  They're more extensive than I hoped so if they're conflicting feel free to push your changes to a fork and I can handle the merge.

src/react-components/src/components/01-atoms/Forms/FormDropdown.tsx
```diff
@@ -15,34 +17,80 @@ export interface FormDropdownProps {
...
-          { options.map((child, key) => {
-            return <option key={key.toString()} aria-selected={child === selectedOption} value={child}>{ child }</option>;
-          }) }
-      </select>
-
-      <Icon decorative={true} name={"arrow"} modifiers={["small"]} />
-    </div>
-  );
+  componentDidMount() {
+    console.log("mounted props", this.props);
```
EdwinGuzman: Remove the console.logs.

src/react-components/src/components/01-atoms/Forms/FormDropdown.tsx
```diff
@@ -15,34 +17,80 @@ export interface FormDropdownProps {
...
+    this.setState({ selectedOption: this.props.selectedOption });
```
EdwinGuzman: I _think_ the right way to set state from props is in the constructor and not in `componentDidMount`, unless if this is now the better way to do this. But, I think it makes more sense to do it in the constructor to not trigger a rerender when the component loads.

src/react-components/src/components/02-molecules/Dropdown/Dropdown.tsx
hokei: There've been so many things moving around (I've been keeping a branch for integration into RN work) that I'll make a separate "integrating into RN PR" that has this file.

src/react-components/src/components/01-atoms/Forms/FormDropdown.tsx
hokei: Yes it does!  I think it's still a good rule of thumb to not put state setting in lifecycle places if you can avoid it and my brain just goofed.

src/react-components/src/components/02-molecules/Helpers/HelperErrorText.tsx
hokei: the difference between this and 
`  const { id, blockName = "input-description", baseClass = "form-item", isError, ...rest } = props;` is purely stylistic, right?

src/react-components/src/components/02-molecules/Dropdown/Dropdown.tsx
hokei: In general I think this is going to be the case working with design system - we should export something not when it's created, but when it gets used by the first product that needs it

src/react-components/src/components/01-atoms/Forms/FormDropdown.tsx
hokei: >  I wouldn't allow components to render anything if they don't have to. 

I like this principle ... It does both the thing that I want (don't break peoples pages if they make boo-boos) and the thing that you want (don't display stuff that doesn't make sense).  I should ... maybe ... use this elsewhere too.

src/react-components/src/__tests__/components/Label-test.tsx
```diff
@@ -23,6 +23,12 @@ describe("Label Test", () => {
     expect(wrapper.find("label")).to.have.lengthOf(1);
   });
 
+  it("if both <label> children and labelText prop is passed, use labelText", () => {
```
EdwinGuzman: Small text change - `are` instead of `is`.

EdwinGuzman (approved):
Looks good.

PR #73, opened 2020-01-08
NO-REF fix link errors

hokei:

- Fixes the `Link`, `IconLink` and `UnderlineLink` errors brought up in https://github.com/NYPL/nypl-design-system/pull/72
- It turns out the original approach is closer to what we want, due to the fact that `react-router` `<Link>`s can be passed into `<UnderlineLink>` and `<IconLink>`.  But hopefully this refactoring fixes some bugs and makes it clearer why it's as complicated as it is.  

### Front End Review:
N/A


src/react-components/src/components/01-atoms/Links/LinkHelper.tsx
```diff
@@ -0,0 +1,57 @@
+// Base Link Helper Class
+import * as React from "react";
+
+export interface LinkHelperProps {
+  className: string;
+  icon?: { element: JSX.Element };
```
EdwinGuzman, line 7: Are there other components that use the ` { element: JSX.Element }` type for an icon? It seems to add an extra layer to create an object for this icon element when it doesn't have any other properties (unless if you expect to add more later?).

src/react-components/src/components/01-atoms/Links/LinkHelper.tsx
```diff
@@ -0,0 +1,57 @@
...
+    if (React.Children.count(this.props.children) > 1) {
+      // There should only be one child
+      // but React.Children.only() can't cast to React.ReactElement for typechecking
+      // Therefore, React.Children.map() must be used.
+
+      throw new Error(`Please only pass one child, got ${this.props.children}`);
+    }
+
+    let link = React.Children.map(this.props.children, (child: React.ReactElement, index) => {
+      if (child.type !== "a") {
```
hokei: This rejects React.Router `<Link>` elements, which moots the point.  Looking into it again

src/react-components/src/components/01-atoms/Links/LinkHelper.tsx
hokei: I don't know if I understand the question -- but, the reason why `icon` needs to passed in here as an element is that we want the anchor tag to wrap both the text and the icon, as per: 
http://themetronome.co/components/mt-18-text-link-w-icon/?fresh=true

As for why it's `JSX.Element` rather than `React.ReactNode` or anything else ... it was the first thing I found that made the compiler happy /hides

hokei:
For the interest of time, I'm gonna forego all the checking and pass the component through.  This means that you will be able to pass links that don't link to anything, but that's for future us to handle.

src/react-components/src/components/01-atoms/Links/LinkHelper.tsx
```diff
@@ -0,0 +1,57 @@
...
+      ...attributes
+    };
+
+    if (React.Children.count(this.props.children) > 1) {
+      // There should only be one child
+      // but React.Children.only() can't cast to React.ReactElement for typechecking
+      // Therefore, React.Children.map() must be used.
+
+      throw new Error(`Please only pass one child, got ${this.props.children}`);
+    }
```
EdwinGuzman: You can cast `this.props.children` as a `React.ReactElement` for `React.Children.only()`. I tried that thinking it would make things easier but then ran into issues when the child was a _string_. So I played around with it for a while and found you can try something like:

```
    let child: React.ReactElement;
    // If the value is a string, then it's only one element but wrap it in a fragment.
    if (typeof this.props.children === 'string') {
      child = <>{this.props.children}</>;
    } else {
      // otherwise it is a react element but could have multiple children. If that's
      // the case, you can check that it's only one element.
      child = this.props.children as React.ReactElement;
      try {
        React.Children.only(child as React.ReactElement)
      } catch(e) {
        // Catching the error because React's error isn't as helpful.
        // This will tell you _where_ the error is happening.
        throw new Error(`LinkHelper, Please only pass on child: ${e}`);
      }
    }
```

If you do this, then you can use `child` like you are doing below without having to do the `.map` since you know it's only one element. And you can return `link` (not `link[0]`) if you assign it `React.createElement` or `React.cloneElement`.

Is this cleaner or easier to understand than your solution? I don't know but this is just an alternative (and I also wanted to find out how `React.Children.only` works). Your solution is fine as is, especially if it's easier for you to understand.

src/react-components/src/components/01-atoms/Links/IconLink.tsx
```diff
@@ -43,6 +43,6 @@ export default class IconLink extends React.Component<IconLinkProps, {}> {
 
     let className = bem(iconLink__base_class, moreLinkModifiers, blockName);
 
-    return <Link url={url} className={className} attributes={attributes} children={children} icon={icon} />;
+    return <LinkHelper url={url} className={className} attributes={attributes} icon={icon}>{children}</LinkHelper>;
```
EdwinGuzman, line 46: I think this is better than `children={children}` 👌

src/react-components/src/__tests__/components/UnderlineLink-test.tsx
```diff
@@ -24,14 +24,14 @@ describe("Link With Underline", () => {
   });
   it("throws an error if text is passed but no url is passed", () => {
     expect(() => Enzyme.mount(<UnderlineLink>Test</UnderlineLink>))
-      .to.throw("IconLink needs prop \"url\"");
+      .to.throw("if children Test has no anchor tag, please pass URL");
   });
   it("Throws error when non-link component passed without URL", () => {
     expect(() => Enzyme.mount(<UnderlineLink><span>Test</span></UnderlineLink>))
-    .to.throw("IconLink needs prop \"url\"");
+    .to.throw("if children [object Object] has no anchor tag, please pass URL");
```
EdwinGuzman: `[object Object]` is hard to understand and I think you can get away with just saying "the children", but I would change `URL` to `please pass a 'url' prop`.

hokei: I'd LOOOOVE to figure out how to print a truncated tostring() because when integrating it we get so many of these kinds of errors and they're very hard to track down.  February.

EdwinGuzman (approved):
Looks good but I would just change the error messages that get thrown. You _don't_ have to update `LinkHelper` with the one child check alternative but just to give you an idea.

src/react-components/src/components/01-atoms/Links/LinkHelper.tsx
hokei: oh my gosh this is so beautiful thank you
My brain's been on a hamster wheel about this all morning

src/react-components/src/components/01-atoms/Links/LinkHelper.tsx
EdwinGuzman: I meant more like, does `icon` have to be an object? Can it just be `icon?: JSX.element`? In the parent components I see where the object is being created but it doesn't seem necessary.

EdwinGuzman: Not a big deal, just curious.

src/react-components/src/__tests__/components/UnderlineLink-test.tsx
EdwinGuzman: Maybe something like?:
```
const children = React.Children.map(this.props.children, child =>
   (child as JSX.Element).type);

throw new Error(`Please only pass one child, got ${children.join(', ')}`);
```

It'll give you `Error: Please only pass one child, got span, span` for `<UnderlineLink url="#passed-in-link"><span>test</span><span>hello</span></UnderlineLink>`.

EdwinGuzman:
> For the interest of time, I'm gonna forego all the checking and pass the component through. This means that you will be able to pass links that don't link to anything, but that's for future us to handle.

Sounds good.

src/react-components/src/components/01-atoms/Links/LinkHelper.tsx
hokei: Yeah, I don't know why it needed to be its own object... I think maybe in the past it had a key or something.

PR #75, opened 2020-01-10
Adding HOC to verify that a component or component object only has on…

EdwinGuzman:
- Introduces a HOC error boundary-like component that verifies that the component passed only has one child. If it does, it renders the component as usual, otherwise, it throws an error.
- Thought this might be helpful because I saw the same pattern that is in `BasicLink` in `SearchResultItem`. I thought it might be nice to reuse the code that checks it. Then I noticed that `SearchResultItem` uses it for a prop's children and not for its own children! That made it trickier (since now the prop is a component object) but can achieve a similar result.
- I haven't really flushed this out yet and don't have tests but tested two scenarios for `BasicLink` and the `headingContent` in `SearchResultItem`:
  - You can test it by playing around with the link-related stories by adding multiple children, or with `SearchResultItem` and passing in multiple children for the `headingContent` prop.

Let me know what you think and I can continue. If there will be more components that need to only render one child, this might be useful (and I also like curry functions/HOCs).


EdwinGuzman:
This is also based on https://github.com/NYPL/nypl-design-system/pull/68/files#r365379499

src/react-components/src/components/01-atoms/Links/BasicLink.tsx
```diff
@@ -61,3 +51,6 @@ export default class BasicLink extends React.Component<BasicLinkProps, {}> {
     return linkElement;
   }
 }
+
+const BasicLinkWError = WithOneChild<BasicLinkProps>(BasicLink);
+export default BasicLinkWError;
```
hokei: Is there a simpler naming convention?  With this one, it doesn't matter as much because it's internal, but I do like consumers being able to use simple names, like `DS.Searchbar` rather than `DS.SearchBarWithError` (also implies that `DS.Searchbar` is useable and that's not true)

hokei:
In general I think this is awesome and going to be super helpful!  It's bad because it feels like a new hammer and now I want to go through and re-evaluate a bunch of the other stuff to see if it can benefit from something like this.  Eg:  can something like `withIcon(link)` use this?  Or is it better still to keep them as separate components because they are thought of as separate things in the design system documentation?   

In terms of folder structure, I think it might be useful to have `/hoc` under `/helpers/hoc` to differentiate it under `/helpers/internalcomponents` Probably a good place to start thinking about organizing the tests folder too (not a part of the PR, just stuff that this is starting to get me thinking about)

hokei (changes requested):
Still needs some work (tests, documentation) since it's a wip but ... looking good!

src/react-components/src/components/01-atoms/Links/BasicLink.tsx
EdwinGuzman: I think it's usually `with...`. I made quick and bad names just as a POC but I'll update.

EdwinGuzman:
Great, if it'll be useful I'll work on it.

EdwinGuzman:
Updated and added some tests. @adriana-alter , not mandatory but I'd like you to review this if you get a chance. As the description says, this is basically to wrap components so that they only allow one child to be passed in. If multiple children are passed, the component will throw an error. Checking for the length of children was a repeated pattern I thought could be solved with this, but it may be more complicated than necessary.

src/react-components/src/helpers/hoc/WithOneChild.tsx
```diff
@@ -0,0 +1,55 @@
+import * as React from "react";
+
+/**
+ * Wrapper component to verify that a component only has one child.
+ * @param Component Component to test, either a React element
+ *   or an object element.
+ */
+const withOneChild = <P extends object>(
+  Component: React.ComponentType<P> | any
```
EdwinGuzman, line 9: This (`ComponentType`) is to allow class or functional components.

src/react-components/src/__tests__/helpers/hoc/WithOneChild-test.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+describe("withOneChild HOC", () => {
+  let wrapper: Enzyme.ReactWrapper<{}, {}>;
+
+  // This DummyComponent can take as many children that are passed
+  // and render them.
+  it("should only modify components when used", () => {
+    const multipleChildrenDummy = (
+      <DummyComponent title="Wizard" age="old">
+        <p>Gandalf</p>
```
hokei, line 34: 😂

adriana-alter:
Happy to review!  I'll take a look tomorrow...or is it time-sensitive?

EdwinGuzman:
It's not. I'm not sure if you've written a HOC before but I want you to take a look because I'd like to introduce it in SimplyE with our next round of updates.

src/react-components/package.json
```diff
@@ -17,7 +17,7 @@
     "dist": "webpack --mode production --progress --display-modules --config webpack.prod.js",
     "watch-webpack": "webpack --mode development --watch --progress --display-modules --config webpack.prod.js",
     "lint": "tslint -c tslint.json src/__tests__/components/*.tsx src/components/**/**/*.tsx src/components/**/**/**/*.tsx src/utils/*.tsx",
-    "test": "rm -rf lib && npm run lint && tsc && mocha --require lib/__tests__/*.js lib/__tests__/**/*.js"
+    "test": "rm -rf lib && npm run lint && tsc && mocha --require lib/__tests__/*.js lib/__tests__/**/*.js lib/__tests__/**/**/*.js"
```
hokei: Can we use `--recursive` here so we're not continuing to pile stuff on?  
https://mochajs.org/#usage

src/react-components/src/helpers/hoc/WithOneChild.tsx
```diff
@@ -0,0 +1,55 @@
...
+  ({ ...initProps }) => {
+    // It's more likely that the component passed will be a
+    // React element so set these as default.
+    let componentChildren = initProps.children;
+    let componentToRender = Component;
+    let name = Component.name;
+    let props = initProps;
+
+    // In case the component is an object!
+    // E.g, a component passed as a prop.
```
hokei, line 20: I'm not completely sure what's happening here -- does this mean that this will work when a component is passed as a prop, like this:  https://github.com/NYPL/nypl-design-system/blob/development/src/react-components/src/components/02-molecules/Cards/EditionCard.tsx#L25 ?  

I'm wondering if we need this - every time a component needs to be passed as a prop it should be wrapped in something else (where the passed component is children).  Perhaps it would be clearer or simpler if we don't allow do one-child checking when a component is passed as a prop.

hokei:
All in all it looks really good to me!

src/react-components/package.json
EdwinGuzman: This is great.

src/react-components/src/helpers/hoc/WithOneChild.tsx
EdwinGuzman: That's right. It was to cover a similar case so it can work: https://github.com/NYPL/nypl-design-system/blob/HOC-error-children/src/react-components/src/components/02-molecules/Cards/SearchResultItem.tsx#L36

You're right, every time a component is passed it does need to be wrapped. It's could be a superfluous check, but I think covering that case _just in case_ is okay to do.

src/react-components/src/helpers/hoc/WithOneChild.tsx
```diff
@@ -0,0 +1,55 @@
...
+    if (Component && typeof Component === "object") {
+      componentChildren = Component.props.children;
+      name = Component.type;
+      props = Component.props;
+      componentToRender = Component.type;
+    }
+
+    // If the passed component is a string, wrap it in a fragment.
+    if (componentChildren && typeof componentChildren === "string") {
+      componentChildren = <>{componentChildren}</>;
```
adriana-alter, line 30: Under what circumstances would the component be a string?

adriana-alter:
I may be missing the basic premise here: what's the rationale for components needing to have only one child, and under what circumstances would someone plausibly be trying to pass in multiple children and need to be stopped?

Anyway: this looks good!  It does seem potentially overcomplicated, but I don't know of a simpler way to do this...

src/react-components/src/helpers/hoc/WithOneChild.tsx
EdwinGuzman: If you have a link wrapper component, for example, that can take in `<b>Click me</b>` or just `"Click me"`. So it would cover the cases where the child is a string or a component.

EdwinGuzman:
Part of that question can be answered by @hokei for the DS. As for the component, you can have a component and a developer can pass in multiple children:

```
<MyComponent>
  <p>one child</p>
  <p>another child</p>
  <p>one too many</p>
</MyComponent>
```

but you expect there to only be one child. A developer should read the documentation, but just in case they don't... then this is a safety check.

For this case, it does some overly complicated but reusable. The pattern itself is useful for more complicated use cases.

adriana-alter:
Okay, that makes sense.  Looks good!

hokei:
@adriana-alter the reason why we want to do one-child checking is for stuff like headers: http://themetronome.co/components/ai-228-h1-headline/?fresh=true or links: http://themetronome.co/components/mt-173-text-link/?fresh=true 
The idea is that the component should only wrap one thing as content.  Naively, this is "one thing" is plain text.  
However, sometimes a consumer is using something like `react-router`, so instead of passing in plain text and url, it has to pass in a `React-Router.Link`.  Or sometimes you have a heading like `Research<em>Now</em>` which gets parsed as two children.  There's a lot of things that can be passed and are conceptually valid in something called "Header" or "Link" and writing rules for them is tedious and brittle, so this is a kind of middle ground - hopefully if you're going around changing the content of components and not really thinking about it, it'll throw an error.  

(This helps simplify a lot of other code and testing, which is nice!)

hokei (approved):

Issue #78, opened 2020-02-12
How often will we expect consumers to want to do experiments driven by product with no UX input?


hokei:
ResearchNow wants to do this:  https://jira.nypl.org/browse/SFR-89

This requires putting in a link in the no-download-links case, which requires actually changing stuff in the design system.  Right now, since it's me, I can do that, but how do we do something like this going forward?  Is it reasonable to say "you can't" or is there a way we can make the design system flexible enough to accommodate something like this?

helenvholmes:
@hokei I can answer this question from a UX perspective, which is... we hope never.

The UX team is really hoping to get NYPL to stop doing user flow testing on code (at least until getting simply prototype flows up and running with the design system is relatively easy).  We believe Figma prototypes are a better alternative as they can be built out much faster. We hope testing on code in the future will be for identifying last nits.

Going to close this out for now.

PR #79, opened 2020-02-18
SFR-89 allows passing in of missing link component to editioncard

hokei:
[ResearchNow Jira Ticket](http://jira.nypl.org/browse/SFR-89)

- in EditionCard, changes component that shows when no links are available to something that can be custom passed by the consumer

### Front End Review:

### DIFF
--- a/src/react-components/CHANGELOG.md
+++ b/src/react-components/CHANGELOG.md
@@ -3,6 +3,10 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 Currently, this repo is in Prerelease.  When it is released, this project will adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
+## PRERELEASE
+### Changed
+- added `noLinkElement` to `EditionCard` to recieve an element
+
 ## [0.0.11] — 2020-01-23
 ### Added 
 - `EditionsList`
--- a/src/react-components/CHANGELOG.md
+++ b/src/react-components/CHANGELOG.md
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 Currently, this repo is in Prerelease.  When it is released, this project will adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 ## PRERELEASE
 ### Changed
-- added `noLinkElement` to `EditionCard` to recieve an element
+- added `noLinkElement` to `EditionCard` to receive an element
 
 ## [0.0.11] — 2020-01-23
 ### Added 
--- a/src/react-components/src/components/01-atoms/Links/UnderlineLink.tsx
+++ b/src/react-components/src/components/01-atoms/Links/UnderlineLink.tsx
@@ -12,7 +12,7 @@ export interface UnderlineLinkProps {
 
 export default class UnderlineLink extends React.Component<UnderlineLinkProps, {}> {
   constructor(props: UnderlineLinkProps) {
-    super(props);
+  super(props);
   }
 
   render(): JSX.Element {
--- a/src/react-components/src/components/01-atoms/Links/UnderlineLink.tsx
+++ b/src/react-components/src/components/01-atoms/Links/UnderlineLink.tsx
@@ -12,7 +12,7 @@ export interface UnderlineLinkProps {
 
 export default class UnderlineLink extends React.Component<UnderlineLinkProps, {}> {
   constructor(props: UnderlineLinkProps) {
-  super(props);
+    super(props);
   }
 
   render(): JSX.Element {
--- a/src/react-components/src/components/02-molecules/Cards/EditionCard.tsx
+++ b/src/react-components/src/components/02-molecules/Cards/EditionCard.tsx
@@ -12,7 +12,8 @@ export type EditionDetails = {
   language: string,
   license: string,
   readOnlineLink: string,
-  downloadLink: string
+  downloadLink: string,
+  noLinkElement?: JSX.Element
 };
 
 export interface EditionCardProps {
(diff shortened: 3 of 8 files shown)


hokei:
@helenvholmes FYI (Related to https://github.com/NYPL/nypl-design-system/issues/78) 
I think changes like this should not require design system fudging.  

I wanna point out this line in particular:
https://github.com/NYPL/nypl-design-system/pull/79/files#diff-5e301e3a41f7c1600f81c3919114706bR78
A general consumer is able to write overrides for this, but as a consumer I have no idea if this was an unexpected case for the `missing-link` span and should be reported up to DS as a bug, or if this was unique to my situation.  

The hope is that consumers should use as few css overrides as possible, but we also need to provide guidance on how to tell when overrides are necessary ...

src/react-components/CHANGELOG.md
```diff
@@ -3,6 +3,10 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 Currently, this repo is in Prerelease.  When it is released, this project will adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
+## PRERELEASE
+### Changed
+- added `noLinkElement` to `EditionCard` to recieve an element
```
EdwinGuzman: small typo for `receive`

src/react-components/src/components/01-atoms/Links/UnderlineLink.tsx
```diff
@@ -12,7 +12,7 @@ export interface UnderlineLinkProps {
 
 export default class UnderlineLink extends React.Component<UnderlineLinkProps, {}> {
   constructor(props: UnderlineLinkProps) {
-    super(props);
+  super(props);
```
EdwinGuzman: Small thing but indentation seems off.

helenvholmes:
@hokei the override to me feels like a DS bug, a use case it should have been able to handle.

I agree that we should have documentation for surfacing issues like this one. Unfortunately with our PM strategy up in the air, the answer to what should be _in_ that documentation is blocked by a conversation I'll be having with Garvita/Jessica/Courtney/Sean/Casey on Friday around how we project-ize DS within NYPL.

src/react-components/src/components/03-organisms/EditionsList/EditionsList.tsx
```diff
@@ -23,7 +23,8 @@ export default function SearchResultsList(props: SearchResultsListProps) {
           editionHeadingElement={edition.editionYearHeading}
           editionInfo={[edition.publisherAndLocation, edition.language, edition.license]}
           readOnlineLink={edition.readOnlineLink}
-          downloadLink={edition.downloadLink} />
+          downloadLink={edition.downloadLink}
+          noLinkElement={edition.noLinkElement} />
```
EdwinGuzman, line 27: Curious, where does this attribute get defined for an edition?

src/react-components/src/components/02-molecules/Cards/EditionCard.tsx
```diff
@@ -12,7 +12,8 @@ export type EditionDetails = {
   language: string,
   license: string,
   readOnlineLink: string,
-  downloadLink: string
+  downloadLink: string,
+  noLinkElement?: JSX.Element
```
EdwinGuzman, line 17: I'd recommend adding inline comments like 
```
/** Element to render when there are no links. */
noLinkElement?: JSX.Element
```
mostly because this prop is a bit hard to read.

src/react-components/src/components/03-organisms/EditionsList/EditionsList.tsx
EdwinGuzman: Is this specific to ResearchNow?

src/react-components/src/components/02-molecules/Cards/EditionCard.tsx
```diff
@@ -28,21 +29,30 @@ export interface EditionCardProps {
...
+    readOnlineLink,
+    downloadLink,
+    noLinkElement = <>Unavailable to read online</>
+  } = props;
   const baseClass = "edition-card";
 
   const getButtonsElement = (readOnlineLink: string, downloadLink: string, baseClass: string) => {
     if (!readOnlineLink && !downloadLink) {
-      return <div className={(bem("missing-links", [], baseClass))}>Unavailable to read online</div>;
+      return <div className={(bem("missing-links", [], baseClass))}>{noLinkElement}</div>;
```
EdwinGuzman: This is just me thinking outloud: I would expect `getButtonsElement` to return just a div of buttons or null. Might it make more sense to have something like `const btns = !readOnlineLink && !downloadLink ? getButtonElement(...) : noLinksDiv;` where `noLinksDiv` is `<div className={(bem("missing-links", [], baseClass))}>{noLinkElement}</div>;`. It's not a big deal and this implementation is fine.

src/react-components/src/components/03-organisms/EditionsList/EditionsList.tsx
hokei: It's real tricky because the original designs are made to be specific to ResearchNow.  "downloadLink" and "readOnlineLink" are also specific to ResearchNow, and that this specificity isn't actually that useful in the real world.

So to answer your question, `noLinkElement` gets defined in the consumer (ResearchNow) ... but @helenvholmes and I have had a conversation about EditionCard, we really need to think through this whole thing more deeply, from a more fundamental level.

EdwinGuzman (approved):
Small changes but overall it looks fine.

src/react-components/src/components/03-organisms/EditionsList/EditionsList.tsx
EdwinGuzman: Yeah, makes sense. From a high-level, does this mean that the server returns a "noLinkElement" value which (I'm assuming) a user defines? And if it's not set or there's no value from the server, the default is `"Unavailable to read online"`? Initially, I was thinking it could use the word `default` in the variable name somewhere, but it seemed there was more to this than I knew about.

hokei: Yeah, that's what it is doing.  `noLinkElement` defaults `Unavailable to read online` but isn't always going to be that.

PR #80, opened 2020-02-20
SFR-135 Changed Editioncard fields to accept elements

hokei:

[ResearchNow Jira Ticket](http://jira.nypl.org/browse/SFR-135)

- EditionCard is a bit more flexible
- Made those changes to ReadOnlineLink and DownloadLink while I'm here
- Should I do the same to CoverURL? 

### Front End Review:

### DIFF
--- a/src/react-components/package-lock.json
+++ b/src/react-components/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@nypl/design-system-react-components",
-  "version": "0.0.13",
+  "version": "0.0.14",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/src/react-components/src/__tests__/components/EditionCard-test.tsx
+++ b/src/react-components/src/__tests__/components/EditionCard-test.tsx
@@ -20,7 +20,7 @@ describe("EditionCard", () => {
     blockName=""
     coverUrl="https://placeimg.com/300/400/arch"
     editionHeadingElement={<a href="edition-link" >2004 Edition</a>}
-    editionInfo={[<span id="pub-span">Publisher</span>, "Written in English", <span>License</span>]}
+    editionInfo={[<span id="pub-span">Publisher</span>, "Written in English", <a id="licenceId" href="/licence-page">License</a>]}
     readOnlineLink="#readOnlineUrl"
     downloadLink="#downloadUrl" />;
 
@@ -30,6 +30,16 @@ describe("EditionCard", () => {
     coverUrl="https://placeimg.com/300/400/arch"
     editionHeadingElement={<a href="edition-link">2004 Edition</a>}
     editionInfo={["Published in New York by Random House", "Written in English", "Under Creative Commons License"]}
+  />;
+
+  let elementLinkEditionCard = <EditionCard
+    id="card#1"
+    blockName=""
+    coverUrl="https://placeimg.com/300/400/arch"
+    editionHeadingElement={<a href="edition-link" >2004 Edition</a>}
+    editionInfo={["Published in New York by Random House", "Written in English", "Under Creative Commons License"]}
+    readOnlineLink={<span id="readOnlineSpan">Reading</span>}
+    downloadLink={<span id="downloadSpan">Download</span>}
   />;
 
   let missingLinkNoLinkElement = <EditionCard
@@ -84,5 +94,15 @@ describe("EditionCard", () => {
     expect(card.find("h3")).to.have.lengthOf(1);
     expect(card.find("h3").find("a")).to.have.lengthOf(1);
     expect(card.find("img")).to.have.lengthOf(1);
+    expect(card.find("#pub-span").text()).to.equal("Publisher");
+    expect(card.find("a#licenceId").text()).to.equal("License");
+  });
+
+  it("Passes call-to-action links if it is given as span", () => {
+    let card = Enzyme.mount(elementLinkEditionCard);
+    expect(card.find("h3")).to.have.lengthOf(1);
+    expect(card.find("h3").find("a")).to.have.lengthOf(1);
+    expect(card.find("img")).to.have.lengthOf(1);
+    expect(card.find("#downloadSpan").text()).to.equal("Download");
   });
 });
--- a/src/react-components/src/__tests__/components/EditionCard-test.tsx
+++ b/src/react-components/src/__tests__/components/EditionCard-test.tsx
@@ -15,6 +15,15 @@ describe("EditionCard", () => {
     readOnlineLink="#readOnlineUrl"
     downloadLink="#downloadUrl" />;
 
+  let elementEditionCard = <EditionCard
+    id="card#1"
+    blockName=""
+    coverUrl="https://placeimg.com/300/400/arch"
+    editionHeadingElement={<a href="edition-link" >2004 Edition</a>}
+    editionInfo={[<span id="pub-span">Publisher</span>, "Written in English", <span>License</span>]}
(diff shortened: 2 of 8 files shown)


src/react-components/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@nypl/design-system-react-components",
-  "version": "0.0.13",
+  "version": "0.0.14",
```
EdwinGuzman, line 3: Did you mean to change the version in this file?

hokei: Oh ugh I think this was merge funk.  But ... I'd love your thoughts on how to do this -- 
So the version was bumped for release (https://www.npmjs.com/package/@nypl/design-system-react-components) 
Should I then push that bump to development directly?

EdwinGuzman: I don't think there's a git flow in the readme but that'd be good to have and follow. I normally do work on a feature or development branch and then merge _that_ branch into `master`. Final changes that go into `master` are then published. Is it normal for the current flow to have a release and then merge back down to the development branch? Why?

EdwinGuzman: Wait, your `development` branch _is_ your production branch (currently). I think this should be documented because it seems confusing now. I know this is going through active development, but I don't know why you would want to make a release on a branch that's not stable, and then push that change into the stable branch.

src/react-components/src/components/02-molecules/Cards/EditionCard.tsx
```diff
@@ -50,14 +50,20 @@ export default function EditionCard(props: React.PropsWithChildren<EditionCardPr
...
+
+  const getButtonsElem = (readOnlineLink: string | JSX.Element, downloadLink: string | JSX.Element, baseClass: string) =>
     <div className={bem("card-ctas", [], baseClass)}>
       {readOnlineLink &&
-        <BasicLink className={bem("card-info-link", [], baseClass)} url={readOnlineLink}>Read Online</BasicLink>
-      }
+        getElem(readOnlineLink, "Read Online") }
       {downloadLink &&
-        <BasicLink className={bem("card-info-link", [], baseClass)} url={downloadLink}>Download</BasicLink>
-      }
```
EdwinGuzman, line 60: This is a very nice refactor.

src/react-components/src/components/02-molecules/Cards/EditionCard.stories.tsx
```diff
@@ -47,6 +47,16 @@ export const editionCardWithSomeEditionInfo = () => <EditionCard
   downloadLink="#downloadUrl">
 </EditionCard>;
 
+export const editionCardWithComponentInfo = () => <EditionCard
+  id="card#1"
+  blockName=""
+  coverUrl="https://placeimg.com/300/400/arch"
+  editionHeadingElement={<a href="edition-link" className={bem("link", [], "heading")} >2004 Edition</a>}
+  editionInfo={[<span>Published in New York</span>, "Under Creative Commons License"]}
```
EdwinGuzman, line 55: 1) Whatever is passed as an edition info will always be wrapped in a `<span>`. So now, for this example, the first info text will be wrapped in two spans and the second info text will be wrapped in one.

2) Also, I do understand that there will be documentation on how to use this component and what to pass as a prop, but that won't stop anyone from passing something like this which is correct html markup:
```
editionInfo={[<div><dl><dt>owl</dt><dd>a type of bird</dd></dl></div>, "Under Creative Commons License"]}
```
Screen Shot 2020-02-20 at 2 35 26 PM

This worries me as more props are now more lenient and can accept almost anything. But, what's the rationale for making these props allow an html element?

src/react-components/package-lock.json
hokei: It's not normal, we just kind of started releasing things "as a test" and then ... before we realized, it's now in a state that definitely can be called "production".  And I agree it needs a real git flow and process because it's in that state now.

src/react-components/src/components/02-molecules/Cards/EditionCard.stories.tsx
hokei: So this change is driven by a need for "Under Creative Commons License" to be a link to a license explanations page.  I'm not sure if there's a more restrictive way to allow something like this, because links can be passed as `<a>` or as `<ReactRouter.Link>`.  And if we allow `<a>` should we allow `<em>` or `<strong>` or anything like that?  Should we allow `<span>` so that it can be styled with a class?  if we allow `<span>` how can we stop people from doing the thing you demonstrated?

I think it comes down to a larger question around what this component should be - is it a card that can only be used for editions of books?  Or is it a card that can display an image on the left, some text in the middle and a call to action on the right?  Or something in between?  How might UX want to repurpose something like this?  

I don't have answers for this - All I know is that when product says, "Can you turn those words into a link" and I go "yes, and this will take a day and require approval from the Design System team" ... that ... might be the kind of thing where, over a long time, be a crazily frustrating amount of friction?

EdwinGuzman: Yeah, that makes sense. You could potentially check that what was passed is either an anchor or a react Link component, but on the larger scheme of things, it seems too specific and a lot of logic for this component to handle. So if you will be lenient like this to make it easier for you later on, I think that's fine...

I think you are missing tests for the new prop values which can now accept an element besides a string, like the downloadLink and readOnlineLink.

I was going to say, you should also test `language` and `license` which can now be elements, but those values are not being rendered directly as props in the EditionCard. I now see that those values are being taken out of the `editionInfo` object, merged together, and passed together as `editionInfo` in the `EditionCard`. You should test the case this reason came up and pass in a link for the license.

src/react-components/package-lock.json
hokei: Regardless, this isn't the place to discuss this, so let's do the broken stuff for now.  I've filed this guy: https://github.com/NYPL/nypl-design-system/issues/81 
The way we're doing it right now is suuuuper broken so yeah let's get something better in.

src/react-components/src/__tests__/components/EditionCard-test.tsx
```diff
@@ -84,5 +94,15 @@ describe("EditionCard", () => {
     expect(card.find("h3")).to.have.lengthOf(1);
     expect(card.find("h3").find("a")).to.have.lengthOf(1);
     expect(card.find("img")).to.have.lengthOf(1);
+    expect(card.find("#pub-span").text()).to.equal("Publisher");
+    expect(card.find("a#licenceId").text()).to.equal("License");
+  });
+
+  it("Passes call-to-action links if it is given as span", () => {
```
EdwinGuzman, line 101: Just to verify, what get's passed as `readOnlineLink` and `downloadLink` won't actually be links, right? So if it's passed as a span, it'll just be a span but not a  call-to-action "link"?

hokei: sigh. yeah.  The name is now bad and misleading, but I don't know if now's a good time to make this breaking change.  
Functionally, it should be a link, or contain something that is a link, but we have no way to verify that that is what it will do

EdwinGuzman: I would write a "TODO" comment for the props in the component so you can be reminded to update it later on.

EdwinGuzman (approved):
Functionality is 🆗 but the naming and element checks could be made stricter later on if it becomes a big issue (specifically outside of the NYPL engineers who might use this and not use it to the best of their knowledge).

PR #90, opened 2020-02-24
Fixing webpack compiler config and updating name for list component.

EdwinGuzman:
- Fixes webpack compilation error due to ts-loader and babel both being configured. ts-loader uses babel so removing it and allowing it to (along with the doc-gen loader) to compile the code. This is to fix a _production_ issue with the storybook:static command (but development runs just fine).

So now the [production site](https://nypl.github.io/nypl-design-system/storybook/storybook-static/react/index.html?path=/story/*) won't be broken:
Screen Shot 2020-02-24 at 11 11 37 AM
 
Also, updated a component name but if that's wrong I can change it back.

### Front End Review:
- [ ] Tested locally using the development and production build.

### DIFF
--- a/src/react-components/src/components/03-organisms/EditionsList/EditionsList.tsx
+++ b/src/react-components/src/components/03-organisms/EditionsList/EditionsList.tsx
@@ -3,13 +3,13 @@ import * as React from "react";
 import bem from "../../../utils/bem";
 import EditionCard, { EditionDetails } from "../../02-molecules/Cards/EditionCard";
 
-export interface SearchResultsListProps {
+export interface EditionsListProps {
   baseClass?: string;
   modifiers?: string[];
   blockName?: string;
   editions: EditionDetails[];
 }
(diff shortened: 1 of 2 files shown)


src/react-components/src/components/03-organisms/EditionsList/EditionsList.tsx
```diff
@@ -3,13 +3,13 @@ import * as React from "react";
 import bem from "../../../utils/bem";
 import EditionCard, { EditionDetails } from "../../02-molecules/Cards/EditionCard";
 
-export interface SearchResultsListProps {
+export interface EditionsListProps {
```
EdwinGuzman, line 6: It seemed like a copy/paste issue, but if it's wrong I can change it back.

helenvholmes (approved):
LGTM! Thank you Edwin!

EdwinGuzman:
Thanks! I meant to add @hokei but thanks for reviewing 😅

PR #97, opened 2020-02-28
NO-REF unified react webpack

hokei:

- Allows design-system-react-components to be released with styles, eliminating the need to pull design-system-styles separately

### Front End Review:


src/react-components/package.json
```diff
@@ -21,7 +21,10 @@
   },
   "keywords": [],
   "dependencies": {
-    "@nypl/design-system-icons": "file:../icons"
+    "@nypl/design-system-icons": "file:../icons",
+    "@nypl/design-system-styles": "file:../styles",
+    "mini-css-extract-plugin": "^0.9.0",
```
EdwinGuzman: This can be a devdep, as well as sass-loader.

storybook/package.json
```diff
@@ -36,5 +37,8 @@
       "last 1 firefox version",
       "last 1 safari version"
     ]
+  },
+  "devDependencies": {
+    "webpack-cli": "^3.3.11"
```
EdwinGuzman: Why do you have to install this for multiple folders?

hokei: aaahhh I did it through Lerna and didn't hoist.  Wish it would ask you which folder you needed it in by default ...

EdwinGuzman:
This works well when testing it (through `npm link`) on a SimplyE app. I would just say the README should be updated to reflect that the design-system-styles package is no longer needed. And now, all you need is `@import "~@nypl/design-system-react-components/dist/styles.scss";` in the app's main scss file.

src/twig/package.json
```diff
@@ -14,9 +14,14 @@
...
-    "puppeteer": "^2.0.0"
+    "puppeteer": "^2.0.0",
+    "webpack-cli": "^3.3.11"
   },
   "publishConfig": {
     "access": "public"
+  },
+  "dependencies": {
+    "extract-text-webpack-plugin": "^3.0.2",
+    "sass-loader": "^8.0.2"
```
EdwinGuzman: This PR is for the react components package, so are these updates necessary? And if so, what are they doing? Or is this similar to the Lerna hoisting you mentioned in a different comment?

EdwinGuzman:
Update looks good... but tests :(

hokei:
yeah still figuring out why it's not finding `node_modules/.staging` files.  I can't repro reliably on my machine ...

EdwinGuzman (approved):

PR #98, opened 2020-03-02
Configuring the a11y storybook addon for the twig instance.

EdwinGuzman:
- Configures the storybook addon-a11y plugin. I'm not sure why but the React configurations picks up the `addons.js` correctly but the twig instance needs a `main.js` file. In any case, any addon can now be used but only enabled the accessibility one for now.

Screen Shot 2020-03-02 at 11 24 40 AM

I do have a question about the breadcrumb storybook: it fails. It can't find the `@nypl/design-system-icons` package. [Example](https://nypl.github.io/nypl-design-system/storybook/storybook-static/twig/index.html?path=/story/breadcrumbs--short-breadcrumbs). It looks like you can remove this block and the story works fine: https://github.com/NYPL/nypl-design-system/blob/development/src/twig/_patterns/02-molecules/menus/breadcrumbs/breadcrumbs.twig#L14
The actual error comes from https://github.com/NYPL/nypl-design-system/blob/development/src/twig/_patterns/01-atoms/images/icons/_icon.twig#L19 . It seems this icon is only used for the mobile responsive view. Was this ever working for this story?

### Front End Review:
- [ ] View local instance of Twig Storybook.

### DIFF
--- a/storybook/.storybook-html/config.js
+++ b/storybook/.storybook-html/config.js
@@ -1,8 +1,13 @@
 import '!style-loader!css-loader!sass-loader!import-glob-loader!@nypl/design-system-styles/style.scss';
-import { configure, load, addDecorator } from '@storybook/html';
+import { configure, addDecorator } from '@storybook/html';
+import { withA11y } from '@storybook/addon-a11y';
 const twig = require('twig');
 import bem from '@nypl/design-system-twig/_twig-components/functions/bem';
 import attach_library from '@nypl/design-system-twig/_twig-components/functions/sb_attach-library';
+
+// Add all the addons when starting up. Right now, only the accessibility
+// addon is configured.
+addDecorator(withA11y);
 
 twig.extendFunction("bem", bem);
 twig.extendFunction("attach_library", attach_library);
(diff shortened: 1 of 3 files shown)


EdwinGuzman:
@hokei I restarted the Travis builds and then the builds/tests pass. This is not ideal but I don't know what's up with puppeteer (which seems to be the cause).

hokei (approved):
It might be because storybook-html and storybook-react have slightly different setup.  
This is great!  But also, might be good to keep in mind that we're completely not sure what's up with storybook-twig - @helenvholmes  just did a bunch of work around the Twig patternlab instances (https://nypl.slack.com/archives/CHYEHCJTW/p1583160684082700)

EdwinGuzman:
That slack link isn't working... Yeah, it seems both instances are different and I don't know much about the Storybook version for html/twig and it seems a few addons are also not supported. I can take a closer look at the set-up later on, but some context on the `attach_library` function would be helpful. Was that pull in from somewhere or created for this project?

helenvholmes:
@EdwinGuzman `attach_library` is a PHP include as far as I understand—allows you to associate JS or PHP code to a template.

PR #99, opened 2020-03-03
Allowing the Button component to render content passed as children.

EdwinGuzman:
Fixes an [issue](https://github.com/NYPL/nypl-design-system/projects/2#card-33030028) found in the backlog. Also adding more inline comments here and there.

- Allows the `Button` component to either render content from its `render` prop or with content passed down as children.

### Front End Review:

### DIFF
--- a/src/react-components/src/components/01-atoms/Button/Button.stories.tsx
+++ b/src/react-components/src/components/01-atoms/Button/Button.stories.tsx
@@ -8,6 +8,38 @@ export default {
   component: Button,
 };
 
-export const buttonFilledIcon = () => <Button callback={action("clicked")} id="button" content="search" type="filled" iconPosition="left" iconName="search-small" iconDecorative={true}/>;
+export const buttonFilledIcon = () =>
+  <Button
+    callback={action("clicked")}
+    id="button"
+    content="search"
+    type="filled"
+    iconPosition="left"
+    iconName="search-small"
+    iconDecorative={true}
+  />
+;
 export const buttonOutline = () => <Button callback={action("clicked")} id="button" content="Hello World" type="outline" />;
 export const buttonWithSpan = () => <Button callback={action("clicked")} id="button" type="outline" content={<span>Style my span!</span>}/>;
+export const buttonIconChildrenString = () =>
+  <Button
+    callback={action("clicked")}
+    id="button"
+    type="filled"
+    iconPosition="right"
+    iconName="search-small"
+    iconDecorative={true}
+  >Search!</Button>
+;
+export const buttonIconChildrenElement = () =>
+  <Button
+    callback={action("clicked")}
+    id="button"
+    type="outline"
+    iconPosition="right"
+    iconName="search-small"
+    iconDecorative={true}
+  >
+    <span>I'm wrapped in a span!</span>
+  </Button>
+;
--- a/src/react-components/src/components/01-atoms/Button/Button.tsx
+++ b/src/react-components/src/components/01-atoms/Button/Button.tsx
@@ -8,14 +8,18 @@ export interface ButtonProps {
   id: string;
   /** The action to perform on the <button>'s onClick function */
   callback: (event: React.MouseEvent) => void;
-  content: string | JSX.Element;
+  /** The content to render inside the the button. An alternative
+   * to passing children elements. */
+  content?: string | JSX.Element;
   attributes?: {};
+  /** Used for BEM css convention. */
   modifiers?: string[];
+  /** Used for BEM css convention. */
   blockName?: string;
   large?: boolean;
   type?: string;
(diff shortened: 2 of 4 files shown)


src/react-components/src/components/01-atoms/Button/Button.stories.tsx
```diff
@@ -8,6 +8,38 @@ export default {
...
+export const buttonIconChildrenString = () =>
+  <Button
+    callback={action("clicked")}
+    id="button"
+    type="filled"
+    iconPosition="right"
+    iconName="search-small"
+    iconDecorative={true}
+  >Search!</Button>
+;
```
hokei, line 33: nit (+L21 above +EOF)

src/react-components/src/components/01-atoms/Button/Button.stories.tsx
```diff
@@ -8,6 +8,38 @@ export default {
...
+export const buttonIconChildrenElement = () =>
```
hokei, line 34: Might be not your wheelhouse but ... 
Who do we think stories are for?  Is it necessary for stories to have text as well as span, when visually they look the same?  (Is it confusing if they look the same visually?) . 
For engineers, is there a way we can document that children can be passed as either, the way that we can with props?

src/react-components/src/components/01-atoms/Button/Button.stories.tsx
EdwinGuzman: I'm not sure I know what this means.

src/react-components/src/components/01-atoms/Button/Button.stories.tsx
EdwinGuzman: I think it's fine so all the possible ways to render something are shown. Visually there's not much of a difference but you can look at the title of the story as well as the story's code. Ideally, developers would read the documentation on the component. In this case, it does say that either way can be passed:

Screen Shot 2020-03-03 at 3 03 04 PM
Screen Shot 2020-03-03 at 3 02 59 PM

src/react-components/src/components/01-atoms/Button/Button.tsx
```diff
@@ -8,14 +8,18 @@ export interface ButtonProps {
   id: string;
   /** The action to perform on the <button>'s onClick function */
   callback: (event: React.MouseEvent) => void;
-  content: string | JSX.Element;
+  /** The content to render inside the the button. An alternative
+   * to passing children elements. */
+  content?: string | JSX.Element;
```
hokei, line 13: do you think this is necessary?  Can we just say you gotta pass it as children?

EdwinGuzman: I don't think so, but I didn't know if you _still_ wanted to support `content`, so I left it there.

hokei: technically this is a breaking change but it's not used in RN so it can be removed.  We can make a new ticket for it though, no need to expand this one.

src/react-components/src/components/01-atoms/Button/Button.stories.tsx
hokei: I guess I was wondering if there was a way we can put the children type in docs the way we have it for proptypes ... 
I don't think it hurts to have them both here, but until we know better how people will use the storybook, it's also not clear what is useful and what isn't

src/react-components/src/components/01-atoms/Button/Button.stories.tsx
hokei: oh uh sorry for my weird shorthand -- I'm not sure which linter you're using?  It's weird to have the semicolon on a separate line here (and in line 21) 
(and you're missing the empty line at end of file)

EdwinGuzman: That's weird. The linter doesn't complain, does it for you?

src/react-components/src/components/01-atoms/Button/Button.stories.tsx
EdwinGuzman: I see. It's just that `this.props.children` is prop that is always automatically there and used for pretty much any component. It's also almost never really described or documented the same way (as in a type or interface) since it's always expected. I think just seeing examples may suffice since most React devs can pickup how to use it.

src/react-components/src/components/01-atoms/Button/Button.tsx
EdwinGuzman: Sure. I'll leave it in here for now then.

hokei (approved):
sorry meant to approve yesterday haha

hokei:
resolves #102

PR #100, opened 2020-03-06
storybook-static fixes

hokei:

- Fixes so that Storybook will build

### Front End Review:

### DIFF
--- a/src/react-components/webpack.prod.js
+++ b/src/react-components/webpack.prod.js
@@ -10,7 +10,8 @@ module.exports = {
   devtool: "source-map",
   entry: {
     app: [
-      "./src/index.ts"
+      "./src/index.ts", 
+      // "./src/styles.scss"
     ]
   },
   output: {
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,9 +12,9 @@
       }
     },
     "@babel/compat-data": {
-      "version": "7.8.5",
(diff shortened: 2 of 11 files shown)


src/react-components/webpack.prod.js
```diff
@@ -10,7 +10,8 @@ module.exports = {
   devtool: "source-map",
   entry: {
     app: [
-      "./src/index.ts"
+      "./src/index.ts", 
+      // "./src/styles.scss"
```
hokei, line 14: so this interferes with importing design system in the consumer app.  I was able to build it without, but ... I'm kind of confused about it

hokei:
ugh this is really frustrating.  The package-lock problem is causing it not to build on travis, but if I revert the package-locks so it can build, it will throw the "cannot find icons" error when trying to `npm link`.  
There's an underlying problem here - perhaps with dependencies?

src/react-components/webpack.prod.js
EdwinGuzman: From this and the removal of `import './styles.scss';`, this means that the scss is not available anymore for the consumer app, right?

hokei:
ugh travis

src/react-components/webpack.prod.js
hokei: because it's being copied over to lib and dist, and then imported directly, it seems to still work.

PR #106, opened 2020-04-01
Hero

helenvholmes:
No ticket for this work, part 1 of 2.

- Abstracts together hero types 1&2 from [this Figma frame](https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Master?node-id=10968%3A5314). 
- This PR does not add types 3&4, and the tests are still insufficient... I've never written tests before so I might need to 1:1 with someone to get those up to snuff.
- The `SectionName` decorator also needs some work but felt out of scope for this ticket.

### Front End Review:
- [ ] View [Hero types in Storybook](http://localhost:9001/?path=/story/hero--hero-primary)
- [ ] Check against the Metronome documentation:
  - [ ] [Exhibition Header](http://themetronome.co/components/oh-64-exhibition-header/?fresh=true)
  - [ ] [Header with Image Right](http://themetronome.co/components/oh-30-header-with-image-right/?fresh=true)


EdwinGuzman:
There are some linting issues which are causing the tests to fail.

src/react-components/src/components/00-base/Placeholder.tsx
```diff
@@ -0,0 +1,15 @@
+
+import * as React from "react";
+import bem from "../../utils/bem";
+
+export interface PlaceholderProps {
+  modifiers?: string[];
+  blockName?: string;
+}
+
+export default function Placeholder(props: React.PropsWithChildren<PlaceholderProps>) {
```
hokei, line 14: I know this is really "do as I say and not as I do" but it would be really nice to have comments for all the components saying what they do.

src/react-components/src/components/00-base/SectionName.tsx
```diff
@@ -0,0 +1,6 @@
+import * as React from "react";
+import bem from "../../utils/bem";
+
+const SectionName = ({ children }) => <div className={bem('main', ['research'])} >{children}</div>;
```
hokei, line 4: I wonder if "research" is a thing that should be hardcoded, or if this is a thing that should be passed as a modifier from the consumer (so a Research product would pass "research" but a Locations product could pass "locations", etc)

src/react-components/src/components/03-organisms/Hero/Hero.tsx
```diff
@@ -0,0 +1,119 @@
...
+import Image from "../../01-atoms/Images/Image/Image";
+import Heading from "../../01-atoms/Text/Headings/Heading";
+import { HeroTypes } from "./HeroTypes";
+
+export interface HeroProps {
+  /** Can be Primary, secondary, tertiary, or 50/50.
+    */
+  heroType?: HeroTypes;
+
+  /** To be deleted after Crystal's ID merge
```
hokei: ID generation is gonna get punted, the Filters ticket is large enough as is.

src/react-components/src/components/00-base/Placeholder.tsx
```diff
@@ -0,0 +1,15 @@
...
+  const { modifiers = [], blockName } = props;
```
EdwinGuzman, line 15: These props aren't being used in this component.

src/react-components/src/components/03-organisms/Hero/Hero.stories.tsx
```diff
@@ -0,0 +1,90 @@
...
+};
+
+export const heroPrimary = () => <Hero
+  heroType={HeroTypes.Primary}
+  heroId="2"
+  heading={<Heading 
+    level={1} 
+    id={"1"} 
+    text={"Hero Primary"}
+    blockName={"hero"}
```
EdwinGuzman, line 22: Is it required for the `Heading` to have a blockName of `"hero"`? If that's the case, is it written down somewhere? Also, what happens if you forget to include this prop in `Heading`?

src/react-components/src/components/03-organisms/Hero/Hero.tsx
```diff
@@ -0,0 +1,119 @@
...
+    */
+  heroId: string;
+
+  /** Required heading element.
+    */
```
EdwinGuzman: For short comments, I think you can write it in a single line and it'll work the same `/** Required heading element. */`.

src/react-components/src/components/03-organisms/Hero/HeroTypes.tsx
```diff
@@ -0,0 +1,6 @@
+export enum HeroTypes {
```
EdwinGuzman, line 1: Very nice.

src/react-components/src/components/03-organisms/Hero/Hero.tsx
```diff
@@ -0,0 +1,119 @@
...
+  let backgroundImage = backgroundImageSrc ? "data-responsive-background-image" : "";
+
+  let backgroundImageStyle = backgroundImageSrc ? {backgroundImage: 'url(' + backgroundImageSrc + ')'} : {};
+
+  function createMarkup() {
+    return {__html: subHeaderText};
+  }
+
+  let contentBoxStyling;
+  if (heroType === "PRIMARY") {
```
EdwinGuzman: Can you do `heroType === HeroType.Primary` instead of the string check? It'll be better to reduce strings in the code and use an existing variable.

src/react-components/src/components/00-base/SectionName.tsx
helenvholmes: That's what I'd like to do but it felt out of scope for this ticket—completely agree though!

src/react-components/src/components/03-organisms/Hero/Hero.stories.tsx
helenvholmes: It is required, and it would fail as a result. How would something like this typically be handled?

src/react-components/src/components/00-base/Placeholder.tsx
helenvholmes: Updated!

src/react-components/src/components/00-base/SectionName.tsx
helenvholmes: (To explain _why_ such a small thing wasn't done here, though, is that I'm not entirely sure what all of the site sections are yet. I could create another `enum` just to get this started and clean this up and add on once I've tracked that info down... I will do that, actually)

src/react-components/src/components/03-organisms/Hero/Hero.tsx
helenvholmes: Updated the variable description!

src/react-components/src/components/03-organisms/Hero/Hero.tsx
helenvholmes: Updated!

src/react-components/src/__tests__/components/Hero-test.tsx
```diff
@@ -0,0 +1,37 @@
...
+  let blurCallback;
+  let changeCallback;
+  let submitCallback;
+
+  beforeEach(() => {
+    changeCallback = stub();
+    blurCallback = stub();
+  });
+
+  it("Generates a Hero with a background image", () => {
```
hokei, line 14: This test is good, but I'd have many more tests.  eg:  What happens if there's a HeroTypes.Primary but the user forgets to pass a background image? (which is optional) 
(that kind of stuff - what combinations of things could a user pass, and what kind of errors would be shown if the user passes in unexpected things?

You don't need to check anything that's checked by PropTypes or Typescript, but anything where it's, "On a Primary hero, an Image is required.  On a Secondary hero, a background image is ignored if passed, etc" I'd write a test for that.

src/react-components/src/components/03-organisms/Hero/Hero.stories.tsx
hokei: if all heros should have blockname "hero", I would figure out how to incorporate it into the component itself rather than expecting users to to remember to pass it.

src/react-components/src/components/03-organisms/Hero/Hero.stories.tsx
```diff
@@ -0,0 +1,90 @@
...
+import SectionName from "../../00-base/SectionName";
+import Placeholder from "../../00-base/Placeholder";
+
+export default {
+  title: "Hero",
+  component: Hero,
+};
+
+export const heroPrimary = () => <Hero
+  heroType={HeroTypes.Primary}
```
hokei, line 17: (may be out of scope) I'd love to figure out a way to make it clear that users are supposed to be only using this enum, and how to do that ... Does Typescript throw a useful error when passing in a string or does it let it through?

src/react-components/src/components/03-organisms/Hero/Hero.tsx
```diff
@@ -0,0 +1,115 @@
+// OH-30 Header with Image Right
+import * as React from "react";
+import bem from "../../../utils/bem";
+import Image from "../../01-atoms/Images/Image/Image";
+import Heading from "../../01-atoms/Text/Headings/Heading";
+import { HeroTypes } from "./HeroTypes";
```
hokei, line 4: Stylistic, but I like to have these small, tightly coupled type classes in this component file rather than a separate one

src/react-components/src/components/03-organisms/Hero/Hero.tsx
```diff
@@ -0,0 +1,115 @@
...
+  /** Required ID for tabfocus. */
+  heroId: string;
+
+  /** Required heading element. */
+  heading: JSX.Element;
+
+   /** Optional subheader that displays underneath the
+    * required heading element.
+    */
+  subHeaderText?: string;
```
hokei: change this to JSX.Element

src/react-components/src/components/03-organisms/Hero/Hero.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  /** Optional details area that contains location data. */
+  locationDetails?: JSX.Element;
+
+  /** Content creators can modify the foreground color
+    * when this component is used on Exhibition pages.
+    */
+  foregroundColor?: string;
```
hokei, line 24: What happens when you pass in a non-hex color?  Can we throw a useful error?

hokei: What happens when you pass in a color and you're not using an Exhibition hero?

src/react-components/src/components/03-organisms/Hero/Hero.tsx
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  return (
+    <div className={bem(heroBaseClass, heroModifiers)} data-responsive-background-image id={heroId} style={backgroundImageStyle}>
+      <div className={bem("content", [], heroBaseClass)} style={contentBoxStyling}>
+        {heading}
+
+        {image}
+
+        <p className={bem("subtitle", [], heroBaseClass)} dangerouslySetInnerHTML={createMarkup()} />
```
hokei: If it's just text, I would use fragment:  
`<>{subtitleText}</>` rather than `dangerouslySetInnerHtml`
In this case since I think it's best that subtitle is a `JSX.Element` you can go {subtitle}

src/react-components/src/components/03-organisms/Hero/Hero.tsx
hokei: This is my bad originally, I thought strings were good enough to do this with, but if you, say, ever have a subtitle with a link in it, or a <b></b> in it, it wouldn't be possible, and it doesn't make sense to go to a PM and say "I can't make part of the subheader italicized"

src/react-components/src/components/03-organisms/Hero/Hero.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  if (backgroundImageSrc && image) {
+    throw new Error(`Please only either backgroundImageSrc or image into Hero, got both`);
+  }
+
+  if (heroType !== "PRIMARY" && locationDetails) {
+    throw new Error(`Please provide locationDetails only to PRIMARY heroTypes`);
+  }
+
+  let heroModifiers = backgroundImageSrc ? ["primary"] : ["secondary"];
```
hokei: is it possible to have a PRIMARY hero type with a "secondary" modifier?  That sounds ... confusing. If not, can we use the HeroType to decide the modifier?

src/react-components/src/components/03-organisms/Hero/Hero.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  let contentBoxStyling;
+  if (heroType === HeroTypes.Primary) {
+    if (foregroundColor && backgroundColor) {
+      contentBoxStyling = {
+        color: foregroundColor,
+        backgroundColor: backgroundColor,
+      };
+    } else if (foregroundColor || backgroundColor) {
+      throw new Error(`Please provide both foregroundColor and backgroundColor to Hero, got only one`);
```
hokei: nit: useful/cheap to say which one you got/is missing, could be helpful.

src/styles/01-atoms/links/link/_link.scss
```diff
@@ -1,72 +1,59 @@
 /// Link
-$color-link: $nypl-blue-regular;
-$color-link--hover: $nypl-blue-dark;
+$color-link: $nypl-red-regular;
```
hokei, line 2: is there guidance on how to override this?

src/react-components/src/components/03-organisms/Hero/Hero.tsx
```diff
@@ -0,0 +1,115 @@
...
+    }
+    else {
+      console.log("Foreground and background not provided");
```
hokei: does this need to be printed every time Hero is used?

hokei (changes requested):
happy to do a walkthrough on tests to better clarify!

src/styles/01-atoms/links/link/_link.scss
helenvholmes: Not currently. This code definitely could use some work but I needed the latest and greatest from Reno for some of the Hero styling. Where do you think that guidance would live? Perhaps in a comment blob at the top of this file?

hokei: some "Overriding Styles" document in Design System, I think.  The idea is that each product should have its style override sheets

src/react-components/src/components/03-organisms/Hero/Hero.tsx
EdwinGuzman: If there are more times, it might be helpful to have a util "Types.tsx" file to hold all the types. Just another option.

src/react-components/src/components/03-organisms/Hero/Hero.tsx
EdwinGuzman: That's true. On the larger scale of this, it should be a written down requirement that this _can_ be italicized. Otherwise, you'd be coding for possible scenarios that might not be implemented.

src/react-components/src/components/03-organisms/Hero/Hero.tsx
EdwinGuzman: `dangerouslySetInnerHTML` can also be thought of as elements that will get rendered but you don't know where they are coming from. For example, maybe an admin added HTML tags to a description they want to render and that info is coming from an API.

src/react-components/src/components/03-organisms/Hero/Hero.stories.tsx
EdwinGuzman: Passing in a string should throw an error since the `heroType` prop is defined as `HeroTypes`. I don't know of a clear way to let users know that it's supposed to be an enum, but maybe reading the component itself is enough?

src/styles/01-atoms/links/link/_link.scss
helenvholmes: I'll create a separate ticket and wiki page for that work, then. It'll be blocked on upcoming color work/`SectionName.tsx` work.

src/react-components/src/components/03-organisms/Hero/Hero.tsx
helenvholmes: @EdwinGuzman, in Reno where this Exhibition header gets used, that's exactly the use case! Sounds like we have a JSX.Element that Crystal's created that would be better to use here, though, so I will switch it.

src/react-components/src/components/03-organisms/Hero/Hero.tsx
helenvholmes: I don't mind it being in a separate file so long as it's grouped within a folder. (Plus I think it keeps `Hero.tsx` from getting too long.) I'm a little more wary of a `Types.tsx` file simply because over time I think it can become a lot of code that doesn't feel as logically grouped as it would with its component.

I'm happy to move the `enum` into `Hero` if you feel strongly about it, Crystal—just want to be sure we're doing it the same way everywhere 😄

src/react-components/src/components/03-organisms/Hero/Hero.tsx
helenvholmes: Good point. I wrote this before implementing the enum and didn't revisit it. This makes sense as cleanup.

src/styles/01-atoms/links/link/_link.scss
helenvholmes: I've opened up #107 to capture log this work. Logistically I'm thinking it will happen after any `Template` work upcoming for RN.

src/react-components/src/components/03-organisms/Hero/Hero.tsx
hokei: I don't feel strongly.  Agreed that we should do it the same way everywhere.  I think button types will need to be factored out soon so we can do it the same way.

src/react-components/src/components/03-organisms/Hero/Hero.stories.tsx
hokei: OK!  In that case all we need is to be exporting `HeroType` in `index.ts`.

src/react-components/src/components/03-organisms/Hero/Hero.stories.tsx
helenvholmes: The idea is that all Headings should have the blockname hero, not all hero components (components have a base class of hero). This would happen anytime we pull in a JSX.Element into another component in a story like this. Is this solved anywhere else @hokei that you know of? I'm operating off my understanding within Reno ([example](https://github.com/NYPL/nypl-d8/blob/development/web/themes/custom/nypl_emulsify/components/_patterns/02-molecules/basic-card/basic-card.twig#L14)) but it could be that when including JSX elements there's a better way to do this.

helenvholmes: Talked about this with @hokei over Slack. Handling this particular problem (how sub-JSX.Elements get the correct blockNames within larger components) is out of scope for this ticket, so I'm going to leave this as-is for now.

src/react-components/src/components/03-organisms/Hero/Hero.stories.tsx
helenvholmes: Exported `HeroTypes` in `index.ts`.

src/react-components/src/components/03-organisms/Hero/Hero.tsx
helenvholmes: Leaving it as-is.

src/react-components/src/components/03-organisms/Hero/Hero.tsx
helenvholmes: I've updated this to use `<BodyText>`.

src/react-components/src/components/03-organisms/Hero/Hero.tsx
helenvholmes: Done

src/react-components/src/components/03-organisms/Hero/Hero.tsx
helenvholmes: Removed

src/react-components/src/components/03-organisms/Hero/Hero.tsx
helenvholmes: Done

src/react-components/src/components/03-organisms/Hero/Hero.tsx
helenvholmes: While I think it would be nice to do some type-checking on colors (e.g., you pass in `#fffffff` by mistake, we could throw an error) I would rather not error on non-hex responses. `#fff`, `#ffffff`, and `white` are all valid CSS colors.

src/react-components/src/components/00-base/SectionName.tsx
hokei, line 4: This seems pretty specific - does it need to be its own component?  If it's supposed to only be used by the ResearchNow hero, can we wrap the ResearchNow Hero Story with  
`<div className={bem('main', ['research'])} ></div> `?

helenvholmes: It will be generic and used for all of the SectionNames stipulated here: https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Master?node-id=674%3A274

See former thread discussing why that work isn't being done yet: https://github.com/NYPL/nypl-design-system/pull/106/files#r401805076

src/styles/03-organisms/headers/_hero.scss
```diff
@@ -0,0 +1,125 @@
+$header-with-image-right--break: 1028px;
```
hokei: where is this being used?

src/react-components/src/__tests__/components/Hero-test.tsx
```diff
@@ -0,0 +1,93 @@
+
```
hokei: nit

hokei (commented):
looks good, just a reminder to add a changelog

src/react-components/src/components/00-base/SectionName.tsx
hokei: ah yes my memory is jogged now, sorry about that.

src/styles/03-organisms/headers/_hero.scss
helenvholmes: Oops, old code—removing

src/styles/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 This repo is in Prerelease. When it is released, this project will adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ========
+## [0.0.13] - 2020-04-09
```
hokei: Minor (? Since there's no release yet?) bump this and call out breaking changes.

hokei:
I'm so sorry:  
Screen Shot 2020-04-10 at 9 35 17 AM
(The image is above the bar) 

Screen Shot 2020-04-10 at 9 35 25 AM
(the image is not)

Maybe the design has changed?

src/react-components/src/components/03-organisms/Hero/Hero.stories.tsx
```diff
@@ -0,0 +1,86 @@
...
+  heading={<Heading
+    level={1}
+    id={"1"}
+    text={"Syncretic Vibrations: Exploring the Mosaic of Blackness through the Melville J. and Frances S.Herskovits Collection"}
+    blockName={"hero"}
+  />}
+  locationDetails={<Placeholder>Placeholder for locationDetails, which doesn't exist yet</Placeholder>}
+  foregroundColor="#ffffff"
+  backgroundColor="#0071ce" // have these default to section colors
+  backgroundImageSrc="https://images.nypl.org/index.php?id=419627&t=w"
```
hokei: This particular image looks like there's a margin-left and margin-right to it.  Subtle, but since these are "reference components" maybe replace it with a clearer image?

src/react-components/src/components/03-organisms/Hero/Hero.stories.tsx
```diff
@@ -0,0 +1,86 @@
...
+/>;
+
+export const ResearchNowHero = () => <SectionName><Hero
```
hokei: Not requesting changes, just some Friday brain-teasers:  
ResearchNow has been renamed to Digital Research Books.  How might we keep something like this up-to-date?

hokei:
at the same screen size: 

Screen Shot 2020-04-10 at 9 42 24 AM
Screen Shot 2020-04-10 at 9 42 29 AM

hokei (commented):
super close!  The JS all looks good, just some visual issues found from pulling the code

src/react-components/src/components/03-organisms/Hero/Hero.stories.tsx
EdwinGuzman: Try our best to not give any components specific names.

src/react-components/src/__tests__/components/Hero-test.tsx
```diff
@@ -0,0 +1,92 @@
+import { expect } from "chai";
+import { stub } from "sinon";
+import * as Enzyme from "enzyme";
+import * as React from "react";
+import * as Mocha from "mocha";
+
+import Image from "../../components/01-atoms/Images/Image/Image";
+import Heading from "../../components/01-atoms/Text/Headings/Heading";
+import BodyText from "../../components/01-atoms/Text/Text/BodyText";
```
EdwinGuzman: You're not using `BodyText` in this file.

src/react-components/src/__tests__/components/Hero-test.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { HeroTypes } from "../../components/03-organisms/Hero/HeroTypes";
+import Hero from "../../components/03-organisms/Hero/Hero";
+
+describe("Hero Test", () => {
+
+  let blurCallback;
+  let changeCallback;
+  let submitCallback;
```
EdwinGuzman: These aren't being used.

src/react-components/src/__tests__/components/Hero-test.tsx
```diff
@@ -0,0 +1,92 @@
...
+      heroType={HeroTypes.Primary}
+      heading={<Heading
+        level={1}
+        id={"1"}
+        text={"Hero Primary"}
+        blockName={"hero"}
+      />}
+      backgroundImageSrc="https://placeimg.com/1600/800/arch">
+    ></Hero>);
+    expect(wrapper.find(".hero")).to.have.lengthOf(1);
```
EdwinGuzman: I would test that you can find an image in this DOM since the description says it has a background image. I'm guessing this is hard to test because a background image is css?

src/react-components/src/components/03-organisms/Hero/Hero.stories.tsx
hokei: This is in stories, though ... I had been thinking one way Design System can move forward is to build out the "story" version of a component and pass it to the team that's supposed to consume it.  This is also the best solution I can think of to do regression testing (eg: if a component needs to change, how do we make sure it doesn't drastically affect existing users of the component?) But when the product changes, there's not really a good way to signal that back to Design System so that they can update the story ...

EdwinGuzman: Yeah, I noticed that after... I guess the person writing the stories should just keep the names up-to-date, and also remove any that are no longer used.

src/react-components/src/components/03-organisms/Hero/Hero.tsx
```diff
@@ -0,0 +1,121 @@
+// OH-30 Header with Image Right
+import * as React from "react";
+import bem from "../../../utils/bem";
+import Image from "../../01-atoms/Images/Image/Image";
+import Heading from "../../01-atoms/Text/Headings/Heading";
```
EdwinGuzman: These components aren't being used in this file.

src/react-components/src/components/03-organisms/Hero/Hero.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  if (backgroundImageSrc && image) {
+    throw new Error(`Please only either backgroundImageSrc or image into Hero, got both`);
+  }
+
+  if (heroType !== HeroTypes.Primary && locationDetails) {
+    throw new Error(`Please provide locationDetails only to PRIMARY heroTypes`);
+  }
+
+  let backgroundImage = backgroundImageSrc ? "data-responsive-background-image" : "";
```
EdwinGuzman: This variable isn't being used in this file.

src/react-components/src/components/03-organisms/Hero/Hero.tsx
```diff
@@ -0,0 +1,121 @@
...
+    image,
+  } = props;
+
+  let heroModifiers;
+
+  if (heroType === HeroTypes.Primary) {
+    heroModifiers = ["primary"];
+  } else if (heroType === HeroTypes.Secondary) {
+    heroModifiers = ["secondary"];
+  }
```
EdwinGuzman, line 66: Should there be a fallback in case PRIMARY or SECONDARY aren't passed?

src/react-components/src/components/03-organisms/Hero/Hero.tsx
```diff
@@ -0,0 +1,121 @@
...
+        color: foregroundColor,
+        backgroundColor: backgroundColor,
+      };
+    } else if (foregroundColor || backgroundColor) {
+      let receivedColor = foregroundColor ? "foregroundColor" : "backgroundColor";
+      throw new Error(`Please provide both foregroundColor and backgroundColor to Hero, only got ` + receivedColor);
+    }
+    else {
+      contentBoxStyling = {};
+    }
```
EdwinGuzman: Not necessary but you can make line 86 have `= {}` as the default. That way you can remove these lines from 97-99.

src/react-components/src/components/03-organisms/Hero/Hero.tsx
```diff
@@ -0,0 +1,121 @@
...
+  } else {
+    if (foregroundColor || backgroundColor) {
+      throw new Error(`Received foregroundColor and/or backgroundColor, but these are only accepted on HeroTypes.Primary`);
+    }
+  }
+
+  return (
+    <div className={bem(heroBaseClass, heroModifiers)} data-responsive-background-image style={backgroundImageStyle}>
```
EdwinGuzman, line 100: I like how you did this responsive styling through the use of `data-*`.

EdwinGuzman:
Update looks good. There are a couple of minor code changes you can do.

In the Hero tests, you caught 2 errors that can be thrown, but there are a total of 5 potential errors. It'd be good to test those cases.

src/react-components/src/components/03-organisms/Hero/Hero.stories.tsx
helenvholmes: Name changes to products don't happen very often. When they do, we should create issues to track down any stories with an old product name and get them updated. I'll do that here.

src/react-components/src/components/03-organisms/Hero/Hero.tsx
helenvholmes: Can't take credit for this, this is just how 4K did it originally :)

src/react-components/src/components/03-organisms/Hero/Hero.tsx
helenvholmes: I've instead made `heroType` required, so it'll now fail if a value from that enum isn't passed (it being optional was a holdover from when I was reworking this code).

helenvholmes:
> I'm so sorry:
> Screen Shot 2020-04-10 at 9 35 17 AM
> (The image is above the bar)
> 
> Screen Shot 2020-04-10 at 9 35 25 AM
> (the image is not)
> 
> Maybe the design has changed?

Have a note to talk about it in [UX's Housekeeping meeting tomorrow](https://docs.google.com/document/d/16C5ZCAkTNt_n75FzDqvueaVjlVgDh9ye6C7O62z_yLo/edit#) to get the designs updated to what's built. The former designs just proved too challenging to build.

> at the same screen size:
> 
> Screen Shot 2020-04-10 at 9 42 24 AM
> Screen Shot 2020-04-10 at 9 42 29 AM

I believe this is working as intended (see [prod here](https://www.nypl.org/books-more/recommendations/staff-picks/adults)) but I think that the true determinate of this is Kathy, who can't do a VQA on this until this PR gets merged or I run a Zoom call with her. I think this particular VQA nit isn't worth blocking on.

src/react-components/src/__tests__/components/Hero-test.tsx
helenvholmes: Yeah, I was struggling with it and left it to solve later... and obviously forgot about it. Figured it out and updated it

EdwinGuzman (approved):
The code looks good. I didn't look too closely at the designs and metronome because I don't feel too comfortable commenting on those updates.

helenvholmes:
@EdwinGuzman long term I think it'd be best to get a multidev environment set up using Heroku or something so Kathy, Ellen, or Marcus can weigh in—I understand the reticence!

helenvholmes:
Marking as `dependent` on @hokei's Filters PR.

src/react-components/CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 Currently, this repo is in Prerelease.  When it is released, this project will adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
+========
+## [0.2.0] - 2020-04-13
```
hokei, line 7: Will also need a package.json update as well

hokei (approved):

helenvholmes:
@hokei Can we look into shipping this next week if I do the integration, or would you prefer we continue prioritizing Template work?

hokei:
Let's ship this first!

helenvholmes:
Merged, although looks like Github being down caused some weirdness where it's not getting recognized. Closing this out.

Issue #109, opened 2020-04-06
CSS Documentation


helenvholmes:
We need documentation to for how we write CSS that links to the BEM documentation.

Should also illustrate how our `bem` utility function works.

```
.hero {
  &__heading {
    &--icon-left {
    }
  }
}
```

See [this thread](https://nypl.slack.com/archives/DM9SJ8Q5Q/p1586196755028100).

Put this in Wiki

helenvholmes:
Closing this out—I believe that in conjunction with #193 and the component flags that reference BEM this is unnecessary (we are using BEM as it's documented on [getbem.com](http://getbem.com/introduction/)).

PR #111, opened 2020-04-10
SFR-625 filters

hokei:
[ResearchNow Jira Ticket](http://jira.nypl.org/browse/SFR-625)

### Breaking Changes
- Large classes now have `Opts` inteface that allows for the component to be built over multiple steps. This affects: 
    - `HelperErrorText`
    - `Label`
    - `TextField`
    - `Button`
- Deprecated `SearchResultsList`, `IconLinkList` and `EditionsList` in favor of `UnorderedList`

### Added: 
- `Checkbox`
- `DateRangeForm` 
- `Input`
- `Accordion` 
- `UnorderedList`

### Changed: 
- `Button`: now has `link` type, allows passing of `buttonType`

## Front End Review:
- [ ] Check against the sub-pages of http://themetronome.co/components/mn-20-filter-sets-desktop/?fresh=true and http://themetronome.co/components/mn-265-filter-sets---mobile-/?fresh=true


hokei:
@scooooooooby Do you know how to do checkbox styling accessibly?  It seems like you can't directly style native html `<input type="checkbox">` elements (is that true or am I looking at old websites), so everywhere I'm seeing, they hide the checkbox and make a new span that looks like one, which seems like a real red flag to me.   Thoughts?

helenvholmes:
@hokei I've been communicating with the team a lot about how challenging it is to style native form elements, and it's been a while since I've had to. [This article](https://css-tricks.com/custom-styling-form-inputs-with-modern-css-features/) looks promising, but I think it would be fine to ship with native styling on these forms as a phase I.

(Also, there's a WIP label on this but my review has been requested—should I hold off for now?)

hokei:
@EdwinGuzman I wanted to ask about this change in approach:  https://github.com/NYPL/nypl-design-system/pull/111/files#diff-bff9e2b12be0b1b5575ad32c2f2b01d2R5 which allows us to do something like https://github.com/NYPL/nypl-design-system/pull/111/files#diff-bdf8ce26f21061d5e77f8cd580e8cc7eR23 - is this really more sane, or am I just rearranging deck chairs on the titanic? The main drawback is that we loose the great proptype checking in the storybook tool - do you know if it's possible to have this type be more useful?  

Screen Shot 2020-04-10 at 9 22 14 AM

hokei:
@scooooooooby Yes please start looking at it! the only thing that isn't done fleshing out the test stubs.

src/react-components/CHANGELOG.md
```diff
@@ -4,6 +4,26 @@ All notable changes to this project will be documented in this file.
...
 Currently, this repo is in Prerelease.  When it is released, this project will adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
+## [0.1.0]
+### Breaking Changes
+- Large classes now have `Opts` inteface that allows for the component to be built over multiple steps. This affects: 
+    - `HelperErrorText`
+    - `Label`
+    - `TextField`
+    - `Button`
+- Deprecated `SearchResultsList`, `IconLinkList` and `EditionsList` in favor of `UnorderedList`
```
helenvholmes, line 14: Nice 😎👏🏻

src/react-components/CHANGELOG.md
```diff
@@ -4,6 +4,26 @@ All notable changes to this project will be documented in this file.
...
+
+### Added: 
+- `Checkbox`
+- `DateRangeForm` 
+- `Input`
+- `Accordion` 
+- `UnorderedList`
```
helenvholmes, line 25: 😱 hot dawg, a lot of work in here

src/react-components/src/__tests__/components/Accordion-test.tsx
```diff
@@ -0,0 +1,47 @@
+
```
helenvholmes: nit

src/react-components/src/__tests__/components/DateRangeForm-test.tsx
```diff
@@ -0,0 +1,39 @@
+
```
helenvholmes: another nit, I think we're both copypastaing from some test file that has a newline at the top of it

hokei: 😂

src/react-components/src/components/01-atoms/Button/Button.stories.tsx
```diff
@@ -42,4 +42,38 @@ export const buttonIconChildrenElement = () =>
...
+    callback={action["clicked"]}
+    id="button"
+    type="link"
+    iconPosition="left"
+    iconName="arrow-xsmall"
+    iconModifiers={["left"]}
+    iconDecorative={true}
+  >
+    <span>button content</span>
+  </Button >;
```
helenvholmes, line 102: Should we add a story for a `submit` button type, just to show how that works? The button used in `DateRangeForm` should be one, no?

src/react-components/src/__tests__/components/DateRangeForm-test.tsx
```diff
@@ -0,0 +1,39 @@
...
+  let container;
+  let callback;
+  before(() => {
+    callback = stub();
+    const container = Enzyme.mount(<DateRangeForm
+      formLabel={<>Publication Year</>}
+
+      fromLabelOpts={{ labelContent: <>From</>, id: "FromLabel" }}
+      fromInputOpts={{ inputId: "fromInput" }}
+      fromHelper={{ content: <>EX. 1901</>, id: "fromyearhelper", isError: false }}
```
helenvholmes, line 19: I was debating whether or not to even leave this comment, but here goes anyway: the designs _do_ have this string exactly as you've mocked it up here, but I've never seen "example" abbreviated this way. I'd love it if in our stories at least if we could mock it up as, "E.x., 1901"

src/react-components/src/__tests__/components/DateRangeForm-test.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+      toLabelOpts={{ labelContent: <>To</>, id: "ToLabel" }}
+      toInputOpts={{ inputId: "toInput" }}
+      toHelper={{ content: <>EX. 2000</>, id: "toYearHelper", isError: false }}
+
+      showError={false}
+      error={{ content: <></>, id: "errorId", isError: true }}
+
+      buttonProps={{ id: "submitButtonId", callback: callback, content: <>Submit</> }}
```
helenvholmes: How do we make buttons `type=submit` now that we're ripped out that assumption? Does that correspond with `type` in the button's prop blob, or is it no longer possible to set a button to be a `submit` button?

src/react-components/src/__tests__/components/HelperErrorText-test.tsx
```diff
@@ -4,12 +4,12 @@ import * as Enzyme from "enzyme";
...
 
-import HelperErrorText from "../../components/02-molecules/Helpers/HelperErrorText";
+import HelperErrorText from "../../components/01-atoms/Forms/HelperErrorText";
 
 describe("HelperErrorText Test", () => {
-  it("Throws error when HElperErrorText is called with no content", () => {
-    expect(() => Enzyme.mount(<HelperErrorText id="helperTextWithLink"
-    isError={false}/>))
-    .to.throw("Helper and Error Text must have content, or else it should not be used");
+  it("Renders HelperErrorText", () => {
```
helenvholmes, line 9: Since error and helper text are the same thing, tests for the in-error state/not in-error state/in both state?? is that possible/ would be good to see stubbed out here.

src/react-components/src/__tests__/components/Input-test.tsx
```diff
@@ -0,0 +1,27 @@
+
```
helenvholmes: nit

src/react-components/src/__tests__/components/Label-test.tsx
```diff
@@ -14,7 +14,7 @@ describe("Label Test", () => {
   });
 
   it("Renders a <label> when passed text as prop", () => {
-    wrapper = Enzyme.shallow(<Label id="label" referenceId="so-lonely" labelText="cupcakes" />);
+  wrapper = Enzyme.shallow(<Label id="label" referenceId="so-lonely" labelContent={<>cupcakes</>} />);
```
helenvholmes: 😭😂

src/react-components/src/components/00-base/Container/Modal.tsx
```diff
@@ -0,0 +1,31 @@
+/* Full-screen modal that appears on top of the body */
+import * as React from "react";
+import bem from "../../../utils/bem";
+
+export interface ModalProps {
+  /** Used by BEM CSS */
```
helenvholmes: Since these comments are probably going to be repeated across every component ever, what do you think of these descriptors:

```
/** Optional modifiers array for use with BEM. See how to work with modifiers and BEM here: http://getbem.com/introduction/ */
modifiers?: string[];

/** Optional blockName for use with BEM. See how to work with blockNames and BEM here: http://getbem.com/introduction/ */
modifiers?: string[];
```

I actually don't love that but it's a start... ideally #109 would exist and we would point them there

src/react-components/src/components/03-organisms/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,50 @@
+
```
helenvholmes: Newline

I've conceptually been grouping Accordions, Modals, and Tab Sets together in something I've called "Content Displays". The idea is that they all hide/show content. We're not really using the atom/molecule/organism structure within Storybook; might be worth discussing if/when/how we might deprecate that model of organization in favor of something else.

[Figma docs for content displays](https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Master?node-id=10734%3A2520)

Subpages for other design groupings:
Screen Shot 2020-04-10 at 10 16 37 AM

hokei: The arrangement of stuff within folders is kind of messy and will touch everything, and cause breaking changes.  I'm thinking this is a good stand-alone task and we should probably do it all at once.

src/react-components/src/components/00-base/Container/Modal.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  componentWillUnmount() {
+    document.body.classList.remove("no-scroll");
+  }
+
+  render() {
+    let baseClass = "modal";
+
+    return (
+      <div tabIndex={0} className={bem(baseClass)}>
```
helenvholmes: Should have `modifiers` and `blockName` in the bem function.

src/react-components/src/components/01-atoms/Button/Button.stories.tsx
```diff
@@ -42,4 +42,38 @@ export const buttonIconChildrenElement = () =>
   >
     <span>I'm wrapped in a span!</span>
   </Button>
-;
\ No newline at end of file
+  ;
```
helenvholmes: I think there's something funky happening with the formatting here

src/react-components/src/components/01-atoms/Button/Button.tsx
```diff
@@ -4,10 +4,37 @@ import bem from "../../../utils/bem";
...
+  /** Used for BEM css convention. */
+  blockName?: string;
+  large?: boolean;
+  type?: string;
+  /** The html button attribute */
+  buttonType?: string;
+  mouseDown?: boolean;
+  /** If an icon is to be rendered, an `iconPosition` prop is required. */
+  iconPosition?: string;
+  iconName?: string;
```
helenvholmes, line 19: I would like to get annotations for all of these as we work on them since I think down the road the larger UX team (and other developers) will be able to click the "More Info" link and see all the available options on a component (particularly useful for IA/Ellen). Even if it feels redundant to us, a description on each of these props will be helpful to others.

src/react-components/src/components/01-atoms/Button/Button.tsx
```diff
@@ -37,12 +66,12 @@ export default class Button extends React.Component<ButtonProps, {}> {
   }
 
   render(): JSX.Element {
-    const { id, callback, content, attributes, modifiers, blockName, type, mouseDown,
+    const { id, callback, content, attributes, modifiers, blockName, type, buttonType = "submit", mouseDown,
```
helenvholmes: I thought we were no longer having `submit` as an assumption on the buttons?

src/react-components/src/components/01-atoms/Forms/Checkbox.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  isSelected?: boolean;
+  onChange?: (event: React.ChangeEvent<HTMLInputElement>) => void;
+}
+
+interface CheckboxProps {
+  /** name of the checkbox */
+  name?: string;
+
+
```
helenvholmes: nit; extra line

src/react-components/src/components/01-atoms/Forms/Checkbox.tsx
```diff
@@ -0,0 +1,61 @@
...
+  /** Unique ID used by checkbox */
+  checkboxId: string;
+  /* The Label that the checkbox is using. */
+  labelOptions: LabelOptions;
+
+  /* The current selected state of the checkbox */
+  isSelected?: boolean;
+  /** The action to perform on the <input>'s onChange function  */
+  onChange: (event: React.ChangeEvent<HTMLInputElement>) => void;
```
helenvholmes, line 22: I think this code isn't working as intended—in Storybook, I've never able to apply the checked state to the checkbox

src/styles/02-molecules/date-range/date-range.scss
```diff
@@ -0,0 +1,3 @@
+.date-range {
+  @include fieldset-reset;
```
helenvholmes, line 2: The spacing in this component is a little tight. I'd maybe add something like:

```
  &__input-description,
  &__button {
    @include space-stack-s;
  }
```

src/react-components/src/components/01-atoms/Forms/DateRangeForm.tsx
```diff
@@ -0,0 +1,88 @@
...
+          inputValue={toInputOpts.inputValue}
+          onInputChange={toInputOpts.onInputChange}
+          helperContentOpts={toHelper}
+        />
+      </div>
+      {showError &&
+        <HelperErrorText isError={true} id={error.id}>{error.content}</HelperErrorText>
+      }
+      {buttonProps &&
+        <Button id={buttonProps.id} callback={buttonProps.callback}>{buttonProps.content}</Button>
```
scooooooooby, line 84: You should also pass in a button `blockName` and `modifiers` if you want to be able to style this (and control the button's visual state based on the larger form state). For example, I'd expect the button to be disabled, potentially, if the error string was showing.

src/react-components/src/components/01-atoms/Forms/DateRangeForm.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+  fromHelper={{ content: <>EX. 1901</>, id: "fromyearhelper" }}
+
+  toLabelOpts={{ labelContent: <>To</>, id: "ToLabel" }}
+  toInputOpts={{ inputId: "toInput" }}
+  toHelper={{ content: <>EX. 2000</>, id: "toYearHelper" }}
+
+  showError={true}
+  error={{ content: <div>The end year should be the same year as or later than the start year</div>, id: "errorId" }}
+
+  buttonProps={{ id: "submitButtonId", callback: callback, content: <>Submit</> }}
```
scooooooooby, line 43: You should also pass in a button `blockName` and `modifiers` if you want to be able to style this (and control the button's visual state based on the larger form state). For example, I'd expect the button to be disabled, potentially, if the error string was showing.

src/react-components/src/components/01-atoms/Forms/DateRangeForm.tsx
helenvholmes: You should also pass in a button `blockName` and `modifiers` if you want to be able to style this (and control the button's visual state based on the larger form state). For example, I'd expect the button to be disabled, potentially, if the error string was showing.

src/react-components/src/__tests__/components/HelperErrorText-test.tsx
hokei: Added something for Errror state, but HelperErrorText can be helper or error, but not both.  If you have both, you want to use two HelperErrorText objects ... wonder if there's a way to make this clear?

src/react-components/src/components/01-atoms/Button/Button.stories.tsx
hokei: In stories, it shouldn't be different.  However, I will add buttonType to the stories to show that they can be passed.

src/react-components/src/__tests__/components/DateRangeForm-test.tsx
hokei: yeahhh I'm gonna use this opportunity to reverse these prop names since button's going through breaking changes.

src/react-components/src/components/00-base/Container/Modal.tsx
hokei: It might be really long to show up in the PropTypes thing in storybook ... Maybe something that's good enough get people googling in the right direction?  
Maybe "Modifiers/Blockname for use with BEM" ? (Optional is already inferred from the type)

src/react-components/src/components/01-atoms/Button/Button.tsx
hokei: this assigns it if it's empty, since it's optional.  We could make buttonType required instead, though.

EdwinGuzman:
> @EdwinGuzman I wanted to ask about this change in approach: https://github.com/NYPL/nypl-design-system/pull/111/files#diff-bff9e2b12be0b1b5575ad32c2f2b01d2R5 which allows us to do something like https://github.com/NYPL/nypl-design-system/pull/111/files#diff-bdf8ce26f21061d5e77f8cd580e8cc7eR23 - is this really more sane, or am I just rearranging deck chairs on the titanic? The main drawback is that we loose the great proptype checking in the storybook tool - do you know if it's possible to have this type be more useful?
> 

That's interesting... There should be something that will you to add an interface as the propType. I can see how that can be annoying and is also incorrect. I'm not sure I have encountered this before to update it but I'll double check.

src/react-components/src/components/01-atoms/Button/Button.tsx
hokei: yeah definitely.  There's a lot of stuff that doesn't have them but new stuff/stuff that's being updated should have them for sure.

src/react-components/src/__tests__/components/HelperErrorText-test.tsx
helenvholmes: I actually don't think you should have both, you're either errored or you're showing helper text. We can illustrate that with a test here.

src/react-components/src/components/00-base/Container/Modal.tsx
helenvholmes: The PropTypes thing shows up as a handy little table—I think it's okay for them to be wordy. We can leave them short if you're prefer though. (And as for the optional thing omg I wrote optional all over the Hero descriptors, I just am a terrible technical writer bleh)

src/styles/03-organisms/list/_list.scss
```diff
@@ -0,0 +1,10 @@
+.unordered-list {
```
helenvholmes: I'd move this to `atoms/text/_text.scss`, unordered lists are just inline elements, feels weird to have 'em all the way up in `organisms/`

src/styles/03-organisms/list/_list.scss
```diff
@@ -0,0 +1,10 @@
...
+  @include list-reset;
+
+  &__list-container {
```
helenvholmes: Looking at the React file this list-container is outside of `.unordered-list`, so these styles aren't ever being applied.

src/styles/02-molecules/input/input.scss
```diff
@@ -0,0 +1,26 @@
+.input {
+  
+  &__label {
+    @include space-inline-xs;
+    display: inline-block;
```
helenvholmes, line 5: Stylistic and we should have a formatter for it, but please add a newline after `@include`s.

src/react-components/src/components/01-atoms/Button/Button.tsx
helenvholmes: I think it would be safe to make the assumption that if `buttonType` wasn't passed, it should default to primary.

src/react-components/src/components/03-organisms/EditionsList/EditionsList.tsx
```diff
@@ -1,4 +1,6 @@
 // OL-24 Editions List
+/** DEPRECATION WARNING, please build with <UnorderedList> component */
```
helenvholmes: Are these deprecation warnings just to make the SFR integration easier? (This is just ignorance on my part. I assume there's a myriad of ways to communicate upcoming deprecations to users of your library?)

src/styles/00-base/containers/modal.scss
```diff
@@ -0,0 +1,18 @@
+
```
helenvholmes: Newline

src/react-components/src/components/03-organisms/EditionsList/EditionsList.tsx
hokei: In the future we'd announce deprecations in patch releases and remove them in major releases.  This is going to be a "major" release so it's mostly for me to remember to delete these files.

src/react-components/src/components/03-organisms/List/UnorderedList.stories.tsx
```diff
@@ -0,0 +1,105 @@
...
+      editionInfo={searchItemData.editionInfo}
+      editionsLinkElement={searchItemData.editionsLinkElement} />
+  ];
+
+export const searchResultsList = () => <UnorderedList id="search-results-list">{searchResults}</UnorderedList>;
+
+const text = <span>Hello</span>;
+let textList: JSX.Element[] = [];
+textList.fill(text, 0, 100);
+export const textListWithScroll = () => <UnorderedList id="text-list-scroll" modifiers={["scroll"]}>{textList}</UnorderedList>;
```
helenvholmes, line 112: Looks like this story might be broken? I'm not seeing anything here locally.

src/react-components/src/components/01-atoms/Forms/Input.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  /** Whether the input is currently in an error state */
+  showError?: boolean;
+}
+
+export default function Input(props: InputProps) {
+
+  const {
+    baseClass = "input",
+    blockName = "form-item",
```
helenvholmes, line 64: I think I blame forms for why we have been historically so confused with BEM

helenvholmes: Idk maybe it's just me

src/styles/02-molecules/input/input.scss
```diff
@@ -0,0 +1,26 @@
+.input {
+  
+  &__label {
+    @include space-inline-xs;
```
helenvholmes, line 3: Label is currently a wrapper around the actual `<label>` element and the required message, so this `space-inline-xs` isn't getting applied where you want it to:

Screen Shot 2020-04-10 at 4 41 29 PM

src/react-components/src/components/01-atoms/Forms/Input.stories.tsx
```diff
@@ -0,0 +1,34 @@
...
+  helperContentOpts={{id: "blah", content: <>This is some helpful text</>}}
+/>;
+export const inputRequiredWithError = () => <Input labelId="blah"
+  labelContent={<>Label</>}
+  inputId="blah"
+  isRequired={true}
+  errorContentOpts={{id: "blah", content: <>Error Message Lorem Ipsum</>}}
+  showError={true}></Input>;
+
+export const inputRequiredWithHelperAndError = () => <Input labelId="blah"
```
helenvholmes, line 28: I think we should delete this story... in fact I don't think you should be able to show helper AND error text at the same time, I don't think it makes sense. That might be an error we want to throw unless you think I'm thinking about this wrong

helenvholmes (commented):
I notice there's no story for Modal—was that intentional? I think one would be a good idea, even if it's just like... a button that opens a gray box and closes it. I think the idea is to illustrate the pattern

I'm a little iffy on the DateRangePicker—I'd love to hear your thoughts on it @hokei. Does this feel like a RN-specific component to you?

Depending on your answer for this, I might consider adding a `value` value so clicking on the "^V" within the form gives you something like "1901" instead of "-1" or "1".

Also looks like `_icons.scss` got deleted—why?

I think we should get in `_accordion.scss` file somewhere in here, as it looks like accordion buttons [fill up all available horizontal space](https://www.figma.com/file/CCCdFt8H7pMqHjkRQlivSw/IA-Templates-Components?node-id=17669%3A8495) (as opposed to outline/secondary buttons, which just have left/right padding). (There are some other stylistic things that are missing from this as well, like the outline extending to the checkboxes below, for example.)

src/react-components/CHANGELOG.md
```diff
@@ -4,6 +4,27 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 Currently, this repo is in Prerelease.  When it is released, this project will adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
+## [0.1.0]
+### Breaking Changes
```
EdwinGuzman, line 8: It's great that you are deprecating multiple components in favor of one component, but if it's a breaking change I think it's safe to completely remove those files. Also, this version bump isn't in `package.json`.

src/react-components/src/components/01-atoms/Forms/Checkbox.tsx
```diff
@@ -0,0 +1,60 @@
+/**
+ * A Form Checkbox component that can be selected and deselected.
+ */
```
EdwinGuzman: It might be better to have these comments right before the component itself rather than at the top of the file. Not a big deal but we should have a convention, especially when this very closely relates to the component.

src/react-components/src/components/01-atoms/Forms/Checkbox.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+import * as React from "react";
+import bem from "../../../utils/bem";
+import Label, { LabelOptions } from "./Label";
+
+export interface CheckboxOptions {
```
EdwinGuzman: You're not using this anywhere.

src/react-components/src/components/01-atoms/Forms/Checkbox.tsx
```diff
@@ -0,0 +1,60 @@
...
+  /* The Label that the checkbox is using. */
+  labelOptions: LabelOptions;
+
+  /* The current selected state of the checkbox */
+  isSelected?: boolean;
+  /** The action to perform on the <input>'s onChange function  */
+  onChange: (event: React.ChangeEvent<HTMLInputElement>) => void;
+}
+
+export default function Checkbox(props: CheckboxProps) {
```
EdwinGuzman, line 29: You don't have tests for this component.

src/react-components/src/components/01-atoms/Button/Button.tsx
```diff
@@ -100,9 +125,9 @@ export default class Button extends React.Component<ButtonProps, {}> {
     }
 
     let btnProps = {
-      id: id,
+      id: "btn-" + id,
       className: bem(button_base_class, buttonModifiers, blockName),
-      type: "submit"
+      type: type
```
EdwinGuzman: Tests for this would be good.

src/react-components/src/__tests__/components/Label-test.tsx
```diff
@@ -24,7 +24,7 @@ describe("Label Test", () => {
   });
 
   it("if both <label> children and labelText prop are passed, use labelText", () => {
-    wrapper = Enzyme.shallow(<Label id="label" referenceId="so-lonely" labelText="cupcakes" >muffins</Label>);
+    wrapper = Enzyme.shallow(<Label id="label" referenceId="so-lonely" labelContent={<>cupcakes</>} >muffins</Label>);
```
EdwinGuzman: A story for this type would be good.

src/react-components/src/components/01-atoms/Forms/Label.tsx
```diff
@@ -2,19 +2,29 @@
...
+  labelContent?: JSX.Element;
+}
+
+interface Props {
+  id: string;
+
+  /** The id of the html element that this Label is describing */
   referenceId: string;
-  labelText?: string;
+  /** The contents of the label.  An alternative to passing children elements */
```
EdwinGuzman: 👍

src/react-components/src/index.ts
```diff
@@ -8,9 +9,15 @@ export { default as Breadcrumbs } from "./components/02-molecules/Menus/Breadcru
...
 export { default as Dropdown } from "./components/02-molecules/Dropdown/Dropdown";
 export { default as EditionCard } from "./components/02-molecules/Cards/EditionCard";
-export { default as IconLinkList } from "./components/03-organisms/IconLinkList/IconLinkList";
+// export { default as IconLinkList } from "./components/03-organisms/IconLinkList/IconLinkList";
 export { default as HeaderImgRight } from "./components/03-organisms/Headers/HeaderImgRight";
 export { default as SearchPromo } from "./components/03-organisms/SearchPromo/SearchPromo";
 export { default as HeaderWithSearch } from "./components/03-organisms/Headers/RNHeaderWithSearch";
-export { default as SearchResultsList } from "./components/03-organisms/SearchResultsList/SearchResultsList";
-export { default as EditionsList } from "./components/03-organisms/EditionsList/EditionsList";
+// export { default as SearchResultsList } from "./components/03-organisms/SearchResultsList/SearchResultsList";
```
EdwinGuzman: So your plan is to eventually remove these but not in this PR? Why?

src/react-components/src/components/03-organisms/Accordion/Accordion.stories.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+
+export default {
+  title: "Accordion",
+  component: Accordion
+};
+const checkboxes = [
+  <Checkbox onChange={action("changed")} checkboxId="checkbox-1" labelOptions={{ id: "checkbox-1-label", labelContent: <>Checkbox 1</> }}></Checkbox>,
+  <Checkbox onChange={action("changed")} checkboxId="checkbox-2" labelOptions={{ id: "checkbox-1-label", labelContent: <>Checkbox 2</> }}></Checkbox>,
+  <Checkbox onChange={action("changed")} checkboxId="checkbox-3" labelOptions={{ id: "checkbox-1-label", labelContent: <>Checkbox 3</> }}></Checkbox>
```
EdwinGuzman, line 17: This is bad because it's showing the same ID as input and it creates an a11y issue:
Screen Shot 2020-04-12 at 5 11 03 PM

src/react-components/src/components/03-organisms/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,50 @@
...
+    buttonOptions.type = "button";
+    buttonOptions.modifiers = ["large"];
+    buttonOptions.iconPosition = "right";
+    buttonOptions.iconName = this.state.isOpen ? "minus" : "plus";
+    buttonOptions.buttonType = ButtonTypes.Outline;
+
+    return (
+      <div className={bem("accordion", modifiers, blockName)}>
+        <Button callback={this.toggleContentShow} {...buttonOptions}></Button>
+        {this.state.isOpen && <div className="accordion-content">
```
EdwinGuzman, line 44: Is this extra div wrapper necessary?

src/react-components/src/__tests__/components/Accordion-test.tsx
```diff
@@ -0,0 +1,40 @@
...
+import Accordion from "../../components/03-organisms/Accordion/Accordion";
+
+describe("Renders Input (closed state)", () => {
+  let container: Enzyme.ReactWrapper<any, Readonly<{}>, React.Component<{}, {}, any>>;
+  before(() => {
+    container = Enzyme.mount(<Accordion
+      id="accordion"
+      buttonOptions={{ id: "accordionBtn", content: <>Click to expand</> }}
+    > <div className="accordion-content">content content</div> </Accordion>);
```
EdwinGuzman, line 13: The `Accordion` component already renders a div with `accordion-content` css class. Now this is rendering twice for these tests. For the "open state" test, this fails `expect(container.find(".accordion-content").length).to.equal(1);`.

src/react-components/src/components/01-atoms/Forms/DateRangeForm.tsx
```diff
@@ -0,0 +1,91 @@
...
+    fromHelper,
+    toLabelOpts,
+    toInputOpts,
+    toHelper,
+    showError,
+    error,
+    buttonProps } = props;
+    if (showError && !error) {
+      throw new Error("Error should be shown, but no Error Content exists");
+    }
```
EdwinGuzman: Nit, the `} = props;` should be on a new line and lines 55-57 are indented by 1 extra tab.

src/react-components/src/components/01-atoms/Forms/DateRangeForm.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+  return (
+    <fieldset className={bem(blockName, [], baseClass)}>
+      <legend>{formLabel}</legend>
+      <div className="input-row">
+        <Input
```
EdwinGuzman, line 64: I think this is fine for a first pass, but you should have some validation for the input fields (like number type and max 4 digits) so users enter reasonable years.
Screen Shot 2020-04-12 at 5 22 53 PM

src/react-components/src/__tests__/components/HelperErrorText-test.tsx
```diff
@@ -1,15 +1,21 @@
...
+    const container = Enzyme.mount(<HelperErrorText id="helperTextWithLink"
+    isError={false}>Text</HelperErrorText>);
+    expect(container.exists("#helperTextWithLink")).to.equal(true);
+  });
+
+  it("Has 'error' modifier if error is passed", () => {
+    const container = Enzyme.mount(<HelperErrorText id="helperTextWithLink"
+    isError={true}>Text</HelperErrorText>);
+    console.log("debug", container.debug());
```
EdwinGuzman: debug statement.

EdwinGuzman: Actually, this test isn't done yet, right? It's not really testing anything that the above test isn't.

src/react-components/src/components/01-atoms/Forms/Input.stories.tsx
```diff
@@ -0,0 +1,34 @@
+import * as React from "react";
+
+import Input from "./Input";
+
+export default {
+  title: "Input",
+  component: Input,
+};
+
+export const input = () => <Input labelId="blah"
```
EdwinGuzman, line 10: These examples don't work because you're adding the same `ID` for both the label and input element, so now there's the a11y issue of id redundancy.

src/react-components/src/components/01-atoms/Forms/Input.tsx
```diff
@@ -0,0 +1,102 @@
...
+        {isRequired && <span className={bem("required-message", [], baseClass)}>
+          Required
+        </span>}
+      </div>
+
+      <TextField id={inputId} {...textFieldOptions}></TextField>
+      {helperContentOpts && <div className={bem("help-text", [], baseClass)}>
+        <HelperErrorText id={helperContentOpts.id} isError={false}>{helperContentOpts.content}</HelperErrorText>
+      </div>}
+      {showError && <div className={bem("help-text", [], baseClass)}>
```
EdwinGuzman: I think you meant `error-text` rather than `help-text`?

src/react-components/src/components/00-base/Container/Modal.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export interface ModalProps {
+  /** Used by BEM CSS */
+  modifiers?: string[];
+
+  /** Used by BEM CSS */
+  blockName?: string;
+}
+
+export default class Modal extends React.Component<ModalProps, {}> {
```
EdwinGuzman, line 14: This is nice but there are no tests or stories for this (even if it's a small component).

src/react-components/src/components/01-atoms/Forms/TextField.tsx
```diff
@@ -35,6 +50,6 @@ export default function TextField(props: TextFieldProps) {
     textFieldProps["aria-required"] = true;
   }
 
-  let transformedTextField = <input {...textFieldProps} placeholder={placeholderText} />;
+  let transformedTextField = <input id={id} {...textFieldProps} placeholder={placeholderText} />;
```
EdwinGuzman, line 53: This is nice but why should the `id` now be optional?

src/react-components/src/components/03-organisms/List/UnorderedList.stories.tsx
```diff
@@ -0,0 +1,105 @@
...
+import UnderlineLink from "../../01-atoms/Links/UnderlineLink";
+import Checkbox from "../../01-atoms/Forms/Checkbox";
+import { action } from "@storybook/addon-actions";
+
+export default {
+  title: "UnorderedList",
+  component: UnorderedList,
+};
+
+const links = [<IconLink url="#url1">Link1</IconLink>, <IconLink url="#url2">link2</IconLink>, <IconLink url="#url3">link3</IconLink>];
```
EdwinGuzman, line 17: Small console error but not an app error: For arrays with components you should add a `key` attribute to not get 
```Warning: Each child in a list should have a unique "key" prop.```
in the console.

File: src/react-components/src/components/03-organisms/List/UnorderedList.stories.tsx (hunk 27 above)
Comment (EdwinGuzman) [reply src/react-components/src/components/03-organisms/List/UnorderedList.stories.tsx:112]: It is broken.

`.fill` works by updating _existing_ values in an array. This is setting each value to `text` but the array is empty. It should be something like:
```
let textList: JSX.Element[] = new Array(100);
textList.fill(text, 0, 100);
```

File: src/react-components/src/components/03-organisms/List/UnorderedList.tsx [hunk 50]
```diff
@@ -0,0 +1,44 @@
+/** Wrapper component for an unordered list of objects */
+import * as React from "react";
+import bem from "../../../utils/bem";
+import { boolean } from "@storybook/addon-knobs";
```
EdwinGuzman, line 3: You don't need this in the component.

EdwinGuzman (commented):
Very nice update. There are some tests missing and minor things here and there, but also review the accessibility tab in Storybook to catch any issues in the examples.

hokei:
@scooooooooby 
re: https://github.com/NYPL/nypl-design-system/pull/111#discussion_r406933034 (I don't know why Github isn't giving me a reply box) can you elaborate?  I'm not sure what's going on here

src/react-components/src/components/01-atoms/Forms/Checkbox.tsx
hokei: Are you using a tool to figure out what isn't being used?  It feels like something my IDE should do but I don't have it set up

EdwinGuzman: Yes. I use vscode but I actually don't remember what extension is being used... or maybe it's a lint configuration. It says the variable isn't used in the file, but then I double checked by looking for it manually in the repo and it didn't come up.

hokei:
> I notice there's no story for Modal—was that intentional? I think one would be a good idea, even if it's just like... a button that opens a gray box and closes it. I think the idea is to illustrate the pattern

Not intentional, adding one in.  

> I'm a little iffy on the DateRangePicker—I'd love to hear your thoughts on it @hokei. Does this feel like a RN-specific component to you? 

If there's nothing else that has two date fields like this (input type = number, etc) then I'm happy to move it to RN.  If it's possible for more than one product to have a picker like this, and if those two products should have their pickers be the same, I think there's an argument to have it here.  There's a few "gotchas" in implementing this, so I think it could save people time in the future if they want something like this.  

> Depending on your answer for this, I might consider adding a value value so clicking on the "^V" within the form gives you something like "1901" instead of "-1" or "1". 
If we assign a straight value, we won't be able to change it unless we get state set up in this component.  In the consumer, the value is hooked up to state properly.  State can exist in design system (I think accordion is a very good obvious example of a place where the design system state makes sense) but we  have to be careful about it because I don't want it to get in the way of the consumer

> Also looks like _icons.scss got deleted—why?

Somehow that file got duplicated.  https://github.com/NYPL/nypl-design-system/blob/development/src/styles/01-atoms/_images/icons/_icons.scss and https://github.com/NYPL/nypl-design-system/blob/development/src/styles/01-atoms/_icons/_icons.scss 
(yes, I did spend an embarrassingly long time trying to figure out why my CSS wasn't recompiling before finding this)

> I think we should get in _accordion.scss file somewhere in here, as it looks like accordion buttons fill up all available horizontal space (as opposed to outline/secondary buttons, which just have left/right padding). (There are some other stylistic things that are missing from this as well, like the outline extending to the checkboxes below, for example.) 

How best do you think we should go about this?  Do you want to make a PR against this branch (or send me files I should add in?) Thanks!

src/react-components/src/__tests__/components/Label-test.tsx
hokei: Reminded me that I really don't like having to pass in content and also checking for children, and that it should be OK just to check for children.  This is a breaking change but this PR is full of them, so I made those updates

src/styles/03-organisms/list/_list.scss
hokei: changed up the list tsx so I think it's applied now.

src/react-components/src/components/01-atoms/Forms/Input.stories.tsx
hokei: Since dropdown currently also can have both helper and error, I think we should split this out into a separate task.  I'm not sure where that should live - I'm happy to make a ResearchNow ticket for it but I don't know how it will be prioritized.  

Mainly, I think this will require a re-thinking of the way HelperErrorText is done - HelperErrorText can't really check number of instances of itself (nor should it, DateRangeForm has two helpers) so I think we might want to change it to a wrapper or HOC that appends helper or error based on certain checks.  

If we can, I would rather not block this ticket pending that work.

src/react-components/src/components/01-atoms/Button/Button.tsx
```diff
@@ -40,9 +33,9 @@ export interface ButtonProps {
   content?: string | JSX.Element;
   /** Additional attributes passed to the button */
   attributes?: {};
-  /** Used for BEM css convention. */
+  /** Optional modifiers array for use with BEM. See how to work with modifiers and BEM here: http://getbem.com/introduction/ */
```
EdwinGuzman, line 32: I don't think you need these for every instance this comes up... It might be better to include this in the "intro" homepage for storybook and in the readme documentation.

src/react-components/src/__tests__/components/Checkbox-test.tsx
```diff
@@ -0,0 +1,22 @@
+import { expect } from "chai";
+import { stub } from "sinon";
+import * as Enzyme from "enzyme";
+import * as React from "react";
+import * as Mocha from "mocha";
+
+import Checkbox from "../../components/01-atoms/Forms/Checkbox";
+
+describe("HelperErrorText Test", () => {
+  it("Renders HelperErrorText", () => {
```
EdwinGuzman: Wrong descriptions.

src/react-components/src/__tests__/components/Accordion-test.tsx
EdwinGuzman: Did this get updated?

src/react-components/src/components/01-atoms/Forms/Input.stories.tsx
helenvholmes: That makes sense. Let's ask for a backlog ticket from Andrew in SFR (since this does get shipped in RN and I think the current implementation is kind of wonky I think it's fair to ask for).

helenvholmes:
> If there's nothing else that has two date fields like this (input type = number, etc) then I'm happy to move it to RN. If it's possible for more than one product to have a picker like this, and if those two products should have their pickers be the same, I think there's an argument to have it here. There's a few "gotchas" in implementing this, so I think it could save people time in the future if they want something like this.

I'm of the same mind. I don't think what's been designed for SFR is more robust than two number pickers but I think there's room for something more robust. I'll surface this with the UX team today in standup.

> I think we should get in _accordion.scss file somewhere in here, as it looks like accordion buttons fill up all available horizontal space (as opposed to outline/secondary buttons, which just have left/right padding). (There are some other stylistic things that are missing from this as well, like the outline extending to the checkboxes below, for example.) 
>> How best do you think we should go about this?  Do you want to make a PR against this branch (or send me files I should add in?) Thanks! 

Accordion is brand-new, so you should just be able to make an `_accordion.scss` file at the same level as `Accordion.tsx` in the `styles` package.

hokei:
> Accordion is brand-new, so you should just be able to make an _accordion.scss file at the same level as Accordion.tsx in the styles package. 

Ah I think I misunderstood you.  Yes, I'll get those styles up!  thanks!

src/react-components/src/components/01-atoms/Button/Button.tsx
helenvholmes: This suggestion came from me, as going forward I'd like to make sure we always have the info chart on components filled out:

Screen Shot 2020-04-14 at 9 59 16 AM

I agree with you though, Edwin, that it would be nice if these were a link to longer documentation that describes BEM/how we're using it/etc. Those docs need to be written. Issue #109 tracks that.

helenvholmes (approved):
Let's follow up on Button and Accordion later and get this merged to cut a release for SFR's sprint.

src/react-components/CHANGELOG.md
```diff
@@ -4,6 +4,31 @@ All notable changes to this project will be documented in this file.
...
+### Added: 
+- `Checkbox`
+- `DateRangeForm` 
+- `Input`
+- `Accordion` 
+- `UnorderedList`
+- `Modal`
+
+### Changed: 
+- `Button`: now has `link` type, allows passing of `buttonType`
```
hokei, line 29: ugh this is wrong.

PR #118, opened 2020-04-17
SFR-625 Styles Cleanup

hokei:

- Fixes some styles issues when qa-ing SFR-625

### Front End Review:

### DIFF
--- a/src/react-components/src/components/02-molecules/Menus/Pagination/Pagination.tsx
+++ b/src/react-components/src/components/02-molecules/Menus/Pagination/Pagination.tsx
@@ -56,7 +56,7 @@ export default function Pagination(props: PaginationProps) {
   return (
     <nav aria-label="Pagination" className={bem(pagination__base_class)}>
       <Button callback={buttonPrevOpts.callback} {...buttonPrevOpts}>{buttonPrevOpts.content}</Button>
-      <div className={bem("dropdown-container", [], pagination__base_class)}><Dropdown labelPosition="left" {...dropdownProps}></Dropdown></div>
+        <Dropdown blockName={pagination__base_class} labelPosition="left" {...dropdownProps}></Dropdown>
       <Button callback={buttonNextOpts.callback} {...buttonNextOpts}>{buttonNextOpts.content}</Button>
   </nav>);
 }
--- a/src/react-components/src/components/02-molecules/Menus/Pagination/Pagination.tsx
+++ b/src/react-components/src/components/02-molecules/Menus/Pagination/Pagination.tsx
@@ -3,7 +3,6 @@ import bem from "../../../../utils/bem";
 import Button, { ButtonOptions } from "../../../01-atoms/Button/Button";
 import Dropdown from "../../../02-molecules/Dropdown/Dropdown";
 import { ButtonIconPositions, ButtonTypes } from "../../../01-atoms/Button/ButtonTypes";
-import ButtonType from "@storybook/addon-knobs/dist/components/types/Button";
 
 export interface PaginationProps {
   paginationDropdownOptions: string[];
@@ -57,7 +56,7 @@ export default function Pagination(props: PaginationProps) {
   return (
     <nav aria-label="Pagination" className={bem(pagination__base_class)}>
       <Button callback={buttonPrevOpts.callback} {...buttonPrevOpts}>{buttonPrevOpts.content}</Button>
-      <Dropdown labelPosition="left" {...dropdownProps}></Dropdown>
+      <div className={bem("dropdown-container", [], pagination__base_class)}><Dropdown labelPosition="left" {...dropdownProps}></Dropdown></div>
       <Button callback={buttonNextOpts.callback} {...buttonNextOpts}>{buttonNextOpts.content}</Button>
   </nav>);
 }
--- a/src/styles/01-atoms/text/text/_text.scss
+++ b/src/styles/01-atoms/text/text/_text.scss
@@ -62,7 +62,6 @@ p,
 
 .unordered-list {
   @include list-reset;
-  list-style-position: outside;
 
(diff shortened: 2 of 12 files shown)


src/react-components/src/components/02-molecules/Menus/Pagination/Pagination.tsx
```diff
@@ -57,7 +56,7 @@ export default function Pagination(props: PaginationProps) {
   return (
     <nav aria-label="Pagination" className={bem(pagination__base_class)}>
       <Button callback={buttonPrevOpts.callback} {...buttonPrevOpts}>{buttonPrevOpts.content}</Button>
-      <Dropdown labelPosition="left" {...dropdownProps}></Dropdown>
+      <div className={bem("dropdown-container", [], pagination__base_class)}><Dropdown labelPosition="left" {...dropdownProps}></Dropdown></div>
```
helenvholmes: Could `Dropdown` allow us to pass these values to it instead of adding them to a div just outside of it?

src/styles/01-atoms/text/text/_text.scss
```diff
@@ -75,6 +74,10 @@ p,
     list-style: initial;
     color: $gray-medium;
   }
+
+  &__list-item {
+    list-style-position: inside;
```
helenvholmes, line 79: Do the designs not hang their punctuation? If so we should talk with Ellen about it, it makes content jump horizontally a lot if not

src/styles/03-organisms/accordion/_accordion.scss
```diff
@@ -3,11 +3,11 @@
...
 
-  > button {
+  &__button {
     border: unset;
   }
 }
 
 .accordion-content {
-  @include space-inset-xs;
+  padding: $space-xs 0;
```
helenvholmes: This change looks like a regression. Now items within the accordion butt right up against the sides: 

Screen Shot 2020-04-20 at 10 56 44 AM

helenvholmes (changes requested):

src/styles/01-atoms/text/text/_text.scss
hokei: I ... am not sure what this means?  I was trying to solve this issue where it looks different in mobile and desktop widths:  

mobile: 
Screen Shot 2020-04-20 at 11 26 14 AM

desktop
Screen Shot 2020-04-20 at 11 26 23 AM

helenvholmes: Ah, I see. I would request that you push back on Ellen for these designs—it's odd that they're being hidden on mobile, and I don't think that unordered lists should hide their bullets in mobile for the DS at least. If it's necessary that they do that should only exist on SFR.

hokei: sorry, I think I wasn't clear - without the change, this is what it looks like.  You can check it out here:  http://researchnow-beta.nypl.org/
With this change, the bullets will no longer hide in mobile, which is what we want.  But if there's a better way to make the bullets not hide in mobile, I'm all up for it!

helenvholmes: Ah, I see what you're saying. I have some thoughts on ways to do this but not anything that fixes it quickly. I'll resolve this, carry on.

src/styles/03-organisms/accordion/_accordion.scss
hokei: hmm so we want it on three sides, then.  But it doesBecause an inset on all of the box is going to cause the scrollbar to float away from the side like this - 
Screen Shot 2020-04-20 at 11 45 18 AM

It seems like this should be in the `unordered-list` style instead of this, but how to make it work?

src/react-components/src/components/02-molecules/Menus/Pagination/Pagination.tsx
hokei: haha I wrote out a whole thing about how custom classes are really tricky here but it turns out https://github.com/NYPL/nypl-design-system/blob/development/src/react-components/src/components/02-molecules/Dropdown/Dropdown.tsx#L60 never has a blockname passed into it and I think that's probably the best place to put these styles, right?

src/styles/03-organisms/accordion/_accordion.scss
helenvholmes: `padding: $space 0 $space $space`

src/react-components/src/components/02-molecules/Menus/Pagination/Pagination.tsx
helenvholmes: Yes, same as button. I'd modify `Dropdown` to take a `blockName` and then do the same as button here

src/styles/03-organisms/accordion/_accordion.scss
hokei: OK!  Still feels weird to have it in `accordion` rather than `unordered-list` since `unordered-list` is the thing that scrolls, not `accordion`, but ... I guess ... future problem?

helenvholmes (approved):

src/styles/03-organisms/accordion/_accordion.scss
helenvholmes: I think it makes sense here still—let's make sure we both are understanding this correctly later

src/styles/03-organisms/accordion/_accordion.scss
```diff
@@ -3,11 +3,11 @@
...
 
-  > button {
+  &__button {
     border: unset;
   }
 }
 
 .accordion-content {
-  @include space-inset-xs;
+  padding: $space-xs 0 $space-xs $space-xs
```
helenvholmes: nit: semicolon at the end

PR #120, opened 2020-04-22
SFR-901 added instance card and changed download button

hokei:

- Creates an Instance card story 
- Adds a download link to the edition card 

### Front End Review:


hokei:
@scooooooooby This work has made it clear to me that we should just have the edition card links accept anything, that we center it and apply styles, rather than keeping that `if (typeof link === "string") { ` block.  I can do that here or a new ticket? 


edit:  after thinking on a bit more I think I will make those changes here since RN needs to deal with a lot of breaking changes anyway, might as well do it all at once.  I'll request review when I'm done.  Sorry for the unnecessary ping.

src/react-components/src/__tests__/components/EditionCard-test.tsx
```diff
@@ -4,40 +4,45 @@ import * as Enzyme from "enzyme";
...
 describe("EditionCard", () => {
   let regularEditionCard = <EditionCard
     id="card#1"
     blockName=""
     coverUrl="https://placeimg.com/300/400/arch"
     editionHeadingElement={<a href="edition-link" >2004 Edition</a>}
-    editionInfo={["Published in New York by Random House", "Written in English", "Under Creative Commons License"]}
-    readOnlineLink="#readOnlineUrl"
-    downloadLink="#downloadUrl" />;
+    editionInfo={[<>Published in New York by Random House</>, <>Written in English</>, <>Under Creative Commons License</>]}
```
helenvholmes, line 17: Are these tags supposed to have content in them?

hokei: these are fragments, which are valid.  They make it possible for us to not have to explicitly check for strings, since if you want to pass in strings, passing in fragments will do the same thing (and you don't have an extra empty div or span)

helenvholmes: ahhh, TIL!

src/styles/02-molecules/edition-card/_edition-card.scss
```diff
@@ -54,23 +54,24 @@
...
         @include space-stack-s;
       }
     }
   }
 
+  &__card-button-link {
+    @include button--filled;
+  }
+
+  &__more-link {
```
helenvholmes: For some reason I'm not seeing a `@space-inline-xs` or whatever it is between the icon and the "Download" string.

This link should also be black according to the specs, although I don't know why. We should reach out to Ellen tomorrow when she's in about updating the documentation for these cards (buttons are also getting reverted back to blue): http://themetronome.co/components/mc-14-edition-card/?fresh=true

src/styles/01-atoms/links/link/_link.scss
```diff
@@ -47,6 +47,8 @@ $color-link--hover: $nypl-blue-dark;
 
   &__icon {
     @include icon;
+    
+    margin-left: $space-xxs;
```
helenvholmes: Please use the `@space-inline` include instead of a `margin-left`.

src/react-components/src/components/02-molecules/Cards/EditionCard.stories.tsx
```diff
@@ -62,16 +64,26 @@ export const editionCardWithEmptyEditionInfo = () => <EditionCard
   coverUrl="https://placeimg.com/300/400/arch"
   editionHeadingElement={<a href="edition-link" className={bem("link", [], "heading")} >2004 Edition</a>}
   editionInfo={[]}
-  readOnlineLink="#readOnlineUrl"
-  downloadLink="#downloadUrl">
+  readOnlineLink={<BasicLink className={bem("card-button-link", [], "edition-card")} url="blah">Read Online</BasicLink>}
+  downloadLink={<IconLink iconName="download" iconPosition="left" url="blah">Download</IconLink>}>
 </EditionCard>;
 
 export const editionCardWithAtypicalImageSize = () => <EditionCard
```
helenvholmes, line 61: I like this story. It makes me think that there we should consider a background on these images, such as [catalog item](http://themetronome.co/components/mmb-4-indi-catalog-item/?fresh=true). (Catalog item is using `$gray-warm-light` for its background, I believe.)

src/styles/02-molecules/edition-card/_edition-card.scss
hokei: If the canonical version here should be blue, we should have it blue here ... but the button is canonically black as of right now (https://nypl.github.io/nypl-design-system/storybook/storybook-static/react/index.html?path=/story/button--button-filled-icon) and if there's been a change about that I'd rather file a new ticket rather than deal with updating all buttons in this ticket.

src/react-components/src/components/02-molecules/Cards/EditionCard.stories.tsx
```diff
@@ -41,19 +43,19 @@ export const editionCardWithSomeEditionInfo = () => <EditionCard
   blockName=""
```
helenvholmes, line 43: The difference between the `editionCardWithSomeEditionInfo` and its following story isn't readily apparent to me other than that one includes double quotes around the Publisher information. What are these two stories intended to communicate?

src/styles/02-molecules/edition-card/_edition-card.scss
helenvholmes: Sorry for getting this muddled about buttons. Yes, I agree with you, if buttons need to be updated we should file a ticket and designs should be updated first.

As for the link, I think you should fix the spacing but I'm fine with leaving it stylistically in every other respect to our default links. While I think what Ellen's done here looks nice I'm not certain the logic for why she's made this link different from all other links on the site.

hokei: ah it's more complicated than this - this is the button-like-link which probably ties into https://github.com/NYPL/nypl-design-system/issues/122.  Anyway, a ball of issues that need to be handled on their own.

src/react-components/src/components/02-molecules/Cards/EditionCard.stories.tsx
hokei: I'd wait till after the card consolidation work to apply this - for now, I think the metronome doesn't reflect this and it might confuse QA on our end.

helenvholmes (changes requested):
Left some comments about visual changes, but this is looking good otherwise!

src/react-components/src/components/02-molecules/Cards/EditionCard.stories.tsx
hokei: this was originally to show passing components, but it's better to consolidate them now since components are always passed.

helenvholmes:
Pushed up some visual nits on the button. If you look at the CSS... it doesn't look great. Makes me think that some of our assumptions on `more-link` have uh, made asses out of us (excuse my French)

I've stolen `0.0.18` on the styles package with the font update so you'll have some merge finagling with your package/CHANGELOG but once those are fixed I say ⛴

hokei:
I definitely look forward to fixing the `.more-link` stuff, it's atrocious.  It's a problem because `.more-link` shouldn't be styled under `card-info-link`(card-info-link should not be aware of what's being passed into it) but ... ughhh it's all a minefield and I don't know how to fix it without re-writing all of `link` and I don't have time for that right now.  

Thanks for helping me get it to look right, I'll bump and release.

helenvholmes (approved):

PR #121, opened 2020-04-23
Makes links blue again, fixes left/right icon stories

helenvholmes:
- Reverts links back to blue
- Reverts `@link` and `@more-link` mixins to what they were pre-`hero`


hokei:
https://nypl.github.io/nypl-design-system/storybook/storybook-static/react/index.html?path=/story/heading--headings-with-bold is this also a regression from the Hero work?  (let's file a ticket if not) 

edit:  I think this might be a regression from different work, let's file an issue.

hokei:
was the "Header Image Right" story and component supposed to be deleted?  (I still see it here.  Merging issues, I assume?)

hokei (changes requested):
changelog and bump as well, please!  Thanks!

helenvholmes:
@hokei yeah, merging issues—some wonky stuff happened when I tried to merge on Tuesday because of Github status issues.

Updated CHANGELOG and package bump comin' right up

helenvholmes:
@hokei I updated the package.json on `react-components` back to a minor number like we talked about... Not certain what happens now that you've run a release, though 😖 really sorry I didn't catch that yesterday, if I need to revert I can

hokei:
I think the main problem with releasing `2.0.0` is if we get to 1.0 release and we want to release 2.0 after that, we'll have to skip and go from 1->3 because 2 already exists.  
I'm not too worried because I think when we do a 1.0 release we'll have figured out a real name for Design System.

hokei:
:shipit:

hokei (approved):

PR #130, opened 2020-05-11
SFR-902 filters and more edition card bugfixes

hokei:

-

### Front End Review:

### DIFF
--- a/src/react-components/src/components/01-atoms/Button/Button.tsx
+++ b/src/react-components/src/components/01-atoms/Button/Button.tsx
@@ -114,10 +114,11 @@ export default class Button extends React.Component<ButtonProps, {}> {
     }
 
     let btnCallback = mouseDown ? { onMouseDown: callback } : { onClick: callback };
+
     return (
       <button
         id={"btn-" + id}
-        className={bem(button_base_class, buttonModifiers, blockName)}
+        className={`${button_base_class} ${bem(button_base_class, buttonModifiers, blockName)}`}
         type={type}
         {...attributes}
         {...btnCallback}>
--- a/src/styles/02-molecules/edition-card/_edition-card.scss
+++ b/src/styles/02-molecules/edition-card/_edition-card.scss
@@ -31,7 +31,6 @@
     @include space-stack-s;
     @include space-inline-s;
 
-    background-color: $gray-warm-light;
     max-width: 150px;
     flex: 0 0 75px;
     align-items: center;
--- a/src/styles/02-molecules/edition-card/_edition-card.scss
+++ b/src/styles/02-molecules/edition-card/_edition-card.scss
@@ -34,7 +34,6 @@
     background-color: $gray-warm-light;
     max-width: 150px;
     flex: 0 0 75px;
-    display: flex;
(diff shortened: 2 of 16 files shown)


src/react-components/src/components/01-atoms/Button/Button.tsx
```diff
@@ -114,10 +114,11 @@ export default class Button extends React.Component<ButtonProps, {}> {
     }
 
     let btnCallback = mouseDown ? { onMouseDown: callback } : { onClick: callback };
+
     return (
       <button
         id={"btn-" + id}
-        className={bem(button_base_class, buttonModifiers, blockName)}
+        className={`${button_base_class} ${bem(button_base_class, buttonModifiers, blockName)}`}
```
hokei, line 121: Is this our implementation of Bem being weird?  Bem won't apply the `button` class unless I explicitly use it.  Or should should all the button classes be @include-ing the button class

src/styles/02-molecules/edition-card/_edition-card.scss
```diff
@@ -34,7 +34,6 @@
     background-color: $gray-warm-light;
     max-width: 150px;
     flex: 0 0 75px;
-    display: flex;
```
hokei, line 37: someday over drinks, ask me about the headaches this caused

hokei:
The publish commit is here because I think I released from this branch instead of development.  I've added a thing to the readme to remind people to check for this, though hopefully we can set up a travis release button after the rip-out-lerna work.

src/react-components/src/components/01-atoms/Button/Button.tsx
helenvholmes: Er, all paths lead to Rome, I guess

For something as atomic as buttons I think that they should probably have the button class applied, and in addition the option to base in blocks. I think what you've done here makes sense.

src/styles/02-molecules/edition-card/_edition-card.scss
helenvholmes: Oh yikes, okay 😣I'm so sorry, I thought this was going to be an easy CSS fix for you!!

hokei: Oh it's definitely not you!  It was a subtle thing that caused a lot of things to go badly, and turns out that removing this line is the thing that fixes it. It was a fun adventure to track down.  The moral of the story here is that in the future we need to QA on physical phones ideally, and Safari at minimum, as well.

helenvholmes: We should probably talk about it, because I'm still seeing a bug that I would have expected this to fix—but it's small, so I'm happy to approve this and follow-up with this later:

Screen Shot 2020-05-12 at 10 27 16 AM

What were the issues we were seeing on physical phones? Can you send me the QA tickets?

hokei: Yeah.  Let's move this to slack.

helenvholmes (approved):
LGTM

PR #136, opened 2020-05-15
Template

### LINKED ISSUE #88: Add story for card_grid.scss
### Component in Question
- [card grid scss](https://github.com/NYPL/nypl-design-system/blob/development/src/styles/base/_card-grid.scss)

### Description of Ticket
Add a story that shows off functionality of card_grid.scss.

- Controls should allow users to add/subtract cards

### LINKED ISSUE #169: Template Documentation
Blocked by #140 

Templates should have documentation about where content is intended for. This is left over from some old code that should help as a starter:

```
export interface TemplateProps {
  /** Enum to control if Template has left/right sidebar or is full width. */
  layoutType: LayoutTypes;

  /** Content intended for inside <header>. */
  headerContent: JSX.Element;

  /** Primary content within <main> element. */
  primaryContent: JSX.Element;

  /** Content intended for inside <footer>. */
  footerContent: JSX.Element;

  /** Heroes appear inside <main> but extend full-width of template. */
  heroContent?: JSX.Element;

  /** Separate block for content above primary content within <main>. */
  topContent?: JSX.Element;

  /** Content intended for inside the sidebar within <main>. */
  sidebarContent?: JSX.Element;
}
```

helenvholmes:
[ResearchNow Jira Ticket](https://jira.nypl.org/browse/SFR-894)

- Adds a template component that can be:
  - full width
  - with left sidebar
  - with right sidebar

### Front End Review:
- [ ] View [the example in Storybook](http://localhost:9001/?path=/story/template--right-sidebar-template)


src/icons/src/svg/arrow-xsmall.svg
```diff
@@ -1 +1,3 @@
-<svg width="14" height="9" xmlns="http://www.w3.org/2000/svg"><path fill-rule="evenodd" clip-rule="evenodd" d="M5.984 7.578L.424 1.514 1.9.163l5.22 5.695L12.341.163l1.475 1.351-5.56 6.064a1.542 1.542 0 01-2.272 0zm1.474-1.351z"/></svg>
+<svg width="8" height="12" fill="none" xmlns="http://www.w3.org/2000/svg">
```
hokei: hmmm why as this changed?  Is this a new file or script weirdness?

helenvholmes: I updated it—designs have changed. Team's moving over to Google's Material icons for the utility icons (arrows, downloads, etc.)... but frankly that isn't all organized yet (and won't be until I do it) so I can rip out this change from this ticket for now. (I have, however, already pushed up the change for arrows in Reno, so I'd like to do it here if I can.)

hokei: sounds good

helenvholmes:
Build is failing... I'll be looking into it tomorrow, realistically.

src/react-components/src/__tests__/components/Template-test.tsx
```diff
@@ -0,0 +1,57 @@
+import { expect } from "chai";
+import { stub } from "sinon";
+import * as Enzyme from "enzyme";
+import * as React from "react";
+import Template from "../../components/Template/Template";
+import { LayoutTypes } from "../../components/Template/LayoutTypes";
+import Placeholder from "../../components/00-base/Placeholder";
```
hokei: I think this isn't importing from the right place

src/react-components/src/components/Template/Template.tsx
```diff
@@ -0,0 +1,86 @@
...
+    rightSidebar = createSidebar(sidebarContent);
+  } else {
+    return;
+  }
+
+  if (layoutType === LayoutTypes.SidebarLeft || LayoutTypes.SidebarRight) {
+    mainModifiers.push("with-sidebar");
+  }
+
+  return (
```
hokei, line 65: In ResearchNow we don't build our pages out with a `header`, `main`, and `footer` on the same page -  we inject the header with the script (https://github.com/NYPL/sfr-bookfinder-front-end/blob/development/src/views/index.ejs#L61) 

This means that we can't use `<Template>` as a component, but we can still make use of all the CSS by calling the classnames directly, using this as an example.  Is that OK, or is this approach something we should re-think?

hokei: I think maybe this component could exist as an example of how you're supposed to build your pages, with the right semantic elements and classes, rather than something that a consumer is supposed to import and use.

helenvholmes: Reno does the same thing, it just does the script injection inside the `header` (so a few lines up from where you have it). Is the issue that I've set this up so the content blocks only take JSX.Elements? Would it be solved if they could take just some compiled HTML?

src/react-components/src/__tests__/components/Template-test.tsx
helenvholmes: will fix

helenvholmes:
@hokei and I talked about the direction of this ticket over Slack. She'd like to take a chance to approach it differently so I'm marking this as blocked until she's had a chance to do so.

helenvholmes:
Closing this out in favor of #167.

PR #139, opened 2020-05-18
SFR-883 Inline SVG icons.

### LINKED ISSUE #128: Inline SVGs instead of shipping them separately
### Description of issue:
Right now we are shipping SVGs separately which means that webpack in the consuming app has to compile them.  SVG compiling is not standard and can be tricky - DS should embed or inline the svgs so that the consumer doesn't need to deal with SVG issues.

hokei:

[ResearchNow Jira Ticket](http://jira.nypl.org/browse/SFR-883)

- Inlines SVG icons to avoid cross origin errors in webkit browsers.  
- Moves icons into the /react-components folder
- Stops using sprite-style icons (inline-loading and sprites don't play well)  

Resolves https://github.com/NYPL/nypl-design-system/issues/128
### Front End Review:


src/react-components/src/components/01-atoms/Images/Icons/Icon.tsx
```diff
@@ -1,46 +1,57 @@
...
+      "role": decorative ? "img" : role,
+      "aria-hidden": decorative,
+      "aria-labelledby": title ? "title-" + name : undefined,
+      "aria-describedby": desc ? "desc-" + name : undefined,
+      "title": title ? `title-${name}` : undefined,
+      "description": desc ? `title-${name}` : undefined
+    };
+    let svg = allSvgs[name];
+    return (
+      <div {...iconProps} dangerouslySetInnerHTML={{ __html: svg }} />
```
hokei, line 53: @scooooooooby is this going to cause problems?  The svg is passed as a string and not natively parseable, so it's a bit more complicated to get `<svg {...iconProps}></svg>` but I can look into it more if this is a thing we definitely need.

src/react-components/src/components/01-atoms/Images/Icons/arrow.svg
```diff
@@ -0,0 +1 @@
+<svg width="24" height="14" xmlns="http://www.w3.org/2000/svg"><path fill-rule="evenodd" clip-rule="evenodd" d="M10.526 12.871L.263 1.676 1.737.324 12 11.52 22.263.324l1.474 1.352L13.474 12.87a2 2 0 01-2.948 0z"/></svg>
```
hokei, line 1: I think one of these things was replaced in Template - note to self to not overwrite that work

src/react-components/src/components/01-atoms/Images/Icons/Icon.tsx
helenvholmes: Perhaps in the future (say, the UX team wants to begin animating icons, or dynamically changing their fills based on component state) but we haven't received any designs with that sort of thing yet. For now I don't think we need to overengineer this and this is great!

src/react-components/src/components/01-atoms/Images/Icons/arrow.svg
helenvholmes: I'll handle the merge conflicts, you don't have to worry about it

src/react-components/src/components/01-atoms/Images/Icons/Icon.tsx
hokei: Sounds good.  It's also not a problem that we're no longer using the spritesheet, right?  I will remove the unused scripts if that's ok

helenvholmes: Yeah, I think now that you can call `Icon` and send it a name, that seems good to me

helenvholmes (approved):
LGTM. Incredible work @hokei, I know it wasn't fun 🎉👏🏻

Going to give @EdwinGuzman a chance to look over this as well before I add the approval label, but lgtm.

src/react-components/src/components/01-atoms/Images/Icons/Icon.tsx
```diff
@@ -1,46 +1,57 @@
...
+import minus from "../../../../../icons/svg/minus.svg";
+import plus from "../../../../../icons/svg/plus.svg";
+import search_small from "../../../../../icons/svg/search-small.svg";
+import search from "../../../../../icons/svg/search.svg";
 
-  export default class Icon extends React.Component<IconProps, {}> {
-    constructor(props: IconProps) {
-      super(props);
-    }
+const allSvgs = {
```
EdwinGuzman, line 12: Nit: you can do the shorthand version for this object.

EdwinGuzman (approved):
Looks good. A SimplyE project is using `svgr` - https://github.com/gregberge/svgr - to turn svg files into react files. An alternative but this works just as well.

hokei:
@EdwinGuzman I tried using `svgr` and couldn't get it to work!  Would love to take a look at it at some point and see what I missed.  Thanks!

PR #165, opened 2020-05-20
Update BEM for UnorderedList

### LINKED ISSUE #117: Bem doesn't seem right for modifiers with no blocknames
### Screenshot of component in your project:
Screen Shot 2020-04-17 at 3 28 05 PM

### Description of issue:
If there is no blockName, the underscores should be removed.

### Description of desired behavior:
A class with only a modifier should appear like so:

`.classname--modifier`

M0nica:
Resolves: https://github.com/NYPL/nypl-design-system/issues/117 

- Moves UnorderedList into the top-level of components
- Adds unit tests for UnorderedList and BEM
- Changes the order that values are being passed into BEM function in UnorderedList (from:  `bem("list-container", modifiers, baseClass)` to `bem(baseClass, modifiers, "list-container")` )

### Front End Review:


src/react-components/src/__tests__/utils/bem.test.ts
```diff
@@ -0,0 +1,32 @@
...
+    it("returns properly formatted baseclass and modifier", () => {
+        expect(bem("list-item", ["scroll"])).to.equal(
+            "list-item list-item--scroll"
+        );
+    });
+});
+
+describe("given a baseclass with no modifier and  blockName", () => {
+    it("returns properly formatted baseclass and blockName", () => {
+        expect(bem("list-item", [], "unordered-list")).to.equal(
```
hokei, line 28: @scooooooooby you know more about bem's intended effects than I do ... Should this be `"unordered_list unordered-list__list-item"` to match the functionality when passed a modifier?  Or should modifier-only be `list-item--scroll` ? 

Also, L13, is it missing a `list_item` or is it missing a `unordered-list`?  (or is it good the way it is?)

src/react-components/src/__tests__/components/UnorderedList-test.tsx
```diff
@@ -0,0 +1,35 @@
...
+import * as Enzyme from "enzyme";
+import * as React from "react";
+
+import UnorderedList from "../../components/List/UnorderedList";
+
+describe("Link With Underline", () => {
+    let wrapper: Enzyme.ReactWrapper<{}, {}>;
+    let textList: JSX.Element[] = [];
+
+    it("Returns proper classes for UnorderedList with modifier", () => {
```
hokei, line 12: These tests are really testing bem, so we probably don't need it here.  

If there aren't UnorderedList tests, I think we want to test on mounting the `<UnorderedList>` and seeing `<ul>`s and `<li>`s

hokei:
Is it possible make `prettier` format everything and make a PR/commit that is just formatting, so it's easier to see what all changed?

helenvholmes:
@hokei I could just go through and save every file in `src/` and submit that just to get it over with 😅

[EDIT]: We now have #166 now. Whoopee 🤖

src/react-components/src/__tests__/components/UnorderedList-test.tsx
M0nica: @hokei I added these tests because the 🐛 bug, in this case, was due to the values being passed into BEM in a different order than expected. These tests are to tests that the unorderedList is sending the values from its props to the BEM function in the appropriate order.

helenvholmes: I agree with @hokei on this one—this test should live elsewhere since all components could run into this issue.

src/react-components/src/__tests__/utils/bem.test.ts
helenvholmes: Based on the order of this function I would expect line 28 to spit out `.unordered-list__list-item`, which is what @M0nica has here.

Line 13 also looks correct to me.

src/react-components/src/__tests__/components/UnorderedList-test.tsx
M0nica: I want to test that the bem function is called with the parameters in the proper order like:    
`unorderedList` sends the values in the right order like: 
```
<div className={bem(baseClass, modifiers, "list-container")}>
         {/**/}
        </div>
```
instead of `bem("list-container", modifiers, baseClass)`. This test is specific to how bem() is called within the OrderedList and without this test we could run into a similar bug in this same component if for some reason the order for how params in unorderedList are passed into the bem function changes

M0nica: Do you all think that bem should accept an object with named parameters instead of just passing in 3 values so it would accept something like

`bem({baseClass, modifiers, blockName: "list-container }` instead of `bem(baseClass, modifiers, "list-container")`? Without this change I think we could run into issues in other components passing the parameters to the bem function in a different order than what is expected.

M0nica:
> @hokei I could just go through and save every file in `src/` and submit that just to get it over with 😅
> 
> [EDIT]: We now have #166 now. Whoopee 🤖

I opened a PR (https://github.com/NYPL/nypl-design-system/pull/168)  to handle listing ass the `.ts` and `.tsx` in `react-components/`

src/react-components/src/__tests__/components/UnorderedList-test.tsx
helenvholmes: So to make sure I'm understanding this correctly, would we do something like this every time we called the bem util?

`bem({baseClass: unordered_list_baseclass, modifiers: unordered_list_modifiers_array, blockName: "list-container"})`

helenvholmes:
I think it makes sense to test the bem util separately from the unordered list, since what's happening with unordered list could potentially happen on any other component.

I think the ticket referenced called for eliminating the `__` unscores that appear when a blockName isn't provided, but a `--` modifier is. Will that be fixed with this PR?

helenvholmes (changes requested):
Discussion in the Conversation tab

M0nica:
> I think it makes sense to test the bem util separately from the unordered list, since what's happening with unordered list could potentially happen on any other component.
> 
> I think the ticket referenced called for eliminating the `__` unscores that appear when a blockName isn't provided, but a `--` modifier is. Will that be fixed with this PR?

Yes, this should resolve this issue as my understanding of the issue is that it was caused by unorderedList passing the properties into bem in a different order than the function was expecting as opposed to a problem with the bem function itself. Due to this confusion re: ordering is why I recommended potentially switching the bem function's input into a single object vs a series of ordered values.

src/react-components/src/__tests__/components/UnorderedList-test.tsx
M0nica: That's what I was proposing unless you have other recommendations or think it's too verbose but I think it could reduce the chance of swapping values when they are being passed into the bem function.

hokei: I think this is a good test - though if we're using it in many places I'd love to see it extracted into a utility, something like (checkBem?) so that all the tests have access to it (doesn't have to happen here, only if we're starting to use it everywhere). (also so if we find a bug in bem and we fix that bug, we only need to fix the tests in one place)

In general, I'd like stuff like this to be tested with visual regression testing, rather than specific class names, which are frustrating to change if there are lots of tests that are looking for specific class names.  (We don't really care that the `unordered-list unordered-list unordered-list--scroll` classes are called, we care that there is a scrollbar).  But setting that up isn't even on the roadmap at this point, so I think this is the closest we can get.

hokei:
This looks good to me, but I'll leave it to Helen to give the sign-off!

helenvholmes (approved):
LGTM!

src/react-components/src/__tests__/components/UnorderedList-test.tsx
helenvholmes: I actually don't oppose _a_ test like the one we have here simply because it helps with understanding BEM imo (Crystal converted the original BEM util from another repo here at NYPL and for a while we were both pretty confused at what the output should be vs. what it was).

But I agree, visual regression testing is something to look for in the future. We can talk about getting it in the backlog at next planning.

PR #170, opened 2020-05-21
Edit modal story

### LINKED ISSUE #112: Modal Story improvements
### Component in question:
[Modal](https://nypl.github.io/nypl-design-system/storybook/storybook-static/react/index.html?path=/story/modal--modal)

### Description of issue:
Right now the Modal story only has the open modal, which is not interactive.  It's better than nothing, but could be more useful.  

### Description of desired behavior:
Modal story should have a clickable button that opens the modal

helenvholmes:
- Closes #112: Edits the modal story to have an open and close button

### Front End Review:
- [ ] View [the example in Storybook](http://localhost:9001/?path=/story/modal--modal)

### DIFF
--- a/src/react-components/src/components/Modal/Modal.stories.tsx
+++ b/src/react-components/src/components/Modal/Modal.stories.tsx
@@ -1,5 +1,7 @@
 import * as React from "react";
 
+import Container from "../Container/Container";
+import Button from "../Button/Button";
 import Modal from "./Modal";
 
 export default {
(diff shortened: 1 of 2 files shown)


helenvholmes:
I'm not sure what I've potentially messed up by editing that `.container` SCSS... @hokei, is that code actively being used in SFR?

hokei:
@scooooooooby 
`.container` is being used in SFR but since we're going to be doing a lot of bugfixes before mobile release, if it's a breaking change that eventually needs to happen let's do them all now.

src/react-components/src/components/Modal/Modal.stories.tsx
```diff
@@ -1,14 +1,52 @@
...
-    <Modal>
-        <div>Modal Content</div>
-    </Modal>
-);
+class ModalStory extends React.Component<{}, { isModalOpen: boolean }> {
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            isModalOpen: false,
```
M0nica, line 17: Should the default for the modal be open?

M0nica (approved):
This looks good to me 👍

src/react-components/src/components/Modal/Modal.stories.tsx
helenvholmes: I'm agnostic /shrug 

Whichever tells the story best! We can switch it if you the story being in the other direction

PR #176, opened 2020-05-21
Decouple icon position and rotation

### LINKED ISSUE #124: Allow icon to be positioned with margin without being rotated
### Description of issue:
Right now the positioning of icons is tied to their rotation.  (see `IconLink` and the `download` link in `EditionCard`. This should not be the case.)

M0nica:
This is related to: https://github.com/NYPL/nypl-design-system/issues/124 

- Decouples icon positioning and rotation this impacts IconLink and its usages, i.e., the download link in EditionCard and pagination.

### Front End Review:


src/styles/CHANGELOG.md
```diff
@@ -9,10 +9,13 @@ This repo is in Prerelease. When it is released, this project will adhere to [Se
 ========
 
 ## [0.0.23]
-- added `Template` styles
+
+-   added `Template` styles
```
M0nica: Do we want prettier to ignore .md files?

src/styles/04-templates/_default.scss
```diff
@@ -3,72 +3,72 @@ $sidebar-width: 25%;
 $sidebar-width-s: calc(25% - calc(4 * #{$space-s}));
 
 .main {
-  @include wrapper;
+    @include wrapper;
```
M0nica: Do we want prettier to ignore .scss files?

src/react-components/CHANGELOG.md
```diff
@@ -7,7 +7,8 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 ========
 
 ## [0.3.6]
-- added `Template` stories
+
```
M0nica: Can we review when the changelog should be bumped vs. just adding the update to the current changelog number?

src/react-components/src/components/Button/Button.tsx
```diff
@@ -17,6 +17,7 @@ export interface ButtonOptions {
     mouseDown?: boolean;
     iconPosition?: ButtonIconPositions;
     iconName?: string;
+    iconRotation?: "rotate-180" | "rotate-270" | "rotate-0" | "rotate-90";
```
hokei: This might be a good time to extract all the iconProps into an `iconOptions` object.  (See `Input.tsx` and how it handles `HelperErrorText`).  I say this because `iconPosition` and `iconRotation` seem like similar props, except that `IconRotation` is a prop of `Icon` and `IconPosition` is a prop handled in `Button`

Also, passing in so many props like this starts to get really frustrating (you must have felt that considering you had to edit literally every story that used a button haha -- now imagine being a consumer who just wants to use pagination, and having to think about which direction the arrow is pointing when you really want DS to take care of that for you?)   The `-Options` objects allow for that object to be initialized, say, with an id, and for stuff to be added on as it gets passed from component to component.

so... uh two questions.  
1) What do you think about this approach?  It's already in `Input` (and `DateRangeForm` which uses `Input`).  I'm up for any other ideas on how to wrangle this props nightmare.  
2) Is this a good place to make that switch over?

I'm happy to chat offline about this on Tuesday also!

src/styles/01-atoms/_icons/_icons.scss
```diff
@@ -3,19 +3,26 @@
...
 
-    &--up {
+    &--up,
+    &--rotate-180 {
+        @include space-inline-xxs;
         transform: rotate(180deg);
     }
-
-    &--right {
+    &--right,
```
hokei: I would love to know if this is being used in places that aren't buttons.  If not, I'd be for removing it.  This would be a breaking change, but since we're making some big changes already in DS, I think it's better to do it now than later.

src/react-components/CHANGELOG.md
hokei: It should be bumped after a release.  Ideally we'd have releases at set times, (sprint end) but ... that's probably not going to happen for now.  

I think the clearest thing to do here is to pre-bump the changelog as a part of release, so that they never need to be bumped (unless there's a breaking change). I'll make a ticket to put that in the readme.  

EDIT: please ignore the above, it was dumb.  But I'm leaving it up for posterity.  

So.  Ideally, we would release by merging to master, and as a part of the merge, Travis would make a "set up next release" commit to development that bumps the changelog.  

Until we set this up, I think the best thing to do is to check https://github.com/NYPL/nypl-design-system/releases.  If the current version is released, bump and start a new version.  If it's not, then append.

src/react-components/src/components/Button/Button.tsx
M0nica: I updated to move the potential rotation  options into the `iconRotationOptions` enum

src/styles/01-atoms/_icons/_icons.scss
M0nica: @hokei I updated the scss to avoid introducing breaking changes to the styling

src/styles/04-templates/_default.scss
helenvholmes: No, I think prettier editing .scss files is fine.

src/styles/CHANGELOG.md
helenvholmes: See .scss comment

src/styles/01-atoms/_icons/_icons.scss
helenvholmes: To answer your question @hokei, links.

helenvholmes (approved):
LGTM.

src/react-components/src/components/Icons/IconTypes.tsx
```diff
@@ -0,0 +1,6 @@
+export enum iconRotationOptions {
```
hokei, line 1: Don't really know where to put this since it's merged but uhhh 
1) I think, since it's `Types`, let's call this enum `iconRotationTypes` to differentiate it from the `-Options` objects (Which means something different in the Design System context) 
2) Since this is something consumers need access to, remember to export it in `index.ts`

PR #180, opened 2020-05-27
add image caption and image credit to Image

### LINKED ISSUE #116: Image should take optional credit + caption line
`Image` should take an optional credit and caption line that are visible to sighted users.

## Ticket tasks
- Add optional credit line
- Add optional caption line
- Update `Image` props to have descriptions for Storybook
- Update `Image` in the file structure to be at the top level

designs

M0nica:
- Adds optional imageCredit and imageCaption to `Image`
- Wraps `Image` in `figure` when imageCredit and imageCaption are passed

### Front End Review:
- [ ] View [Image in Storybook](http://localhost:9001/?path=/story/image--image-with-optional-captionand-credit)
- [ ] View [Hero in Storybook](http://localhost:9001/?path=/story/hero--hero-primary)

### DIFF
--- a/src/components/Image/Image.stories.tsx
+++ b/src/components/Image/Image.stories.tsx
@@ -8,7 +8,7 @@ export default {
     component: Image,
 };
 
-export const ImageWithCaption = () => [
+export const ImageWithOptionalCaptionandCredit = () => [
     <Image
         src="https://placeimg.com/400/200/arch"
         isDecorative={true}
--- a/src/components/Image/Image.stories.tsx
+++ b/src/components/Image/Image.stories.tsx
@@ -2,45 +2,59 @@ import * as React from "react";
 
 import Image from "./Image";
 import { action } from "@storybook/addon-actions";
+import { text, boolean, select } from "@storybook/addon-knobs";
 
 export default {
     title: "Image",
     component: Image,
 };
 
-export const ImageWithOptionalCaptionandCredit = () => [
-    <Image
-        src="https://placeimg.com/400/200/arch"
-        isDecorative={true}
-        imageCaption={"Image Caption"}
-    />,
-];
+let showCaption, showCredit, circleMask;
 
-export const ImageWithCredit = () => [
-    <Image
-        src="https://placeimg.com/400/200/arch"
-        isDecorative={true}
-        imageCaption={"Image Credit"}
-    />,
-];
+const imageRatios = {
+    "1x1": "https://placeimg.com/100/100/animals",
+    "2x1": "https://placeimg.com/200/100/animals",
+    "3x4": "https://placeimg.com/150/200/animals",
+    "4x1": "https://placeimg.com/400/100/animals",
+    "4x3": "https://placeimg.com/200/150/animals",
+    "16x9": "https://placeimg.com/400/225/animals",
+};
 
-export const ImageWithCreditAndCaption = () => [
-    <Image
-        src="https://placeimg.com/400/200/arch"
-        isDecorative={true}
-        imageCaption={"Image Caption"}
-        imageCredit={"Image Credit"}
-    />,
-];
+export const ImageWithOptionalCaptionandCredit = () => (
+    <>
+        {boolean("Show Caption", true)
+            ? (showCaption = true)
+            : (showCaption = false)}
+        {boolean("Show Credit", true)
+            ? (showCredit = true)
+            : (showCredit = false)}
+        <Image
+            src={select(
+                "Initial Selected Option",
+                imageRatios,
+                imageRatios["2x1"]
+            )}
+            isDecorative={true}
+            imageCaption={
+                showCaption
+                    ? text("Image Caption", "Deserted Islander Relocator")
+                    : null
+            }
+            imageCredit={
+                showCredit ? text("Image Credit", "Taken by Tom Nook") : null
+            }
+            modifiers={boolean("Apply Circle Mask", false) ? ["circle"] : null}
+        />
+    </>
+);
 
-export const twoByOne = () => [
-    <Image src="https://placeimg.com/200/100/arch" isDecorative={true} />,
-    <Image src="https://placeimg.com/400/200/arch" isDecorative={true} />,
-    <Image src="https://placeimg.com/1600/800/arch" isDecorative={true} />,
-    <Image src="https://placeimg.com/2000/1000/arch" isDecorative={true} />,
-];
-
-export const threeByFour = () => [
-    <Image src="https://placeimg.com/150/200/arch" isDecorative={true} />,
-    <Image src="https://placeimg.com/300/400/arch" isDecorative={true} />,
-];
+ImageWithOptionalCaptionandCredit.story = {
+    name: "Image with Optional Caption and Credit",
+    parameters: {
+        design: {
+            type: "figma",
+            url:
+                "https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Master?node-id=11896%3A45379",
+        },
+    },
+};
--- a/src/components/Image/Image.stories.tsx
+++ b/src/components/Image/Image.stories.tsx
@@ -8,6 +8,31 @@ export default {
     component: Image,
 };
 
+export const ImageWithCaption = () => [
+    <Image
+        src="https://placeimg.com/400/200/arch"
+        isDecorative={true}
+        imageCaption={"Image Caption"}
+    />,
+];
+
+export const ImageWithCredit = () => [
+    <Image
+        src="https://placeimg.com/400/200/arch"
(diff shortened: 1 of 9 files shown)


src/styles/01-atoms/_images/image/_image.scss
```diff
@@ -1,18 +1,30 @@
...
-  @include image;
+    @include image;
 }
 
 figure {
-  margin: 0; //override normalize
+    margin: 0; //override normalize
 }
 
 .figure__caption {
```
M0nica: It seems like this `.figure__caption` was unused.

src/react-components/src/components/Image/Image.tsx
```diff
@@ -34,5 +38,23 @@ export default function Image(props: ImageProps) {
...
     };
 
-    return <img {...imageProps} />;
+    return (
+        <>
+            <img {...imageProps} />
+            {(imageCaption || imageCredit) && (
+                <div className={bem(image_base_class, [], "annotation")}>
+                    {imageCaption && (
+                        <p className={bem(image_base_class, [], "caption")}>
```
M0nica: Do these paragraphs need `bem()` classes even if they're being selected based on their position in the annotation div?

src/styles/01-atoms/_images/image/_image.scss
helenvholmes: A lot of the CSS was pulled over from another project that was started before this one that uses a different tech stack and has different components. Once we're done with our Story clean-up milestone we should talk about removing unused CSS (I think one thing that will make that task a lot easier is getting this .scss files with their relevant component, which we want to tackle first).

src/react-components/src/components/Image/Image.tsx
helenvholmes: Giving all three (wrapper, child, second child) their own BEM classes allows at the integration level for a project to decide to style them differently if they so choose with a clear hook to do so. I think what you've done here makes sense.

src/react-components/src/components/Image/Image.tsx
```diff
@@ -34,5 +38,23 @@ export default function Image(props: ImageProps) {
...
+                            {imageCaption}
+                        </p>
+                    )}
+                    {imageCredit && (
+                        <p className={bem(image_base_class, [], "heading")}>
```
helenvholmes: I would use `credit` as a blockName over `heading` for this.

helenvholmes (changes requested):
I did a bad job writing the original ticket. There was some documentation I didn't link to I should have passed along:

[Image with Caption and Credit](http://themetronome.co/components/ai-177-image-unrestricted-with-caption-and-credit-line/?fresh=true)

I think that will explain better some of the SCSS you're already seeing in `_image.scss`.

M0nica:
> I did a bad job writing the original ticket. There was some documentation I didn't link to I should have passed along:
> 
> [Image with Caption and Credit](http://themetronome.co/components/ai-177-image-unrestricted-with-caption-and-credit-line/?fresh=true)
> 
> I think that will explain better some of the SCSS you're already seeing in `_image.scss`.

Thanks for sharing this link. It looks like credit needs to be prepended with "credit: ". In metronome the styles are not italicized. In the design system do we still want the text to be italicized?

src/react-components/src/components/Image/Image.tsx
M0nica: I renamed this blockName to `credit`.

helenvholmes:
@M0nica Let's not prepend credit with "Credit:"—that's not going to be universally desired.

src/react-components/src/components/Image/Image.tsx
```diff
@@ -34,5 +38,23 @@ export default function Image(props: ImageProps) {
...
+            {(imageCaption || imageCredit) && (
+                <div className={bem(image_base_class, [], "annotation")}>
+                    {imageCaption && (
+                        <p className={bem(image_base_class, [], "caption")}>
+                            {imageCaption}
+                        </p>
+                    )}
+                    {imageCredit && (
+                        <p className={bem(image_base_class, [], "credit")}>
+                            Credit: {imageCredit}
```
helenvholmes: See comment in the Conversation tab

src/react-components/src/components/Image/Image.tsx
```diff
@@ -34,5 +38,23 @@ export default function Image(props: ImageProps) {
         alt: isDecorative ? "" : altText,
     };
 
-    return <img {...imageProps} />;
+    return (
+        <>
+            <img {...imageProps} />
+            {(imageCaption || imageCredit) && (
+                <div className={bem(image_base_class, [], "annotation")}>
```
helenvholmes: In the event that this has an image caption and/or credit, instead of a div this should become a `figcaption` and the entire component should be wrapped in a `figure`

src/styles/01-atoms/_images/image/_image.scss
```diff
@@ -1,18 +1,30 @@
...
-  font-style: italic;
+    font-style: italic;
+}
+
+.annotation__image {
```
helenvholmes: Can we use the `.figcaption` class above instead of the `.annotation__image` class you've added? It'll better describe the element we're styling once you've made the requested changes in `Image.tsx`

M0nica:
> @M0nica Let's not prepend credit with "Credit:"—that's not going to be universally desired.

Okay, aside from the credit thing are there additional changes that should be made to the current PR? I tried to update to reflect the metrolink but am still not clear on what was decided re: italics

src/styles/01-atoms/_images/image/_image.scss
```diff
@@ -1,18 +1,30 @@
...
+    p {
+        font-style: italic;
+    }
+    p:nth-child(odd) {
```
helenvholmes: These overrides are a little odd. I'd recommend going forward with a `span` or `div` for the credit and caption instead, and then you won't have to override the `p` styling.

Also as an FYI we have helpers to restrict spacing to a strict taxonomy. You can find all of those defined in the "space/" folder under `styles/`.

`@include space-stack-none`

helenvholmes (changes requested):

helenvholmes:
@M0nica let's move forward with italics to style these.

src/react-components/src/components/Image/Image.tsx
M0nica: Thanks for providing that context is helpful! I will definitely err towards including BEM classes.

src/react-components/src/components/Image/Image.tsx
M0nica: I updated to use `figure` and `figcaption`. I _think_  up to two `figcaptions` can be included in a `figure` as [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figcaption) says "<figcaption> element must be its first or last child.". Is that the structure you had in mind?

src/styles/01-atoms/_images/image/_image.scss
M0nica: Thanks for pointing out the spacing variables. I have updated to use `@include space-stack-none`

src/react-components/src/components/Image/Image.tsx
helenvholmes: What we did on the other project was wrap the two divs in the figcaption, so you ended up with something like this:

```
figure
  figcaption
    div1, caption
    div2, credit
```

helenvholmes:
Hey @M0nica, are you still actively working on this PR?

src/react-components/src/components/Image/Image.tsx
```diff
@@ -34,5 +38,31 @@ export default function Image(props: ImageProps) {
         alt: isDecorative ? "" : altText,
     };
 
-    return <img {...imageProps} />;
+    return (
+        <>
+            <img {...imageProps} />
```
helenvholmes: The image would need to be inside the `figure` if there are imageCaption or imageCredit, but no `figure` if there aren't.

helenvholmes (changes requested):
Really close!

src/components/Image/Image.stories.tsx
```diff
@@ -2,20 +2,59 @@ import * as React from "react";
...
+                imageRatios["2x1"]
+            )}
+            isDecorative={true}
+            imageCaption={
+                showCaption
+                    ? text("Image Caption", "Deserted Islander Relocator")
+                    : null
+            }
+            imageCredit={
+                showCredit ? text("Image Credit", "Taken by Tom Nook") : null
```
hokei, line 44: 😆

hokei:
I am loving these knobs!

hokei (approved):

helenvholmes (approved):

PR #182, opened 2020-05-27
Combines Various Link Components

### LINKED ISSUE #122: Links should have LinkTypes just like Buttons do
### Description of issue:
We have UnderlineLink, IconLink, and Links that look like buttons (Edition Card).  We should apply treatment the way that we do to Buttons, where we use "ButtonTypes".

helenvholmes:
Closes #122.

**This PR does the following:**
- Combines `IconLink`, `UnderlineLink`, and `BasicLink` into `Link`
- Updates test suite
- Creates enum for link types (default, button, and "action")

### Front End Review:
- [ ] View [link examples in Storybook](http://localhost:9001/?path=/story/link--passed-in-link)
- [ ] View other places that use link:
  - [ ] [Section Title](http://localhost:9001/?path=/story/section-title--heading-no-link)
  - [ ] [Edition Card](http://localhost:9001/?path=/story/edition-card--edition-card-with-full-data)
  - [ ] [Search Result Item](http://localhost:9001/?path=/story/search-result-item--search-result-item)
  - [ ] [Search Result List](http://localhost:9001/?path=/story/unorderedlist--search-results-list)


src/react-components/src/components/Link/Link.tsx
```diff
@@ -1,11 +1,30 @@
-// Base Link Helper Class
 import * as React from "react";
 import { uid } from "react-uid";
+import { LinkTypes, LinkIconPositions } from "./LinkTypes";
 
 export interface LinkProps {
```
hokei, line 8: since you're working on it, this is probably a good time to add `LinkOptions` to be exported and not export `LinkProps`

helenvholmes: Thanks for catching that. I'll review all your docs and maybe hit you up with some questions, I only glanced at everything yesterday.

src/react-components/icons/svg/arrow-xsmall.svg
```diff
@@ -1 +1,3 @@
-<svg width="14" height="9" xmlns="http://www.w3.org/2000/svg"><path fill-rule="evenodd" clip-rule="evenodd" d="M5.984 7.578L.424 1.514 1.9.163l5.22 5.695L12.341.163l1.475 1.351-5.56 6.064a1.542 1.542 0 01-2.272 0zm1.474-1.351z"/></svg>
\ No newline at end of file
+<svg width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg">
```
hokei: FYI: The viewbox additions will block RN's release.  However since we don't release on Fridays, the earliest we'll need it is Monday.  I might end up extracting them to their own ticket if it becomes the last thing that's blocking, we can keep in touch about this on Slack

hokei:
Just a reminder that this is a breaking change and should be marked as such!

src/react-components/src/components/Link/Link.tsx
```diff
@@ -0,0 +1,122 @@
...
+            if (iconPosition === LinkIconPositions.Left) {
+                iconRotation = iconRotationTypes.rotate90;
+            } else if (iconPosition === LinkIconPositions.Right) {
+                iconRotation = iconRotationTypes.rotate270;
+            }
+        }
+
+        iconProps = {
+            name: "arrow_xsmall",
+            modifiers: [iconPosition, iconRotation],
```
hokei: can modifiers never be passed to icon?

src/react-components/src/components/Link/Link.tsx
```diff
@@ -0,0 +1,122 @@
...
+            blockName: "more-link",
```
hokei: there's gotta be a better name for this

src/react-components/src/components/Link/Link.tsx
```diff
@@ -0,0 +1,122 @@
...
+            decorative: "true",
+        };
+
+        if (iconPosition) {
+            if (iconPosition === LinkIconPositions.Left) {
```
hokei: this conditional doesn't seem like it's doing anything

src/react-components/src/components/Link/Link.tsx
```diff
@@ -0,0 +1,122 @@
...
+    modifiers?: string[];
+
+    /** BEM specifiers */
+    blockName?: string;
+
+    /** Additional attributes, such as rel=nofollow, to pass to the <a> tag */
+    attributes?: {};
+}
+
+export default class Link extends React.Component<LinkProps, {}> {
```
hokei: since there's no state change I think you can make this into a functional component

src/react-components/src/components/Link/Link.tsx
helenvholmes: It probably be "action-link" imo, but this chunk of work was taking long enough as it was—I would prefer a follow-up for renaming this

src/react-components/src/components/Link/Link.tsx
helenvholmes: Not for this action-link "< link" "link >" lock-up, not currently... I can update that though

hokei: I think it's more do you think people can pass custom modifiers to icon within link or is it only going to be both `iconPosition` and `iconRotation` ? (Also, what happens if you pass one but not the other?)

helenvholmes: I think I want to err on the side of extensibility. But to answer the question, probably only ever `iconPosition` and `iconRotation`. 

I'm starting to think that I'm conflating two separate linkTypes together, though, which is why this is confusing... maybe splitting these apart into `actionLink`, `backLink`, and `forwardLink` would be less confusing. Back and forward would create these iconBlocks without the consumer putting in any extra effort and `action` would just display the icon as it was received.

src/react-components/src/__tests__/components/Link-test.tsx
```diff
@@ -0,0 +1,98 @@
...
+        expect(
+            wrapper.find(".more-link__icon").hasClass("more-link__icon--left")
+        ).to.equal(true);
+    });
+
+    it("Can pass in right icon", () => {
+        wrapper = Enzyme.mount(
+            <Link href="#whatever">
+                content
+                <Icon
```
hokei: So this (and all the stories) are wrapping the icon rather than passing in an icon (using `iconPosition` and `iconRotation`).  Should `iconPosition` and `iconRotation` be used at all in `Link` or should it always be passed in like this?

helenvholmes: Since I've added the other linkTypes I think the consumer shouldn't have to pass these in to Link, so I've updated this.

src/react-components/src/components/Link/Link.tsx
```diff
@@ -0,0 +1,95 @@
...
+            } else if (child.type === "a") {
+                return child.props.children;
+            } else {
+                return child;
+            }
+        }
+    );
+
+    if (
+        props.linkType === LinkTypes.Action ||
```
hokei: Is there a fallback for if linkType doesn't get passed? (You can add it as `linkType={LinkTypes.Action}` in L27

src/react-components/src/components/Link/Link.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+    /** BEM specifiers */
+    blockName?: string;
+
+    /** Additional attributes, such as rel=nofollow, to pass to the <a> tag */
+    attributes?: {};
+}
+
+export default function Link(props: React.PropsWithChildren<LinkProps>) {
+    let { href, modifiers, blockName, attributes } = props;
```
hokei: Should these be `const`s?

hokei: shouldn't `tslint` have caught that?

src/react-components/src/components/Link/Link.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+    navigationIconProps = {
+        name: "arrow_xsmall",
+        modifiers: [iconPosition, iconRotation, modifiers],
+        blockName: "more-link",
+        decorative: "true",
+    };
+
+    if (props.linkType === LinkTypes.Backwards) {
+        iconLeft = <Icon {...navigationIconProps} />;
```
hokei, line 77: I'm kind of confused as to how `<Icon>` works in `Link`.  Is it generated or is it passed in?  It looks from the stories that `<Icon`> is passed in `props.children`.  does `iconLeft` or `iconRight` ever get set?

src/react-components/src/components/Link/Link.tsx
```diff
@@ -0,0 +1,95 @@
...
+    } else if (props.linkType === LinkTypes.Forwards) {
+        iconRight = <Icon {...navigationIconProps} />;
+    }
+
+    elementChildren.map((child) => {
+        return React.cloneElement(child, { key: uid(child) });
+    });
+
+    let className = bem(link_base_class, props.modifiers, props.blockName);
```
hokei: you set these in L27 so you don't have to  call `props` here.

src/react-components/src/components/Link/Link.stories.tsx
```diff
@@ -0,0 +1,54 @@
+import * as React from "react";
+
+import Link from "./Link";
+import { LinkTypes, LinkIconPositions } from "./LinkTypes";
```
hokei: LinkIconPositions doesn't seem to be used in this file

src/react-components/src/__tests__/components/Link-test.tsx
```diff
@@ -0,0 +1,90 @@
+import { expect } from "chai";
+import { stub } from "sinon";
+import * as Enzyme from "enzyme";
+import * as React from "react";
+import * as Mocha from "mocha";
+
+import Link from "../../components/Link/Link";
+import { LinkTypes, LinkIconPositions } from "../../components/Link/LinkTypes";
```
hokei: LinkIconPositions isn't used in this file

src/react-components/src/components/Link/LinkTypes.tsx
```diff
@@ -0,0 +1,12 @@
+export enum LinkTypes {
+    Action = "action",
+    Backwards = "backwards",
+    Button = "button",
+    Default = "default",
+    Forwards = "forwards",
+}
+
+export enum LinkIconPositions {
```
hokei: considering that `LinkIconPositions` is only used internally as a constant in `Link` and never passed in, I think it doesn't need to be here --   It can just be assigned as string - in `Link.tsxL63` and `L66` There's no need to do validation against a set of allowed values because it's just an internal variable used internally.

src/react-components/src/__tests__/components/Link-test.tsx
```diff
@@ -0,0 +1,90 @@
...
+            <Link href="#passed-in-link" linkType={LinkTypes.Backwards}>
+                content
+            </Link>
+        );
+        expect(
+            wrapper.find(".more-link__icon").hasClass("more-link__icon--left")
+        ).to.equal(true);
+    });
+
+    it("Can pass in right icon", () => {
```
hokei: An icon isn't being passed here, I think this test is better called `LinkType Forwards generates right icon`

hokei (changes requested):
Icon handling in this PR continues to confuse me ... 
It seems like we should decide either to allow icons to be passed in, like 

```
                <Link href="#passed-in-link">
                    <Icon
                        name="download"
                        blockName="more-link"
                        modifiers={["left"]}
                        decorative={true}
                    ></Icon>
                    Download
                </Link>
``` 

Or if icons should be generated, like 

```
            <Link href="#passed-in-link" linkType={LinkTypes.Forwards}>
                content
            </Link>
```


Right now, the design system has stories in the first format but tests in the second, and it's confusing.  The second option means the design system has all the control - but it means that if a consumer wants to switch out the icon, they'll have to hack something together.  That being said, I don't think we can forbid the first option, but I think if we want to keep good control over the way links look in the design system (aka a consumer cannot swap out an icon at will, etc) we should change all the stories to be using the second format.  (Maybe have one story where the link is passed in, to say that it's possible). 

Alternately, we can eliminate the second way altogether and ask the consumer to know which icon they want to pass in every time.  I /don't/ think that's what we want, but it's an option ...

helenvholmes:
@hokei so we should have stories for both methods, although not having tests for both is an issue—that can be updated if they're not there.

The first method (see below) is only intended for "see more >" "< go back"-esque links so the consumer doesn't have to futz with the rotation and icon selection. I think it's easier to do all that for the consumer.

```
<Link href="#passed-in-link" linkType={LinkTypes.Forwards}>
  content
</Link>
```

We could set this up with options so you can set individual properties on a Link's icon(s) for other options, but I think it's limiting and makes Link more complicated than it needs be, since you could just do it this way:

```
<Link href="#passed-in-link">
  <Icon
    name="download"
    blockName="more-link"
    modifiers={["left"]}
    decorative={true}
  ></Icon>
  Download
</Link>
```

You will definitely need consumers to pass in icons other than just < and >, because edition cards are using the download icon, but extending the "more-link" styles.

I don't think having both options is bad (I intended `LinkType.Forward` + `LinkType.Backward` as helpers), so long as it doesn't feel confusing to the consumer how to use them. Would it be helpful to Zoom about a strategy going foward?

hokei:
aaah now that you've explained, I understand it a lot better, thank you!  This seems like a good strategy to me. 
I think something like this has made it clear how much we need the documentation plugin - having this explanation on the Links story page would be incredibly useful.  
So to reiterate, in general we want people to pass in the icons, but for the arrows, since they are used so often, `LinkType.Forward` and `LinkType.Backward` exist to be helpers?

src/react-components/src/__tests__/components/Link-test.tsx
hokei: There should be a different passed in link test.

helenvholmes:
@hokei yep, that understanding is correct.

src/react-components/src/components/Link/Link.tsx
helenvholmes: `tslint` didn't, but I'll update

src/react-components/src/__tests__/components/Link-test.tsx
helenvholmes: A different passed in link test?

src/react-components/src/__tests__/components/Link-test.tsx
```diff
@@ -0,0 +1,90 @@
...
+                    decorative={true}
+                    iconRotation={iconRotationTypes.rotate0}
+                ></Icon>
+            </Link>
+        );
+        expect(wrapper.find("a")).to.have.lengthOf(1);
+        expect(wrapper.find("Icon")).to.have.lengthOf(1);
+    });
+
+    it("Can generate a back link", () => {
```
hokei: I'd call this `Generated back link has icon` since it's explicitly checking for the icon class. (same with below)

src/react-components/src/components/Link/Link.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+import Icon from "../Icons/Icon";
+import { iconRotationTypes } from "../Icons/IconTypes";
+
+export default {
+    title: "Link",
+    component: Link,
+};
+
+export const passedInLink = () => (
+    <Link linkType={LinkTypes.Default}>
```
hokei, line 14: This doesn't have an underline, is that intentional? The rest of the links have underlines

hokei (changes requested):
getting really close! Just some last small things

src/react-components/src/components/Link/Link.stories.tsx
helenvholmes: I haven't been able to figure this out at all!!... something is applying `text-decoration: none`, but where, I am not clear on... it's definitely not intentional

helenvholmes: Ah, it's not properly applying the `href` prop anymore, that's what's up

src/react-components/src/components/Link/Link.tsx
```diff
@@ -0,0 +1,108 @@
...
+export default function Link(props: React.PropsWithChildren<LinkProps>) {
+    const {
+        href,
+        modifiers,
+        blockName,
+        attributes,
+        linkType = LinkTypes.Default,
+    } = props;
+
+    const passedProps = { href, attributes };
```
helenvholmes: @hokei There is probably a better way to do this, but I was getting an error in the console for passing extra stuff like "props.linkType" into the DOM element, so this was how I got around it 😬

src/react-components/src/components/Link/Link.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+    let icon;
+
+    let elementChildren = React.Children.map(
+        props.children,
+        (child: React.ReactElement) => {
+            if (typeof child === "string" && !props.href) {
+                throw new Error("Link needs prop 'href'");
+            } else if (child.type === "a") {
+                childProps = child.props;
```
helenvholmes, line 44: This is also why that one link didn't have an underline... wasn't getting its `href`

src/react-components/src/__tests__/components/Link-test.tsx
hokei: Hmm ... actually, I think that's not necessary since the functionality is shown in stories and it's really only testing react.

src/react-components/src/components/Link/Link.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+    const passedProps = { href, attributes };
+    passedProps.href = props.href;
+    passedProps.attributes = props.attributes;
+
+    let link_base_class = "link";
+
+    let childProps = {};
+
+    let icon;
```
hokei: this is not used

src/react-components/src/components/Link/Link.tsx
hokei: This is probably because the word `props` is reserved.  If you change `props` to `linkProps` the warning goes away and you can delete L34-37

src/react-components/src/components/Link/Link.tsx
```diff
@@ -0,0 +1,108 @@
...
+                childProps = child.props;
+                return child.props.children;
+            } else {
+                return child;
+            }
+        }
+    );
+
+    if (
+        props.linkType === LinkTypes.Action ||
```
hokei: You should not need to call `props` within the body of the function at all, because this variable is set on L32.  
(This isn't as subtle as it seems, forcing the `const` on it so we always know whether or not it's been changed is one of the great benefits of TS)

src/react-components/src/components/Link/Link.tsx
```diff
@@ -0,0 +1,108 @@
+import * as React from "react";
+import bem from "../../utils/bem";
+import { uid } from "react-uid";
+import { LinkTypes } from "./LinkTypes";
+import { element } from "prop-types";
```
hokei: this is not used

src/react-components/src/components/Icons/Icon.tsx
```diff
@@ -1,6 +1,8 @@
 import * as React from "react";
 import bem from "../../utils/bem";
 
+import { iconRotationTypes } from "./IconTypes";
```
hokei: Nit: Capitalize the I so it matches the other Types convention

src/react-components/src/index.ts
```diff
@@ -17,15 +16,14 @@ export { default as Heading } from "./components/Heading/Heading";
 export { default as Hero } from "./components/Hero/Hero";
 export { HeroTypes } from "./components/Hero/HeroTypes";
 export { iconRotationTypes } from "./components/Icons/IconTypes";
-export { default as IconLink } from "./components/IconLink/IconLink";
 export { default as Image } from "./components/Image/Image";
 export { default as Input } from "./components/Input/Input";
 export { default as Label } from "./components/Label/Label";
+export { default as Link } from "./components/Link/Link";
```
hokei, line 22: export `LinkTypes` as well

src/react-components/src/components/Link/Link.tsx
helenvholmes: I sorta figured once it worked in other places... thanks for catching my messes 😭

src/react-components/src/components/Link/Link.tsx
helenvholmes: TIL!

helenvholmes:
@hokei you are truly a saint Crystal thank you for the 32nd review 🙏🏻

hokei (approved):
🎉  🎉 
This was pretty gnarly work, will be happy to get it in!  Thank you!

Issue #206, opened 2020-06-02
Create a more generic Card component


helenvholmes:
- optional 2 columns or 3
- optional top content
- probably optional bottom content
- all of that is generic jsx

kristojorg:
One question here: what would the card itself enforce if all of these are optional? Spacing between the columns? Would it be more useful to simply have a `Card` component that does the border and maybe margins, and then takes whatever children you want to give it?

hokei:
I think there's a strong case that `card` should be a wrapper, like `Unordered List` and `Modal` are ... 
Depending on how they're used though, it might be good for `card` to have helpers, such as `two-column`, `three-column`, etc 

I think the best way to move forward here could be to create a `Card` component that is just a wrapper, and refactor `EditionCard` to use the `Card component`.  How to remove `EditionCard` from design system is a thing that would come later, when we have a better sense of which patterns are very often used (and therefore should belong here) and which patterns are one-offs.

helenvholmes:
i think it makes sense to keep some three column catalog item component from the NYPL perspective—while the styling has been different every time, the component seems to be cropping up over and over. With a generic card that could probably just be captured as stories.

helenvholmes:
@kristojorg, is it okay if we assign this to you?

kristojorg:
That makes sense, but would you rather then make a `Card` component and then some kind of `Columns` component that takes `Column` children and applies the right spacing? EG:

```
<Card>
  <Columns >
    <Column size="1">content</Column>
    <Column size="3">content</Column>
  </Columns>
</Card>
```

kristojorg:
Yea I can contribute the `Card` component at least, but not sure at what point in my development I'll get to it

hokei:
I think it might be confusing to have `Column` be a separate component, since it'll only be used in `Card`.  It would imply that `Card` has an internal grid system which might be true, but also might get pretty complicated.  I expect NYPL will only have one or two commonly used card layouts, and importantly, I don't think that we'd have different two-column cards that used differing internal column widths.  

I think it might be ok for Card to take prop `content` of an array of JSX elements.  If you put in one JSX element, it'll fill up the whole thing, if you put in two or three, it'll split it accordingly, and if you want to put in more than that, or do anything fancy, you can do it yourself by wrapping everything into one JSX element.  

`<Card content=[<JSX.Element>, <JSX.Element>] />`

kristojorg:
[...]

PR #207, opened 2020-06-02
Addresses #205: Changes array for EditionInfo to JSX.Element

### LINKED ISSUE #205: Update EditionCard to take JSX.Element > array of strings
### Component in question:
- EditionCard

### Description of issue:
EditionCard currently takes an array of strings to generate the middle items of the card. Taking a JSX element would allow for this card to be more versatile.

helenvholmes:
- Changes EditionInfo inside EditionCard to accept a JSX.Element instead of an array

### Front End Review:
- [ ] View the [EditionCard stories](http://localhost:9001/?path=/story/edition-card--edition-card-with-full-data)

### DIFF
--- a/src/react-components/src/__tests__/components/EditionCard-test.tsx
+++ b/src/react-components/src/__tests__/components/EditionCard-test.tsx
@@ -15,11 +15,13 @@ describe("EditionCard", () => {
             blockName=""
             coverUrl="https://placeimg.com/300/400/arch"
             editionHeadingElement={<a href="edition-link">2004 Edition</a>}
-            editionInfo={[
-                <>Published in New York by Random House</>,
-                <>Written in English</>,
-                <>Under Creative Commons License</>,
-            ]}
+            editionInfo={
+                <>
+                    <p>Published in New York by Random House</p>
+                    <p>Written in English</p>
+                    <p>Under Creative Commons License</p>
+                </>
+            }
             readOnlineLink={
                 <Link
                     blockName="edition-card"
@@ -49,13 +51,16 @@ describe("EditionCard", () => {
             blockName=""
             coverUrl="https://placeimg.com/300/400/arch"
             editionHeadingElement={<a href="edition-link">2004 Edition</a>}
-            editionInfo={[
-                <span id="pub-span">Publisher</span>,
-                <>"Written in English"</>,
-                <a id="licenceId" href="/licence-page">
-                    License
-                </a>,
-            ]}
+            editionInfo={
+                <>
+                    <span id="pub-span">Publisher</span>,
+                    <p>"Written in English"</p>,
+                    <a id="licenceId" href="/licence-page">
+                        License
+                    </a>
+                    ,
+                </>
+            }
             readOnlineLink={
                 <Link
                     blockName="edition-card"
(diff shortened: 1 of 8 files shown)


hokei:
A note:  This is another breaking change that would require a version bump.  Right now this would be merged around the same time as `Link` so I can make all of those changes in DRB at once, but we should be careful going forward on how to do breaking changes thoughtfully.

src/react-components/src/components/EditionCard/EditionCard.tsx
```diff
@@ -23,7 +23,7 @@ export interface EditionCardProps {
     coverUrl: string;
 
     editionHeadingElement: JSX.Element;
```
hokei: For this to help Kristo's work, `editionHeadingElement` has to be made optional

hokei:
I'd put a story up for a header-less edition card and have it also be an example of passing in something else (maybe an icon? may a link?) into the body of the card.  

edit: actually, we probably shouldn't because it's not explicitly ok by the design system.

src/react-components/src/__tests__/components/EditionCard-test.tsx
```diff
@@ -70,11 +75,13 @@ describe("EditionCard", () => {
...
             coverUrl="https://placeimg.com/300/400/arch"
             editionHeadingElement={<a href="edition-link">2004 Edition</a>}
-            editionInfo={[
-                <>Published in New York by Random House</>,
-                <>Written in English</>,
-                <>Under Creative Commons License</>,
-            ]}
+            editionInfo={
+                <>
+                    <p>Published in New York by Random House</p>,
```
hokei, line 80: is `<p>` the right tag to use here?  Maybe they should just be spans.

hokei (changes requested):
see comments inline

edit:  Just the changelog and optional editionHeadingElement thing

helenvholmes:
@hokei I know when it was just us there was a desire to get through as many breaking changes as soon as possible after our conversation with Kristo—is that no longer the case?

hokei:
@scooooooooby  There was no changelog update here so I just wanted to be on the same page re: yes, this is a breaking change, and yes, we are making this breaking change intentionally and have thought about the work this generates for SFR.  

(if you update the changelog as a part of PR creation, I think that would make it clear and I can stop nagging about it every time haha)

helenvholmes:
@hokei yes! And sorry that you have been having to do that, it has been in the backlog to update Travis to say something when I inevitably forget over and over 😥

hokei:
haha no worries, Imma just nag over and over.  (This can be released as 0.4.0 since there's not going to be a release between the link stuff and when this gets in - my hope is that 0.4.0 can be the no-Lerna release but that depends on how SFR deploy goes )

helenvholmes:
@hokei, Kristo left a comment in #206 I've also been considering, but would be a more involved update than what I've done here. Do you think I should address #206 with this work now or bundle this breaking change with the Link breaking change with a potential future breaking change?

hokei:
I think the right way to go is to create a `Card` component - which, sadly, makes this work moot.  This wouldn't represent a breaking change for `EditionCard` but when we remove `EditionCard` that would get bundled with other breaking changes

helenvholmes:
@hokei I think you're right. I'm going to close this out and update #206 to capture that work.

PR #212, opened 2020-06-04
Styles

### LINKED ISSUE #187: Integrate styles with their corresponding components

helenvholmes:
- Integrates relevant SCSS files with their component folders
- Moves the base styles into a `styles/` folder within `src/react-components/`


storybook/.storybook-react/config.js
```diff
@@ -1,8 +1,8 @@
-import '!style-loader!css-loader!sass-loader!import-glob-loader!@nypl/design-system-styles/style.scss';
-import { configure, addDecorator } from '@storybook/react';
-import { withInfo } from '@storybook/addon-info';
-import { withKnobs } from '@storybook/addon-knobs';
-import { withA11y } from '@storybook/addon-a11y';
+import "!style-loader!css-loader!sass-loader!import-glob-loader!@nypl/design-system-react-components/style.scss";
```
hokei, line 1: import-glob-loader is not needed here anymore, right?

hokei:
is it okay to remove import-glob-loader from package json?  
Also, changelog.  

Were any style names changed?  I think RN overrides specific styles, so I'll want to know to be able to make changes over there.

storybook/.storybook-react/config.js
helenvholmes: We're still globbing the base styles, but I can remove this as a dependency by independently declaring those as well.

helenvholmes:
@hokei Updated CHANGELOG.

No classNames have been changed, no. Sounds like if we want to do that it will constitute as a breaking change for DRB, so let's plan those separately when we identify them.

We're still glob-importing the base styles, but if you'd like to have the same strategy for all of `style.scss` I can work toward removing it in this PR.

hokei:
remove `design-system-styles` from `storybook` dependencies

hokei (approved):

storybook/.storybook-react/config.js
hokei: I forgot about base styles - I think globbing for base styles is fine then, since it all lives in the same place.

helenvholmes: I forgot too until you mentioned the dependency, hah

PR #213, opened 2020-06-04
Remove Lerna, use TSDX

kristojorg:
Hi all,

This is a PR to generally make this package easier to develop and consume. It does seemingly a lot, but the changes to the actual components are pretty minimal. Here is a list of what I've done: 
1. Remove Lerna, flatten the package directory. The assumption now is that this repo will only be managing a single npm package, and styles will be bundled with the react components.
1. Uses [txds](https://github.com/jaredpalmer/tsdx) to abstract much of the build configuration. It has a lot of lovely features and means we don't have to worry much about managing build config anymore. It's the officially recommended package by Typescript for TS packages.
1. Replaces tslint with eslint. Tslint was deprecated at the beginning of the year in favor of eslint with a typescript parser. I recently did this in `circulation-patron-web` as well, and have copied over the eslint config from there. It's not perfect, but includes `jsx-a11y` rules as well as the react and typescript presets. Finally, eslint is now in charge of code formatting as well, via the `prettier` plugin. The prettier rules are in package.json. I didn't alter them from before.
1. Replace Mocha with [Jest](https://jestjs.io/). This is the test runner used by TSDX, and it's generally faster and more modern than Mocha. It is also an assertion library and provides mocking functionality out of the box, which allowed me to remove `chai` and `sinon`. I did have to change some test syntax, but it is almost the same API, just slight differences. 
1. TSDX uses [Rollup](https://rollupjs.org/) under the hood instead of webpack, which is generally simpler and has good defaults. It is popular for bundling packages (as opposed to applications). In any case, we shouldn't need to interact with it much since TSDX configures it for us. 
1. As part of migrating to eslint, I did need to fix many eslint errors, especially removing unused variables and escaping html entities (like apostrophes). 
1. I removed `tabIndex={0}` on the modal component since it is not an interactive element so that won't have any effect (eslint informed me of this).
1. I removed many class constructors that simply called `super(props)`, as that is automatically done if not constructor is provided, so they weren't needed (another thing react/eslint complained about).
1. I removed the js files inside of `src/styles` and also the slick css. Let me know if this is still being used and I'll add it back.
1. I began updating the readme, but it's not quite done.
1. I added a Github Workflow for CI (linting, testing, building). We will want to add another one to automatically release and publish merges to master and (possibly) development. 
1. I updated the storybook config to work with the new structure and new storybook configuration.
1. I updated the root `package.json` name, but it should be changed to whatever you'd like to call this from now on.

### To be done
- Determine strategy for bundling styles in production. We could either (a) import modules per-component, (b) import `style.scss` in the index, or (c) provide `@nypl/ds/styles` to the consuming application to import on their own. Option a is probably the best as it means you're not importing styles unless you use them, but option c will be the fastest to implement.
- Determine package name (are we keeping `@nypl/design-system-react-components` or something shorter since there is not only one package eg `@nypl/ds`).
- Configure publishing workflow via Github Actions for both npm package and storybook
- Update README
- Update CHANGELOG


src/__tests__/helpers/hoc/WithOneChild.test.tsx
```diff
@@ -39,15 +34,15 @@ describe("withOneChild HOC", () => {
         );
         wrapper = Enzyme.mount(multipleChildrenDummy);
 
-        expect(wrapper.find("p")).to.have.lengthOf(2);
-        expect(wrapper.prop("title")).to.equal("Wizard");
-        expect(wrapper.prop("age")).to.equal("old");
+        expect(wrapper.find("p")).toHaveLength(2);
```
hokei, line 37: Was this an autoformatter thing?  I think we should keep this using the old Chai testing syntax since there's not a real need to change it.

kristojorg: This is just part of moving from chai assertion library to jest. We can continue using chai if you prefer, but it would be adding a second assertion library on top of the one included with jest

.eslintrc.js
```diff
@@ -0,0 +1,94 @@
+module.exports = {
```
hokei, line 1: What would be super helpful to me is if you could let me know which files were auto-generated, and where you've made app-specific changes.

.github/workflows/main.yml
```diff
@@ -0,0 +1,32 @@
...
+      - name: Begin CI...
+        uses: actions/checkout@v2
+
+      - name: Use Node 13
+        uses: actions/setup-node@v1
+        with:
+          node-version: 13.x
+
+      - name: Cache and install node modules
+        uses: bahmutov/npm-install@v1
```
hokei: what is this?

src/__tests__/helpers/hoc/WithOneChild.test.tsx
hokei: we weren't using jest before, is there any reason why it should be added now? 
edit: actually let's address this in a different comment chain.

.eslintrc.js
kristojorg: Hmmm I'm not totally sure how to do that. I did use a codemod to migrate from chai to jest, and I used the `lint --fix` option to auto fix some issues. Let me write a comment to try and guide you through the changes a bit

.github/workflows/main.yml
kristojorg: It's a github action to install and cache `node_modules`: https://github.com/bahmutov/npm-install

It is also being used in `circulation-patron-web`

kristojorg:
RE moving to Jest: It certainly isn't totally necessary, but jest was included as part of the tsdx package. For context on the differences, Jest is a one-stop testing library, you just bring your own rendered (in this case Enzyme). It includes the assertion library and mocking library built in, whereas mocha is just a test runner, and then you use chai and sinon for assertions and mocking respectively. 

The main reason I did the switch was because TSDX comes with Jest out of the box. We could undo that though and revert to using mocha, chai and sinon if you prefer that.

kristojorg:
### How to read through this PR
There are a few levels of review that you should think about for this:
1. Do you like tsdx? Does it suit your needs? Does the directory structure work for you? Is this how you would like to bundle the app? Would you prefer to configure your own webpack/rollup/parcel setup instead of abstracting that away?
1. Does switching to Jest work for you all? That's not completely necessary and with some work, TSDX can be set up with Mocha instead.
1. Does moving to Eslint work for y'all? If so, are these lint rules suitable?
1. Is this how you would like to bundle the styles and set up the package directory?

Once you've answered those questions, it makes sense to read through the actual changes. There were a few levels of changes I had to make:
1. Moving things around to adopt the new directory structure
1. Migrating to jest with a codemod
1. Fixing typescript errors (the current tsconfig is stricter than the previous)
1. Fixing lint errors both automatically and manually.
1. Updating storybook config

hokei:
hey Kristo so this is really great work, but it is a lot of changes to happen at once.  I completely understand the benefits of this new work, but a lot of the stuff, especially on tooling, is a lot to change eventually.  

Items 1 and 2 are great, and makes sense to put in this PR to bundle together.

`Replaces tslint with eslint. Tslint was deprecated at the beginning of the year in favor of eslint with a typescript parser. I recently did this in circulation-patron-web as well, and have copied over the eslint config from there. It's not perfect, but includes jsx-a11y rules as well as the react and typescript presets. Finally, eslint is now in charge of code formatting as well, via the prettier plugin. The prettier rules are in package.json. I didn't alter them from before.` 
This is great, but I'd prefer this to be in a different PR.  

`Replace Mocha with Jest. This is the test runner used by TSDX, and it's generally faster and more modern than Mocha. It is also an assertion library and provides mocking functionality out of the box, which allowed me to remove chai and sinon. I did have to change some test syntax, but it is almost the same API, just slight differences.` 

I completely agree with preferring Jest.  However, Mocha is the library that is used by most of the organization, and I think there would need to be a bigger conversation within NYPL before adopting Jest, considering that Design System is a thing that all of the NYPL engineers need to be able to work with.  I'm reluctant to make that change unilaterally and I'm more reluctant to do it here.  Is TSDX fundamentally incompatible with Mocha or is there a way to make that work?  

`TSDX uses Rollup under the hood instead of webpack, which is generally simpler and has good defaults. It is popular for bundling packages (as opposed to applications). In any case, we shouldn't need to interact with it much since TSDX configures it for us.`

Have we removed the need for webpack configuration then?  How is it doing things like glob importing or svg inlining (does TSDX do that automatically?)  I'm worried here that we're swapping out a configuration tool that we know and hate (webpack) for another configuration tool that we don't know (tsdx) and don't know if we'll hate because we don't know it.  It's an awful lot of change.  

Also, does that mean that storybook building, which continues to use webpack, is even more different from the package?  Ideally we want the storybook to use as much of the same toolchain as possible so that if something works in storybook we can be reasonably sure it works the same way when exported as a package.  

`As part of migrating to eslint, I did need to fix many eslint errors, especially removing unused variables and escaping html entities (like apostrophes).
I removed tabIndex={0} on the modal component since it is not an interactive element so that won't have any effect (eslint informed me of this).
I removed many class constructors that simply called super(props), as that is automatically done if not constructor is provided, so they weren't needed (another thing react/eslint complained about).`

All these sound ok, but also sound like they're related to the `tslint-eslint` work.  It would go better in that PR.  

`I removed the js files inside of src/styles and also the slick css. Let me know if this is still being used and I'll add it back.` This sounds good.  I'd also direct you to this PR: https://github.com/NYPL/nypl-design-system/pull/212 which will be merged before this PR is

`I added a Github Workflow for CI (linting, testing, building). We will want to add another one to automatically release and publish merges to master and (possibly) development.` 
This is also great and clearly needed, can you extract that into another PR?  

`I updated the storybook config to work with the new structure and new storybook configuration.` 
I'll have to circle back to this - I'm not completely sure what's going on here.  

`I updated the root package.json name, but it should be changed to whatever you'd like to call this from now on.`
Let us continue publishing on `@nypl/design-system-react-components`.  Our `1.0` release will be on another name, but until then, this holds all of the beta work-in-progress in one place.

.eslintrc.js
hokei: thank you!  the comment is super useful too!

kristojorg:
I think I should have mentioned that I'm not necessarily expecting y'all to accept this PR. I meant it initially more as a proof of concept for how I would do this, and a helpful jumping off point if you agree with the strategy. 

**Eslint**
In terms of keeping the eslint changes in a separate PR, I normally would have done that also except that tsdx came bundled with eslint set up to work with typescript, so making it work with tslint would have been a bunch of extra time. Seemed cleaner to just include the switch to eslint since it's part of tsdx.

**Mocha/Jest**
I hear you about needing input from the whole team. TSDX isn't incompatible with Mocha, it's just working around the preconfigured options a bit. In any case, if you want to eventually merge this PR I can simply revert those changes and get it to work with Mocha. 

**package name**
I can make that change if you want me to continue this PR

**github CI**
I can extract this, but it would probably be better to hold off until lerna is stripped out because I'm not sure exactly how to make it work with lerna yet, and it doesn't make sense to spend time on that if we're going to change it right away

hokei:
**Github CI**

We can definitely remove this and hold off until this work is done.  There's so much going on here that it's going to be much easier to split stuff out as much as we are able.

kristojorg:
Alright, I removed the github workflow for now. We can add it back in later

kristojorg:
Closing this in favor of @hokei 's `remove-bootstrap` branch and separate PRs to switch to `eslint` and possibly `tsdx`

PR #217, opened 2020-06-08
Fixes #202: Retire container component

### LINKED ISSUE #202: Move Decorators Out of Export
### Components in question:
- Container
- SectionName

### Description of issue:
- Move to a Decorators folder
- Remove from export
- Update corresponding stories
- Will be a breaking change (DRB is using `SectionName`)

helenvholmes:
- Removes the container component in favor of the Template stories


helenvholmes:
@hokei I'm thinking that instead of retiring this, perhaps we want this as a private-to-Storybook decorator (along with say, `SectionName`). Thoughts?

helenvholmes:
I'm going to close this out in favor of figuring out where decorators should go.

PR #218, opened 2020-06-08
SFR-937 remove lerna

### LINKED ISSUE #132: Remove Lerna

hokei:
- Removes Lerna and flattens the file structure.  
- Outstanding work:  Update Readme.

### Front End Review:


src/components/BasicLink/BasicLink.stories.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from "react";
```
helenvholmes: `BasicLink` got removed with the `Link` work

src/components/IconLink/IconLink.stories.tsx
```diff
@@ -0,0 +1,37 @@
+import * as React from "react";
```
helenvholmes: Same with IconLink, got removed with Link work

src/components/UnderlineLink/UnderlineLink.stories.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from "react";
```
helenvholmes: Also removed with Link work

helenvholmes (commented):
Left some inline comments about some components that snuck back in, let's not reintroduce them.

After a fresh npm install, getting some errors:

Screen Shot 2020-06-08 at 2 40 20 PM

Screen Shot 2020-06-08 at 2 40 36 PM

.storybook-react/webpack.config.js
```diff
@@ -0,0 +1,39 @@
...
+const packageConfig = require("../webpack.config.js");
+const MiniCssExtractPlugin = require("mini-css-extract-plugin");
+
+// Export a function. Accept the base config as the only param.
+module.exports = async ({ config, mode }) => {
+    // `mode` has a value of 'DEVELOPMENT' or 'PRODUCTION'
+    // You can change the configuration based on that.
+    // 'PRODUCTION' is used when building the static version of storybook.
+
+    //Add plugins
```
helenvholmes, line 12: Nit: we seem to be switching between a space after the slashes and no space

.storybook-react/webpack.config.js
```diff
@@ -0,0 +1,39 @@
...
+    config.plugins.push(
+        new MiniCssExtractPlugin({
+            filename: "styles.css",
+            chunkFilename: "[id].css",
+        })
+    );
+
+    //Remove SVG config from storybook
```
helenvholmes: Can you explain this config to me? Currently I can't make heads or tails of it

README.md
```diff
@@ -16,66 +16,30 @@ This repo is versioned with [Lerna](https://github.com/lerna/lerna).
 Follow these steps for setup:
 
 1. Pull the repo `git clone https://github.com/NYPL/nypl-design-system.git`
-2. Install Lerna globally `npm i -g lerna`
-3. Install all the packages `npm run install-all`
-4. Run `npm start`.
+2. Install the module `npm install`
```
helenvholmes, line 17: This file describes this PR pretty neatly 👏🏻🎉

README.md
```diff
@@ -16,66 +16,30 @@ This repo is versioned with [Lerna](https://github.com/lerna/lerna).
...
+There should be no need to run the static Storybook instances. If the static Storybook npm script is run, make sure to not commit the `storybook/storybook-static` directory.
 
-When change are merged into `master` (currently using `development`), Travis CI will automatically build the two Twig and React Storybook instances as static sites and deploy them to the `gh-pages` branch. If this needs to be tested locally, run `npm run storybook:static` and then view the `storybook/index.html` file in a browser. No server needs to be running since this is a static web site.
+When change are merged into `master` (currently using `development`), Travis CI will automatically build the Storybook instance as a static site and deploy to the `gh-pages` branch. If this needs to be tested locally, run `npm run storybook:static` and then view the `storybook/index.html` file in a browser. No server needs to be running since this is a static web site.
```
helenvholmes: Nit: changes

Re: branches, are we switching up our branching strategy? I'm fine with it but maybe we should describe current state and leave out future state until we're ready to make the change.

src/components/Links/Link.tsx
```diff
@@ -0,0 +1,67 @@
+// Base Link Helper Class
```
helenvholmes: Nit: I think we can remove L1

helenvholmes (commented):
Tiny nits/questions, but mostly wow wow wow 🎉🙏🏻🎉

README.md
hokei: 🎉 🎉

.storybook-react/webpack.config.js
hokei: Storybook comes with a webpack config that it uses.  This file is transforming that config into a config that actually displays our stuff.  
So this line is getting rid of the default storybook svg config and L21 is telling it to treat SVGs the way we want to.

README.md
```diff
@@ -16,66 +16,30 @@ This repo is versioned with [Lerna](https://github.com/lerna/lerna).
...
 
-Make sure to include `glob-importer` in your (sass-loader)[https://github.com/NYPL/nypl-design-system/blob/development/storybook/.storybook-react/webpack.config.js#L2].
+1. Install the design system `npm install @nypl/design-system-react-components`
+2. Import the styles in your main css file `@import '~@nypl/design-system-react-components/dist/app.css';`
+3. Consult Storybook for the list of available components and props that they require.
+4. Usage example
 
-### Importing React Components
+```
+import * as DS from '@nypl/design-system-react-components';
```
kristojorg, line 51: I may be wrong, but would importing this way make tree shaking impossible? As in, if we `import * as DS` instead of `import { Link } from`, will we always be including the whole bundle, even if we only use one component?

package.json
```diff
@@ -1,15 +1,38 @@
...
+        "prepare": "npm run build",
+        "start": "node node_modules/@storybook/react/bin/index -p 9001 -c .storybook-react",
+        "static": "node node_modules/.bin/build-storybook -c .storybook-react -o ./storybook-static/react",
+        "test": "rm -rf lib && npm run lint && tsc && cp src/styles.scss lib && mocha --require lib/__tests__/setupEnzyme.js --recursive lib/**/*.test.js",
         "format": "prettier --write 'src/**/*.{ts, tsx, js,jsx,json}'",
-        "pretty-quick": "pretty-quick"
+        "pretty-quick": "pretty-quick",
+        "prepublishOnly": "npm run test && npm run dist",
+        "build": "rm -rf lib && tsc",
+        "dist": "rm -rf dist && webpack --mode production --progress --display-modules --config webpack.config.js",
```
kristojorg, line 21: What is the difference between `npm run build`, `dist` and `prepare`?

kristojorg: It looks like running `npm link` runs `dist`, which builds into `/lib`, but the package.json lists main as `dist/design-system-react-components.min.js`

kristojorg (changes requested):
In the README it still says it is versioned with Lerna (I couldn't comment on the line, bc it wasn't changed :/ )

I made a couple little comments as well, but otherwise this looks good!

package.json
hokei: Thanks for bringing this up, I agree this should be more clear.  

`build` is building the ts files into `lib`.  The tests are running from `lib`. `dist` is packaging everything for publish.  I'm not quite sure why the intermediate `lib` step needs to happen though - when I set it up I think I was just happy to have it working, so I'm happy to rethink it. 
 
`npm link` should be run with `npm run watch-webpack` for local development, I will update the readme to reflect this.

`prepare` isn't doing anything - really `build` should be run in `prepublishOnly` so I'll have it reflect that

README.md
hokei: To be honest, I haven't given much thought to tree shaking.  Looking at it in webpack, it seems like there's a some other things that I'll have to check in this library to be able to label it with `"sideEffects": false` so it's currently not being treeshaken at all.  

There's also not currently (and there are no plans to) support for something like 
`import { Heading, Link } as DS` which means that we would have to do something like 
`import { Heading as DsHeading, Link as DsLink }` (the aliasing is important because you can imagine, something like `Link` is going to be easily confused with `ReactRouter.Link` )

I think this might need a whole other thought process around it (perhaps if/when we migrate to Rollup which supports treeshaking much more out-of-box).  

In the meantime, I will update the readme to use `import { Link } from` because it is more correct, but add a note that people might need aliases.

package.json
kristojorg: Ah I see. There should be a way to make Mocha parse the typescript I assume if you want to update that. But it's not the end of the world either way

helenvholmes (approved):
LGTM 🎉

README.md
kristojorg: Okay sounds good!

kristojorg (approved):
😎

PR #225, opened 2020-06-11
SFR-940 Bugfixes for DRBB integration

hokei:
- When SFR tried to pull the latest design system, a number of things broke (which was intended) but in a way where it had to be fixed inside design system (which was unintended).  This PR fixes those things.

### Front End Review:
- Check the storybook to make sure that the DRBB fixes didn't have unintended side effects.

### DIFF
--- a/src/components/Hero/_Hero.scss
+++ b/src/components/Hero/_Hero.scss
@@ -57,6 +57,7 @@
 
         .hero__heading {
             @include space-stack-none;
+            margin-top: 0;
 
             z-index: 0;
             position: relative;
--- a/src/components/Link/Link.stories.tsx
+++ b/src/components/Link/Link.stories.tsx
@@ -42,9 +42,8 @@ export const backwardsLink = () => (
 
 export const actionLinkWithDownloadIcon = () => (
     // To Pass in an icon and its link, make sure that the link tag wraps the icon.
-    // The passed-in content must be properly styled.
     <Link linkType={LinkTypes.Action}>
-        <a href="#passed-link" style={{ display: "flex" }}>
+        <a href="#passed-link">
             <Icon
                 name="download"
                 blockName="more-link"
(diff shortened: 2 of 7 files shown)


src/components/Hero/_Hero.scss
```diff
@@ -57,6 +57,7 @@
 
         .hero__heading {
             @include space-stack-none;
+            margin-top: 0;
```
helenvholmes, line 60: Doesn't look like `.hero__heading` is applying any top margin, so I'm not convinced DS CSS is causing the discrepancy—I might take a look in SFR, see what's happening exactly 🤔

src/components/Link/Link.stories.tsx
```diff
@@ -41,14 +41,18 @@ export const backwardsLink = () => (
...
-        <Icon
-            name="download"
-            blockName="more-link"
-            modifiers={["left"]}
-            decorative={true}
-            iconRotation={IconRotationTypes.rotate0}
-        ></Icon>
-        Download
+    // To Pass in an icon and its link, make sure that the link tag wraps the icon.
+    // The passed-in content must be properly styled.
```
helenvholmes: Man, I super don't like this, this is the whole point of the DS—can we maybe apply flex to all subchildren instead of L45+L47?

src/components/Hero/_Hero.scss
helenvholmes: I think after some investigation this is happening in the `dgx-footer`. Can you remove this line and override it in SFR?

src/components/Link/Link.stories.tsx
hokei: seems like there was an underlying issue here that I fixed.  This should not be necessary anymore.

helenvholmes: Let's make sure to remove the comment, then

helenvholmes (changes requested):
Nit on that styling comment, but otherwise looks good to go

helenvholmes (approved):
👍🏻

PR #226, opened 2020-06-11
Closes #219: Makes ids optional

### LINKED ISSUE #219: Remove mandatory IDs
This applies across many components. Unless there is a specific reason that IDs are required (why are they required?), we should make these props optional. It becomes a pain pretty quickly when using the components to have to pass IDs, especially when rendering components dynamically. If there is a reason these are required, though, I'm all ears.

helenvholmes:
-  Makes IDs optional


kristojorg (approved):
looks good to me! 
? ? ? ?  ?
 ?  ? ? ?? 
   ?  ? ? ?
      ?  ?
        ?

Question marks all the way down

PR #227, opened 2020-06-12
Fixes some style imports

helenvholmes:
- Fixes some style imports causing Travis to fail


hokei:
whyyy did this work before this aaaah (rhetorical statement of frustration)

hokei (approved):

PR #230, opened 2020-06-17
Card

### LINKED ISSUE #206: Create a more generic Card component

helenvholmes:
Issue number: #206 

- (breaking change) Changes `EditionCard` to `Card`, which takes (all optional):
  - header
  - footer
  - image
  - middle content
  - calls to action area
- Changes `Button` from black by default to blue
- Updates EditionCard styles to what is currently reflected in [Metronome](http://themetronome.co/components/mc-14-edition-card/?fresh=true)
- Adds audiobook icon from Google Material icons
- Separates generic card CSS from story-specific CSS for Edition Card and the Lyrasis card to show how the styles get extended
- STILL OUTSTANDING: I'd like to separate story-specific CSS into a separate bundle that isn't included in our file CSS file, but that felt like too much to tackle here.

### Front End Review:
- [ ] View [Card(s) in Storybook](http://localhost:9001/?path=/story/card--card-with-full-data)
- [ ] View [ResearchResultItem in Storybook](http://localhost:9001/?path=/story/search-result-item--search-result-item)
- [ ] View [UnorderedList in Storybook](http://localhost:9001/?path=/story/unorderedlist--search-results-list-with-edition-cards)


src/components/Card/Card.tsx
```diff
@@ -0,0 +1,81 @@
+import * as React from "react";
+import bem from "../../utils/bem";
+
+interface CardProps {
+    /** Uh id  */
+    id?: string;
+
+    /** ClassName that appears in addition to "card" */
+    extraClassName?: string[];
```
kristojorg: Can we simply call this `className`? It think it's assumed that it'll be added to the existing styles

src/components/Card/Card.tsx
```diff
@@ -0,0 +1,81 @@
...
+    footer?: JSX.Element;
+
+    /** First column within the card */
+    image?: JSX.Element;
+
+    /** Main content within the card */
+    content?: JSX.Element;
+
+    /** Calls to Action for the card, such as "download" on a card for a book */
+    ctas?: JSX.Element;
```
kristojorg: I think that these should be typed as `React.ReactNode`s so that they can accept anything React would actually accept for rendering purposes. See an explanation [here](https://github.com/typescript-cheatsheets/react-typescript-cheatsheet/blob/cc2ec67e0b36b7e838df8f2a54b4cd84df6c78cd/README.md#useful-react-prop-type-examples).

I generally prefer `React.ReactNode` for anytime I'm passing components around

kristojorg (changes requested):
This looks so good!! I'm about to build the list view in CPW and will use this : ). Excited for it. 

One question: Does the card work with _just_ content? No heading, footer, image or ctas? If not, should all of those be optional? If so, does it make sense to use `children` for the `content` so that we can just:
```js
<Card>
  <Some Content />
</Card>
```
Up to you, but I find passing children to be easier to read than passing JSX as a prop generally, especially if that content is going to be a large block of JSX.

Also made a couple notes on types and a prop name : )

🎉

src/components/Card/Card.tsx
helenvholmes: I think I had just mistakenly assumed it'd be a protected variable, but it looks like that isn't the case so sure :)

CHANGELOG.md
```diff
@@ -6,6 +6,20 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+
+-   `EditionCard` is now `Card`
+
+### Adds
+
+-   Icon for audiobook
+
+### Changes
+
+-   Buttons from black by default to blue
```
hokei: this is probably breaking too, since I think RN will now have to override buttons styles.

src/components/Card/Card.test.tsx
```diff
@@ -0,0 +1,238 @@
...
+                            ></Icon>
+                            Download
+                        </Link>
+                    </div>
+                </div>
+            }
+        ></Card>
+    );
+
+    it("Generates a Card with an optional header, footer, image, middle content, and CTAs", () => {
```
hokei: I'd drop optional in the test description here since it's not really testing the "optionality".

src/components/Card/Card.tsx
```diff
@@ -0,0 +1,81 @@
+import * as React from "react";
+import bem from "../../utils/bem";
+
+interface CardProps {
+    /** Uh id  */
```
hokei: haha

CHANGELOG.md
helenvholmes: @hokei Let's follow up with Ellen on this one—I'm fairly certain this is the new intended style for ResearchNow. If that isn't being reflected on your template pages in Metronome than we should get those updated.

(I mean I'm happy to mark it as a breaking change, I'm just saying that it _shouldn't_ be for RN.)

src/components/Card/Card.tsx
helenvholmes: 🤦🏻‍♀️ I saw this and Kristo's comments and was like helen pls

src/components/SearchResultItem/SearchResultItem.test.tsx
```diff
@@ -47,15 +66,15 @@ describe("SearchResultItem", () => {
     // let shortHeadingText = "hello";
     // let longHeadingText = `hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello `;
```
hokei: Since you're here -- this can be deleted since text shortening will not happen in design system.  Might be a small other ticket to not pollute card stuff.

src/components/Card/Card.tsx
kristojorg: Keep it!

src/components/SearchResultItem/SearchResultItem.test.tsx
helenvholmes: I can do this small change now, I was trying not to touch too much else but huge blocks of commented out code bothers me too

src/components/Card/Card.tsx
helenvholmes: Crystal will leave comments about how she's feeling in the stories and [I think it's so funny](https://nypl.github.io/nypl-design-system/storybook/storybook-static/react/?path=/story/link--passed-in-link) hahaha

src/components/Card/Card.tsx
```diff
@@ -0,0 +1,82 @@
...
+    heading?: React.ReactNode;
+
+    /** Footer that appears within the card but below the three columns within it */
+    footer?: React.ReactNode;
+
+    /** First column within the card */
+    image?: React.ReactNode;
+
+    /** Main content within the card */
+    content?: React.ReactNode;
```
kristojorg: I think you need to remove this if you're switching to children : )

src/components/Card/Card.tsx
```diff
@@ -0,0 +1,82 @@
+import * as React from "react";
+import bem from "../../utils/bem";
+
+interface CardProps {
+    /** ID  */
+    id?: string;
+
+    /** ClassName that appears in addition to "card" */
+    ClassName?: string[];
```
kristojorg: should be lowercase `className` like when passing to a dom element

kristojorg (changes requested):
couple nits

src/components/Card/Card.tsx
helenvholmes: 🤦🏻‍♀️ thank you

src/components/Card/Card.tsx
```diff
@@ -6,7 +6,7 @@ interface CardProps {
     id?: string;
 
     /** ClassName that appears in addition to "card" */
-    ClassName?: string[];
+    className?: string[];
```
kristojorg: Okay one more thing... I think this should be typed like a standard `className` on an html element. So just a string. If someone wants to add more than one class name they can add them to the string like normal html `className="class1 class2"`

kristojorg (changes requested):
sorry one last thing here

src/components/Card/Card.tsx
helenvholmes: Ohhhhh, solid. I'll make the change, thanks for all the great reviews

kristojorg (approved):
yeehaw!

PR #233, opened 2020-06-18
Fixes #204: Adds content bottom to Template stories

### LINKED ISSUE #204: Update stories with `content-bottom` component
### Component in question:
- Template 

### Screenshot of issue:
- 
Screen Shot 2020-06-02 at 11 56 11 AM 

vs

Screen Shot 2020-06-02 at 11 58 20 AM

### Description of issue:
- It looks like in the figma that the pagination is a content-bottom stretching across the two columns, mirroring content-top, rather than a full-width footer.  If this is the case with all the templates (and not just ResesarchNow) we should update the left-sidebar-template story to have a content-bottom section.

helenvholmes:
Issue number: #204

- Adds content bottom to Template stories

### Front End Review:
- [ ] View [Template in Storybook](http://localhost:9001/?path=/story/template--right-sidebar-template)


src/components/Template/Template.stories.tsx
```diff
@@ -91,3 +114,14 @@ export const RightSidebarTemplate = () => (
         </footer>
     </div>
 );
+
+RightSidebarTemplate.story = {
```
hokei, line 118: I think there needs to be some work (https://storybook.js.org/docs/addons/using-addons/ ?) to make the addons appear?

helenvholmes: What do you mean? Beyond what Monica has already done? Seems to be working to me:

Screen Shot 2020-06-19 at 2 23 03 PM

hokei: This doesn't show up for me at all -- I've pulled this branch and run npm install.  
I see that Monica's PR is merged - does it show up for you here as well?  https://nypl.github.io/nypl-design-system/storybook/storybook-static/react/?path=/story/accordion--accordion-with-checkboxes  

Is there a button I need to click?

helenvholmes: It won't show designs for stories unless the designs have been added. We haven't added designs for all of our stories. A better example would be [Button](https://nypl.github.io/nypl-design-system/storybook/storybook-static/react/?path=/story/button--button-filled-icon), where we have added designs. You should see a Figma embed when you click on the "Design" panel.

helenvholmes: You're not seeing the Design panel on `gh-pages` or locally?

hokei (approved):

PR #244, opened 2020-06-22
add missing dep

kristojorg:
- adds the `he` dep which is used but was missing from dependencies


helenvholmes (approved):
Woof, thanks for finding this

Issue #147, opened 2020-05-20
Publish to npm via travis, preferrably on a production branch.


helenvholmes:
## Tasks
- Setting up a production branch
- tag and push releases with package.json and changelog updates

PR #245, opened 2020-06-23
CI/CD

kristojorg:
Issue number: #147 

- Adds a CI workflow to lint, typecheck and test code on all pushes and pull requests.
- Adds a release workflow to tag, release, publish to npm, and upload to github pages


.gitignore
```diff
@@ -4,3 +4,4 @@ public
 .DS_Store
 lib
 dist
+storybook-static
```
helenvholmes, line 7: Since we're adding this to the `.gitignore`, can you also remove the last section about `storybook-static` from the README?

.github/workflows/release.yml
```diff
@@ -0,0 +1,86 @@
+# This workflow will tag, release, publish to npm, and upload to gh pages.
+name: Production Release
+
+# workflow will only run when a pull request to the release branch is closed.
+on:
+    push:
+        branches: [release]
```
helenvholmes, line 7: Does this mean that instead of a `master`, we have a branch that matches the name of the release? Is that the idea?

helenvholmes (changes requested):
I have a few elementary questions that I think could be answered with some documentation. @kristojorg, as part of this PR, can you also update [this wiki page](https://github.com/NYPL/nypl-design-system/wiki/How-to-Run-a-Release) on How to Run a Release?

.gitignore
hokei: Is the issue that https://github.com/NYPL/nypl-design-system/pull/71 resolves no longer relevant?

.github/workflows/release.yml
hokei: Yep that's exactly what it is!

.github/workflows/release.yml
```diff
@@ -0,0 +1,86 @@
...
+                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+              with:
+                  tag_name: ${{ steps.autotagger.outputs.tagname }}
+                  release_name: ${{ steps.autotagger.outputs.tagname }}
+                  # use the body of the PR commit as the release body
+                  body: $commitmsg
+                  draft: false
+                  prerelease: false
+
+    publish_to_npm:
```
hokei, line 47: Do you know if it will need permissions to publish inside the @nypl org?

.github/workflows/release.yml
helenvholmes: Do these branches exist in perpetuity in addition to the tagged releases?

.gitignore
helenvholmes: Looks so because it'll be running those two other scripts is my guess

kristojorg:
Yes I will! I also am not entirely sure this is bug free yet. Sadly I don’t know how to test it without merging and running the workflow

hokei:
yeah, merge and tweak -- I think that's just how this goes!

hokei (approved):

hokei (changes requested):
sorry, missed Helen's request for changes above

.github/workflows/release.yml
kristojorg: Oh actually I think this is a little different. We won't have a `master` branch, but instead have a single `release` branch, which is the currently released code. The difference between calling it `master` or `release` is just semantic. There _won't_ be a separate branch for every release (like `release-v0.6.5`), which is what I think you were referring to. The purpose of marking/saving the state of the code when a release was made is what git tag is for. So if we ever needed to go back and branch off from an old release, we would simply start from the tag for that release. 

Long story short, we'll have:
- `release`
- `development` (default)
- `feature-branches...`

And a tag for every version:
- `v0.6.3`
- `v0.6.4`
- etc

.github/workflows/release.yml
kristojorg: Yep, we need to add a secret to the repo called `NPM_TOKEN`

.gitignore
kristojorg: Yea I just read through that PR. It _should_ now be possible to gitignore this folder, but I will have to double check once we merge this that it worked (I think it will)

kristojorg:
Okay I made an update to the wiki page. Let me know what you think!

.github/workflows/release.yml
helenvholmes: Ahh, that makes sense. Thanks for the explanation Kristo!

helenvholmes (approved):
Since we're not sure about whether or not we can gitignore .storybook-static I think it's fine to leave it in the README for now and follow-up after we've tested the workflow. Thanks for this hard work Kristo!

kristojorg:
@hokei I think I still need you to approve before I can merge since you also requested changes : )

helenvholmes:
Looks like the new checks are running on other PRs, so let's try and get this merged and ironed out to not delay parallel work :)

hokei (approved):

Issue #220, opened 2020-06-09
SearchBar props and types


kristojorg:
### Component in question:
- SearchBar
- Also Input

### Description of issue:
I would like to use the SearchBar in CPW, however there are a few changes that I think would be great to make before I did that:

1. Remove the mandatory IDs (opened another issue for this with all components)
2. rename `searchBarAriaLabel` and `searchBarAriaLabelledBy` to simply use the default `aria-label` and `aria-labelledby` names. This is a trend I think we should follow in general, using the native prop names where applicable. 
3. rename `searchValue` to simply `value`. Same idea as above
4. rename `placeHolder` to `placeholder`
5. rename the handlers from `searchChangeHandler` and `selectChangeHandler` to `onChange` and `onSelectChange`.
6. Refactor the error states into typescript. There are currently errors thrown for incorrect props, but those requirements can be expressed in typescript instead. 

Those are the more minor changes, but there is a larger rethinking of this component I would like to propose as well: I want to consider refactoring the `SearchBar` API to simply be a component composed from an `Input`, `Select` and `Button`. 

The `SearchBar` component doesn't introduce any logic per-se, it's mainly styling. I think it would be easier to let the consuming app compose one themself from smaller components. This way the consumer doesn't have to get to know the specifics of this API and how it passes props through to children. An example API composing a `SearchBar` from `Input`, `Button`, and `Select`:

```js
const SearchBar = () => {
  const [value, setValue] = React.useState("");
  const [selectValue, setSelectValue] = React.useState("keyword");
  const handleChange = (e) => {
    setValue(e.target.value);
  }
  const handleSelectChange = (e) => {
    setSelectValue(e.target.value)
  }
  const handleSearch = () => {
    api.search(value, selectValue);
  }
  
  return (
    <InputGroup aria-label="Search">
      <InputLeftAddon >
        <Select aria-label="Search Type" onChange={handleSelectChange}>
      </InputLeftAddon>
      <Input placeholder="Enter a search term..." onChange={handleChange} />
     <InputRightAddon >
       <Button aria-label="Submit search" onClick={handleSearch}>Search</Button>
     </InputRightAddon>
    </InputGroup>
  );
}
```

This is inspired by [Chakra UI](https://chakra-ui.com/input#left-and-right-addons), where the `InputLeftAddon` and `InputRightAddon` position their children correctly with the input so it appears similar to how the NYPL one does. 

[...]

PR #253, opened 2020-06-24
Select API Changes

helenvholmes:
Part of #220.

- `FormDropdown` and `Dropdown` are now `Select`
- Select stories now use knobs to hide/show `Label` and `HelperErrorText` as part of its story
- Changes `referenceId` to `htmlFor` on `Label`
- Moves `input` and `select` CSS from `_Textfield.scss` to their corresponding .scss files
- Makes updates to `HelperErrorText` and `Select` based on the Metronome documentation ([HelperErrorText](http://themetronome.co/components/mt-51-helper-and-error-text-/?fresh=true), [Select](http://themetronome.co/components/af-121-form-dropdown/?fresh=true))

### Front End Review:
- [ ] View [Select in Storybook](http://localhost:9001/?path=/story/select--select-with-optional-label)


src/components/Select/Select.tsx
```diff
@@ -0,0 +1,115 @@
+import * as React from "react";
+import bem from "../../utils/bem";
+import Icon from "../Icons/Icon";
+
+export interface SelectProps {
```
kristojorg, line 5: should this also take a `className` prop? I sort of feel that almost all components should

src/components/Select/Select.tsx
```diff
@@ -0,0 +1,115 @@
...
+    /** Populates the selects' options */
+    options: string[];
+
+    /** Sets whatever string you pass it as the default selected */
+    selectedOption?: string;
+
+    onSelectBlur: (event: React.FormEvent) => void;
+
+    /** Passes selects' current value to the React state handler */
+    onSelectChange: (event: React.FormEvent) => void;
```
kristojorg: My personal preference would be to name these `onBlur`, `onChange`, etc to keep the API as close to the HTML api as possible

src/components/Select/Select.tsx
```diff
@@ -0,0 +1,115 @@
...
+    isRequired: boolean;
+
+    /** When passed, will populate the aria-label on the select */
+    ariaLabel?: string;
+
+    /** When true, disables the select */
+    disabled?: boolean;
+
+    /** Populates the selects' options */
+    options: string[];
```
kristojorg: I would let the consumer render these as one usually does with a `<select>`, so passed as children. The API would then be:
```
<Select>
  <option>my option</option>
  <option>another!</option>
</Select>
```
which is the same as using regular HTML:
```
<select>
   <option>option 1</option>
</select>
```

This is especially because a select might have an option with a value that is different from its display name:
```
<Select>
  {mediaType.map(mediaType => (
    <option key={mediaType.id} value={mediaType.id}>{mediaType.name}</option>
  )}
</Select>
```

kristojorg (approved):
I made some API comments, take them or leave them but otherwise I approve!

PR #260, opened 2020-06-29
Refactors Input

helenvholmes:
Relevant issue number: #220 

- Refactors `Textfield` into `Input`, dropping the other `Input` component
- Removes `DateRangeForm` 

### Front End Review:
- [ ] View [Input in Storybook](http://localhost:9001/?path=/story/input--input)
- [ ] View [Input Group in Storybook](http://localhost:9001/?path=/story/input--input-group)


src/components/Input/Input.stories.tsx
```diff
@@ -1,60 +1,152 @@
...
-        labelContent={<>Label</>}
-        inputId="blah"
-        isRequired={false}
-    ></Input>
+    <>
+        {boolean("Show Label", true) ? (showLabel = true) : (showLabel = false)}
+        {showLabel && (
+            <Label
+                htmlFor="inputID"
+                optReqFlag={select(
```
hokei, line 25: Changing this knob doesn't change anything visually right now

src/components/Input/Input.tsx
```diff
@@ -1,129 +1,92 @@
...
+    /** Populates the aria-label on the select */
+    ariaLabel?: string;
 
-    isRequired?: boolean;
+    /** Populates the aria-labelledby on the select */
+    ariaLabelledBy?: string;
 
-    helperContentOpts?: HelperErrorTextOptions;
-    errorContentOpts?: HelperErrorTextOptions;
+    /** Will add 'aria-required: true' to input */
```
hokei, line 15: Should it also add a flag/label?

src/components/Input/Input.stories.tsx
```diff
@@ -1,60 +1,152 @@
...
+                id={"label"}
+            >
+                {text("Input Label", "Choose your islander name: ")}
+            </Label>
+        )}
+        <Input
+            id="inputID"
+            ariaLabel="Input Label"
+            helperTextId={"helperText"}
+            required={boolean("Input Required", false)}
```
hokei, line 39: Is there anything that checks whether this value is the same as the one in the label, or is that a thing we trust the user to not mess up?

src/components/Input/Input.stories.tsx
```diff
@@ -1,60 +1,152 @@
...
+            placeholder={text("Input Placeholder", "CoolPerson99")}
+            type={select("Input Type", InputTypes, InputTypes.text)}
+        ></Input>
+        {boolean("Show Helper Text", true)
+            ? (showHelperText = true)
+            : (showHelperText = false)}
+        {showHelperText && (
+            <HelperErrorText
+                isError={boolean("Helper Text Errored", false)}
```
hokei, line 48: This knob doesn't change anything in the story either

hokei:
The `proptypes` part of the info panel no longer shows up.  I think this might be a separate ticket, because it affects all components, not just this one.

src/components/Input/Input.test.tsx
```diff
@@ -3,82 +3,157 @@ import * as Enzyme from "enzyme";
...
             ></Input>
         );
     });
 
-    it("Renders label for inputId", () => {
-        expect(container.find("label").exists()).to.equal(true);
-        expect(container.find("label").prop("htmlFor")).to.equal("input");
-    });
-    it("Renters TextField", () => {
+    it("Renters Input", () => {
```
hokei: nit

src/components/Input/Input.test.tsx
```diff
@@ -3,82 +3,157 @@ import * as Enzyme from "enzyme";
...
-        expect(container.find("label").prop("htmlFor")).to.equal("input");
-    });
-    it("Renters TextField", () => {
+    it("Renders Input", () => {
         expect(container.find("input").exists()).to.equal(true);
     });
-    it("Renders Helper", () => {
-        expect(container.find("#input-helper").exists()).to.equal(true);
+
+    it("Label points to its associated input", () => {
```
hokei: This test can be removed because it's just testing `htmlFor="inputID"` which is just html functionality.

src/components/Input/Input.tsx
```diff
@@ -1,129 +1,92 @@
...
 
-    showError?: boolean;
-}
-interface InputProps {
-    /** Optional baseClass for use with BEM. See how to work with blockNames and BEM here: http://getbem.com/introduction/ */
-    baseClass?: string;
-    /** Optional blockName for use with BEM. See how to work with blockNames and BEM here: http://getbem.com/introduction/ */
+    /** className you can add in addition to 'input' */
+    className?: string;
+
```
hokei, line 20: does it make sense to put a catch-all attributes here like https://github.com/NYPL/nypl-design-system/blob/development/src/components/Link/Link.tsx#L26 since `input` is going to be flexible and general as well?

hokei: edit:  huuh link attributes doesn't actually get used.  The idea I think is that if you want to attach something to the base html tag, like `<a>` or `<input>` it will give you a way to do that.

src/components/Input/Input.stories.tsx
```diff
@@ -1,60 +1,152 @@
...
+                {text("Input Label", "Choose your islander name: ")}
+            </Label>
+        )}
+        <Input
+            id="inputID"
+            ariaLabel="Input Label"
+            helperTextId={"helperText"}
+            required={boolean("Input Required", false)}
+            placeholder={text("Input Placeholder", "CoolPerson99")}
+            type={select("Input Type", InputTypes, InputTypes.text)}
```
hokei, line 41: Is this in preparation to get the `input` to be an element that supports all inputs?  Or is this still going to only support text or number still?  

Maybe I'm misunderstanding the purpose of knobs - are they not supposed to all be expressive?  Considering that we may be sharing the storybook outside of engineering, how can we communicate which knobs produce expressive and useful results, and which don't?

src/components/SearchBar/_SearchBar.scss
```diff
@@ -33,7 +33,7 @@ $search-bar--breakpoint: 856px;
     }
 
     &__input {
-        @include input;
+        // @include input;
```
hokei: these comments intentional?

hokei (changes requested):
Lots of great work here!

src/components/Input/Input.stories.tsx
helenvholmes: It does, but it's obscured by the "Show Info" storybook button... every time I try to work on storybook-docs it's a nonstarter since it's more complicated than just dropping in it, but damn that ugly blue button really cramps my vibes

src/components/Input/Input.stories.tsx
helenvholmes: There isn't. I think a warning would be a helpful thing to add here so I'll add that in

[EDIT]: actually, unsure how to check this, since they're part of two separate components that aren't composed into a larger component anymore. I think the answer is that we gotta trust the user.

src/components/Input/Input.stories.tsx
helenvholmes: I'm seeing it add red styling to the HelperText to show it's error'd—funnily enough though, this is actually good UX feedback...

src/components/Input/Input.tsx
helenvholmes: I think it makes sense to add that in addition to className—the className you want added to the bem function specifically.

helenvholmes: I assume the link attributes are just getting passed right on through via the `linkProps`?

src/components/Input/Input.stories.tsx
helenvholmes: Looking through the Metronome documentation, I'm not seeing any wording that the DS.Input should only accept text/numbers—do you know where you heard that requirement?

I'm not sure expressive and useful are the right words here—maybe "designed" and "not designed". All of the knobs here are still useful in that they can show UX what is and isn't designed, and they are all expressive in that they are all valid HTML and options that the HTML spec gives developers.

That said I think we should eliminate the other InputTypes all the same if it's going to cause confusion. I'll rip the extras out for now but I'd like to leave it as an enum so we can add to it over time.

helenvholmes: (Ended up also leaving in the `password` type... if the designers want anything other than the default I want them to be thinking about that before we get to any Header refactor.)

helenvholmes:
I'm not sure what's going on with the propTypes not showing up 🤔 I'll file a bug

[EDIT] Created #261 for this.

src/components/Input/Input.stories.tsx
hokei: > I'm not seeing any wording that the DS.Input should only accept text/numbers—do you know where you heard that requirement?

I didn't -- it was more me wondering, is `DS.Input` specific to stuff you type on your keyboard, or should it also encompass radio and checkbox ?  The changes to `InputTypes` makes this clear now.

hokei:
I wonder if this component should be renamed to `TextInput` to differentiate it from all the other types of input?  In particular, it really feels like `InputTypes.tsx` would be more accurately named `TextInputTypes.tsx`

src/components/Input/Input.tsx
helenvholmes: Label's a separate component now, so the story shows them being used together (which is how we mostly want consumers to build forms, and what most of the DS designs call for).

helenvholmes:
Based on our Slack conversation, sounds like we're going to leave the name as-is for now.

hokei (approved):

PR #262, opened 2020-07-01
IconNames enum, improved Button stories, Button API change

### LINKED ISSUE #214: Update Button Props
### Component in question:
- Button

### Description of issue:
- We currently use `callback` as the action to perform when the button is clicked (or otherwise activated). It seems like it would be a simpler API to leave the callback with the default browser name so consumers would pass an `onClick` which can then be caught and applied `onMouseDown` when appropriate.
- The `id` field shouldn't be required unless it is needed for some reason.

helenvholmes:
Issue number: #214 

- Removes `ButtonOpts` from Accordion
- Makes all SVG widths `100%`
- Adds viewport properties to any SVGs that were missing it
- Addresses #214

### Front End Review:
- [ ] View [Button in Storybook](http://localhost:9001/?path=/story/button--button)
- [ ] View [Accordion in Storybook](http://localhost:9001/?path=/story/accordion--accordion-with-checkboxes)


src/components/Accordion/Accordion.tsx
```diff
@@ -34,19 +38,21 @@ export default class Accordion extends React.Component<
...
+        ButtonProps.iconPosition = ButtonIconPositions.JustifyRight;
+        ButtonProps.iconName = this.state.isOpen ? "minus" : "plus";
+        ButtonProps.iconModifiers = ["medium"];
+        ButtonProps.buttonType = ButtonTypes.Secondary;
 
         return (
             <div className={bem("accordion", modifiers, blockName)}>
-                <Button callback={this.toggleContentShow} {...buttonOptions}>
-                    {buttonOptions.content}
+                <Button onClick={this.toggleContentShow} {...ButtonProps}>
```
hokei: I know this was here before, but I think we should be explicit about props instead of passing them all in like this - IMO it's more clear and less likely to have side effects

src/components/Accordion/Accordion.tsx
```diff
@@ -1,22 +1,26 @@
...
+import Button from "../Button/Button";
 import bem from "../../utils/bem";
 import { ButtonTypes, ButtonIconPositions } from "../Button/ButtonTypes";
 
 export interface AccordionProps {
+    /** ID that other components can cross reference for accessibility purposes */
     id?: string;
 
+    /** String labelling the button that opens the accordion */
+    accordionLabel?: string;
```
hokei: we're sure about this, right?  We explicitly want to exclude `<em>` or `<strong>` or spans or icons?

helenvholmes: Good call... I actually think it should accept a variety of things since I think SCC might use an Accordion for the Subject Header Explorer (where the accordion innards look like columns to me)

src/components/Accordion/Accordion.tsx
helenvholmes: I'm not sure I understand—what do you mean by being explicit, and what does unintended circumstances does it avoid?

src/components/Accordion/Accordion.stories.tsx
```diff
@@ -31,9 +31,9 @@ const checkboxes = [
 export const accordionWithCheckboxes = () => (
     <Accordion
         id="accordion"
-        buttonOptions={{ id: "accordionBtn", content: <>Click to expand</> }}
+        accordionLabel="Click to expand"
```
hokei, line 34: the checkboxes look broken in this story

src/components/Button/Button.stories.tsx
```diff
@@ -1,140 +1,56 @@
...
-    >
-        Hello World
-    </Button>
-);
-
-buttonOutline.story = {
-    name: "Button Outline",
+let showIcon;
+
+export const button = () => (
```
hokei, line 18: man I love knobs

src/components/Accordion/Accordion.stories.tsx
helenvholmes: Checkboxes have been lookin' broke for a while, I'm not sure when that bug got introduced 😣 I'll fix it here though

src/components/Pagination/Pagination.tsx
```diff
@@ -61,7 +61,7 @@ export default function Pagination(props: PaginationProps) {
 
     return (
```
hokei, line 61: The dropdown in the pagination story has broken

hokei (changes requested):
export `IconName` in `index.ts` as well

src/components/Accordion/Accordion.tsx
hokei: Like going 
`<button id={blah} className={blah} onClick={blah} >` 

instead of assigning them above and passing it as `{...ButtonProps}` 

Personally I find it easier to read, but also it makes sure that you're not passing in anything that you don't intend to.  I think we should only use the spread syntax to make it clear that we aren't completely sure what we're passing in (like attributes).

src/components/Pagination/Pagination.tsx
helenvholmes: I imagine this was broken with #253 originally. Because this will require reworking the component I'm going to request we file a new ticket for the work instead of trying to do it here (#220 is blocked on this work, which Location Finder needs more immediately).

[EDIT] Nah you're right I was just tired 5pm on a Monday

hokei:
Double checking that all the searchbar related stuff is broken here and that's ok right?

hokei (approved):

helenvholmes:
@hokei "okay" is maybe the wrong word, but it's known and we'll be running a release once fixed. Next time we have this many interconnected breaking API changes in a single sprint I'll have a separate branch for all the breaking changes instead of merging to dev.

PR #265, opened 2020-07-02
Fix #238: Removes references to specific projects for Card

### LINKED ISSUE #238: Remove references to specific projects in `Card`
### Component in question:
- `Card`

### Description of issue:
- Based on the conversation with @hokei yesterday, we're going to take out references in the stories to the names of specific products

helenvholmes:
Issue number: #238

- Removes references to specific projects for Card
- Removes `EditionCard` styling

### DIFF
--- a/src/components/Card/Card.stories.tsx
+++ b/src/components/Card/Card.stories.tsx
@@ -1,6 +1,7 @@
 import * as React from "react";
 import bem from "../../utils/bem";
 import { action } from "@storybook/addon-actions";
+import { text, boolean, select } from "@storybook/addon-knobs";
 
 import Card from "./Card";
 import Heading from "../Heading/Heading";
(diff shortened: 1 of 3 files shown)


src/components/Card/Card.stories.tsx
```diff
@@ -17,299 +18,91 @@ export default {
...
+                        text={text("Header Content", "Optional Header")}
+                    />
+                ) : (
+                    false
+                )
+            }
+            image={
+                showImage ? (
+                    <Image
+                        src={select("Image Ratio", imageRatios, imageRatios[2])}
```
hokei, line 64: Changing the image ratio doesn't seem to change the size of the container, just the content - is that intended?  Is this a change to the card from when it was designed in DRB?)

hokei (changes requested):
Just have that question, but otherwise it looks good!

src/components/Card/Card.stories.tsx
helenvholmes: Yeah, that was intended. DRB should have always had this functionality imo, I think the fact that it didn't was an oversight

hokei: This is a thing we're gonna have to communicate to Sean, otherwise it'll be considered a bug ... I think we're pretty used to seeing it the other way.  The Metronome that I built DRB off of had the border sticking to the book, I think.

hokei (approved):

src/components/Card/Card.stories.tsx
hokei: This is also the kind of thing that I think would be nice to put in a PM-facing changelog.

helenvholmes: [Here's a link to those designs](https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Master?node-id=17167%3A58131). I don't see a border around the book, but it is a change. Should definitely be called out since it's a change we want QA to be aware of, maybe that's what we surface with PMs?

helenvholmes: I've updated the CHANGELOG with the change as well, since that should have been in there separately from the note about `.edition-card` styles going away.

PR #274, opened 2020-07-13
Searchbar

helenvholmes:
Issue number: #220 

- Refactors `UnorderedList` into `List`
- Refactors `SearchBar` into a `<form>` wrapper that accepts children

### Front End Review:
- [ ] View [the List in Storybook](http://localhost:9001/?path=/story/list--list)
- [ ] View [the Searchbar in Storybook](http://localhost:9001/?path=/story/searchbar--search-bar)


src/components/Input/Input.tsx
```diff
@@ -74,22 +71,15 @@ export default function Input(props: InputProps) {
...
-        inputProps["aria-labelledby"] = labelId;
-    } else if (helperTextId && !labelId) {
-        inputProps["aria-labelledby"] = helperTextId;
-    } else if (labelId && helperTextId) {
-        inputProps["aria-labelledby"] = labelId + " " + helperTextId;
-    }
-
     let transformedInput = (
-        <input id={`input-${id}`} {...inputProps} placeholder={placeholder} />
+        <input id={id} {...inputProps} placeholder={placeholder} />
```
hokei: are you adding attributes to this here or in the attributes PR? (I don't see it in the attributes pr)

helenvholmes: Whoops! Attributes PR. I've pushed it up over there.

src/components/List/List.stories.tsx
```diff
@@ -0,0 +1,137 @@
...
+    parameters: {
+        design: {
+            type: "figma",
+            url:
+                "https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Master?node-id=16115%3A304",
+        },
+    },
+};
+
+const exampleCard = (
```
hokei, line 67: ordered list cards are selectable in knobs, but don't have a separate style

src/components/List/List.test.tsx
```diff
@@ -0,0 +1,34 @@
+import { expect } from "chai";
+
+import * as Enzyme from "enzyme";
+import * as React from "react";
+
+import List from "./List";
+import { ListTypes } from "./ListTypes";
+
+describe("Link With Underline", () => {
```
hokei: Can you add at test for what happens if you pass in a bunch of `<span>`s into a `<List>` component?

hokei: And maybe figure out if that should be tackled in DS, and how?

src/components/SearchBar/SearchBar.stories.tsx
```diff
@@ -1,86 +1,161 @@
...
+        {boolean("Form Disabled", false)
+            ? (formDisabled = true)
+            : (formDisabled = false)}
+        {boolean("Show Select", true)
+            ? (showSelect = true)
+            : (showSelect = false)}
+        {boolean("Show Helper Text", true)
+            ? (showHelperText = true)
+            : (showHelperText = false)}
+        <SearchBar
```
hokei, line 96: Screen Shot 2020-07-14 at 6 47 50 PM

This looks off

src/components/SearchBar/SearchBar.tsx
```diff
@@ -1,159 +1,53 @@
...
-            id={searchBarId}
+            className={bem(searchbar__base_class, modifiers, blockName)}
+            id={id}
             role="search"
-            aria-label={searchBarAriaLabel}
-            aria-labelledby={searchBarAriaLabelledBy}
+            aria-label={ariaLabel}
+            aria-labelledby={ariaLabelledBy}
+            onSubmit={props.onSubmit}
+            {...props.attributes}
```
hokei: Assign these to consts in the L27 block so that we don't have to call `props` directly.

src/components/SearchBar/SearchBar.tsx
```diff
@@ -1,159 +1,53 @@
...
-    errorMessage?: string;
-    selectBlurHandler?: (event: React.MouseEvent) => void;
-    selectChangeHandler?: (event: React.MouseEvent) => void;
-    searchSubmitHandler: (event: React.MouseEvent) => void;
-    searchChangeHandler?: (event: React.FormEvent) => void;
+
+    /** Populates aria-label on the form */
+    ariaLabel?: string;
+
+    /** Populates aria-labelledby on the form */
```
hokei: With these comments, it would be nice to use it as guidance for engineers.  So like, something like "ID of the label for the search input" 

(Also would be a helpful reminder, I can never remember what the difference is between aria-label and aria-labelledby haha)

hokei (changes requested):
Mostly looks good, a few straggler issues to address

src/components/SearchBar/SearchBar.tsx
helenvholmes: This is the Metronome documentation for this component, which is pretty vague:

```
Is a form with attributes role="search" and aria-label or aria-labelledby that is meaningful for the template in which it is used, eg "Search Research Catalog" or "Search NYPL.org"
```

I figured that without additional context about the template the SearchBar is being used in, we probably can't give much guidance about what the `aria-labelledby` would be populated by, but I think we can still improve on the definitions by using the language from [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute) defining the difference between the two. What do you think?

src/components/HelperErrorText/HelperErrorText.tsx
```diff
@@ -35,14 +36,15 @@ export default function HelperErrorText(
...
+        attributes,
         id,
         blockName,
-        baseClass = "helper-text",
         isError,
         ariaLive = "polite",
         ariaAtomic = true,
     } = props;
 
+    let baseClass = "helper-text";
```
hokei, line 38: Is `props.baseClass` used at all here, then?  If not, take it out of `HelperErrorTextProps`

helenvholmes: I'm not seeing it in `HelperErrorTextProps`... do you mean `HelperErrorTextOptions`?

helenvholmes: Ended up removing `HelperErrorTextOptions`—looks like they could be safely deprecated at this point.

src/components/SectionTitle/SectionTitle.tsx
```diff
@@ -69,14 +70,13 @@ export default class SectionTitle extends React.Component<
 
         let headingModifiers = linkUrl ? ["has-link"] : [];
         return (
-            <div className={bem(baseClass, modifiers, blockName)}>
+            <div className={bem(baseClass, modifiers, blockName, [className])}>
```
hokei, line 73: SectionTitle is eventually being deprecated right

hokei (approved):
:tada: :tada: phew this was a lot, lots of great work here!

src/components/SectionTitle/SectionTitle.tsx
helenvholmes: This component? Yeah, I think we can safely bundle it as a `Heading` story.

Issue #273, opened 2020-07-13
`HelperErrorText` and its aria-* props


EdwinGuzman:
### Component in question:
- `HelperErrorText`

### Description of issue:
- `HelperErrorText` should read its `ariaLive` and `ariaAtomic` props and also allow `assertive` as an option.

EdwinGuzman:
Fixed in #275. Closing.

Issue #261, opened 2020-07-01
PropTypes area in "Show Info" are missing


helenvholmes:
<img width="1188" alt="No PropTypes appear" src="https://user-images.githubusercontent.com/1720093/86249552-6f401c00-bb7d-11ea-9854-fda570c4fcc4.png">

After hitting the "Show Info" button in Storybook, the PropTypes tables for the component(s) no longer appear (as show in the screenshot above).

helenvholmes:
https://storybook.js.org/docs/configurations/typescript-config/

EdwinGuzman:
Fixed in #275. Closing.

PR #275, opened 2020-07-14
Minor updates to form-related components.

EdwinGuzman:
Issue numbers: #273 and #261 

- Adds some changes so I could get `Label`, `Input`, `HelperErrorText`, and `Checkbox` to work in the Library Card app. Specifically:

* The `HelperErrorText` component wasn't exported.
* The Library Card app uses refs to focus on input elements for specific errors. So if there are 10 input elements on the page and the 6th one has an error because the email (for example) is invalid, the app creates a dynamic anchor error link to that input. But this doesn't seem to work if the `Input` component doesn't implement `forwardRef` to pass down the `ref` but not as a prop.
* Used the `attributes` prop in the `Input` component.
* Fixed the issue with the prop table not showing up in the "Info" tab in Storybook. The webpack config was removed...
* Updated the stories for some of the components.

Let me know if I'm missing anything else for this PR. I'll create a draft PR in the Library Card app that uses these changes.

### Front End Review:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@nypl/design-system-react-components",
-    "version": "0.6.2",
+    "version": "0.6.3",
     "description": "Design System React Components",
     "repository": {
         "type": "git",
--- a/src/components/HelperErrorText/HelperErrorText.tsx
+++ b/src/components/HelperErrorText/HelperErrorText.tsx
@@ -23,7 +23,7 @@ interface HelperErrorTextProps {
     /** Optional baseClass for use with BEM. See how to work with blockNames and BEM here: http://getbem.com/introduction/ */
     baseClass?: string;
     /** Added prop when HelperText is errored */
-    ariaLive?: boolean;
+    ariaLive?: "polite" | "off" | "assertive";
     /** Added prop when HelperText is errored */
     ariaAtomic?: boolean;
 }
@@ -34,14 +34,21 @@ interface HelperErrorTextProps {
 export default function HelperErrorText(
     props: React.PropsWithChildren<HelperErrorTextProps>
 ) {
-    const { id, blockName, baseClass = "helper-text", isError } = props;
+    const {
+        id,
+        blockName,
+        baseClass = "helper-text",
+        isError,
+        ariaLive = "polite",
+        ariaAtomic = true,
+    } = props;
 
     let modifiers = [];
-    let ariaLive, ariaAtomic;
+    let announceAriaLive = false;
 
     if (isError) {
         modifiers.push("error");
-        ariaLive = true;
+        announceAriaLive = true;
     }
 
     if (props.modifiers) {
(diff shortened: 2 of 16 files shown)


helenvholmes (approved):
This looks really good to me! Thanks for the updates and sticking to our AC-examples 😂

I did pop off some of the stuff from `Input` story #1 that was carried into #3, I think it's helpful to not have everything under the sun in the knobs panel if it's being illustrated in a different story.

I want @hokei to take a look at the `ref`s work you've done here too so I'm going to ask you don't merge just yet even though you have approval ✨

webpack.config.js
```diff
@@ -33,7 +33,14 @@ module.exports = {
         rules: [
             {
                 test: /\.tsx?$/,
-                loader: "ts-loader",
+                use: [
+                    {
+                        loader: "ts-loader",
+                    },
+                    {
+                        loader: "react-docgen-typescript-loader",
```
hokei, line 41: 😮

src/components/HelperErrorText/HelperErrorText.tsx
```diff
@@ -23,7 +23,7 @@ interface HelperErrorTextProps {
     /** Optional baseClass for use with BEM. See how to work with blockNames and BEM here: http://getbem.com/introduction/ */
     baseClass?: string;
     /** Added prop when HelperText is errored */
-    ariaLive?: boolean;
+    ariaLive?: "polite" | "off" | "assertive";
```
hokei, line 26: Do we want this to be knob-able?

helenvholmes: Yeah, I think we should... I can push up that change

[EDIT] We don't have knobs for HelperErrorText yet! Let's do that separately from this work.

src/components/Input/Input.tsx
```diff
@@ -89,8 +90,15 @@ export default function Input(props: InputProps) {
     }
 
     let transformedInput = (
-        <input id={`input-${id}`} {...inputProps} placeholder={placeholder} />
+        <input
+            id={`input-${id}`}
+            {...inputProps}
+            placeholder={placeholder}
+            ref={ref}
```
hokei, line 97: With attributes added into inputProps, such that anyone can pass any extra prop and have it attached to `<input>`, could you technically go 

<DS.Input attributes={{ref: newRef}} /> and should do that instead, 

Or is it still something we want to do to pass refs into all the form components?  (I think ReactRouter allows explicit passing of refs, so I'm up for either).

src/components/HelperErrorText/HelperErrorText.tsx
EdwinGuzman: I had trouble setting up a knob for this because of the type.

src/components/Input/Input.tsx
EdwinGuzman: It makes sense and it is technically possible to do the above, but `ref` isn't an HTML input attribute and is mostly used for React. So I would like to keep that separate. I think it also makes sense to keep it separate since it's using forwardRef for this purpose.

EdwinGuzman:
I pushed another commit that also adds the forwardRef implementation to the `Checkbox` component.

EdwinGuzman:
The Library Card PR using this change is up: https://github.com/NYPL/nypl-library-card-app/pull/48

Whenever both can be reviewed, that'd be great.

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@nypl/design-system-react-components",
-    "version": "0.6.2",
+    "version": "0.6.3",
```
hokei, line 3: this no longer should be bumped manually

Issue #115, opened 2020-04-16
All components should take attributes blob


helenvholmes:
[See this doc for additional context](https://docs.google.com/document/d/1eUx4hJPnNmzFZwlYx-_uRS0i4uw_0Pmq1hSOsAwLdHQ/edit)

Different projects will want to add additional attributes to our components. All components in the DS library should be able to take additional attributes from the projects they are integrated into.

PR #276, opened 2020-07-14
Adds `attributes?: {}` and `className?` to all components

helenvholmes:
Issue number: #115

- Adds `attributes?: {}` to all components
- Adds `className?: string` to all components
- Deprecates `Container`
- Organizes all the props in all the components alphabetically (I don't feel strongly about this or anything, it was just a decision—different components were organizing props based on different criteria, and this felt like a good organizational decision /shrug)
- Adds prop descriptions to some components that didn't have them (some are on the roadmap to be revisited soon so I left them for later, like `Pagination`)


src/components/Card/Card.tsx
```diff
@@ -2,41 +2,36 @@ import * as React from "react";
 import bem from "../../utils/bem";
 
 interface CardProps {
-    /** ID  */
-    id?: string;
-
+    /** Additional attributes passed to the card */
+    attributes?: {};
```
hokei: Where are attributes used in `card` ? 

In general, if something is passed in the interface but not passed in the class itself (L27-36 below) it's not being used in the class.  I think a lot of classes here have attributes declared but not used

src/components/Checkbox/Checkbox.tsx
```diff
@@ -3,21 +3,22 @@ import bem from "../../utils/bem";
 import Label, { LabelOptions } from "../Label/Label";
 
 interface CheckboxProps {
-    /** name of the checkbox */
-    name?: string;
-
+    /** Additional attributes passed to the checkbox */
+    attributes?: {};
```
hokei, line 7: These are unused

src/components/Button/Button.tsx
```diff
@@ -23,52 +23,37 @@ export interface ButtonOptions {
...
 
 interface ButtonProps {
-    /** ID that other components can cross reference for accessibility purposes */
-    id?: string;
-
-    /** The action to perform on the <button>'s onClick function */
-    onClick: (event: React.MouseEvent | React.KeyboardEvent) => void;
-
     /** Additional attributes passed to the button */
     attributes?: {};
```
hokei, line 27: These aren't used in `button`

hokei:
I think we should be really clear on what `attributes` are really doing.  
I see in https://github.com/NYPL/nypl-design-system/pull/276/files#diff-eaf22cccaf0b1e13dc39209235b7229fR37
that what this is doing is turning 

```
<DS.Modal 
extraProp={extraValue} /> 
```

into 
```
<Modal>
<div extraProp={extraValue} > { rest of modal content ] </div>
</Modal>
```

Is that functionally different than this?
```
<Modal>
<div className="modal" tabIndex=0 >
<div extraProp={extraValue} > { rest of modal } </div>
</div>
</Modal>
```

which is what is (currently) supported going 
```
<Modal> <div extraProp={extraValue} > { rest of modal } </div> </Modal>
```
I'm legitimately asking.  In my experience extra divs are clutter, but if there's a reason why we have to put the props on the same div that's getting the tabindex, etc.  that changes things.  

Contrast that with `Link` where if you need to attach something to the `a` element you have no way to do so.  You can't pass an anchor into another anchor.

Does that make sense?

hokei (changes requested):
I think this needs another pass 1) to make sure that when attributes are passed in, they're actually used and 2) to make sure that attributes are only on elements where a consumer using the component can't pass the attribute onto a thing that they control, and pass that whole thing in as `jsx`.  

For the un-refactored components, we can keep them on the divs, especially if those divs are things you can't control.

src/components/Hero/Hero.tsx
```diff
@@ -110,9 +109,12 @@ export default function Hero(props: React.PropsWithChildren<HeroProps>) {
 
     return (
         <div
-            className={bem(heroBaseClass, heroModifiers)}
+            className={bem(heroBaseClass, heroModifiers, blockName, [
+                className,
+            ])}
             data-responsive-background-image
             style={backgroundImageStyle}
+            {...attributes}
```
hokei: (I think) this is an example of a div where this makes sense.  You aren't really allowed to pass whatever you want into `<Hero>`, it's supposed to generate the div for you, so if you want more functionality this is your only choice.

src/components/Button/Button.tsx
helenvholmes: When you say they're not used in `button`, what do you mean? Meaning L157 isn't doing anything?

src/components/Checkbox/Checkbox.tsx
helenvholmes: Removing them from the wrapping div into the checkbox `input` based on our conversation yesterday—is that what you mean by unused?

src/components/Card/Card.tsx
helenvholmes: Whoops! Thanks for catching that—after adding in so many of these it's inevitable stuff would fall through the cracks :)

src/components/Button/Button.tsx
helenvholmes: So weird—I don't see them in Github's highlighted changes, but I am applying them on L157... so weird 🤔

src/components/Hero/Hero.tsx
helenvholmes: Is this different from what `Modal` is doing? `Modal` also generates a wrapping div for you that you otherwise cannot add attributes to.

src/components/Button/Button.tsx
hokei: wait whaaaaaaaaatttttt

src/components/Checkbox/Checkbox.tsx
hokei: aaaah either I was blind yesterday or github is now showing highlighted changes that it wasn't showing yesterday.  It looked like same problem here: https://github.com/NYPL/nypl-design-system/pull/276#discussion_r454680818 but I don't see the problem anymore

src/components/Hero/Hero.tsx
hokei: I think the difference is that in `Modal` you're expected to pass your own content that you control.  In `Hero` you are supposed to pass 

```
<Hero
  backgroundImageSrc="https://placeimg.com/1600/800/arch"
  heading={<Heading blockName="hero" id="1" level={1} text="Hero Primary"/>}
  heroType="PRIMARY"
/>
```

Importantly there's no way to pass 
```
<Hero>
{Custom Hero with Custom content}
</Hero>
```

Whereas in Modal you are always expected to pass 

```
<Modal> {Custom Content} </Modal>
```

hokei: That means if you wanted to put a customProp, you could go 

`<Modal> <div customProp=customFunction /> </Modal>` 

but with Hero where would you put it?

src/components/Accordion/Accordion.tsx
```diff
@@ -4,20 +4,20 @@ import bem from "../../utils/bem";
...
     accordionLabel?: React.ReactNode;
-
+    /** Additional attributes passed to the accordion */
+    attributes?: {};
+    /** Optional blockName for use with BEM. See how to work with blockNames and BEM here: http://getbem.com/introduction/ */
+    blockName?: string;
+    /** className you can add in addition to 'input' */
+    className?: string;
+    /** ID that other components can cross reference for accessibility purposes */
+    id?: string;
```
EdwinGuzman, line 14: I don't think this is being used in this component.

src/components/Button/Button.tsx
EdwinGuzman: :gasp:

EdwinGuzman (commented):
I think overall this looks good but it could use some tests to make sure that the attributes passed in the `attributes` prop get passed down and work as intended.

My only concern about this is opening it up to any kind of attribute could potentially lead to bad patterns we don't want others to use. For example, maybe this is passed down to a component:
```
attributes={{ onClick: () => {...} }}
```
and that component renders a `div`. `onClick` should not be used for `div`s. Maybe some documentation to point users or remind them to pass down accessible attributes for the right DOM element.

I was thinking about this when I was doing it for the `Input` element but it's okay for that since that component can use an `onChange` or an `onBlur`. Maybe each `attributes` declaration can be more specific about it's type - rather than `{}` it can be something like `HTMLDivElement`. I haven't tested it but it _might_ work well.

src/components/Checkbox/Checkbox.tsx
helenvholmes: You're not crazy on all of them! Some of them I pushed up with [cfb38a9](https://github.com/NYPL/nypl-design-system/pull/276/commits/cfb38a952bf4c69d203e4c1b71d4095c789c615f) but some yeah, Github isn't highlighting changes 😱

src/components/Button/Button.tsx
helenvholmes: No me gusta ☹️

src/components/Hero/Hero.tsx
helenvholmes: I see what you're saying with Modal—you can always pass in customized children, so you'd probably never need to use attributes. But isn't the whole point of attributes to acknowledge that we can't anticipate everyone's needs? By adding attributes to Modal, in the event that someone _does_ need to add an attribute to the Modal where `tabindex` is, they can. It also means everything is extensible in the same way more or less. It could be I'm being influenced by PatternLab's philosophy here (they universally add attributes to everything without making distinctions). Is there harm to doing it this way? (brb, gonna go read Edwin's comments.)

helenvholmes:
> I think overall this looks good but it could use some tests to make sure that the attributes passed in the `attributes` prop get passed down and work as intended.

This makes sense. I imagine we'll want it in every test which makes me sad, but it makes sense 😂 I think it'll go quickly

> My only concern about this is opening it up to any kind of attribute could potentially lead to bad patterns we don't want others to use. For example, maybe this is passed down to a component:
> 
> ```
> attributes={{ onClick: () => {...} }}
> ```
> 
> and that component renders a `div`. `onClick` should not be used for `div`s. Maybe some documentation to point users or remind them to pass down accessible attributes for the right DOM element.

It does and will introduce this possibility. Perhaps each component's `attributes{}` documentation points to the MDN documentation for the relevant HTML tag?

hokei:
> My only concern about this is opening it up to any kind of attribute could potentially lead to bad patterns we don't want others to use

I think as of now we've given up on Design System trying to prevent you from doing invalid things.  We guarantee that the examples in the Storybook are well formed, and then if you want to do something weird to your component, that's on you.  (eg: we allow you to embed images into buttons, or embed a Hero into a Card if you want, but ... why)  

Personally my view towards it is that it's OK to give engineers the ability to hack together solutions as long as they know that that's what they're doing (Like, passing in attributes is an implicit acknowledgement of, "Hi, You are outside the anticipated use realm of design system")

EdwinGuzman:
@scooooooooby I don't think you may need to test _every_ single component that uses attributes. I think some of the general attributes can be generalized for more than one component since the pattern is the same but the more complicated ones like event handlers would be good.

@hokei I do agree that we should give engineers the ability to do what they want, but it may not be useful if a third party is making an app _for_ NYPL. In that case we would need some enforcement or guidelines (which I'm sure we'll have anyway).

helenvholmes:
> @scooooooooby I don't think you may need to test _every_ single component that uses attributes. I think some of the general attributes can be generalized for more than one component since the pattern is the same but the more complicated ones like event handlers would be good.

Phew. Okeydoke.

> @hokei I do agree that we should give engineers the ability to do what they want, but it may not be useful if a third party is making an app _for_ NYPL. In that case we would need some enforcement or guidelines (which I'm sure we'll have anyway).

We have other methods for guidelines that I think are better than trying to enforce it in the code (BRDs, UX documentation, VPATs).

src/components/Hero/Hero.tsx
hokei: I wonder if it might be better to only add attributes to components that can take event handlers.  When working in DRB I have only found that I needed attributes in `<input>` or `<button>` or `<a>` 

Like, the extra className handles the most common case to put anything in `<div>` right?  

I say this because once we add `<div>`s stuff gets complicated?  
Let's go back to the Hero example - it compiles down to 

```
<div class="hero hero--primary" 
data-responsive-background-image="true" s
tyle="background-image: url(&quot;https://placeimg.com/1600/800/arch&quot;);"
{attributes will go here if this is merged as is}> <---
<div class="hero__content"
{But what if I want attributes here? where would I pass that?}> <---
<h1 class="hero__heading" id="1" 
{This is ok because the header is passed in as a separate element}>Hero Primary</h1>
</div>
</div>
```

helenvholmes: I see what you're saying-basically that if you're not passing attributes to the "lowercase" component, as we've been calling it, it isn't as easily identifiable where those attributes are getting added. By not including them on such components, we feel the API is more clear. Am I summing this up well?

hokei: yup!  That's exactly it.  
I think to keep it simple, let's only put attributes in `lowercase` components for now.  If people need them for divs, we will revisit once we have clear use cases. If we ever have components with more than one "lowercase" component in them, that's a sign we need to allow them to be passed in independently.  (I don't think this is the case with any of the refactored components)

src/components/HelperErrorText/HelperErrorText.tsx
```diff
@@ -2,30 +2,29 @@ import * as React from "react";
...
-    blockName?: string;
-    baseClass?: string;
 }
 
 interface HelperErrorTextProps {
+    /** Added prop when HelperText is errored */
+    ariaAtomic?: boolean;
+    /** Added prop when HelperText is errored */
+    ariaLive?: boolean;
+    /** Additional className to add to the helperErrorText */
```
hokei, line 18: Nit, add "Optional" like we do with blockNames

src/components/Input/Input.tsx
```diff
@@ -3,44 +3,32 @@ import bem from "../../utils/bem";
...
+    id?: string;
     /** Adds the 'disabled' prop to the input when true */
     disabled?: boolean;
-
-    /** Populates the value of the select */
-    value?: string | number;
-
+    /** Helper for modifiers array; adds 'errored' styling */
+    errored?: boolean;
+    /** Modifiers array for use with BEM. See how to work with modifiers and BEM here: http://getbem.com/introduction/ */
```
hokei, line 22: Nit: add "Optional"

src/components/SearchBar/SearchBar.tsx
```diff
@@ -30,28 +24,28 @@ export default function SearchBar(
...
+            className={bem(searchbar__base_class, modifiers, blockName, [
+                className,
+            ])}
             id={id}
             role="search"
             aria-label={ariaLabel}
             aria-labelledby={ariaLabelledBy}
-            onSubmit={onSubmit}
-            {...attributes}
+            onSubmit={props.onSubmit}
```
hokei: use onSubmit specified in L32 instead of calling props directly

src/components/SearchBar/SearchBar.tsx
```diff
@@ -30,28 +24,28 @@ export default function SearchBar(
...
         >
-            {children}
+            {props.children}
```
hokei, line 48: same as above

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -10,7 +10,14 @@ function isTextBreadcrumb(obj: breadcrumb | JSX.Element): obj is breadcrumb {
 }
 
 export interface BreadcrumbProps {
+    /** Optional blockName for use with BEM. See how to work with blockNames and BEM here: http://getbem.com/introduction/ */
+    blockName?: string;
+    /** Breadcrumb links as an array */
     breadcrumbs: breadcrumb[] | JSX.Element[];
+    /** className you can add in addition to 'input' */
```
hokei, line 17: specify that this is optional

src/components/Checkbox/Checkbox.tsx
```diff
@@ -3,37 +3,36 @@ import bem from "../../utils/bem";
...
     blockName?: string;
-    /** Optional modifiers array for use with BEM. See how to work with modifiers and BEM here: http://getbem.com/introduction/ */
-    modifiers?: string[];
-
     /** Unique ID used by checkbox */
     checkboxId?: string;
-    /* The Label that the checkbox is using. */
-    labelOptions: LabelOptions;
-
+    /** ClassName that appears in addition to "checkbox" */
```
hokei, line 12: specify optional

hokei (changes requested):
It's pretty much good, just some nitty things.  

With the optionality, either we should say in all classNames, blockNames, modifiers that it's optional, or we should say it in none of them.  It's confusing if some blockNames specifically call out that it's optional, and others don't.

helenvholmes:
@hokei I've removed the Optional language, since as you've said it's in the propTypes table

src/components/Button/Button.tsx
```diff
@@ -23,52 +23,37 @@ export interface ButtonOptions {
...
-    /** The name of the icon.  Corresponds with the name of the icon's svg file */
-    iconName?: string;
-
-    /** Optional amount of degrees to rotate icon */
-    iconRotation?: IconRotationTypes;
-
-    /** Optional modifiers array for use with BEM. See how to work with modifiers and BEM here: http://getbem.com/introduction/ */
-    iconModifiers?: string[];
-
+    /** Optional additional className for use with BEM. See how to work with blockNames and BEM here: http://getbem.com/introduction/ */
```
hokei: nit

src/components/Placeholder/Placeholder.tsx
```diff
@@ -2,8 +2,10 @@ import * as React from "react";
 import bem from "../../utils/bem";
 
 export interface PlaceholderProps {
-    modifiers?: string[];
+    /** Optional blockName for use with BEM. See how to work with blockNames and BEM here: http://getbem.com/introduction/ */
```
hokei: nit

hokei (approved):
last nits, but you're good!  :tada:

PR #287, opened 2020-07-22
Select component update

### LINKED ISSUE #285: Updated Select component
### Component in question:
- Select

### Description of issue:
- The `Select` component is missing a few properties that will allow it to work inside any form. Specifically, there should be a `name` prop and it should allow for React `ref`s as props to use with other libraries.

EdwinGuzman:
Issue number: #285 

- This PR updates the `Select` by giving it a `name` prop and implement it with `ForwardRef`. All HTML form elements should have a `name` attribute so that form submissions can pick up the value. The `forwardRef` is for libraries that need it, such as `react-hook-form` or for any other ref that a dev wants to pass.

There are some components that use the `Select` component so I had to update those components simply by adding a `name` prop.

I converted the `Select` component to a functional component and used the `useState` hook to keep track of the selected value.

### Front End Review:

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -23,8 +23,9 @@ export interface SelectProps {
     labelId?: string;
     /** Modifiers array for use with BEM. See how to work with modifiers and BEM here: http://getbem.com/introduction/ */
     modifiers?: string[];
-    /**  */
+    /** The name of the select element to use in form submission */
     name: string;
+    /** Passes selects' current value to the React state handler */
     onBlur?: (event: React.FormEvent) => void;
     /** Passes selects' current value to the React state handler */
     onChange?: (event: React.FormEvent) => void;
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -56,7 +56,9 @@ const Select = React.forwardRef<HTMLSelectElement, SelectProps>(
             if (target) {
                 setSelectedOption(target.value);
             }
-            additionalChange && additionalChange(event);
+            if (additionalChange) {
+                additionalChange(event);
+            }
         };
         const {
             ariaLabel = null,
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -1,6 +1,5 @@
-import * as React from "react";
+import React, { useState } from "react";
 import bem from "../../utils/bem";
-import Icon from "../Icons/Icon";
 
 export interface SelectProps {
     /** When passed, will populate the aria-label on the select */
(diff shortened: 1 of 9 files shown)


src/components/Select/Select.tsx
```diff
@@ -11,44 +10,54 @@ export interface SelectProps {
...
-        additionalChange(event);
-    }
-
-    render() {
+/**
+ * Select
+ * A component that renders a `select` DOM element along with its `option`
+ * children.
+ */
+const Select = React.forwardRef<HTMLSelectElement, SelectProps>(
```
hokei, line 41: Is there an argument to just go through all the form elements in one go and enable ref forwarding in all of them?  Or is it ok to just add them as we need them?  (this isn't a breaking change so I guess there's not an urgent need to get them all in at once ... ) Which are the components that this would affect?

src/components/Select/Select.tsx
```diff
@@ -85,18 +95,22 @@ export default class Select extends React.Component<
 
         return (
             <select
+                name={name}
                 {...selectProps}
```
hokei: I know this wasn't your work, but since you're here anyway I've been trying to move away from this and pass in props explicitly (since it's easier to read) - can you make that change here?

src/components/Select/Select.tsx
```diff
@@ -85,18 +95,22 @@ export default class Select extends React.Component<
...
+                ref={ref}
+                onBlur={(e) => onSelectChange(e, onBlur)}
+                onChange={(e) => onSelectChange(e, onChange)}
             >
                 {React.Children.map(children, (child, key) =>
                     React.cloneElement(child as React.ReactElement<any>, {
                         "aria-selected":
-                            children[key].props.children ===
-                            this.state.selectedOption,
+                            children[key].props.children === selectedOption,
```
hokei, line 107: Perhaps a better question for @scooooooooby : 
So from the stories, we're supposed to use this like 
```
<Select>
      <option aria-selected={true}>
                {text("Option One", optionsGroup[0])}
            </option>
</Select>
```

but is ` aria-selected={true}` actually doing anything?  Isn't this line overriding `aria-selected={true}` or is something else going on here?

src/components/Select/Select.tsx
EdwinGuzman: Right now, the `Pagination` and `SearchBar` components import `Select` but this update doesn't affect it. I could go through the rest of the components but since the PR is for the select element, I just focused on that. I figured if more components need to be updated that can be a different PR.

src/components/Select/Select.tsx
EdwinGuzman: So remove `...selectedProps` and put each prop individually?

src/components/Select/Select.tsx
EdwinGuzman: I was a bit unsure about this at first, but then I noticed that it was updating `aria-selected` correctly. I can't say I fully know how this works.

src/components/Select/Select.tsx
hokei: yup!

src/components/Select/Select.tsx
hokei: Oh I mean like would you recommend a design system ticket that's like "go through all the components and add refs to places that need it" or is it ok to do ad-hoc

EdwinGuzman: I think it might be good to do ad-hoc because being able to use and test it in other components makes it easier to know that it's working as it should. I don't want to introduce a bug updating components and not testing it outside of the repo.

src/components/Select/Select.tsx
EdwinGuzman: Done.

src/components/Select/Select.tsx
helenvholmes: `aria-selected` that you pass in the `option` sets the first selected state that gets updated onChange from the user, if I'm remembering correctly.

Our a11y checker gets upset if they're not applied to all of the `option`s, which when we moved to passing them in as children I had to add manually... I'm sure someone smarter than me could figure out how to automate it for the consumer with `React.cloneElement` though 😂

helenvholmes (approved):

Issue #281, opened 2020-07-20
SCSS variables aren't exposed to consumers


helenvholmes:
### Description of issue:
- SCSS variables are compiled down to their values in the final `.css` file, meaning consumers don't have access to those values. We should upgrade to using CSS variables so they're exposed in the final `.css` file.

PR #297, opened 2020-07-28
CSS variables

helenvholmes:
Issue number: #281 

- Changes SASS variables to CSS Variables
- Sets up linter to work with CSS Variables
- Adds stories for CSS Variables

### Front End Review:
- [ ] View [Breakpoints in Storybook](http://localhost:9001/?path=/story/breakpoints--breakpoints)
- [ ] View [Colors in Storybook](http://localhost:9001/?path=/story/colors--colors-brand)
- [ ] View [Typography in Storybook](http://localhost:9001/?path=/story/typography--typefaces)
- [ ] View [Spacing in Storybook](http://localhost:9001/?path=/story/spacing--spacing)


src/components/Colors/Colors.stories.tsx
```diff
@@ -0,0 +1,254 @@
+import * as React from "react";
```
EdwinGuzman, line 1: This page is 🔥

src/components/Colors/Colors.stories.tsx
```diff
@@ -0,0 +1,254 @@
...
+import getCSSVariable from "../../helpers/getCSSVariable";
+import UIDocCard from "./UIDocCard";
+import Card from "../Card/Card";
+import List from "../List/List";
+import Heading from "../Heading/Heading";
+import { ListTypes } from "../List/ListTypes";
+
+export default {
+    title: "Colors",
+    decorators: [(storyFn) => <div style={{ padding: "2%" }}>{storyFn()}</div>],
```
EdwinGuzman: I forgot how but you can make this a global decorator for all stories so you don't have to duplicate this in every story.

EdwinGuzman: Try adding this to `.storybook-react/config.js`.

```
export const CenterDecorator = storyFn => <div style={styles}>{storyFn()}</div>;
addDecorator(CenterDecorator);
```

src/components/Colors/Colors.stories.tsx
```diff
@@ -0,0 +1,254 @@
...
+};
+
+const uiVariables: { [k: string]: string } = {};
```
EdwinGuzman: `{ [k: string]: string }` comes up a lot throughout this PR. It'd be good to have this as a typescript `interface` probably in `/src/interfaces.js` and import it wherever you need it.

src/helpers/getCSSVariable.tsx
```diff
@@ -0,0 +1,7 @@
+export default function getCSSVariable(name: string) {
+    let propertyValue = "";
+    propertyValue = getComputedStyle(document.documentElement).getPropertyValue(
+        name
+    );
```
EdwinGuzman: You can make the variable `const` and at the end of line 5 do `|| "";`. So if `getComputedStyle` doesn't return a value, then the fallback is the empty string.

src/components/Colors/UIDocCard.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from "react";
+import Card from "../Card/Card";
+
+interface UIDocCardProps {
+    /** backgroundColor of the doc card */
+    backgroundColor: string;
+}
+
+export default function UIDocCard(
```
EdwinGuzman, line 13: Is this meant just to be used for Storybook?

src/helpers/CSSVariablesHelper.tsx
```diff
@@ -0,0 +1,34 @@
+let cssVariables: {
+    [k: string]: string;
+} = {};
+
+let allCSS = [].slice
```
EdwinGuzman: This is just for the stories, right?

EdwinGuzman: I think you can just directly call `[].slice....` to start off the process without having to assign it to a variable you're not going to use.

I also suggest to make the inner `[].slice` call a separate function for readability.

src/components/Colors/Colors.stories.tsx
```diff
@@ -0,0 +1,254 @@
...
+
+for (const [key, value] of Object.entries(grayScaleVariables)) {
+    grayscaleDocs.push(
+        <li style={{ textAlign: "center" }}>
+            <UIDocCard backgroundColor={value}>
+                {key}: {value}
+            </UIDocCard>
+        </li>
+    );
+}
```
EdwinGuzman: Not necessary for this PR, but the code for these two loops are very similar. You can make this into a function where you pass in the object and the array (ie `makeUIDocCard(grayScaleVariables, grayscaleDocs)`).

src/styles/01-colors/_colors-brand.css
```diff
@@ -0,0 +1,23 @@
+/** Brand Colors */
+:root {
+    /** What's On is used for Exhibitions & Events. */
+    --section-whats-on-primary: #8a1a9c;
+    --section-whats-on-secondary: #7c1097;
```
EdwinGuzman, line 5: Not necessarily a question about the PR but about the colors: as a dev, is it acceptable to use any of these colors for anything that's not for these specific pages? In other words, can these also be used as utility colors?

src/components/Colors/Colors.stories.tsx
helenvholmes: 🙇‍♂️ thank you!

Do you know I _unset_ it for certain stories? (`Hero` and `Breadcrumbs` should span full width, for example.)

EDIT: JK I can look this up

src/styles/01-colors/_colors-brand.css
helenvholmes: My _current_ understanding is no, so i think it would be good to express that in the notes somehow. (NYPL theming is a moving target unfortunately.) I'm also thinking that forward-facing, we'll want to expose these values not as NYPL names but maybe as more generic `--theme-1-primary`, `--theme-1-secondary`, etc. for external apps like SimplyE and Lyrasis. Kinda tempted to bundle that in with this work so we can have one big change all at once, but I don't think it would be breaking to do that in a follow-up. I'll log a ticket

EDIT: Filed #304

src/components/Colors/UIDocCard.tsx
helenvholmes: Yeah, it is... should I maybe add Storybook to the name of it so that's clear?

EdwinGuzman:
Two things:

1) The older `nypl` colors are no longer available. Are those coming back? I am currently doing:
```
&.available {
    // From DS
    // $nypl-green-dark: #497629
    color: #497629;
  }
  &.unavailable {
    // From DS
    // $nypl-red-regular: #d0343a;
    color: #d0343a;
```
but there's no css variable for either of those colors.

2) The bundle doesn't work. When I ran `npm run watch-webpack` or `npm run dist`, the bundle wasn't picked up correctly by the Library Card app. It gave me 404s on all the css files. This is because of how `@import ".... .css"` works. Apparently importing .css files does not work the same way as importing .scss files.

When you import a .css file it converts it to an absolute file path that breaks when importing it into another app.
```
@import "./styles/01-colors/_colors-brand.css";
```

gets converted by webpack to 
```
@import url(/Users/edwinguzman/projects/nypl-design-system/src/styles/01-colors/_colors-brand.css);
```

The consuming app doesn't know what to do with that:
Screen Shot 2020-07-29 at 11 10 44 AM

I tried to:
- update the webpack bundler so it knows about `.css` files but there's no entry point for css files so this doesn't seem to work.
- try to minify the css file and then combine it with the scss bundle but this is harder than I thought it would be and I didn't want to spend too much time on this... but it might be the better solution.

What does work:
- convert the `.css` files to `.scss` files. The minified bundle works well in the Library Card app and I can do something like:
```
#patronFirstName-label {
  color: var(--ui-success);
}
```

The linter didn't give me any issues but I didn't try to run tests or commit (since that's when the linter probably picks up if this is wrong). As it stands, this needs a bit of work but I'm not sure what fix you want to implement or look into.

EdwinGuzman (changes requested):
Marking as request changes due to broken minified file.

src/components/Colors/UIDocCard.tsx
EdwinGuzman: No. Just a comment above the function is enough.
```
/**
 * UIDocCard
 * Component only used for Storybook....
 */
export default function UIDocCard( .....
```

helenvholmes:
@EdwinGuzman thanks for all the info on minification. I'll work on that here as well. Sounds like the easiest is to convert them back to `.scss`, which shouldn't be challenging. (She says, optimistically.)

To answer your question about `$nypl-color-brightest`, this is a breaking change where these are going away, being replaced by the `--section-sectionName-level` variables. The snippet you've linked to would need to be updated to

```
$nypl-green-dark: #497629 --> var(--ui-success);
$nypl-red-regular: #497629 --> var(--ui-error);
```

Bad news is breaking change. Good news is you'll be able to hook into variables now instead of copying hex values.

.storybook-react/config.js
```diff
@@ -1,4 +1,5 @@
 import "!style-loader!css-loader!sass-loader!import-glob-loader!../src/styles.scss";
+import * as React from "react";
```
hokei, line 2: is this used?

hokei:
> Sounds like the easiest is to convert them back to .scss,
This sounds really strange to me, aren't we hopefully working towards a world where we can drop `scss`? 

A quick google is not turning up anything that allows sharing of css variables directly, between modules, but I'll keep looking?  It seems weird that no one would need this ... 
Alternately, do you think this is a reason to start looking into postcss or something in the "export these values as js" approach rather than an "export these values as sass" approach?

helenvholmes:
@hokei yeah, we should definitely be thinking about PostCSS for this. You're right, switching them back to `.scss` is not going to work now that I've dived in, and was a gut reaction to doing the easiest thing... c'mon, Helen.

I'll see if I can get that working to load CSS files.

.storybook-react/config.js
EdwinGuzman: It's needed whenever you use jsx in a file.

EdwinGuzman:
Yeah, me converting the `.css` files back to `.scss` files was just to test and make sure that it was the `@import` statements that were not working. It was also the easiest solution to the problem but I didn't dive too deep into a solution. I briefly did see something about postcss but the fix seemed too long and I had other things to review/do :(

helenvholmes:
Unfortunately it means that there's no way to catch unknown values... which is gonna be real rough when developing. So it looks like PostCSS is what we'll have to do.

helenvholmes:
Turns out there was a simpler answer.

According to this [StackOverflow question](https://stackoverflow.com/questions/7111610/import-regular-css-file-in-scss-file), by omitting the file extension libsass (node-sass) will properly handle the imports.

What a strange little edgecase. Maybe explicitly calling out `.css` is libsass' way of saying "Let the consumers' loaders handle it". Idk. Anyway, I've tested this against a [demo app](https://github.com/scooooooooby/bookbox) I have and everything is looking okay now.

EdwinGuzman (approved):
Tested and working in LCA.

PR #298, opened 2020-07-28
Fix #295

### LINKED ISSUE #295: Remove width: 100% on svg
### Component in question:
- global css

### Description of issue:
- The global css used by the DS sets all svg to have `width: 100%`. This is messing up my local styles in CPW. In my opinion, we should try to stay clear of using global selectors like that unless necessary, and ideally we would isolate the NYPL design system css to solely the imported components. You could still provide a global css setting for if a consumer specifically wants that though.

helenvholmes:
Issue number: #295

- Removes the `width: 100%` on SVGs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 Currently, this repo is in Prerelease. When it is released, this project will adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 ========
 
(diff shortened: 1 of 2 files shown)


kristojorg:
Does removing it here without adding it back in a more specific selector not break something? If this works for you then it works for me!

kristojorg (approved):
So long as this doesn't break something for you, I'm good with it : )

helenvholmes:
@kristojorg Fairly certain this got added while I was testing something—but you're right, to be safe I'll bump minor (our major).

PR #307, opened 2020-07-30
Autosuggest component

### LINKED ISSUE #278: Autosuggest

EdwinGuzman:
Issue #278.

- This is a WIP and POC to test out the [`react-autosuggest`](https://github.com/moroshko/react-autosuggest) library. I did not make it very generic but generic enough for two small examples. It requires a few function configurations/preferences but seems easy to manipulate. A lot more can be added like a scrollbar window so the suggestions don't render below the input, but that can be added later (as well as types and more generic inputs). There's no styling but this is mainly for functionality to see how we all feel about this.

The main configuration for this is that it reuses `Label`, `Input`, and `HelperErrorText` components from DS.

Let me know what you think and how we can continue forward for both the Location Finder and Library Card app.

### Front End Review:
- [ ] View [the Autosuggest example in Storybook](http://localhost:9001/?path=/story/autosuggest--autosuggest-library)


src/components/Autosuggest/Autosuggest.stories.tsx
```diff
@@ -0,0 +1,109 @@
...
+    <span>
+        {suggestion}
+        <Icon name={IconNames["check"]} decorative={true}></Icon>
+    </span>
+);
+
+// When suggestion is clicked, Autosuggest needs to populate the input
+// based on the clicked suggestion. Teach Autosuggest how to calculate the
+// input value for every given suggestion.
+const getSuggestionValueFish = (suggestion) => suggestion;
```
helenvholmes: Fish?

helenvholmes: Oh, like fishing... for the value? 🎣

src/components/Autosuggest/Autosuggest.tsx
```diff
@@ -0,0 +1,83 @@
+import React, { useState } from "react";
+import Autosuggest from "react-autosuggest";
+import Label from "../Label/Label";
+import Input from "../Input/Input";
+import { InputTypes } from "../Input/InputTypes";
+import HelperErrorText from "../HelperErrorText/HelperErrorText";
+
```
hokei: If we go forward with this, please define proptypes.

src/components/Autosuggest/Autosuggest.tsx
```diff
@@ -0,0 +1,83 @@
...
+const AutosuggestComp = ({
+    defaultValue,
+    childRef,
+    list,
```
hokei: is list going to be ok with just being a list of strings?  We won't ever want an autosuggest list with half-bolded elements, etc?

src/components/Autosuggest/Autosuggest.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+    const renderInputComponent = (inputProps) => {
+        return (
+            <>
+                <Label
+                    htmlFor="input-autosuggest"
+                    id="input-autosuggest-label"
+                    optReqFlag={isRequired ? "Required" : ""}
+                >
+                    <span>{label}</span>
```
hokei: nit:  Fragment instead?

src/components/Autosuggest/Autosuggest.tsx
```diff
@@ -0,0 +1,83 @@
...
+    const onChange = (event, { newValue }) => setValue(newValue);
+
+    // Autosuggest will call this function every time suggestions need to be
+    // updated. `getSuggestions` must be passed.
+    const onSuggestionsFetchRequested = ({ value }) =>
+        setSuggestions(getSuggestions(value, list));
+
+    const onSuggestionsClearRequested = () => setSuggestions([]);
+
+    // Autosuggest will pass through all these props to the Input component.
```
hokei: This looks like it might go against the composeability principles - we really want a compound component to allow the user to pass in their own `<Label>`, `<Input>` and `<HelperErrorText>` rather than having a component that's doing them for you (for example, what happens if the user doesn't want a label?)  

To this end, I think this should be a standalone `<Input-autosuggest>` component that doesn't reference `<Input>` (or they both reference the same helpers)

src/components/Autosuggest/Autosuggest.stories.tsx
```diff
@@ -0,0 +1,109 @@
...
+const getSuggestionsFish = (value, list) => {
+    const inputValue = value.trim().toLowerCase();
+    const inputLength = inputValue.length;
+
+    return inputLength === 0
+        ? []
+        : list.filter((l) => l.indexOf(inputValue) !== -1);
+};
+
+const fishList = [
```
hokei: Clearly I haven't played enough animal crossing, I don't know half these fish

src/components/Autosuggest/Autosuggest.tsx
hokei: There's a part of me that really wants 
`<Input autosuggest={true}         
       getSuggestionValue={blah}
        getSuggestions={blah2}
        list={listOfThings}  />` but it might lead us back to the "too many props" problem we had before

src/components/Autosuggest/Autosuggest.stories.tsx
helenvholmes: I haven't played since the diving update. Am I even an AC stan at this point? Yeesh.

EdwinGuzman:
I can update this just to show _how_ to use `auto-suggest` rather than expose a wrapper component.

src/components/Autosuggest/Autosuggest.stories.tsx
EdwinGuzman: 👎

src/components/Autosuggest/Autosuggest.tsx
EdwinGuzman: I like `<>` over `<React.Fragment>`....

src/components/Autosuggest/Autosuggest.tsx
EdwinGuzman: It can be anything. I didn't write an interface for this though it probably would've been something like `:any`.

src/components/Autosuggest/Autosuggest.tsx
helenvholmes: I think @hokei means in favor of `<span>`?

(As a note, `<>` will generate a `<div>`. Divs are block elements and will go to the next line. Spans are inline elements and will not.)

EdwinGuzman:
@hokei @scooooooooby I added this repo to an NYPL Netlify account so we can have PR/feature preview demo sites. Check the "Details" link in the `netlify/stoic-murdock-c7f044/deploy-preview — Deploy preview ready!` link. That's the autogenerated name but I think we can update it if we want to, but it's just a PR/demo link so it's not that important.

Now we can say, "review https://deploy-preview-307--stoic-murdock-c7f044.netlify.app/?path=/story/autosuggest--autosuggest-library".

src/components/Autosuggest/Autosuggest.tsx
EdwinGuzman: I thought that fragments don't generate the extra nodes, so it shouldn't generate a `<div>`.

helenvholmes:
@EdwinGuzman holy cow!! I'd been putting off multidevs indefinitely because I thought they were going to be really complicated to set up. Thank you for doing this, this is really neat!

> Check the "Details" link in the `netlify/stoic-murdock-c7f044/deploy-preview — Deploy preview ready!` link. 
Took me a second to understand this, but neat. That's pretty darn easy. I wonder if we can autogenerate this into the PR template...

> That's the autogenerated name but I think we can update it if we want to, but it's just a PR/demo link so it's not that important.
It'd be nice to rename this "nypl-design-system" or something easier to remember than "stoic-murdock-c7f044". I think my Netlify account is squatting that... let me dump it so you can have it.

The autosuggest component itself looks really nice too! Pinging @nypl-wluisi—you've already written the CSS for this component, is that correct? Can we get you to push that up to this PR?

[EDIT] Looks like nypl-design-system.netlify.app is free.

nypl-wluisi:
This is what I have for the CSS. It looks ok on the frontend, but kind of messy. We're going to clean it up more next week:

```
// React Autosuggest
.react-autosuggest__container {
  position: relative;
  display: inline;
}

.react-autosuggest__input {
  width: 618px;
  height: 34px;
  padding: 10px;
  font-size: 16px;
  border: 1px solid #aaa;
}

.react-autosuggest__input--focused {
  outline: none;
}

.react-autosuggest__input--open {
  border-bottom-left-radius: 0;
  border-bottom-right-radius: 0;
}

.react-autosuggest__suggestions-container {
  display: none;
}

.react-autosuggest__suggestions-container--open {
  display: block;
  position: absolute;
  width: 618px;
  border: 1px solid #aaa;
  background-color: #fff;
  font-family: Helvetica, sans-serif;
  font-weight: 300;
  font-size: 16px;
  border-bottom-left-radius: 4px;
  border-bottom-right-radius: 4px;
  z-index: 2;
}

.react-autosuggest__suggestions-list {
  margin: 0;
  padding: 0;
  list-style-type: none;
}

.react-autosuggest__suggestion {
  cursor: pointer;
  padding: 10px;
}

.react-autosuggest__suggestion--highlighted {
  background-color: #ddd;
}

.auto-suggest-bottom {
  padding: .5em;
}
```

@EdwinGuzman @scooooooooby

helenvholmes:
@nypl-wluisi Thanks Bill!

@EdwinGuzman, I've pushed up Bill's CSS with some tweaks.

@hokei, I think this is ready for another round of review.

EdwinGuzman:
Thanks. The updated CSS looks good!

helenvholmes (approved):
This looks great to me! Thanks for this great work, Edwin.

nypl-wluisi:
@hokei @scooooooooby @EdwinGuzman This looks good to me. My only question would be since `react-autosuggest` is only used for demo purposes in the storybook, can we exclude that code from the JS bundle that DS generates for other apps to consume? It might already do this, and if it doesn't, then maybe we merge now and circle back on this later?

nypl-wluisi (approved):
Looks good!

PR #316, opened 2020-08-04
Adds logos for NYPL, Brooklyn, and Queens

### LINKED ISSUE #314: Add Brooklyn and Queens library svg logos
### Component in question:
- Icons

### Description of issue:
- A request has been made to add Brooklyn and Queens library svg logos from here: https://www.nyclibraries.org/schools/ and include them in the Location Finder app.

helenvholmes:
Issue number: #314

- Adds logos for Brooklyn, Queens, and NYPL

### Front End Review:
- [ ] View [the example in Storybook](https://deploy-preview-316--stoic-murdock-c7f044.netlify.app/?path=/story/icon--icons)


nypl-wluisi (commented):
This looks good. My only question would be if we should include logo SVGs as part of the Icon component? Will there be more than 3 logos included in the future? If so, should there be a Logo component?

src/components/Icons/IconTypes.tsx
```diff
@@ -15,6 +15,9 @@ export enum IconNames {
     close = "close",
     download = "download",
     headset = "headset",
+    logo_brooklyn = "logo_brooklyn",
```
hokei: I think instead of putting this in IconTypes let's make a new Type object called Logo.  Most of these icons are interchangeable, like "Hey you could put a search button magnifying glass here, or you could put a plus sign", but I can't think of any situation where you'd put the NYPL logo where you put a "close" button and vice versa.  So let's make a new type where people can choose between the three logos and nothing else.

helenvholmes:
@hokei Bill brings up a good point about a potential Logo component—we have multiple different NYPL logo lock-ups alone. Does this merit its own component?

helenvholmes:
I've added some designs to the `Logos` story to show off the different options we might want/need out of a potential Logo component.

hokei:
I think, if we are thinking of components as things that people will not want to know the internals of, logo makes a lot of sense for a component.  Even if it's just a wrapper, it's a great way for DS to go "Hey. Use this.  Don't think about what's inside".

hokei (approved):
It looks good to me.  If we want a `<logo>` component let's put that in a different ticket

helenvholmes:
Filed #320 for a future Logo component.

Issue #317, opened 2020-08-05
Add "radio" type to the `Input` component


EdwinGuzman:
### Component in question:
- Input

### Description of issue:
- The `Input` component should allow a "radio" type and accept the required attributes for radio buttons.
- Additional component/question: @scooooooooby do you think it'll be good to have a `RadioInput` component as well?

It's what I'm doing now for LCA and I think it makes sense but I would include the `fieldset` and `legend` to make the whole component (which at this point is more like a molecule) accessible.

```
<fieldset>
  <legend id="radiobutton-location">I live, work, go to school, or pay property taxes at an address in:</legend>
  <input type="radio" id="input-location-nyc" class="input " value="nyc" name="location" aria-checked="true" checked=""> 
  <label id="radiobutton-location_nyc" for="location-nyc" class="label ">New York City (All five boroughs)</label>
  <input type="radio" id="input-location-nys" class="input " value="nys" name="location" aria-checked="false">
  <label id="radiobutton-location_nys" for="location-nys" class="label ">New York State (Outside NYC)</label>
  <input type="radio" id="input-location-us" class="input " value="us" name="location" aria-checked="false">
  <label id="radiobutton-location_us" for="location-us" class="label ">United States (Visiting NYC)</label>
</fieldset>
```

Sources:
https://designsystem.digital.gov/components/form-controls/#radio-buttons
https://a11y-style-guide.com/style-guide/section-forms.html#kssref-forms-radio-buttons
https://webaim.org/techniques/forms/controls#radio

helenvholmes:
We should definitely add a radio type to `Input`, that makes total sense to me.

I think for the particular the example you've included we would want another story over a brand new component that is built up of those HTML tags like we did with Autosuggest. I'm envisioning a new `Input` story that's exactly what you've included here that just uses the `Input` component.

EdwinGuzman:
That works.

helenvholmes:
Closed via #319.

PR #319, opened 2020-08-05
Radio input

EdwinGuzman:
Issue number: #317 

- Adds a "radio" input type, an example in Storybook on how to create an accessible fieldset with a group of radio button inputs, and tests.

@scooooooooby this is missing styling... but I'm not sure what the intended styling is so making this a draft for now.

### Front End Review:
- [ ] View [the example in Storybook](https://deploy-preview-319--stoic-murdock-c7f044.netlify.app/?path=/story/input--radio-input)


src/components/Input/Input.stories.tsx
```diff
@@ -176,3 +176,52 @@ export const inputAttributes = () => (
         </HelperErrorText>
     </>
 );
+
+export const radioInput = () => {
```
EdwinGuzman, line 180: If you ever want a `Radio` component in the future, this is pretty much it (after exposing some props).

src/components/Input/Input.tsx
```diff
@@ -59,7 +59,6 @@ const Input = React.forwardRef<HTMLInputElement, InputProps>((props, ref?) => {
         value: value,
         "aria-label": ariaLabel,
         "aria-labelledby": ariaLabelledBy,
-        ...attributes,
```
helenvholmes, line 62: Why was this deleted?

helenvholmes:
Re: styles... I'm recommending we use the browser defaults for these bad boys for now, since we otherwise don't have designs. I see some issues with radios being 100% width though, so I'll futz with that tomorrow morning.

src/components/Input/Input.tsx
EdwinGuzman: It was also added directly to the `input` element, so effectively they were added twice.

EdwinGuzman:
I added minor styling to the LCA but some styling is coming from the Design Toolkit - https://github.com/NYPL/nypl-library-card-app/pull/62/files#diff-af21424f94550a36f66eef0e847ea5f6

Screen Shot 2020-08-05 at 5 47 33 PM

hokei:
Is this ready to be in the design system at all?  When we have designs, won't those styles be added to the design system and automatically be used by this component, which would retroactively cause breaking changes to who ever was using it?

helenvholmes:
@hokei I don't think there ever will be designs for radios in the DS that depart from the browser defaults—that's been my recommendation to the team, and they seem fine with it. So I don't think it's inappropriate to bring it in.

I think that we _should_ ship this with styles to make radios float left of the label.

helenvholmes (approved):
Added some CSS. Looks good to me.

EdwinGuzman:
I'm going to make a small update to the story because it's not working properly.

EdwinGuzman:
Okay, merging now. Thanks!

PR #323, opened 2020-08-06
Pagination

### LINKED ISSUE #269: Pagination API Changes
### Component in question:
- `Pagination`

### Description of issue:
- `Pagination` is still using `dropdownOptions`, which has been deprecated with #253. Pagination needs to be upgraded with the next release

helenvholmes:
Issue number: #269 

- Updates `Pagination` to use sub components as children and props

### Front End Review:
- [ ] View [the example in Storybook](https://deploy-preview-323--stoic-murdock-c7f044.netlify.app/?path=/story/pagination--pagination)

### DIFF
--- a/src/components/Pagination/Pagination.stories.tsx
+++ b/src/components/Pagination/Pagination.stories.tsx
@@ -3,8 +3,12 @@ import { action } from "@storybook/addon-actions";
 import { withDesign } from "storybook-addon-designs";
 
 import Button from "../Button/Button";
-import { ButtonTypes } from "../Button/ButtonTypes";
+import { ButtonTypes, ButtonIconPositions } from "../Button/ButtonTypes";
+import Icon from "../Icons/Icon";
+import { IconNames, IconRotationTypes } from "../Icons/IconTypes";
+import Label from "../Label/Label";
 import Pagination from "./Pagination";
+import Select from "../Select/Select";
 
 export default {
     title: "Pagination",
@@ -24,7 +28,11 @@ export const pagination = () => (
         previousPage={
             <Button
                 buttonType={ButtonTypes.Secondary}
-                onClick={action("back clicked")}
+                onClick={action("Previous clicked")}
+                iconDecorative={true}
+                iconName={IconNames.arrow}
+                iconPosition={ButtonIconPositions.Left}
+                iconRotation={IconRotationTypes.rotate90}
             >
                 Previous
             </Button>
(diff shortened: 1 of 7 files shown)


src/components/Pagination/Pagination.stories.tsx
```diff
@@ -1,20 +1,78 @@
...
+import Select from "../Select/Select";
 
 export default {
     title: "Pagination",
     component: Pagination,
+    decorators: [withDesign],
 };
 
 export const pagination = () => (
     <Pagination
```
hokei, line 21: I'm like "I can't believe we didn't do this from the start" every time I see this

src/components/Pagination/Pagination.stories.tsx
```diff
@@ -1,20 +1,78 @@
...
-        currentValue={"1 of 4"}
-        onSelectChange={action("selectChange")}
-        onSelectBlur={action("selectBlur")}
-    />
+        previousPage={
+            <Button
+                buttonType={ButtonTypes.Secondary}
+                onClick={action("Previous clicked")}
+                iconDecorative={true}
+                iconName={IconNames.arrow}
```
hokei, line 27: Button still needs to be refactored?

src/components/Pagination/Pagination.tsx
```diff
@@ -1,67 +1,34 @@
...
-export default function Pagination(props: PaginationProps) {
+export default function Pagination(
+    props: React.PropsWithChildren<PaginationProps>
+) {
     const {
         blockName,
+        children,
         className,
-        currentValue,
+        id,
```
hokei: is id used?

hokei: same with blockName, className, modifiers

hokei: Should also add tests that the component does apply id, className, etc through if they're passed in.

src/components/Pagination/Pagination.stories.tsx
```diff
@@ -1,20 +1,78 @@
 import * as React from "react";
 import { action } from "@storybook/addon-actions";
+import { withDesign } from "storybook-addon-designs";
```
hokei, line 3: Is the lack of knobs intended (Eg consumers aren't ever supposed to use this component with button/label text that's different?)

hokei (changes requested):
Looks great!  A few minor things.  

I know this is the `<Select>` component but it's relevant here - I think it couldn't hurt to make `<Select>` throw a console.warn when you try to pass it more than 7 options.  So that if it's going of the page, etc.  the consumer knows they need to do some extra work / use the not-yet-built giant-paginator.

src/components/Pagination/Pagination.tsx
helenvholmes: I've removed id, but blockName, className, and modifiers allow consumers to change aspects of Pagination for their app and shouldn't be removed.

I think that making sure blockName, className, and modifiers are applied to components might belong better with the BEM helper, OR we should add that test to every component in the library.

hokei: if a blockName gets passed in here, where is it applied?  Sorry I feel like I'm missing something very obvious

helenvholmes: L33, no?

src/components/Pagination/Pagination.stories.tsx
helenvholmes: I'll add these in!

helenvholmes:
@hokei That's a good suggestion—added!

src/components/Pagination/Pagination.stories.tsx
helenvholmes: Yeah, looks like it

src/components/Pagination/Pagination.tsx
hokei: GAAAAH github hid literally just these two lines T_T.  Yes is perfect.

hokei (approved):
Optional:  A test that checks that the warn gets thrown.  But looks great.

helenvholmes:
@hokei that is a great suggestion... that for the life of me I can't figure out how to do. 

I've mocked out an "it-should" statement and will file an issue if someone else wants to try and take that on (I suspect there are a number of people who would find such a thing quite easy).

Issue #335, opened 2020-08-12
Icon component name prop


EdwinGuzman:
### Component in question:
- Icon

### Description of issue:
- The `Icon` component's story doesn't have good type checking since it renders all the logos in a loop, so it missed that the `LogoNames` enum should be defined for the `name` prop. Otherwise, trying to use a logo name for an svg throws an error:

<img width="826" alt="Screen Shot 2020-08-11 at 8 42 54 PM" src="https://user-images.githubusercontent.com/1280564/89962757-b35e1c00-dc13-11ea-9c32-399c02a2a81c.png">

PR #336, opened 2020-08-12
Allowing Icon 'name' prop to take in a LogoNames enum type.

EdwinGuzman:
Issue number: #335 

- It updates the `name` prop in the `Icon` component to allow the `IconNames` enum type or the `LogoNames` enum type.

### Front End Review:


EdwinGuzman:
Updated the PR with a "negative" version of the NYPL logo to be used on non-white backgrounds:

Screen Shot 2020-08-12 at 3 16 11 PM

https://deploy-preview-336--stoic-murdock-c7f044.netlify.app/?path=/story/icon--logos

src/components/Icons/_Icons.scss
```diff
@@ -47,6 +47,10 @@
         height: 18px;
         width: 18px;
     }
+
+    &--negative {
```
hokei: Should this be coded into the styling of the icon itself, or should it be already implemented on the client side and the only thing that's using it is Storybook?

src/components/Icons/Icon.stories.tsx
```diff
@@ -11,13 +11,17 @@ export default {
 };
 
 const iconRow = function (array, icon, modifiers = "") {
+    let mods = [modifiers];
+    if (icon.indexOf("_negative") !== -1) {
```
hokei, line 15: if the black background should be in Design System (which I am not sure it should be?  Is it possible that we want the white icon on a background that isn't black?)  I think we should ask the user to pass a "black-background" or "negative-icon" modifier in directly rather than doing this off of the name of the icon.

hokei (changes requested):
The "name" prop stuff looks good to me, but I think the implementation of the white icon styling needs to be re-thought.  

So this is what I think should happen here:  

The white icon:  Lives inside design system. There is no background color. 
Users:  Call something like 
```
<Banner background-color={background-color}>
<DS.Icon name={LogoNames.logo_nypl_negative} /> 
</Banner>
```

And the only thing is that `Logo.stories` needs an extra `.sb-dark-background` class

What are your thoughts?

EdwinGuzman:
Yes, I forgot to update this after messing around with it more in LCA. But the negative definitely shouldn't be part of the modifiers and a wrapper component works best for the stories. So I'll re-implement that.
Example of it in non-black background:
Screen Shot 2020-08-13 at 2 32 14 PM

src/components/Icons/Icon.stories.tsx
helenvholmes: We definitely don't want the black background on the white icon, so I'm going to resolve this

EdwinGuzman: The last update should resolve it, but now the example color for the background is still black. This is just not part of the modifiers anymore.

helenvholmes (approved):
LGTM! 👍🏻

src/components/Icons/Icon.stories.tsx
helenvholmes: I think that's great 👍🏻

hokei (approved):

Issue #338, opened 2020-08-13
Link component update


EdwinGuzman:
### Component in question:
- Link

### Description of issue:
- Implement `forwardRef` so libraries such as next/link can take advantage of it.

PR #339, opened 2020-08-13
Link component update

### LINKED ISSUE #309: linkType should become type
### Component in question:
- `Link`

### Description of issue:
- `Link` currently accepts a `linkType` prop. This should be a `type` prop instead to accommodate for more HTML-ish prop names.
- Update: It's not that the `linkType` prop name should be updated, but the props are being passed directly from the parent to the child and I think a new object should be created. Right now, the `a` element is created through `...linkProps` but there's also a separate `...attributes` prop for any other attributes. This seems redundant.

EdwinGuzman:
Issue number: #309 #338

- This updates the `Link` component to not render custom props in the generated anchor element, updates `linkType` to `type`, updates the stories with some explanations, adds an example using it with the Nextjs `Link` component, and implements `forwardRef` to use with other libraries.

Note: I branched this branch off `logo-icons` by accident so leaving that branch as the base for now so you don't see unnecessary changes.

### Front End Review:

### DIFF
--- a/src/components/Accordion/Accordion.stories.tsx
+++ b/src/components/Accordion/Accordion.stories.tsx
@@ -12,7 +12,7 @@ export default {
     component: Accordion,
 };
 const checkboxes = [
-    <li>
+    <li key="one">
         <Checkbox
             onChange={action("changed")}
             checkboxId="checkbox-1"
@@ -22,7 +22,7 @@ const checkboxes = [
             }}
         ></Checkbox>
     </li>,
-    <li>
+    <li key="two">
         <Checkbox
             onChange={action("changed")}
             checkboxId="checkbox-2"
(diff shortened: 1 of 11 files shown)


src/components/Accordion/Accordion.stories.tsx
```diff
@@ -12,7 +12,7 @@ export default {
     component: Accordion,
 };
 const checkboxes = [
-    <li>
+    <li key="one">
```
EdwinGuzman, line 15: Not strictly part of this PR but it bothered me seeing this warning every time Storybook opened.

helenvholmes (changes requested):
These are some great updates, thanks Edwin. I pushed up some `<p>` tags and edited the wording on the NextJS link description (which, coincidentally, might mean that now it's incorrect—please check it).

Looks like the NextJS story is broken though, and I'm unclear on why—Link's `href` prop is optional but that's what it's throwing the error on.

EdwinGuzman:
@scooooooooby thanks for catching that. I updated the example with a better dummy/example "nextjs" wrapper that does a better job of cloning and properly rendering (even for an example) the `Link` component.

helenvholmes (approved):
Looks great!

PR #343, opened 2020-08-18
Fix #11: CDN

### LINKED ISSUE #11: CDN Solution
We know that we'll want some sort of CDN solution for people who want to pull in the design library whole-hog. We'll need to understand:

- [ ] Is this just CSS, or also JS? Are those shipped separately?
- [ ] How is react code vs. vanillaJS shipped in the CDN?

helenvholmes:
Issue number: #11

- Adds CDN information to the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,18 @@ function NewComponent(props) {
 }
 ```
 
+### CDN
+
+You can also use the Design System in your project with our CDN:
+
+`<script src="https://unpkg.com/@nypl/design-system-react-components/dist/design-system-react-components.min.js"></script>`
+
+`<link href="https://unpkg.com/@nypl/design-system-react-components/dist/styles.css">`
+
+If you need to reference a particular version you can do do by including the version number in the URL:
+
+`<script src="https://unpkg.com/@nypl/design-system-react-components@0.9.1/dist/design-system-react-components.min.js"></script>`
+
 ## Design System and Product Requirements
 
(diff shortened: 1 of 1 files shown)


hokei:
do we have a site anywhere where this is demonstrated?  If it breaks, is there something that will alert us in an automated way?

README.md
```diff
@@ -62,6 +62,18 @@ function NewComponent(props) {
 }
 ```
 
+### CDN
```
hokei, line 65: 😮 there doesn't need to be code changes? it's just ... here? I never knew!

helenvholmes: Neither did I, Edwin completely to the rescue. Now we know!

helenvholmes:
> do we have a site anywhere where this is demonstrated? If it breaks, is there something that will alert us in an automated way?

@hokei We don't have a site where this is demonstrated, but we could! I'm spinning something up on Codepen. I'm not done with it but I'll push up reference to it in this PR and ping you when it's ready.

Here is the site for unpkg: https://unpkg.com/
I haven't gone much deeper than that to figure out what happens if it breaks. We could make reference in the README that it's an at-your-own-risk situation since a third-party will be serving up assets for your site, but I think that's a given for any CDN-type situation. The internet's a pile of poorly stacked irregular rocks at the best of times

helenvholmes:
I'm going to close this out until someone is ready to take a closer look at getting the CDN working with something like Codepen/a github.io static site again.

PR #348, opened 2020-08-24
Status Badge

### LINKED ISSUE #322: Status Badge Component
### Description of issue:
- Location Finder needs a "Status Badge" component. At least 2 variations are needed:

alert-status
hours-effected-today

We have a similar component in the D8 Pattern Lab: https://dev-d8.nypl.org/themes/custom/nypl_emulsify/pattern-lab/public/?p=molecules-location-info-closed

"Articles & Databases" figma designs also have a similar "Status Badge": https://www.figma.com/file/t1URLPU6mAsxUX0cyEJblX/Articles-%26-Databases%3A-sketches?node-id=20%3A0

helenvholmes:
Issue number: #322

- Adds a Status Badge component as stories

The API for this (particularly if you check out the attached designs) is still a little crunchy—looks like the designs for this have grown organically over time. I've filed [RENO-1644](https://jira.nypl.org/browse/RENO-1644) to work with Kathy on cleaning that up. This will solve Location Finder's immediate needs.

### Front End Review:
- [ ] View [the example in Storybook](https://deploy-preview-348--stoic-murdock-c7f044.netlify.app/)

### DIFF
--- a/src/components/StatusBadge/_StatusBadge.scss
+++ b/src/components/StatusBadge/_StatusBadge.scss
@@ -1,9 +1,9 @@
 .status-badge {
+    border-radius: 2px;
     color: var(--ui-black);
     display: block;
     font-style: italic;
-    margin-bottom: calc(var(--space-m) / 2);
-    padding: calc(var(--space-s) / 2);
+    padding: var(--space-xs);
     text-align: center;
 
     &--inactive {
--- a/src/components/StatusBadge/_StatusBadge.scss
+++ b/src/components/StatusBadge/_StatusBadge.scss
@@ -1,55 +1,16 @@
 .status-badge {
     color: var(--ui-black);
     display: block;
-    margin-bottom: var(--space-m) / 2;
-    padding: var(--space-m) / 2;
+    font-style: italic;
+    margin-bottom: calc(var(--space-m) / 2);
+    padding: calc(var(--space-s) / 2);
     text-align: center;
 
-    &--cancelled {
+    &--inactive {
         background-color: var(--ui-gray-light);
     }
 
-    &--errored {
-    }
     &--warning {
-    }
-
-    a:focus,
-    a:hover {
-        text-decoration: underline;
+        background-color: #f9e08e;
     }
 }
-
-// See http://refills.bourbon.io/components/#flashes.
-// $base-spacing: 1.5em !default;
-// $statuses: (
-//     "warning": $nypl-yellow-light,
-//     "error": $nypl-red-light,
-//     "status": $nypl-green-light,
-// ) !default;
-
-// @each $status-type, $color in $statuses {
-//     .status--#{$status-type} {
-//         background-color: $color;
-//         color: darken($color, 60%);
-//         display: block;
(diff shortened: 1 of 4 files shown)


src/components/StatusBadge/_StatusBadge.scss
```diff
@@ -0,0 +1,16 @@
+.status-badge {
+    color: var(--ui-black);
+    display: block;
+    font-style: italic;
+    margin-bottom: calc(var(--space-m) / 2);
+    padding: calc(var(--space-s) / 2);
```
hokei: Is this padding intended?  With other components, we've had it stick to its container - we should probably be consistent

hokei:
It looks good, though it does make me wonder whether we want to separate out the "stories only" components with the actual "has React" components.   Not for this ticket, just yet another thing to start floating in the back of our heads ...

src/components/StatusBadge/_StatusBadge.scss
helenvholmes: I'm not sure what you mean by padding sticking to its container with other components... are you maybe referring to the global decorator I added? That has caused every story to have some space from the edges of the browser: https://github.com/NYPL/nypl-design-system/blob/development/.storybook-react/config.js#L12

It was a good comment though, I don't think I should have added that marginBottom (that's a holdout from the CSS I copied over from PatternLab), and it looks like I was missing a border-radius when I looked at the designs more carefully 😅

helenvholmes:
@hokei What do you think consumers would get out of organizing components to make that distinction more clear?

nypl-wluisi:
@scooooooooby This looks great to me. There is one UX change on location finder kathy requested we remove some of the padding and make the font smaller, so it looks like this now:

location-finder-status-badge

you can see this in action here, if you search for "Countee Cullen Library":

https://qa-scout.nypl.org/location-finder 

I prefer what's in the DS now on this PR, but you might need to sync up w/ Kathy on the UX tweak?

helenvholmes:
@nypl-wluisi I actually think the better choice here is to match LF. I've opened up a ticket for not having the relevant Figma/Metronome documentation for this component yet here: https://jira.nypl.org/browse/RENO-1644

hokei:
> @hokei What do you think consumers would get out of organizing components to make that distinction more clear?

Hmmm I kind of assumed it would be useful for consumers to know what can be implemented with just DS CSS and what needs DS React code.  But ... maybe that wouldn't be that useful?

src/components/StatusBadge/_StatusBadge.scss
hokei: guhhhh ignore me.  The components used to stick to their container but now there is a global decorator which my brain forgot about, dumb brain.  So the padding threw me and then I thought this was the padding ...   Glad that you did find real things to fix though haha

helenvholmes:
@hokei I think the most helpful thing is for the "Show Info" panel to be copypasta-able, imo—so long as they are, I don't think we need to expose the difference.

hokei (approved):

PR #354, opened 2020-08-24
Universal Focus

### LINKED ISSUE #353: Universal focus style

helenvholmes:
Issue number: #353 

- Adds in focus styling
- Creates an `.nypl-ds` namespace for global CSS settings

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ function NewComponent(props) {
 }
 ```
 
-Some things in the Design System, such as the universal focus styling and the box-model the DS uses, are namespaced to the `.nypl-ds` class. To include those in your app, add `.nypl-ds` to whichever wrapper tag makes sense in your application (e.g., `<html>`, `<body>`, or `<div class="app">`).
+Some things in the Design System, such as the universal focus styling and the box-model the DS uses, are namespaced to the `.nypl-ds` class. To include those in your app, add `.nypl-ds` to whichever wrapper tag makes sense in your application (e.g., `<div class="app">` or `<div class="container">`).
 
 Please note that the NYPL Header and Footer should be _outside_ of the `.nypl-ds` wrapper class.
 
--- a/README.md
+++ b/README.md
@@ -62,6 +62,8 @@ function NewComponent(props) {
 }
 ```
 
+Some things in the Design System, such as the universal focus styling, are namespaced to the `.nypl-ds` class. To include those in your app, add `.nypl-ds` to whichever wrapper tag makes sense in your application.
+
 ## Design System and Product Requirements
 
 The NYPL Design System provides accessible stories, but real live data can necessitate additional accessibility requirements beyond what we're committed to in our generic, extensible components. To ensure your products' final result is accessible, please adhere to the accessibility requirements put together by NYPL's accessibility coordinator on [Metronome](http://themetronome.co/).
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ function NewComponent(props) {
 }
 ```
 
-Some things in the Design System, such as the universal focus styling, are namespaced to the `.nypl-ds` class. To include those in your app, add `.nypl-ds` to whichever wrapper tag makes sense in your application.
+Some things in the Design System, such as the universal focus styling and the box-model the DS uses, are namespaced to the `.nypl-ds` class. To include those in your app, add `.nypl-ds` to whichever wrapper tag makes sense in your application.
 
 ## Design System and Product Requirements
 
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ function NewComponent(props) {
 }
 ```
 
-Some things in the Design System, such as the universal focus styling and the box-model the DS uses, are namespaced to the `.nypl-ds` class. To include those in your app, add `.nypl-ds` to whichever wrapper tag makes sense in your application.
+Some things in the Design System, such as the universal focus styling and the box-model the DS uses, are namespaced to the `.nypl-ds` class. To include those in your app, add `.nypl-ds` to whichever wrapper tag makes sense in your application (e.g., `<html>`, `<body>`, or `<div class="app">`).
 
 ## Design System and Product Requirements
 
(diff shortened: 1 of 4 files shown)


hokei (approved):

src/styles/base/_05-focus.scss
```diff
@@ -0,0 +1,34 @@
...
+    outline: 2px solid var(--ui-focus);
+    outline-offset: 2px;
+
+    @if $darkBackground {
+        outline-color: var(--ui-white);
+    }
+}
+
+// Apply :focus behavior to focusable elements only (For IE 11)
+input,
```
hokei: Looking through your comment here:  https://github.com/NYPL-Simplified/circulation-patron-web/pull/119#issuecomment-680907809

Does this mean that anybody that uses DS is getting these styles, because all CSS is automatically pulled in?  

Even though I think that visible focus is a desirable thing, I do want to give people the option to use their own focus styles if they want to.  To that end, it would make more sense to put these under a classname rather than on the elements themselves.

hokei (changes requested):
requesting changes again because of that question above ...

src/styles/base/_05-focus.scss
helenvholmes: That's what I'm thinking too. I think I'm going to use `.nypl-ds` as a namespace. I'll make those updates and get this in.

src/styles/base/_04-base.scss
```diff
@@ -1,9 +1,11 @@
-html {
-    box-sizing: border-box;
-}
+.nypl-ds {
+    html {
```
EdwinGuzman: I think you want `&html { ` or rather `html.nypl-ds` unless if you are thinking of putting a wrapper div around `html` which doesn't sound right.

helenvholmes: 🤦🏻‍♀️ thank you Edwin

README.md
```diff
@@ -62,6 +62,8 @@ function NewComponent(props) {
 }
 ```
 
+Some things in the Design System, such as the universal focus styling, are namespaced to the `.nypl-ds` class. To include those in your app, add `.nypl-ds` to whichever wrapper tag makes sense in your application.
```
helenvholmes: There are some things in the DS that will look awfully funny (`Input` comes to mind) without this declaration, but redeclaring the box model on every component feels very backward... I suppose just so people are aware that visual issues in some consumer apps might arise if people don't add this wrapper class to their application

README.md
```diff
@@ -62,6 +62,8 @@ function NewComponent(props) {
 }
 ```
 
+Some things in the Design System, such as the universal focus styling and the box-model the DS uses, are namespaced to the `.nypl-ds` class. To include those in your app, add `.nypl-ds` to whichever wrapper tag makes sense in your application.
```
hokei: > to whichever wrapper tag makes sense 
Can we be more specific than this?  Something like, "to include all `.nypl-ds` styles, add `.nypl-ds` to the html tag"?  

I think if people want to pick and choose they can actually dig into it, but giving people an easy quick start would be handy

EdwinGuzman:
Code looks good but just pointing this out:

Footer in production
Screen Shot 2020-08-31 at 12 22 32 PM

Footer in dev
Screen Shot 2020-08-31 at 12 21 48 PM

But this is specific and I know the footer has specific footer focus styling. Maybe it can have specific styling from DS in the future.

Also, not directly related to this PR but I noticed this in production and doesn't change with this PR
Screen Shot 2020-08-31 at 12 21 57 PM

README.md
helenvholmes: Added some more specific langugage

helenvholmes:
> Code looks good but just pointing this out:
> 
> Footer in production
> Screen Shot 2020-08-31 at 12 22 32 PM
> 
> Footer in dev
> Screen Shot 2020-08-31 at 12 21 48 PM
> 
> But this is specific and I know the footer has specific footer focus styling. Maybe it can have specific styling from DS in the future.

Does this mean that the DS styles for focus are somehow overriding the footer styles for focus?

> Also, not directly related to this PR but I noticed this in production and doesn't change with this PR
> Screen Shot 2020-08-31 at 12 21 57 PM
Nice catch. Can you file a bug against the [RENO project](https://jira.nypl.org/projects/RENO/issues/RENO-1017?filter=allopenissues) so it can get properly categorized within DXP for this Edwin?

EdwinGuzman:
> Does this mean that the DS styles for focus are somehow overriding the footer styles for focus?

Yes

Ignore the white background in the alert... it seems to be only in the Library Card app. It's in production but I don't know why (production hasn't been deployed in a long time). I'll fix that myself in development.

helenvholmes:
> > Does this mean that the DS styles for focus are somehow overriding the footer styles for focus?
> 
> Yes

Are you wrapping the footer inside of the `.nypl-ds` class?

EdwinGuzman:
The footer is indeed inside the wrapper with the `.nypl-ds` class.

helenvholmes:
> The footer is indeed inside the wrapper with the `.nypl-ds` class.

This is why. I should probably add to the `README` that the NYPL footer and header should be outside of this class.

(Eventually I'd like for the header and footer to be on the Design System, but until that happens DS styles are going to interact poorly with header and footer.)

README.md
```diff
@@ -62,6 +62,10 @@ function NewComponent(props) {
 }
 ```
 
+Some things in the Design System, such as the universal focus styling and the box-model the DS uses, are namespaced to the `.nypl-ds` class. To include those in your app, add `.nypl-ds` to whichever wrapper tag makes sense in your application (e.g., `<html>`, `<body>`, or `<div class="app">`).
+
+Please note that the NYPL Header and Footer should be _outside_ of the `.nypl-ds` wrapper class.
```
EdwinGuzman, line 67: This is fine but note that above it says it can go in `<body>` but the header and footer _need_ to go in the body element. I don't think this is a big deal yet and we also don't know how other apps will use it.

EdwinGuzman (approved):

PR #364, opened 2020-08-27
Template Changes

### LINKED ISSUE #342: Spacing on Main element
### Component in question:
- Full Width Template

### Description of issue:
-  There are certain screen widths (in the desktop breakpoint) where the padding on the Main element no longer applies and the page items hit the edge of the screen.
- Additionally whitespace is added on either side of the content-top when it sound be a break-out.

main-padding

helenvholmes:
Issue number: #342

- Removes the bug referenced in #342 where items in content at certain breakpoints would be right up against the browser window
- Adds knobs to Template stories
- Removes padding that was added to the `breakout` mixin, which was causing empty whitespace around `Hero`s when added

### Front End Review:
- [ ] View [the example in Storybook](https://deploy-preview-364--stoic-murdock-c7f044.netlify.app/?path=/story/template--template)


helenvholmes:
Looks like the CI for Netlify isn't running—anyone know why that's changed? cc @EdwinGuzman

EdwinGuzman:
No, it's working. Maybe it didn't show up right away at first.

Also, are you meant to have a left and right sidebar at the same time? I see this when both are rendered:
Screen Shot 2020-08-27 at 1 22 22 PM

helenvholmes:
> No, it's working. Maybe it didn't show up right away at first.
Oh phew! That was it.

> Also, are you meant to have a left and right sidebar at the same time? I see this when both are rendered:
> Screen Shot 2020-08-27 at 1 22 22 PM
No, people shouldn't be using both, but I was having trouble finding an elegant way to express that with Storybook's knobs...

src/styles/base/_03-mixins.scss
```diff
@@ -38,15 +38,6 @@
     max-width: #{$container-max-width};
     padding: #{$v-padding} #{$h-padding};
     width: 100%;
-
-    @include breakpoint(max-width $container-max-width) {
```
misselbeck, line 42: Looking at this same wrapper mixin in the D8 repo, I think this will work by removing `max-width` from the value being passed into the breakpoint function.

`@include breakpoint($container-max-width)` instead of...
`@include breakpoint(max-width $container-max-width)`

That way this only kicks in _after_ the width reaches the `$xl` break.

helenvholmes: Unfortunately this will mess with the breakout mixin—looks like it will apply padding at all the lower sizes, whereas we want `content-top` to have no spacing for things like various `Hero` variants.

misselbeck: The opposite would be true, no?

In D8, the following applies the padding calculation at (min-width: 1152px):
`@include breakpoint($container-max-width)`

d8-wrapper

In the DS, the wrapper mixin originally had the following:
`@include breakpoint(max-width $container-max-width)`

That would reverse the media query logic...applying the padding calculation only to smaller size screens.

It's possible that padding calculation is not needed, and removing it completely is fine (as done in the PR). But should we then remove it from the mixins in D8 also?

helenvholmes: > The opposite would be true, no?

I've had some bad merges working on this ticket, so it could be that I'm the one turned around here. I was seeing a padding getting applied on items that had the `breakout` mixin on smaller sizes, which was messing with the point of that mixin.
 
> It's possible that padding calculation is not needed, and removing it completely is fine (as done in the PR). But should we then remove it from the mixins in D8 also?

I would like to keep these two things in parity if possible, but one thing of note is that d8 is using `1152px` as that max breakpoint and we're using `1280px`. I would feel more confident if I understood why that media query was adding padding in the first place...

misselbeck:
I don't see `box-sizing: border-box` being applied to the layout elements. Is it because the * selector doesn't apply to items the storybook iframe?

src/styles/base/_03-mixins.scss
misselbeck: Agreed. I also am not sure what is the point of kicking in this padding at the largest breakpoint.

src/styles/base/_03-mixins.scss
```diff
@@ -58,8 +49,6 @@
 @mixin breakout($v-padding: var(--space)) {
     margin-left: calc(-50vw + 50%);
```
misselbeck, line 50: The breakout mixin in D8 also adds 0.5rem of padding to these margins:
```
margin-left: calc(0.5rem + calc(-50vw + 50%));
margin-right: calc(0.5rem + calc(-50vw + 50%));
```
I think it is because the viewport calculation doesn't take into account the scrollbar (if one is present), triggering an x overflow?

helenvholmes:
> I don't see `box-sizing: border-box` being applied to the layout elements. Is it because the * selector doesn't apply to items the storybook iframe?

I just pushed up some changes that should resolve this—we moved the `box-sizing` declaration under a className (`.nypl-ds`) in 0.12.0* to resolve some spacing issues with the NYPL footer.

[EDIT] got the release wrong

src/styles/base/_03-mixins.scss
helenvholmes: You mean a browser scrollbar? Is this specifically for older browsers? Sounds so odd, because I'm not seeing that on my machine...

misselbeck: It's an odd one, but there seems to be a difference when a browser presents an overlay scrollbar vs the standard scrollbar.

Overlay Scrollbar

overlay-scrollbar

Non-overlay Scrollbar

non-overlay-scrollbar

The screenshots are from Chrome, but I was able to reproduce the same in Firefox also.

What is interesting, is that I only see a non-overlay scrollbar when viewing a browser window in my external monitor. This leads to me believe that it might be a setting related to pixel density?

Either way, the non-overlay scrollbar gutter does not get calculated in the VW unit. [This article](https://cloudfour.com/thinks/breaking-out-with-viewport-units-and-calc/#one-big-dumb-caveat) seems to confirm that.

I think that might be why the breakout mixin in D8 adds the extra 0.5rem to the calc.

Anyway, I'd be fine with removing that extra 0.5rem as you did here, as I don't think we can control the non-overlay vs overlay scrollbar, as far as I can tell.

How do you want to account for the differences in the wrapper/breakpoint mixins in the Design System vs D8? Should we make the changes there, as well?

misselbeck (approved):

src/styles/base/_03-mixins.scss
helenvholmes: Ahhh, thanks for writing this up, it helped make it really clear.

~I think instead _I'm_ going to add back in the margins here—otherwise it sounds like I'd be introducing a regression. I will also add a comment, because wowza, what an odd one~

Actually, on further reflection, I'm not going to do this—it messes up everyone who _doesn't_ have the non-overlay scrollbar, which seems... worse.

Instead I've used this fix as identified in the article: 

```
.nypl-ds {
  overflow-x: hidden;
}
```

😪

PR #369, opened 2020-08-31
Storybook 6

### LINKED ISSUE #340: Upgrade Storybook
http://localhost:9001/?path=/settings/about

### LINKED ISSUE #359: Test needed for Storybooks "Show Info"
Screen Shot 2020-08-25 at 10 11 04 AM

"Show Info" consistently breaks without people being aware of it. We should add a test of some sort to know if/when our work is breaking the Storybook show info panel.

EdwinGuzman:
Issue number: #340 and possibly #359 

- Upgrades to Storybook 6 along with addons and minor modifications to _some_ stories. The task is to upgrade to S6 but I figured individual components can/should be updated individually since there may be new features or updates for each.

This new version of Storybook understands the `forwardRef` implementation so a component's source doesn't break (check specifically the `Checkbox` stories).

Upgrade notes:
### Config
- `.storybook-react` is now `.storybook`. That's the convention and the only reason we had `.storybook-react` was because we also had the twig storybook in this repo.
- All the older config files are now pretty much a part of `.main.js`. This includes `.babelrc`, `addons.js`, `config.js`, and `webpack.config.js`.
- There's also a `.preview.js` file which isn't normally needed but we need it for styling. I haven't found a way to import the scss without _yet_.

### Packages
- Typescript is the baked into Storybook 6 so we don't need babel/typescript packages or configurations, so these are removed now:
  - preset-env
  - preset-react
  - preset-typescript
  - awesome-typescript-loader
- Updated packages:
  - addon-a11y
  - addon-actions
  - addon-knobs
  - addon-links
  - `@storybook/react`
  - storybook-addon-designs -> There's no v6 version of this package but it still works.
- Upgrades:
  - addons and addon-info are not a part of addon-essentials
- New:
  - addons-docs -> This also shows the source, info, and props table.

Screen Shot 2020-08-31 at 11 10 21 AM

### Stories tab
- Design, Accessibility, Knobs, Actions -> They all work like before.
- Controls -> This is new and does the same thing that Knobs does but you configure it through Args (more on that later). So it's more baked in now.
- Knobs/Actions -> These are going to be deprecated in version 7 but I kept them here because they are important for _some_ stories. Maybe we want to rethink how those stories are built and I think MDX can help solve that (more on that later). For now, this is not a problem to have but I suggest not using them in _new_ stories.
- There's a new background tester to check dark mode:
Screen Shot 2020-08-31 at 11 53 21 AM
- And a responsive viewer:
Screen Shot 2020-08-31 at 11 53 29 AM

### Stories syntax
1) This _must_ be changed to remove syntax warnings:
```
breadcrumbs.story = {
    name: "Breadcrumbs",
    parameters: { ... },
}
```

is now 

```
breadcrumbs.storyName = "Breadcrumbs";
breadcrumbs.parameters = { ... };
```

2) The following is _not_ necessary but using it allows the "controls" tab to show up and allows for dynamic UI updates. We might as well upgrade because it provides a lot of what we want but it means updating all the components to use Storybook's "args".
```
// Set up the component as a template with Storybook's new `arg` syntax.
const Template: Story<ComponentPropsType> = (args) => <Component {...args} />;

// Now we can bind that template with any number of stories we want to export.
export const StoryNameForComponent = Template.bind({});

// And each binded template has its own args property which shows up in the "Controls" tab
StoryNameForComponent.args = {
    id: "checkbox",
    text: "some text",
    isSelected: false,
    ...
};
```

A good example is the `Checkbox` component. 
Screen Shot 2020-08-31 at 11 10 00 AM

### Summary

- Easier and smaller config!
- We should move to the Args syntax - https://storybook.js.org/docs/react/writing-stories/args
- Check the `Accordion`, `Checkbox`, and `Icon` stories for the Args syntax. I only updated these components for now.
- A new feature I'd like to try and should be a new ticket is using MDX for stories - https://storybook.js.org/docs/react/writing-docs/mdx - but a topic for discussion at a different time.
- Knobs/Actions are going to be deprecated so some stories need to be reworked eventually.

(Long PR note but writing everything down because I want to blog about this upgrade soon - so notes for me, too).

### Front End Review:


helenvholmes (changes requested):
Wow. _Super_ cool. 

I'm marking this with Planning to talk about it tomorrow—should we update all of the stories to merge this in, or merge it and update the stories over time to use args/deprecate knobs? Curious what you and @hokei think.

Only nit, we don't need to update the CHANGELOG for this since storybook changes aren't shipped with the library itself. Once those are taken out consider this approved from me 💯💫🌟

EdwinGuzman:
Updated. I think it's easier to update stories as we keep developing. It would also allow everyone to make the update and get to know the new API.

helenvholmes (approved):
🎉 Great work Edwin, this is very cool

PR #371, opened 2020-08-31
Removes Icon props from Button

### LINKED ISSUE #328: Update Button API
### Component in question:
- `Button`

### Description of issue:
- `Icon` should be passed as child instead of via props

### LINKED ISSUE #337: Button doesn't need an `onClick`
### Component in question:
- Button

### Description of issue:
- Currently, the `onClick` prop is required, but it's not necessary if the button is of type `submit`. For `form`s, the `onSubmit` attribute will be triggered by a button that is of type `submit`. Requiring the `onClick` prop calls the submit function twice which can be bad.
The `onClick` prop can be optional but an error can be thrown whenever `type="submit"` is _not_ used.

helenvholmes:
Issue number: #328 | #337

- Makes `Button`'s `onClick` property options
- Removes all icon props from `Button`. Button now uses `Icon` as a child

### Front End Review:
- [ ] View [the example in Storybook](https://deploy-preview-371--stoic-murdock-c7f044.netlify.app/?path=/story/button--button)

### DIFF
--- a/src/components/Button/Button.stories.tsx
+++ b/src/components/Button/Button.stories.tsx
@@ -1,7 +1,7 @@
 import * as React from "react";
 
 import Button from "./Button";
-import { ButtonTypes, ButtonIconPositions } from "./ButtonTypes";
+import { ButtonTypes } from "./ButtonTypes";
 import Icon from "../Icons/Icon";
 import { IconRotationTypes, IconNames } from "../Icons/IconTypes";
 import { action } from "@storybook/addon-actions";
--- a/src/components/Button/Button.stories.tsx
+++ b/src/components/Button/Button.stories.tsx
@@ -14,11 +14,17 @@ export default {
     decorators: [withDesign],
 };
 
-let showIcon;
+let showIconLeft;
+let showIconRight;
 
 export const button = () => (
     <>
-        {boolean("Show Icon", true) ? (showIcon = true) : (showIcon = false)}
+        {boolean("Show Icon on Left", true)
+            ? (showIconLeft = true)
+            : (showIconLeft = false)}
+        {boolean("Show Icon on Right", false)
+            ? (showIconRight = true)
+            : (showIconRight = false)}
         <Button
             onClick={action("clicked")}
             id="button"
@@ -26,7 +32,7 @@ export const button = () => (
             type="submit"
             disabled={boolean("Disabled", false)}
         >
-            {showIcon && (
+            {showIconLeft && (
                 <Icon
                     name={select("Icon", IconNames, IconNames.search_small)}
                     decorative={true}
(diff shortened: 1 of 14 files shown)


src/components/Button/Button.stories.tsx
```diff
@@ -24,22 +25,14 @@ export const button = () => (
             buttonType={select("Button Type", ButtonTypes, ButtonTypes.Primary)}
             type="submit"
             disabled={boolean("Disabled", false)}
```
hokei, line 33: Hovering disabled in the stories doesn't seem to give the right result.

helenvholmes: What's the effect you're expecting? Do buttons that are disabled typically have rollover states?

hokei: I guess I would expect it to stay disabled looking, not look like you can suddenly click it again?  I think this works on the primary button, but not any of the other buttons.  
link button disabled: 
Screen Shot 2020-09-01 at 9 24 38 AM

link button disabled hover: 
Screen Shot 2020-09-01 at 9 24 41 AM

hokei:
What's the difference between the search and search_small icons? Possibly for a different ticket.

src/components/Button/Button.stories.tsx
```diff
@@ -24,22 +25,14 @@ export const button = () => (
...
-                    ? select("Icon", IconNames, IconNames.search_small)
-                    : null
-            }
-            iconDecorative={true}
         >
+            {showIcon && (
+                <Icon
+                    name={select("Icon", IconNames, IconNames.search_small)}
+                    decorative={true}
+                    modifiers={["small", "icon-left"]}
```
hokei, line 39: Would an icon position knob be helpful here?

src/components/SearchBar/SearchBar.stories.tsx
```diff
@@ -3,9 +3,10 @@ import * as React from "react";
 import SearchBar from "./SearchBar";
 import Select from "../Select/Select";
 import HelperErrorText from "../HelperErrorText/HelperErrorText";
-import { IconRotationTypes, IconNames } from "../Icons/IconTypes";
 import Button from "../Button/Button";
 import { ButtonTypes, ButtonIconPositions } from "../Button/ButtonTypes";
```
hokei: is ButtonIconPositions used anymore?

src/components/Button/Button.stories.tsx
helenvholmes: You're right, I was just checking on primary... I'll get that updated

src/components/SearchBar/SearchBar.stories.tsx
helenvholmes: No, we should probably get rid of it all together

helenvholmes:
> What's the difference between the search and search_small icons? Possibly for a different ticket.

These both exist within the Figma library and should probably be cleaned up. (A holdover from moving to Google Material Design icons.) The original thinking is that you'd use "small" (which is thicker) on dark backgrounds or in smaller locations... but I think we should eliminate it, it's just confusing. That said I agree that it's a separate ticket since it includes a conversation with Kathy

EdwinGuzman (approved):

Issue #308, opened 2020-07-30
Running List of Snapshot Bugs


helenvholmes:
- Icons in Button will be rendered, but not appear visually
- Button text should always be centered, even if the button is larger than 100% width

helenvholmes:
Related to #345.

helenvholmes:
Closed in favor of the [planning doc](https://docs.google.com/document/d/1wTUo7PCZMmkjRZ96WXQ-M3gUKJGTPxwV7MDf8nCrtwo/edit#).

PR #381, opened 2020-09-03
Button content centering fix

### LINKED ISSUE #376: Button needs justify-content
Space-between should probably only be for Accordion

helenvholmes:
Issue number: #376 

- Fixes a bug introduced in 0.13.0 where content within buttons was not centered if button was made, say, width 100%

### Front End Review:


hokei:
Is there a way we can check for regressions on something like this, that's hopefully not famously brittle visual regression testing?

hokei (approved):

hokei:
Is it possible to add a resizeable button in stories?  I think, in stories right now, we don't have a button that is bigger than its content

helenvholmes:
> Is there a way we can check for regressions on something like this, that's hopefully not famously brittle visual regression testing?

I think it would be good to add tests for regressions like these, although I'm not sure the best way to do so. I've added a note to #308. I think more generally what I'd want out of a test suite like this is to be able to describe a series of visual states that I think the component should handle, and what it should be doing in those states...

> Is it possible to add a resizeable button in stories? I think, in stories right now, we don't have a button that is bigger than its content

I think this is a good idea, so I'm going to file a ticket for it. A lot of our stories will need to be tweaked now that we're on Storybook 6, so I'm going to do it in a follow-up. #385

PR #389, opened 2020-09-10
Accordion Changes

### LINKED ISSUE #195: Improve Accordion stories
### Component in question:
- Accordion

### Description of issue:
- It'd be nice to show even one other story of the Accordion with different inner content using knobs. See [the Metronome designs](http://themetronome.co/components/oc-272-accordion-(-extended-checkbox-set-pattern)-/?fresh=true) for some examples of what else we could include.

helenvholmes:
Fixes #195

- Adds additional `Accordion` story and styling (FAQ set)
- Adds hover styling to `Accordion` button
- Changes `className` to BEM as `.accordion__content` instead of second class `.accordion-content`

### Front End Review:
- [ ] View [the Accordion in Storybook](https://deploy-preview-389--stoic-murdock-c7f044.netlify.app/?path=/story/accordion--accordion-as-faq-set)

### DIFF
--- a/src/components/Accordion/Accordion.stories.tsx
+++ b/src/components/Accordion/Accordion.stories.tsx
@@ -4,7 +4,7 @@ import Accordion, { AccordionProps } from "./Accordion";
 import { action } from "@storybook/addon-actions";
 import { withDesign } from "storybook-addon-designs";
 
-import Checkbox from "../Checkbox/Checkbox";
+import { BasicCheckbox } from "../Checkbox/Checkbox.stories";
 import List from "../List/List";
 import { ListTypes } from "../List/ListTypes";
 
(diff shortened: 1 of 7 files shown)


EdwinGuzman:
The code seems pretty good on a first glance. The checkbox story doesn't work well because each checkbox has the same label id so when you click on a label, only the first checkbox is checked/unchecked.

I would also try not to use `lodash` especially if it's only for a story and not for a component. I _think_ it's a big bundle but we can double-check how big it really is.

helenvholmes:
> The code seems pretty good on a first glance. The checkbox story doesn't work well because each checkbox has the same label id so when you click on a label, only the first checkbox is checked/unchecked.

Noted, I'll probably swap this out with another component then to keep it simple.

> I would also try not to use `lodash` especially if it's only for a story and not for a component. I _think_ it's a big bundle but we can double-check how big it really is.

Do you feel this way even though it's only included as a devDependency? It wouldn't be included with the package, I don't think.

src/components/Accordion/Accordion.stories.tsx
```diff
@@ -1,67 +1,97 @@
...
 import { Meta, Story } from "@storybook/react/types-6-0";
+import range from "lodash/range";
 import Accordion, { AccordionProps } from "./Accordion";
 import { action } from "@storybook/addon-actions";
+import { withDesign } from "storybook-addon-designs";
 
-import Checkbox from "../Checkbox/Checkbox";
-import List from "../List/List";
-import { ListTypes } from "../List/ListTypes";
+// import List from "../List/List";
```
EdwinGuzman: You can remove these comments.

src/components/Accordion/Accordion.stories.tsx
```diff
@@ -1,67 +1,97 @@
...
-                labelContent: <>Checkbox 3</>,
-            }}
-        ></Checkbox>
-    </li>,
-];
-const childrenList = (
-    <List type={ListTypes.Unordered} id="checkbox-list">
-        {checkboxes}
-    </List>
+const AccordionTemplate = ({ count, children, ...args }) => (
```
EdwinGuzman: I'd call this something else, like `AccordionListTemplate` since it's a list of `Accordion` components and not the component itself. Maybe the returned HTML should be ul/li of `Accordion`s.

EdwinGuzman:
Cool, the code and update look great but just had a minor comment.
I checked the dist file and both (from this branch and `development`) are the same size so it seems that lodash isn't being picked up in the dist, and probably because it's only used in the stories as well. That's no longer a concern.

helenvholmes:
> Cool, the code and update look great but just had a minor comment.
> I checked the dist file and both (from this branch and `development`) are the same size so it seems that lodash isn't being picked up in the dist, and probably because it's only used in the stories as well. That's no longer a concern.

Rad, I was suddenly worried I misunderstood `devDependencies` vs. regular dependencies. I'll work through those other nits and get those changes up today

EdwinGuzman (approved):

PR #394, opened 2020-09-14
Fix #325

### LINKED ISSUE #325: Spacing Stories Improvement
- Add spacing example to spacing story with a component, colored margin (controlled by knobs) (show code for this)
- Do the same thing in the Template component so you can see everything change via knobs

helenvholmes:
Fixes #325 

- Adds a spacing story that shows a list of cards with a control for `marginBottom` 

### Front End Review:
- [ ] View [the spacing example in Storybook](https://deploy-preview-394--stoic-murdock-c7f044.netlify.app/?path=/story/spacing--spacing-in-context)

### DIFF
--- a/src/components/Accordion/Accordion.stories.tsx
+++ b/src/components/Accordion/Accordion.stories.tsx
@@ -21,7 +21,7 @@ export default {
 const AccordionListTemplate = ({ count, children, ...args }) => (
     <List type={ListTypes.Unordered} modifiers={["no-list-styling"]}>
         {range(count).map((i) => (
-            <li>
+            <li key={i}>
                 <Accordion {...args}>{children}</Accordion>
             </li>
         ))}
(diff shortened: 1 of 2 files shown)


src/components/Accordion/Accordion.stories.tsx
```diff
@@ -21,7 +21,7 @@ export default {
 const AccordionListTemplate = ({ count, children, ...args }) => (
     <List type={ListTypes.Unordered} modifiers={["no-list-styling"]}>
         {range(count).map((i) => (
-            <li>
+            <li key={i}>
```
hokei, line 24: I think it's not advised to use the iterator for keys (https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318).  Is it reasonable to ask accordion children to have ids?

hokei: edit: oh my bad, I thought this was accordion.tsx rather than accordion.stories.tsx.  It's ok here since the data is static and in the story -- ignore me =)

hokei (approved):
this is super nifty!

PR #413, opened 2020-10-06
Checkbox Component Can't Be Used as a Controlled Component

### LINKED ISSUE #395: Checkbox Component Can't Be Used as a Controlled Component
### Description of issue:
- The Checkbox component does not use the `checked` prop, which is needed to use a checkbox as a controlled component (typically you'd use controlled components when using redux, but it varies). 

The problem is you shouldn't use `checked` and `defaultChecked` props together. `defaultChecked` is for uncontrolled components, `checked` is for controlled components.

One possible way to handle this is how the Salesforce DS does it, with a conditional, to only use checked if it's defined:

```
<input
  disabled={props.disabled}
  /* A form element should not have both checked and defaultChecked props. */
  {...(props.checked !== undefined
    ? { checked: props.checked }
    : { defaultChecked: props.defaultChecked })}
  ... other stuff ...
/>
```

### Other DS checkboxes for reference:
Salesforce: https://github.com/salesforce/design-system-react/blob/master/components/checkbox/index.jsx#L250
React Aria: https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/checkbox/src/useCheckbox.ts

- Checkbox

nypl-wluisi:
Fixes #395 

- Allows checkbox to be used as a controlled component, as well as an uncontrolled component
- Removes the `isSelected` prop from checkbox component.
- Replaces with either `checked` (controlled) or `defaultChecked` (uncontrolled) props

### To Do 12/1/2020
- [ ] Fix space after "checkbox" class name, currently renders like `<div class="checkbox ">`
- [x] Wire up storybook "controls" to update state of controlled checkbox. 

### Front End Review:


helenvholmes:
@wluisi @nypl-wluisi Hey Bill, what's the status on this?

src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -1,33 +1,93 @@
...
 };
 
 const Template: Story<CheckboxProps> = (args) => <Checkbox {...args} />;
 
-export const BasicCheckbox = Template.bind({});
-BasicCheckbox.args = {
+// Uncontrolled checkbox.
+export const UncontrolledCheckbox = Template.bind({});
+
+// Uncontrolled checkbox set props
```
hokei, line 25: Does it represent a breaking change to remove the uncontrolled checkbox entirely?  Controlled components are best practice, so it would be nice for us to encourage their usage by having the example only be things that are best practice.

src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -1,33 +1,93 @@
...
+const ControlledCheckboxTemplate: Story<CheckboxProps> = (args) => {
+    const [isChecked, setChecked] = useState<boolean>(false);
+
+    const onChange = (event) => {
+        setChecked(!isChecked);
+    };
+
+    return (
+        <Checkbox
+            {...args}
```
hokei, line 55: what additional args are being passed?  Can we `args.checked` directly instead of in a spread like this?  Because it's documentation I'd like things to be as explicit as possible, but I'm not sure if this is best practice somewhere else.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -11,8 +11,10 @@ export interface CheckboxProps {
     checkboxId?: string;
     /** ClassName that appears in addition to "checkbox" */
     className?: string;
-    /* The current selected state of the checkbox */
-    isSelected?: boolean;
+    /* For a controlled component only. If checked is not defined, the state of the uncontrolled native `input` component will be used. */
+    checked?: boolean;
+    /* For uncontrolled component only. The initial value of an uncontrolled checkbox component. */
```
hokei, line 16: ah I see, this is gonna be a breaking change.  Let's keep this, but put in the comments here (and in the stories) that we should be using the controlled version, and that `defaultChecked` will be deprecated in the next major release.

EdwinGuzman: You mentioned it's a best practice to use controlled components only and that's mostly true. The React team does recommend using controlled components but they also recognized that when using form elements with other libraries, you may need to use uncontrolled components and `ref`. This is the use case I ran into when using `react-hook-form`. For that library, a `ref` is passed to subscribe to the library to keep track of data. There are alternatives for this and I can change the `Checkbox` component to be controlled, but this would be a bigger manual change.

I would keep both versions and not deprecate the usage of uncontrolled components for those cases where it's useful.

EdwinGuzman:
For controlled checkboxes with a `checked` prop, an `onChange` prop is also needed or this error shows up:

Screen Shot 2020-12-02 at 10 15 33 AM

This is coming from my app but I think it'll be good to have this dependency in this component as well to better know where the error is coming from.

src/components/Checkbox/Checkbox.tsx
nypl-wluisi: Yeah we want to allow either `checked` (controlled) or `defaultChecked` (uncontrolled) props, as there are use cases for both. The breaking change is removing the `isSelected` prop.

hokei: Thanks for the clarification - let's keep both.  I know it's generally the responsibility of the app developer to know stuff that's written in React documentation, but it would be helpful to have a few lines in the comments that point a new developer to which version they should be using.

src/components/Checkbox/Checkbox.stories.tsx
nypl-wluisi: I think we can resolve this question as its covered below.

src/components/Checkbox/Checkbox.stories.tsx
nypl-wluisi: I think just checked is needed. Most of the examples from the storybook documentation have the args spread: https://storybook.js.org/docs/react/essentials/controls -- thats where I got this from. I'm fine with doing either. Ive made some changes to get this to work with controls, so there's also this in the storybook example now that explains what we're doing with the args:

```javascript
// Allows storybook controls to toggle checked state.
useEffect(() => {
  setChecked(args.checked);
}, [args.checked]);
```

nypl-wluisi:
@EdwinGuzman so sounds like we should just make the `onChange` prop required?

EdwinGuzman:
Is it possible to only make it required or check if it's passed if `checked` is passed? Otherwise, that might be the way to go.

nypl-wluisi:
> Is it possible to only make it required or check if it's passed if `checked` is passed? Otherwise, that might be the way to go.

I can try a few things. Are there any use cases where you'd have an uncontrolled checkbox, but want to do something using `onChange` prop?

EdwinGuzman:
I haven't ran into that situation. Right now, for the uncontrolled checkbox, I'm passing a `ref` that handles getting the data so I have no need to pass the `onChange` prop. I imagine an empty function might be okay but I would want to test it first.

nypl-wluisi:
> I haven't ran into that situation. Right now, for the uncontrolled checkbox, I'm passing a `ref` that handles getting the data so I have no need to pass the `onChange` prop. I imagine an empty function might be okay but I would want to test it first.

@EdwinGuzman 
Ok, I think we can make the `onChange` prop optional, and only available if `checked` prop is used, so something like this:

```
{...(props.checked !== undefined
  ? {
      checked: props.checked,
      "aria-checked": props.checked,
      onChange: props.onChange
    }
  : { defaultChecked: props.defaultChecked })}
```

Not sure if we can make `onChange` required only if `checked` prop is used tho. how does this sound?

EdwinGuzman:
Yeah, that looks good. I think @hokei did something in some components where it checks to see if both props are passed and it spits out a warning saying that if one prop is passed, then a second is also required.

hokei:
I just checked and threw warnings explicitly in the app - It was mostly A11Y hints.  I think it's nice to be explicit, but I don't think it's necessary.  In situations like this, I fall back on, "Is it easy for the developer to think that this is working when really it isn't" (if yes, throw warnings/errors) and "Will it crash the app if it doesn't work" (If yes, try not to crash the app if at all possible, so catch/throw warning/provide a fallback)

nypl-wluisi:
@EdwinGuzman if you get a chance, can you re-test this in your app? I made the onChange prop optional and only avail if checked prop is used. There's also one small issue, but I think it only happens in storybook, where there's an extra space after the class name:

ds-checkbox-class-extra-space

Im using this DS branch on location finder, and I don't get the extra space after the class name.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -33,21 +35,27 @@ let Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(
             blockName = "",
             checkboxId,
             labelOptions,
-            isSelected = false,
+            checked,
+            defaultChecked,
```
EdwinGuzman, line 39: These two values are not being used because on lines 54, 55, and 58, you're using `props.checked` and `props.defaultChecked`.

EdwinGuzman:
The update looks good but `onChange` is a required prop so I do get an error about it:
Screen Shot 2020-12-21 at 10 05 17 AM

Otherwise, the pattern to add it to the `input` whether the `checked` or `defaultChecked` prop was passed makes sense.

One minor thing, I also see the extra space in the class name but it may be related to how BEM works.

nypl-wluisi:
> The update looks good but `onChange` is a required prop so I do get an error about it:
> Screen Shot 2020-12-21 at 10 05 17 AM
> 
> Otherwise, the pattern to add it to the `input` whether the `checked` or `defaultChecked` prop was passed makes sense.
> 
> One minor thing, I also see the extra space in the class name but it may be related to how BEM works.

Do you get the extra space in the class name in your app or only in storybook?

EdwinGuzman:
It happens in the app and in [storybook](https://deploy-preview-413--stoic-murdock-c7f044.netlify.app/?path=/story/checkbox--controlled-checkbox).

src/components/Checkbox/Checkbox.tsx
nypl-wluisi: This should be fixed now. I removed the `props.` and just using `checked` and `defaultChecked` directly.

nypl-wluisi:
> It happens in the app and in [storybook](https://deploy-preview-413--stoic-murdock-c7f044.netlify.app/?path=/story/checkbox--controlled-checkbox).

Yeah it might be something in the bem util function. I noticed it happens on other components as well like `List`.

nypl-wluisi:
> The update looks good but `onChange` is a required prop so I do get an error about it:
> Screen Shot 2020-12-21 at 10 05 17 AM
> 
I set `onChange` to be optional now.

EdwinGuzman (approved):
The update looks good to me.

Issue #416, opened 2020-10-07
Mobile Input Group stlying


EdwinGuzman:
### Description of issue:
- The mobile styling for the `.input-group` class doesn't fit the expected mobile design.
Example of what displays now:
<img width="305" alt="Screen Shot 2020-10-07 at 2 04 21 PM" src="https://user-images.githubusercontent.com/1280564/95370254-7f674600-08a6-11eb-80eb-fa2af6b5fc85.png">

Example of how it should display:
<img width="263" alt="Screen Shot 2020-10-07 at 2 04 03 PM" src="https://user-images.githubusercontent.com/1280564/95370265-842bfa00-08a6-11eb-86d8-e33a3340f6e9.png">

### Component in question (if applicable):
- `Input`

PR #417, opened 2020-10-07
Adding a hidden type and updating the mobile styles for the  class.

### LINKED ISSUE #415: Hidden Input
### Description of issue:
- The `Input` element can use an `InputTypes` of `"hidden"`.

```
  <Input
    type={InputTypes.hidden}
    id={id}
    attributes={{
       ["aria-hidden"]: true,
       name: "hiddenInput",
    }}
    defaultValue="some value"
    ref={ref}
  />

// would render something like:

  <input
    type="hidden"
    aria-hidden={true}
    name="hiddenInput"
    defaultValue="some value"
    ref={ref}
  />
```

- `Input`

EdwinGuzman:
Fixes #415 and #416 

- This adds a `"hidden"` InputType and renders a hidden `input` element from it with `aria-hidden` set to true.
- This also updates the mobile design for the `.input-group` class so that input fields are stacked rather than side-to-side.

Note: the fields bleed into the right edge but that's storybook styling.
current:
Screen Shot 2020-10-07 at 2 55 41 PM

update:
Screen Shot 2020-10-07 at 2 55 20 PM

### Front End Review:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 Currently, this repo is in Prerelease. When it is released, this project will adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 ========
 
+## 0.18.1
+
+### Breaking Changes
+
+-   For `Input` components that use the `.input-group` class, on mobile those input elements are now stacked on top of each other rather than side-to-side.
+
+### Updates
+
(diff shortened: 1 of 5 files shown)


EdwinGuzman:
Ugh, yes github-actions bot, I did forget to update the changelog this time around.

helenvholmes (approved):
Broken storybook styling should get fixed with some work I'm doing in a new component branch. (`.nypl-ds` class just needs to be added to Storybook so all children have `box-sizing: border-box;`, which will fix this.)

Looks great Edwin, thank you.

PR #420, opened 2020-10-23
GH Hosted Storybook Cleanup

helenvholmes:
- Adds `.nypl-ds` class to our Storybook as a wrapper
- Fixes issue in Accordion sets where first item in set had a `-1px` margin-top

### DIFF
--- a/src/components/Accordion/_Accordion.scss
+++ b/src/components/Accordion/_Accordion.scss
@@ -29,7 +29,7 @@
     }
 
     // Eliminates the double border on accordion sets in ul-lis
-    li & {
+    li:not(:first-of-type) & {
         margin-top: -1px;
     }
 }
--- a/.storybook/preview-body.html
+++ b/.storybook/preview-body.html
@@ -1,0 +1,12 @@
+<!--  .storybook/preview-body.html -->
+
+<script>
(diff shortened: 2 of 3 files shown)


src/components/Accordion/_Accordion.scss
```diff
@@ -29,7 +29,7 @@
     }
 
     // Eliminates the double border on accordion sets in ul-lis
-    li & {
+    li:not(:first-of-type) & {
         margin-top: -1px;
```
hokei, line 33: why is there even a negative margin?  haha

hokei (approved):

helenvholmes:
Screenshots, since this one is subtle for the Accordion:

Pre-changes: 
Pre changes

Post changes: 
- `.nypl-ds` class is making certain we're applying our type settings
- we're removing the `1px` margin on the first-of-type for accordion sets, so it's no longer blank white at the top
Post changes

src/components/Accordion/_Accordion.scss
helenvholmes: There's a negative margin on the other items in the list (that are not first) to prevent the double-border in accordion sets.

PR #425, opened 2020-10-28
Upgrades Storybook

### LINKED ISSUE #422: Various issues with Docs tab
### Description of issue:
The docs tab on our stories has a couple of regressions since when? unclear:

- Canvas tab doesn't have the `.nypl-ds` class applied, so basic things like type-settings and box-sizing settings aren't being applied.
- "Code" button is grayed out for all components, showing inactive "No code available"—this means everyone using the documentation needs to go into the `.stories.tsx` files themselves, which isn't a deal breaker but a definite bummer.

- [An example page to see above bugs](https://nypl.github.io/nypl-design-system/storybook-static/?path=/docs/modal--modal)

helenvholmes:
Fixes #422 

- Upgrades storybook to latest patch version
- Applies `.nypl-ds` class to docs page as well as canvas

### Front End Review:


package.json
```diff
@@ -44,13 +44,13 @@
...
         "@babel/plugin-proposal-export-default-from": "^7.5.2",
         "@babel/plugin-transform-arrow-functions": "^7.2.0",
-        "@storybook/addon-a11y": "^6.0.20",
-        "@storybook/addon-actions": "^6.0.20",
-        "@storybook/addon-docs": "^6.0.20",
-        "@storybook/addon-essentials": "^6.0.20",
-        "@storybook/addon-knobs": "^6.0.20",
-        "@storybook/addon-links": "^6.0.20",
-        "@storybook/react": "^6.0.20",
+        "@storybook/addon-a11y": "^6.0.27",
```
hokei, line 47: shouldn't all of this already get pulled in for anyone who is using it because caret versioning?

hokei (approved):

package.json
helenvholmes: Possibly! I just ran the command that Storybook was throwing me a warning for. It probably means that our hosted Storybook (since it re-installs dependencies each time) has the latest and greatest already...

Issue #419, opened 2020-10-22
Modal is wider than screen-width


siberry:
### Description of issue:
- Model extends past width of screen. Cannot scroll to see rest of content

### Component in question (if applicable):
- Modal

helenvholmes:
Hey @siberry, I'm not seeing this in our [modal story](https://nypl.github.io/nypl-design-system/storybook-static/?path=/story/modal--modal) but that is possibly a result of our modal story being too bare-bones. Can you link to an instance where this is happening for you in a QA environment? Thanks!

siberry:
I used the modal in this open pr: https://github.com/NYPL-discovery/discovery-front-end/pull/1510
These lines fixed the issue:
[styling modal](https://github.com/NYPL-discovery/discovery-front-end/pull/1510/files#diff-bee91a76598956e9a02823e906efba0ab36014a54452cd27d545bd3e951fd9faR103)
[styling on `.no-scroll` coming from Design System](https://github.com/NYPL-discovery/discovery-front-end/pull/1510/files#diff-bee91a76598956e9a02823e906efba0ab36014a54452cd27d545bd3e951fd9faR114)

This is not deployed yet. I can share a screenshot of the issue on Slack. If you determine we should prevent this from happening with the component out-of-the-box, this could be a good issue for me to pick up :)

helenvholmes:
@EdwinGuzman The CSS you've added for the `.no-scroll` class should definitely come out of the box for sure.

For the `box-sizing` you've added for the modal, is that in _addition_ to the `.nypl-ds` wrapper on the application? Because that seems... odd

EdwinGuzman:
I did not add a `box-sizing` rule to the modal, unless if that question is meant for @siberry ?

You can see an example of the issue on QA LCA - https://qa-librarycard.nypl.org/library-card/new  (I did not update this with the fix and you'll see the page's content shift to the left after filling out the address fields and clicking on next).

helenvholmes:
> I did not add a `box-sizing` rule to the modal, unless if that question is meant for @siberry ?

Whoops, that was! Sorry for the confusion.

> You can see an example of the issue on QA LCA - https://qa-librarycard.nypl.org/library-card/new (I did not update this with the fix and you'll see the page's content shift to the left after filling out the address fields and clicking on next).

And just to clarify, the `position: inherit` fixed this behavior?

EdwinGuzman:
Yes, exactly.

[...]

PR #426, opened 2020-11-10
Make overflow hidden on `<body>` for modal

siberry:
Related to discussion in #419

- Add `overflow: hidden;` to the `<body>` to prevent the vertical scroll from displaying for the content behind the modal.

### Front End Review:
- [ ] View [the example in Storybook](https://nypl.github.io/nypl-design-system/storybook-static/?path=/story/modal--modal)

^ this story does not show the modal over content that extends past the bottom edge of the container. Any thoughts on adding another modal story?

### DIFF
--- a/src/components/Modal/_Modal.scss
+++ b/src/components/Modal/_Modal.scss
@@ -1,6 +1,7 @@
 .no-scroll {
     position: fixed;
     top: -100vh;
+    overflow: hidden;
 }
 
 .modal {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 Currently, this repo is in Prerelease. When it is released, this project will adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 ========
 
+## 0.18.3
+
+### Changes
+
+-   Add `overflow: hidden` to `.no-scroll` styling on `<body>` for modal to prevent scroll element for content behind modal being displayed
+
 ## 0.18.2
 
 ### Changes


helenvholmes (approved):
Looks great Serena, feel free to merge and I can drop 15 on your calendar for running the release

helenvholmes:
> ^ this story does not show the modal over content that extends past the bottom edge of the container. Any thoughts on adding another modal story?

I think it might be nice just to extend the existing story? Just add a bunch of additional content?

Let's not block this work on that, though. I'll open another issue for it.

[EDIT] Opened #428 for this.

EdwinGuzman:
Can we wait a bit until this is merged? I am finding that I still have the issue where the page shifts to the left and I'm still investigating.

Without `position: inherit`:
Screen Shot 2020-11-17 at 4 58 58 PM

With `position: inherit`:
Screen Shot 2020-11-17 at 5 01 09 PM

src/components/Modal/_Modal.scss
```diff
@@ -1,6 +1,7 @@
 .no-scroll {
     position: fixed;
     top: -100vh;
```
EdwinGuzman, line 3: @scooooooooby  what was the idea behind this rule?

EdwinGuzman:
I removed the 
```
.no-scroll {
  position: inherit !important;
}
```
rule in the Library Card app and opted to give `body` a width of 100%. I think in general this might not be a problem since the modal does cover the content underneath. For this app, I brought the opacity down to 60% so you can still see the page but not interact with it because there's a loader displaying. Maybe there should be a component specific to a loader for page requests and transitions.

In any case, I think this PR is good to go and will say the issues I encountered are app-specific.

src/components/Modal/_Modal.scss
helenvholmes: I can't remember, I might not have written this (despite possibly showing up in the `git blame`). @hokei, do you remember?

helenvholmes:
@siberry, looks like you're good to go. Can you merge this?

PR #432, opened 2020-11-18
Styling the wrappers for the Autosuggest stories

### LINKED ISSUE #429: Autosuggest story on Storybook hard to see
### Description of issue:
- For the [Autosuggest stories](https://nypl.github.io/nypl-design-system/storybook-static/?path=/story/autosuggest--autosuggest-library), the suggestions are hidden by the container, making this component hard to decipher. You have to scroll within the container to see them.

- Autosuggest

- https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Main?node-id=11689%3A423

EdwinGuzman:
Fixes #429 

- Adds padding and a min-height to the wrappers of the Autosuggest component so that the suggestions show up in the Storybook stories.
- Also a minor update to two different stories and how they are configured.

### Front End Review:
- [ ] View [the Autosuggest example in Storybook](https://deploy-preview-432--stoic-murdock-c7f044.netlify.app/?path=/story/autosuggest--autosuggest-fish)

### DIFF
--- a/src/components/Autosuggest/Autosuggest.stories.tsx
+++ b/src/components/Autosuggest/Autosuggest.stories.tsx
@@ -106,15 +106,17 @@ export const autosuggestLibrary = () => {
     };
 
     return (
-        <Autosuggest
-            suggestions={suggestions}
-            onSuggestionsFetchRequested={onSuggestionsFetchRequested}
-            onSuggestionsClearRequested={onSuggestionsClearRequested}
-            getSuggestionValue={getSuggestionValue}
-            renderSuggestion={renderSuggestion}
-            inputProps={inputProps}
-            renderInputComponent={renderInputComponent}
-        />
+        <div style={{ padding: "5px", minHeight: "400px" }}>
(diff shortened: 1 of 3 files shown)


helenvholmes (commented):
Should a `min-height` be added onto the `.autosuggest {}` class instead of on the outer wrapper?

EdwinGuzman:
I was trying to add a min-height to the wrapper of the story and just realized it's on the wrapper of the example component. So I'll update it. It's not the Autosuggest component itself that needs it but if there's nothing else on the page, you can't see the suggestion dropdown. I added other dummy components below it and the suggestion dropdown appeared, but the page itself still cuts it off and begins to scroll.

EdwinGuzman:
Okay, updated. So now the component itself doesn't need that rule but it's the wrapper that needs it, but only because nothing else is rendered on the page so there's not enough space for the suggestions dropdown to display.

helenvholmes (approved):
LGTM! 👍🏻

PR #436, opened 2020-11-18
Adding content for the Modal story.

### LINKED ISSUE #428: Add additional content to force scroll on Modal story
### Description of issue:
- Modals get funky when the content under them scrolls. It would be good for there to be more in the content body of our `Modal` story.

- `Modal`

EdwinGuzman:
Fixes #428

- This adds more content to the `Modal` component's story example for when the modal is closed and opened. Also fixed some accessibility warnings from the `SearchBar` story example.

### Front End Review:
- [ ] View [the Modal example in Storybook](https://deploy-preview-436--stoic-murdock-c7f044.netlify.app/?path=/story/modal--modal)

### DIFF
--- a/src/components/Modal/Modal.stories.tsx
+++ b/src/components/Modal/Modal.stories.tsx
@@ -4,7 +4,6 @@ import Button from "../Button/Button";
 import Modal from "./Modal";
 import { ExampleCard } from "../Card/Card.stories";
 import { searchBar } from "../SearchBar/SearchBar.stories";
-import { skeletonLoader } from "../SkeletonLoader/SkeletonLoader.stories";
 
 export default {
     title: "Modal",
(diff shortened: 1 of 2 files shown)


helenvholmes:
I'd modify the story to not have the skeleton loader outside of the modal and _inside_ the modal—at first I thought this was a bug. Maybe instead of the skeleton loader we can just have some ipsum lorem text?

helenvholmes (approved):
Looks good to me! 👍🏻

siberry:
I think this will be very helpful for the story! I wonder if the modal content is longer than the height of the screen, will `overflow: hidden` (the change I just made) prevent the bottom content from being seen? (I'm thinking we may need refine that change.)

siberry (approved):

PR #438, opened 2020-11-18
Optionally adding an id attribute to the Input component

### LINKED ISSUE #433: Empty `id` prop renders `undefined` for `Input`
### Description of issue:
- The `id` prop is optional but it is still added to the id for the `input` element that gets rendered as `input-${id}`. But, if it's optional and no value is passed, the `Input` component renders `<input id="input-undefined" .... />`.

If there are multiple `Input` components on a page and no id values are passed to any of them, there will be duplicate id issues. Example from `react-axe`:

Screen Shot 2020-11-18 at 12 04 19 PM

Solution:
Make the `id` required.

- `Input`

EdwinGuzman:
Fixes #433 

- Conditionally renders an `id` for the `Input` component. The `id` is an optional prop but if one isn't added, the `Input` component still renders an `id` as `id="input-undefined"`.

Does this update require a changelog update? It's not changing the API but it is updating what gets rendered.

Another approach I was thinking of doing was making the `id` required so it always renders an id. What do you think?

### Front End Review:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 Currently, this repo is in Prerelease. When it is released, this project will adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 ========
 
+## 0.18.4
+
+### Updates
+
+-   Updates the `Input` component to conditionally render an `id` attribute if an `id` prop value is passed to it.
+
 ## 0.18.3
 
 ### Changes
--- a/src/components/Input/Input.tsx
+++ b/src/components/Input/Input.tsx
@@ -54,6 +54,7 @@ const Input = React.forwardRef<HTMLInputElement, InputProps>((props, ref?) => {
     }
 
     let inputProps = {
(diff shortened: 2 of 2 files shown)


helenvholmes:
> Does this update require a changelog update? It's not changing the API but it is updating what gets rendered.

I think so. I'd bump patch since it's fixing a bug fix.
 
> Another approach I was thinking of doing was making the `id` required so it always renders an id. What do you think?

Crystal and I moved away from required props at @kristojorg's suggestion over the summer... additionally this would be a more disruptive release if we went that route. I think the proposed solution you have in the PR is a good one, let's move forward with that.

kristojorg:
Question - why not just not render the id if one isn't defined? I believe you'll get browser errors if you have multiple elements with the same ID , which would happen here if you don't pass that prop to the component right?

EdwinGuzman:
Yes, if you don't pass in the `id` prop and have multiple `Input` components, you'd get duplicate id errors since `id-undefined ` will be used for all the components. This PR fixes that by making it so that the id isn't rendered if it is not passed in.

kristojorg:
Ah okay I see, awesome 👍

helenvholmes (approved):
Looks great!

PR #440, opened 2020-11-19
Tabs

### LINKED ISSUE #410: Tabs
### Description of issue:
- [link to designs](https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Main?node-id=12910%3A52412)

- New component | [documentation](https://inclusive-components.design/tabbed-interfaces/)

adriana-alter:
Okay, here's the basic implementation of the tabs (plus tests); resolves https://github.com/NYPL/nypl-design-system/issues/410.  I'm holding off on CSS-ing them up or otherwise complicating them until I know whether this is actually the structure you'd like.  (Also, I realize the sample content in Storybook is super boring; let me know what would be more helpful!  Want me to just copy over the content from the Figma design, or...?)

So, it is visually very plain just for the time being, but it works (and has the accessibility features):
Screen Shot 2020-11-19 at 1 20 04 PM


CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 Currently, this repo is in Prerelease. When it is released, this project will adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 ========
 
+## 0.18.4
```
EdwinGuzman: @scooooooooby , the `Tabs` component is still being worked on by UX so maybe we shouldn't release this WIP? Maybe it doesn't really matter since DS isn't `1.x.x` yet.
Or maybe the code can be merged but not exported until it's fully ready.

src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,129 @@
+import * as React from "react";
+
+export interface TabsProps {
+    items: {
+        [key: string]: JSX.Element;
+    };
+    uniqueId: string;
```
EdwinGuzman: I think the convention is to use the native attribute names as props, so this should be `id`.

CHANGELOG.md
helenvholmes: That makes sense to me. Let's hold off and get it all correct in Main first.

adriana-alter: To be clear, does this mean I'm indefinitely holding off on doing more work on this, or...?

helenvholmes: @adriana-alter For the time being, yes. I'll work on getting #410 properly triaged for you today.

adriana-alter: Okay, thanks :)

helenvholmes: Might actually happen the week after Thanksgiving now that I'm in it—I'll need Kathy to approve the designs first, and I don't know what her day is like. For now you're off the hook though Adriana, thanks for getting us started here!

adriana-alter: Happy Thanksgiving, and looking forward to CSS-ing (or whatever else you end up needing)! 😃

EdwinGuzman:
I think the JIRA ticket in the SimplyE board can possibly be closed with the caveat that it's a two-step process. Bringing in the code, which was the original task, is done.

There's that issue with the `uniqueId` prop but if we want to fix those later on along with the style update, that's fine with me.

adriana-alter:
Fixed the `uniqueId` thing, and will close the SimplyE ticket!

src/components/Tabs/Tabs.stories.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from "react";
+import Tabs from "./Tabs";
+
+export default {
+    title: "Tabs",
+    component: Tabs,
+};
+
+const tabItems = {
+    "1": <p>Some content for the first tab item...</p>,
```
hokei: Instead of assigning these to props, it would be nice if there was a way to pass them as children.  It makes for a more easy to read, and more flexible component.  There's a way this is being done in List ( https://github.com/NYPL/nypl-design-system/blob/development/src/components/List/List.tsx ) Or, creating a subcomponent called tabItem so that you have a structure that looks like 

```
<Tabs props={}>
<TabItem> {Content 1} </TabItem> 
<TabItem> { Content 2} </TabItem>
(etc) 
</Tabs>
```

adriana-alter: Good idea; I'll take care of it in the next round :)

helenvholmes:
Blocked by [UX-122](https://jira.nypl.org/projects/UX/issues/UX-122?filter=allopenissues).

EdwinGuzman:
@bigfishdesign13 I'm going to close this. I reviewed the Research Catalog tabs component and it is not generic enough to reuse that code. Another reason I'm going to close this is because Chakra has it's own `Tabs` component that we can use - https://chakra-ui.com/docs/disclosure/tabs.

I think that'll be a good first component we use from the Chakra UI integration.

PR #441, opened 2020-11-23
CSS Accordion

### LINKED ISSUE #403: Accordion content rendering
### Description of issue:
- Not an issue per se, but since the content of an accordion is mounted/unmounted when the button is clicked, for form fields the input values get lost. So, if an accordion renders a set of input fields and a user fills it out but then closes the accordion, the values are lost. When the accordion is opened again, the input fields are blank.

Here's an example. There are address-related input fields that, if the accordion it is rendered in is closed, will be lost.

Screen Shot 2020-09-29 at 2 15 57 PM

Possible ways to handle this:

- What if the content is visible through css rather than mounting/unmounting? We'd have to make the visibility accessible/unaccessible.
- What if an optional "onClick" prop is passed to the component so that we can insert a hook? That way, if the accordion is closed, we can insert a function to handle whatever else we need to handle.

- Accordion

EdwinGuzman:
Fixes #403 

- Updates the `Accordion` component to open and close just with CSS. Instead of using local state and a button, there is now an input element and a label element that apply CSS rules to open/close the content area with display `none` or `inline-block`.

One reason why I wanted this change was that closing the modal unmounted the content area and that caused form values to disappear. So now the content gets rendered all the time.

Some problems with this approach:
* The a11y linter complained that the checkbox did not have the right default value so I had to add `value="empty checkbox"`. This seems unnecessary but the linter complained and now I'm wondering if there'll be any consequences to this. I tried a placeholder which seems less risky but that did not work.
* The a11y linter also complained that it needed the right `aria-*` props and I had to add `aria-checked={false}`. This doesn't seem so bad except that there's no way to update this when the checkbox component _is_ checked since the point of this was trying _not_ to use javascript.
* Lastly, I'm having a hard time testing this with enzyme. It's much easier with jest using something like "check to see if this text string is visible to the user" but that's not available through enzyme. Also, it seems you can't check for CSS rules or DOM element styles like "check if `display` is `none`", etc.

So now, would this be worth continuing and merging?

### Front End Review:

### DIFF
--- a/src/components/Accordion/_Accordion.scss
+++ b/src/components/Accordion/_Accordion.scss
@@ -3,32 +3,21 @@
     color: inherit;
     width: 100%;
 
-    input {
-        position: absolute;
-        opacity: 0;
-        z-index: -1;
-
-        &:checked {
-            + .accordion__label .icon--minus {
-                display: inline;
-            }
-            + .accordion__label .icon--plus {
-                display: none;
-            }
-            ~ .accordion__content {
-                display: inline-block;
-            }
-        }
+    &__content {
+        display: none;
+        padding: var(--space-xxs) var(--space-xxxl) var(--space-xs)
+            var(--space-s);
+        transition: all 0.35s;
     }
 
     &__label {
         @include button--outline;
 
-        border: none;
-        justify-content: space-between;
-        font-weight: 500;
-        cursor: pointer;
         align-items: center;
+        border: none;
+        cursor: pointer;
+        font-weight: 500;
+        justify-content: space-between;
 
         .icon--minus {
             display: none;
@@ -44,11 +33,24 @@
         }
     }
 
-    &__content {
-        padding: var(--space-xxs) var(--space-xxxl) var(--space-xs)
-            var(--space-s);
-        display: none;
-        transition: all 0.35s;
+    input {
+        opacity: 0;
+        position: absolute;
+        z-index: -1;
+
+        &:checked {
+            + .accordion__label .icon--minus {
+                display: inline;
+            }
+
+            + .accordion__label .icon--plus {
+                display: none;
+            }
+
+            ~ .accordion__content {
+                display: inline-block;
+            }
+        }
     }
 
     // Eliminates the double border on accordion sets in ul-lis
--- a/src/components/Accordion/_Accordion.scss
+++ b/src/components/Accordion/_Accordion.scss
@@ -1,48 +1,37 @@
-.tab {
-    &-label {
-        display: flex;
-        justify-content: space-between;
-        padding: 1em;
-        font-weight: bold;
-        cursor: pointer;
-    }
-}
-
-input {
-    position: absolute;
-    opacity: 0;
-    z-index: -1;
-}
-.tab-label .icon--minus {
-    display: none;
-}
-
-input:checked {
-    + .tab-label .icon--minus {
-        display: inline;
-    }
-    + .tab-label .icon--plus {
-        display: none;
-    }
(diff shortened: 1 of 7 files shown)


helenvholmes:
> * The a11y linter complained that the checkbox did not have the right default value so I had to add `value="empty checkbox"`. This seems unnecessary but the linter complained and now I'm wondering if there'll be any consequences to this. I tried a placeholder which seems less risky but that did not work.
> * The a11y linter also complained that it needed the right `aria-*` props and I had to add `aria-checked={false}`. This doesn't seem so bad except that there's no way to update this when the checkbox component _is_ checked since the point of this was trying _not_ to use javascript.

Quoting this to cc Zach.

zpeyton:
Hi @EdwinGuzman and @scooooooooby The linter in this case is a bit overzealous. Standard `<input type="checkbox"` does not need a default value and also do not need aria-* props. Screen readers know how to handle these built in UI components. Only if you make a custom checkbox like, `<div role="checkbox" aria-checked="false">` do you need to use javascript to update the aria attributes with the state change.

EdwinGuzman:
Thanks, @zpeyton .

Updated the PR: removed strict tslint a11y rules and cleaned up the component and test a bit. It's hard to do visual testing with Enzyme and this is one reason why I think we should move to Jest/react-testing-library. Those provide a better way to test components visually.

The test for the `Accordion` is not very in-depth right now but we do have Storybook for interactive UI testing. I tried using `window.getComputedStyle` and it partially worked but did not return the style I was looking for, specifically "none" or "inline-block" for the `display` property.

src/components/Accordion/_Accordion.scss
```diff
@@ -3,15 +3,35 @@
...
+            ~ .accordion__content {
+                display: inline-block;
+            }
+        }
+    }
+
+    &__label {
         @include button--outline;
 
-        align-items: center;
```
helenvholmes, line 9: Small nit, but by removing this the icon is no longer centered vertically within the accordion label.

helenvholmes (changes requested):
One small change request, but this looks great—thanks Edwin.

src/components/Accordion/_Accordion.scss
EdwinGuzman: Nice catch. Updated.

src/components/Accordion/_Accordion.scss
```diff
@@ -3,15 +3,36 @@
...
         border: none;
-        display: inline-flex;
-        font-weight: 500;
         justify-content: space-between;
-        width: 100%;
+        font-weight: 500;
+        cursor: pointer;
+        align-items: center;
```
helenvholmes: Craziest nit but your linter should be bothering you that this should be the first property based on our linting rules—is it not telling you that? https://github.com/NYPL/nypl-design-system/blob/development/.stylelintrc.json#L41

EdwinGuzman: I installed `stylelint` in my vscode so that error shows up now. It looks like stylelint is not setup in the project correctly because that error should come up when running `npm test`.

Although I fixed it for this file, there are errors in other files. This should be a separate ticket not just to fix those files but to make that linter part of the entire process.

I ran `npx stylelint "**/*.scss"` and got this output:

Screen Shot 2020-12-01 at 1 03 28 PM

helenvholmes: Oh nice, guess I was only getting those from my editor directly—so lot of stuff was getting missed.

I've filed #450 to track this.

helenvholmes (approved):
Looks good!

PR #442, opened 2020-11-23
RENO-1881: Update Pagination component (in progress)

### LINKED ISSUE #437: Revise Pagination Component
### Description of issue:
- Update `Pagination` component to use newer design.

New design looks like this:
new-pagination

Existing Pagination is flexible and allows for components to be passed as props to build the Pagination as needed:

```jsx
<DS.Pagination
  nextPage={NextButton(pageNumber, pageCount)}
  previousPage={PrevButton(pageNumber)}
>
  <DS.Label
    htmlFor="paginationSelect"
    id="paginationLabel"
  >
    Page{' '}
  </DS.Label>
  <DS.Select
    ariaLabel="Pagination Label"
    id="paginationSelect"
    labelId="paginationLabel"
    name="Pagination Select"
    onBlur={onPageChange}
    onChange={onPageChange}
    selectedOption={`${pageNumber} of ${pageCount}`}
  >
    {paginationDropdownOptions.map((option) => (
      <option>
        {option}
      </option>
    ))}
  </DS.Select>
</DS.Pagination>
```

For the new design variation, we could do similar:

```jsx
<DS.Pagination
  nextPage={<DS.Link>Next</DS.Link>}
  previousPage={<DS.Link>Previous</DS.Link>}
>
  {paginationPages.map((page) => (
    <DS.Link>
      {page}
    </DS.Link>
  ))}
</DS.Pagination>
```

- [ ] Update `Pagination` component
- [ ] Update Storybook integration
- [ ] Update basic existing tests

- Pagination

- [Metronome](http://themetronome.co/components/mn-10-pagination/?fresh=true)

- [Figma Variation for Longer Sets](https://www.figma.com/file/CCCdFt8H7pMqHjkRQlivSw/IA---Templates-%2B-Components?node-id=41654%3A60531)
- [Figma Variation for A-Z]()

misselbeck:
Fixes #437 

I took a first pass (rough draft) at this based on the Figma design. There seems to be some inferred logic that the component would need to handle, such as navigating between items, applying an active state, and passing back the selected item to the consuming application to apply to its own app.

I came up with a component that takes in a page count and a callback function as props.

### Props

| Name  | Type | Description
| ------------- | ------------- |------------- |
| `pageCount`  | `Number`  |  The total number of items. |
| `onPageChange` | `Function`  | The method to call when an item is selected.<br>Exposes the selected page as an argument.

### Example Usage
```
<DS.Pagination
  pageCount={10}
  onPageChange={pageUpdate}
/>
```

### Result

nypl-pagination

### To-Do

**Truncation**

Based on the Figma designs, there is also some logic needed to truncate long lists, using ellipses as placeholders, and showing just the first, last, active page + 2 siblings at a time on each side (?).

figma

This looks like it changes a bit if you are at the beginning or the end of the list, in which it only shows one set of ellipses (either just after the first item, or just before the last item).

I couldn't quite work out the logic in the first pass. After a little investigation, it seems like maybe we want something similar to how the Pagination works in the material-ui project.

Does the logic here look like it would meet the Figma requirements?
https://material-ui.com/components/pagination/#pagination-ranges

I can try to implement something similar in this component (though it looks tricky).

**Previous / Next**

Based on the Figma designs, the Previous button should not display if at the first item, and the Next button should not display if at the last item.

I have this working, but adding/removing the buttons makes the pager "jump." As a page selection most likely will not trigger a page refresh in a React app, maybe it would be better to just have the Previous/Next buttons just become disabled as needed, rather than removing completely?

### Note

I have not touched the Storybook/Tests for this component yet. They are outdated in this branch at the moment.

[Pagination in Storybook](https://deploy-preview-442--stoic-murdock-c7f044.netlify.app/?path=/story/pagination--pagination)


src/components/Pagination/Pagination.tsx
```diff
@@ -1,42 +1,142 @@
...
+            this.props.onPageChange({ selected: selectedItem });
+        }
+    };
+
+    getPageElement(index) {
+        const { selected } = this.state;
+
+        return (
+            <li key={index + 1}>
+                <Button
```
hokei: I think you want to be using `<Link>` here instead of `<Button>`, because the underlying native element you want is an `<a>` element rather than a `<button>` element.  (Same goes for "next" and "previous")

src/components/Pagination/Pagination.tsx
```diff
@@ -1,42 +1,142 @@
...
+            <nav
+                aria-label="Pagination"
+                className={bem(
+                    pagination__base_class,
+                    this.props.modifiers,
+                    this.props.blockName,
+                    [this.props.className]
+                )}
+            >
+                <ul>
```
hokei: ~~This is kind of a nit because I'm not sure it matters, ultimately, but `<ul>` doesn't seem semantically correct here.  You can't re-order the `<li>` items at will.  I'm seeing that Amazon is using `ul` anyway, while Google is using `table`, and both are weird to me.  I think I like 
https://react.semantic-ui.com/addons/pagination/
the best, where it's just a div with a bunch of links in it.  This is not tabular data, so table seems weird.  `ol` might be fine? Thoughts?~~

edit: I'm a dumb person who got ol and ul mixed up cuz clearly I can't spell.  `<ul>` is clearly fine here.  Sorry, ignore this.

src/components/Pagination/Pagination.tsx
misselbeck: I made this change. All the page items are now using the `<Link>` component.

misselbeck:
**To Do**
- [x] Add a Storybook entry
- [x] Add truncation logic and display
- [x] Write a test file

helenvholmes:
@misselbeck, I've updated your PR description to add the multidev preview. You can grab that link in the future by scrolling in the checks area to `netlify/stoic-murdock...etc` and hitting the 'Details' link.

helenvholmes:
> Does the logic here look like it would meet the Figma requirements?
> https://material-ui.com/components/pagination/#pagination-ranges
> 
> I can try to implement something similar in this component (though it looks tricky).

We styled our designs off of Google's pagination (since they're so good at handling large search result sets like this), so yes! I would go even further and say that if our designs are _missing_ something, this is the component we'll be looking to for guidance. Best of luck, I think we're all aware this component is going to be pretty tricky. ☹️
 
> **Previous / Next**
> 
> Based on the Figma designs, the Previous button should not display if at the first item, and the Next button should not display if at the last item.
> 
> I have this working, but adding/removing the buttons makes the pager "jump." As a page selection most likely will not trigger a page refresh in a React app, maybe it would be better to just have the Previous/Next buttons just become disabled as needed, rather than removing completely?

That's a good suggestion, so I've created [UX-162](https://jira.nypl.org/browse/UX-162) to make this update.

helenvholmes:
An odd thing happening on the multidev is that the knobs for this component are appearing across of the other components on the multidev... super weird

src/components/Pagination/Pagination.stories.tsx
```diff
@@ -1,99 +1,29 @@
...
-                {text("Option 7 Text", "7 of 7")}
-            </option>
-        </Select>
-    </>
-);
-
-const PaginationTemplate = ({ children, ...args }) => (
-    <Pagination {...args}>{children}</Pagination>
+const PaginationTemplate: Story<PaginationProps> = (args) => (
+    <Pagination {...args}></Pagination>
```
hokei: Is it a Storybook requirement to pass the props in with spread syntax?  If not, I would prefer to pass the args in explicitly so that anyone looking at this documentation knows what the valid and expected args are!

misselbeck: It looks like the other components using a Storybook template are doing it this way. I can investigate if there is another approach.

misselbeck:
> An odd thing happening on the multidev is that the knobs for this component are appearing across of the other components on the multidev... super weird

I can't replicate this. Let me know if this is ongoing. I can revisit my story file for any anomalies.

misselbeck:
Added the truncation logic.

To Do:

- [x] Write a test file
- [x] Get linting to work!

I noticed that the `.prettierrc` settings are adding semicolons to my code (as well as 4 tab indentation, etc). But then when running `npm run lint`, it complains about the semicolons. This also fails during the CI/Lint (push).

Just me? Any tricks to getting this setup properly with VS Code?

nypl-wluisi:
@misselbeck Whenever you get a chance, can you also create a PR off the location finder repo that uses this branch and updates the pagination on location finder? You see how I did similar here: https://github.com/NYPL/dxp-react-search/pull/81/files

In your `package.json` you can point the DS to use this github branch, so something like:
`"@nypl/design-system-react-components": "git+https://git@github.com:NYPL/nypl-design-system.git#RENO-1881-update-pagination-component"`

And you'd have to update this component:
https://github.com/NYPL/dxp-react-search/blob/development/src/components/location-finder/Locations/LocationsPagination.js

We can probs track the work in another ticket, but just doing a basic sanity check that this component will work in LF would be helpful.

helenvholmes:
> I noticed that the `.prettierrc` settings are adding semicolons to my code (as well as 4 tab indentation, etc). But then when running `npm run lint`, it complains about the semicolons. This also fails during the CI/Lint (push).
> 
> Just me? Any tricks to getting this setup properly with VS Code?

Grrr, this is mysterious. You might need to crowd-source solutions in #designs-system, I've never seen something like this :/

helenvholmes:
> > I noticed that the `.prettierrc` settings are adding semicolons to my code (as well as 4 tab indentation, etc). But then when running `npm run lint`, it complains about the semicolons. This also fails during the CI/Lint (push).
> > Just me? Any tricks to getting this setup properly with VS Code?
> 
> Grrr, this is mysterious. You might need to crowd-source solutions in #designs-system, I've never seen something like this :/

Did some research and stumbled across [this thread](https://github.com/palantir/tslint/issues/2431). I think we might need to update our `tslint.json` L14 to `"semicolon": [true, "always", "strict-bound-class-methods"],`. @misselbeck, wanna make that change and see if it fixes `npm test` for you?

misselbeck:
> > > I noticed that the `.prettierrc` settings are adding semicolons to my code (as well as 4 tab indentation, etc). But then when running `npm run lint`, it complains about the semicolons. This also fails during the CI/Lint (push).
> > > Just me? Any tricks to getting this setup properly with VS Code?
> > 
> > 
> > Grrr, this is mysterious. You might need to crowd-source solutions in #designs-system, I've never seen something like this :/
> 
> Did some research and stumbled across [this thread](https://github.com/palantir/tslint/issues/2431). I think we might need to update our `tslint.json` L14 to `"semicolon": [true, "always", "strict-bound-class-methods"],`. @misselbeck, wanna make that change and see if it fixes `npm test` for you?

This looks like it fixes the issue. Thanks for tracking that down! Should I commit it? Here or in another branch?

helenvholmes:
> This looks like it fixes the issue. Thanks for tracking that down! Should I commit it? Here or in another branch?

I say ship it! 🚀 let's do it here, why not.

misselbeck:
> @misselbeck Whenever you get a chance, can you also create a PR off the location finder repo that uses this branch and updates the pagination on location finder? You see how I did similar here: https://github.com/NYPL/dxp-react-search/pull/81/files
> 
> In your `package.json` you can point the DS to use this github branch, so something like:
> `"@nypl/design-system-react-components": "git+https://git@github.com:NYPL/nypl-design-system.git#RENO-1881-update-pagination-component"`
> 
> And you'd have to update this component:
> https://github.com/NYPL/dxp-react-search/blob/development/src/components/location-finder/Locations/LocationsPagination.js
> 
> We can probs track the work in another ticket, but just doing a basic sanity check that this component will work in LF would be helpful.

This is a good idea. I'll start on that.

helenvholmes:
@misselbeck based on our conversation in this thread I've updated the `Pagination` designs to not shift the next/previous buttons and adhere more closely to [Google Material Design's component](https://material-ui.com/components/pagination/). The designs for that are [here](https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Main?node-id=17226%3A932).

src/components/Pagination/Pagination.stories.tsx
misselbeck: Args now passed in specifically, as requested.

misselbeck:
After [wiring up the Pagination in the Location Finder app](https://github.com/NYPL/dxp-react-search/pull/84), I think it's best to go with what @hokei suggested on Slack, and not rely on state and let the consuming app pass the currentPage. All the previous functionality works just as well doing this way, and it avoids potential  issues with keeping the state in sync after re-renders, etc.

- Re-worked the component to remove state and added a new prop for the currentPage
- Updated the Pagination Story
- Wrote some tests

src/components/Pagination/_Pagination.scss
```diff
@@ -11,13 +11,32 @@
         }
     }
 
-    .label {
-        @include space-stack-none;
+    ul {
```
helenvholmes: Can we give these block modifiers instead of styling the elements themselves? 

I might call this `&__numbers`, for example.

Same comments for L19 + L22.

src/components/Pagination/Pagination.tsx
```diff
@@ -1,42 +1,209 @@
...
+        };
+
+        pageAttributes["aria-label"] = item ? item : null;
+
+        return (
+            <li key={item}>
+                <Link
+                    attributes={{ ...pageAttributes }}
+                    className={currentPage === item ? "selected" : null}
+                    href="#"
```
helenvholmes, line 85: Does this being specified here prevent users from updating the route in the url bar?

src/components/Pagination/Pagination.stories.tsx
```diff
@@ -1,99 +1,30 @@
...
-            <Icon
-                name={IconNames.arrow}
-                decorative={true}
-                iconRotation={IconRotationTypes.rotate270}
-                modifiers={["small", "icon-right"]}
-            />
-        </Button>
-    ),
-    children: selectOptions,
+    pageCount: 100,
```
helenvholmes, line 38: For whatever reason I'm not able to click around to different pages on Storybook.

helenvholmes (changes requested):
Some notes on the SCSS but mostly questions.

src/components/Pagination/_Pagination.scss
misselbeck: I updated the component and styles to make better use of the BEM function.

src/components/Pagination/Pagination.tsx
misselbeck: It doesn't affect the url bar. There is an event handler on the click event that prevents that.

src/components/Pagination/Pagination.stories.tsx
misselbeck: I recently re-worked the component to remove having its own state, and instead added a new prop for the currentPage. There are some further details in the comment a little further above, but basically, after testing the Pagination component in the Location Finder, it seemed better to let the consuming application keep track.

Because the component no longer has its own state, you have to use the Storybook control for currentPage to simulate navigating through the Pagination.

I could look into adding a useState hook into the Storybook file so the example would be clickable. Should I give that a shot?

helenvholmes: > I recently re-worked the component to remove having its own state, and instead added a new prop for the currentPage. There are some further details in the comment a little further above, but basically, after testing the Pagination component in the Location Finder, it seemed better to let the consuming application keep track.
On the same page here—this is definitely the right direction for this component.

> Because the component no longer has its own state, you have to use the Storybook control for currentPage to simulate navigating through the Pagination.
> 
> I could look into adding a useState hook into the Storybook file so the example would be clickable. Should I give that a shot?
I would really appreciate you giving this a shot—I know that I wanted to understand how to click around in the component and see it in action. I suspect I won't be the only one confused.

misselbeck: I was able to add the hooks to the Story, so that the Pagination component can be clicked in the example and it will update accordingly.

misselbeck:
Let me know if there are any further changes needed.

src/components/Pagination/_Pagination.scss
```diff
@@ -11,13 +11,32 @@
...
+    &__link {
+        line-height: 1.15;
+        padding: 8px;
+        text-decoration: none;
+
+        &.selected {
+            color: var(--ui-black);
+        }
+
+        &[role="button"] {
```
helenvholmes: Perhaps this should be moved into `_Link.scss`, since it seems like something you'd want to happen universally.

src/components/StatusBadge/StatusBadge.test.tsx
```diff
@@ -10,15 +10,16 @@ describe("StatusBadge Test", () => {
 
     it("Shows status badge", () => {
         wrapper = Enzyme.shallow(
-            <StatusBadge level={"low"} statusBadgeText={"Registration Required"} />
+            <StatusBadge
```
helenvholmes, line 13: Are these `StatusBadge` changes something you've done, @misselbeck, or something showing up from a funky merge?...

helenvholmes (approved):
👏🏻

This looks incredible. Thank you Matt. I left a couple of comments—one about where some CSS should live and some questions about some changes to `StatusBadge`, but both are pretty minor.

src/components/StatusBadge/StatusBadge.test.tsx
misselbeck: I didn't work on the `StatusBadge `component at all. It looks like these files were included as part of a merge from `development` that had a conflict with the CHANGELOG.

misselbeck: Prettier formatting differences, perhaps?

helenvholmes: Ahh, got it.

PR #444, opened 2020-11-25
RENO-1882 - Revise Status Badge Component

### LINKED ISSUE #434: Revise Status Badge Component
### Description of issue:
- Revise the StatusBadge to be an actual component that accepts props, and could be used like this:

```
// Renders gray
<StatusBadge type={'notice'}>Registration Required</StatusBadge>

// Renders yellow
<StatusBadge type={'alert'}>Cancelled</StatusBadge>
```

- [ ] Add the StatusBadge component
- [ ] Integrate with Storybook
- [ ] Write basic tests

Looks like this:

statusBadge

- Status Badge

- [Metronome](http://themetronome.co/components/mt-345-status-badge/?fresh=true)

- [Figma](https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Main?node-id=11256%3A0)

knezmilos:
Fixes #434 

- Adds StatusBadge Component
- Creates basic tests

### Front End Review:
- [ ] View [the example in Storybook](https://deploy-preview-444--stoic-murdock-c7f044.netlify.app/?path=/story/status-badge--status-badge)

### DIFF
--- a/src/components/StatusBadge/StatusBadge.stories.tsx
+++ b/src/components/StatusBadge/StatusBadge.stories.tsx
@@ -11,16 +11,6 @@ export default {
     decorators: [withDesign],
 };
 
-/*export const statusBadge = () => (
-    <StatusBadge level={select(
-        "Status of Badge",
-        ["low", "medium", "high"],
-        "low"
-    )}>
-        {text("Status Badge Text", "Location is temporarily closed")}
-    </StatusBadge>
-);*/
-
 const StatusBadgeTemplate: Story<StatusBadgeProps> = (args) => (
     <StatusBadge {...args} />
 );
@@ -30,7 +20,7 @@ export const statusBadge = StatusBadgeTemplate.bind({});
 // The `args` allow these props to be updated in the UI through the
 // "Controls" tab.
 statusBadge.args = {
-    statusBadgeText: "Click to expand",
+    statusBadgeText: "Location is temporarily closed",
 };
 
 statusBadge.storyName = "Status Badge";
--- a/src/components/StatusBadge/StatusBadge.stories.tsx
+++ b/src/components/StatusBadge/StatusBadge.stories.tsx
@@ -1,25 +1,28 @@
 import * as React from "react";
+
+import StatusBadge from "./StatusBadge";
 import bem from "../../utils/bem";
 import { withDesign } from "storybook-addon-designs";
 import { select, text } from "@storybook/addon-knobs";
 
 export default {
     title: "Status Badge",
+    component: StatusBadge,
     decorators: [withDesign],
 };
 
-export const StatusBadge = () => (
-    <div
-        className={bem("status-badge", [
-            select("Status of Badge", ["inactive", "warning"], "warning"),
-        ])}
-    >
+export const statusBadge = () => (
+    <StatusBadge level={select(
+        "Status of Badge",
+        ["low", "medium", "high"],
+        "low"
+    )}>
         {text("Status Badge Text", "Location is temporarily closed")}
-    </div>
+    </StatusBadge>
 );
 
-StatusBadge.storyName = "Status Badge";
-StatusBadge.parameters = {
+statusBadge.storyName = "Status Badge";
+statusBadge.parameters = {
     design: {
         type: "figma",
         url:
--- a/src/components/StatusBadge/StatusBadge.stories.tsx
+++ b/src/components/StatusBadge/StatusBadge.stories.tsx
@@ -1,9 +1,9 @@
 import * as React from "react";
 
-import StatusBadge from "./StatusBadge";
+import { Story } from "@storybook/react/types-6-0";
+import StatusBadge, { StatusBadgeProps } from "./StatusBadge";
 import bem from "../../utils/bem";
 import { withDesign } from "storybook-addon-designs";
-import { select, text } from "@storybook/addon-knobs";
 
 export default {
     title: "Status Badge",
@@ -11,7 +11,7 @@ export default {
     decorators: [withDesign],
 };
 
-export const statusBadge = () => (
+/*export const statusBadge = () => (
     <StatusBadge level={select(
         "Status of Badge",
         ["low", "medium", "high"],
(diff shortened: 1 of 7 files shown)


src/components/StatusBadge/StatusBadge.tsx
```diff
@@ -0,0 +1,52 @@
+import * as React from "react";
+import bem from "../../utils/bem";
+
+export interface StatusBadgeProps {
+    /** BlockName for use with BEM. See how to work with blockNames and BEM here: http://getbem.com/introduction/ */
+    blockName?: string;
+    /** ClassName that appears in addition to "status-badge" */
+    className?: string;
+    /** Inner text of the element */
+    text?: string;
```
helenvholmes: Let's remove this in favor of the children—no need to have both.

src/components/StatusBadge/StatusBadge.tsx
```diff
@@ -0,0 +1,52 @@
...
+    /** Type of badge **/
+    type?: string;
```
helenvholmes: I'd use a string literal here to expose the [types available](https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Main?node-id=11256%3A0) to developers more clearly. See more info [here](https://github.com/typescript-cheatsheets/react/blob/cc2ec67e0b36b7e838df8f2a54b4cd84df6c78cd/README.md#basic-prop-types-examples).

`type?: ""waiting" | "success";`

src/components/StatusBadge/StatusBadge.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+export default function Heading(props: React.PropsWithChildren<StatusBadgeProps>) {
+    const {
+        blockName,
+        className,
+        text,
+        type,
+    } = props;
+
+    let status_badge_base_class = "heading";
```
helenvholmes: I'd make this `.status-badge` to match what's in the [SCSS file](https://github.com/NYPL/nypl-design-system/blob/development/src/components/StatusBadge/_StatusBadge.scss).

src/components/StatusBadge/StatusBadge.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+    if (!props.children && !text) {
+        throw new Error("Status Badge has no children, please pass prop: text");
```
helenvholmes: Let's throw a warning here instead and remove the part about passing a prop since users should pass a child instead.

src/components/StatusBadge/StatusBadge.tsx
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    let statusBadgeModifiers = ["warning"];
```
helenvholmes: Looking at [Kathy's designs](https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Main?node-id=11256%3A0) maybe type and the modifiers associated with it could become levels—low, medium, high. I'm finding the warning/notice/etc. language confusing, especially as it doesn't explain which is more important than the other. What do you think?

helenvholmes (changes requested):
Because the new component isn't hooked up to the `.stories` file yet, we can't see the work you've done here. Please update `StatusBadge.stories.tsx` as well so we can see your changes. You can get more info on stories [here](https://storybook.js.org/docs/react/writing-stories/introduction).

src/components/StatusBadge/StatusBadge.tsx
knezmilos: I agree.  I was going to rename them to "notice" and "alert", but as the design now has 3 levels, it makes more sense to be low, medium, high.  I will make that adjustment

nypl-wluisi: I think we should also change the prop name from `type` to `level` ? Type has a certain meaning in javascript that might make this confusing. What do you think? @scooooooooby @knezmilos

helenvholmes: I agree, good suggestion @nypl-wluisi.

knezmilos:
Thank you for all the suggestions, fixes are made accordingly.  Please check.
@scooooooooby @nypl-wluisi

src/components/StatusBadge/StatusBadge.stories.tsx
```diff
@@ -1,25 +1,28 @@
...
+    </StatusBadge>
 );
 
-StatusBadge.storyName = "Status Badge";
-StatusBadge.parameters = {
+statusBadge.storyName = "Status Badge";
+statusBadge.parameters = {
     design: {
         type: "figma",
         url:
```
helenvholmes, line 30: Can you swap this URL out with `https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Main?node-id=11256%3A0`? Thanks!

src/components/StatusBadge/StatusBadge.stories.tsx
```diff
@@ -1,25 +1,28 @@
 import * as React from "react";
+
+import StatusBadge from "./StatusBadge";
 import bem from "../../utils/bem";
 import { withDesign } from "storybook-addon-designs";
 import { select, text } from "@storybook/addon-knobs";
```
helenvholmes: Instead of knobs, can you try out these stories using controls? https://storybook.js.org/docs/react/essentials/controls 

Storybook 6 recommends using controls so we're trying to not introduce any more code using knobs. If you'd like some examples of what those looks like within our codebase, Edwin's set up `Checkbox` to use controls.

src/components/StatusBadge/StatusBadge.tsx
helenvholmes: Let's still make this change.

src/components/StatusBadge/StatusBadge.tsx
```diff
@@ -0,0 +1,57 @@
...
+            blockName,
+            className,
+            text,
+            level,
+        } = this.props;
+
+        let status_badge_base_class = "status-badge";
+
+        if (!this.props.children && !text) {
+            throw new Error("Status Badge has no children");
```
helenvholmes: Let's change this to a warning, and remove all references to the `text` prop.

helenvholmes (changes requested):
Looking good, we're getting close!

src/components/StatusBadge/StatusBadge.tsx
```diff
@@ -0,0 +1,49 @@
+import * as React from "react";
+import bem from "../../utils/bem";
+
+export interface StatusBadgeProps {
```
nypl-wluisi, line 4: Should we make some of these props required, i.e. `statusBadgeText` and `level` ? Would you have a badge without a level or text?

helenvholmes: I don't think `statusBadgeText` will be a prop since it will instead be children. I agree that `level` makes sense as a required prop, though—or we should have it default to say, `low`.

knezmilos: In "StatusBadge.tsx", it's already defaulting to "low" and if `level` is passed it will use that value.

helenvholmes: Perfect, you've already anticipated us 😅 thanks Milos.

helenvholmes (approved):
This looks good to me!

nypl-wluisi (changes requested):
@knezmilos This looks great. One last change before we can merge. Can you switch the StatusBadge component to be a functional component instead of a class component? Functional components are the direction React is headed, so for all new DS components we're going to use them.

nypl-wluisi (approved):
This looks good!

PR #457, opened 2020-12-02
Updates hero padding on primary type

helenvholmes:
- Updates `Hero`'s primary option to have a larger top/bottom padding on the inner content via [Reno-1670](https://jira.nypl.org/browse/RENO-1670)

### Front End Review:
- [ ] View [the example in Storybook](https://deploy-preview-457--stoic-murdock-c7f044.netlify.app/?path=/story/hero--hero-primary)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,15 +6,21 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 Currently, this repo is in Prerelease. When it is released, this project will adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 ========
 
+## 0.18.6
+
+### Changes
+
+-   Updates `Hero`'s primary option to have a larger top/bottom padding on the inner content via [Reno-1670](https://jira.nypl.org/browse/RENO-1670)
+
 ## 0.18.5
 
-### Updates
+### Changes
 
 -   Converts the Accordion component to open and close through CSS rather than through Javascript.
 
 ## 0.18.4
 
-### Updates
+### Changes
 
 -   Updates the `Input` component to conditionally render an `id` attribute if an `id` prop value is passed to it.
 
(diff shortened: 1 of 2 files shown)


helenvholmes:
To give some context here, Sharon emailed me yesterday to increase the padding and Kathy identified the ticket where this work happened in Reno ([1670](https://jira.nypl.org/browse/RENO-1670)).

nypl-wluisi (commented):
This looks good. Are there any small tweaks we should make now to cleanup storybook? Seems like this component is using controls, instead of knobs, which is good. Maybe on the "Hero 5050" we can update the controls to use the color picker for `backgroundColor` and `foregroundColor`, instead of text field? "Hero Primary" already has this setup for those props.

helenvholmes:
> This looks good. Are there any small tweaks we should make now to cleanup storybook? Seems like this component is using controls, instead of knobs, which is good. Maybe on the "Hero 5050" we can update the controls to use the color picker for `backgroundColor` and `foregroundColor`, instead of text field? "Hero Primary" already has this setup for those props.

I'm working on a second PR to clean up how Hero.Secondary works along with some of this controls work—Sharon brought up this one explicitly so I wanna get it in quick.

EdwinGuzman (approved):
Works well in the LCA.

EdwinGuzman:
I just noticed this is under "Web-Based Card Creator Support". Does this mean that Sharon (et al.) want this change deployed soon as well?

nypl-wluisi (approved):

helenvholmes:
> I just noticed this is under "Web-Based Card Creator Support". Does this mean that Sharon (et al.) want this change deployed soon as well?

That was certainly where she noticed the bug. I would imagine a redeployment with this work in it getting deployed along a reasonable timeline would be hoped for

EdwinGuzman:
Yeah, that's fine, just finding that out through this PR is a bit odd to me but nbd.

helenvholmes:
> Yeah, that's fine, just finding that out through this PR is a bit odd to me but nbd.

I hear you on that. I'm going to ask Marcus as UX lead to file a ticket for you against the Library Card app—that should have been done from the beginning anyway

Issue #449, opened 2020-11-30
`set-env` in Github Actions erroring out


helenvholmes:
### Description of issue:
-  Our Github Actions is requesting that we change `ACTIONS_ALLOW_UNSECURE_COMMANDS` to true now when we run our release action. Seems... superfluous?

<img width="1097" alt="Screen Shot 2020-11-30 at 1 57 27 PM" src="https://user-images.githubusercontent.com/1720093/100651977-47afd400-3314-11eb-9f81-33daf7d75338.png">

[Related failing action](https://github.com/NYPL/nypl-design-system/actions/runs/400963495)

The reset of the action is completing, however—still looks like everything is publishing to NPM as expected.

cc'ing @kristojorg, as you set these up in the first place. Maybe you understand what the fix is?

- [ ] Fix error
- [ ] Remove duplicate `npm lint` (it's already running in `npm test`)

PR #458, opened 2020-12-02
Stylelint

### LINKED ISSUE #450: Set up stylelint to work with npm test
### Description of issue:
- Our stylelint rules should run on `npm test`. The command is `npx stylelint "**/*.scss"`.

EdwinGuzman:
Fixes #450 

- Adds `stylelint` to the `npm test` script workflow.

The original issues fixed in this PR:
Screen Shot 2020-12-02 at 12 14 02 PM

The one that took the most time is this one because it's not a helpful message at all!
Screen Shot 2020-12-02 at 12 11 25 PM

It turns out the error was in the `_Template.scss` file. A css variable was being interpolated but it doesn't need to be. (Another useless message because it's not that specific line that is the issue but the next line).
Screen Shot 2020-12-02 at 12 11 31 PM

On a related not but maybe not for this PR, the Github Actions flow runs `npm lint` and then `npm test`. The test command, `npm test`, internally runs `npm lint` so it's running twice. Not sure if that's intended. I added the `stylelint` command into the `npm run test` command but it can be added into the `lint` command instead. Thoughts?

### Front End Review:
- [ ] View [the example in Storybook](https://deploy-preview-458--stoic-murdock-c7f044.netlify.app/)

### DIFF
--- a/src/components/Label/_Label.scss
+++ b/src/components/Label/_Label.scss
@@ -7,7 +7,7 @@
     justify-content: space-between;
 
     &__required-helper {
-        font-size: var(--type-size--1);
+        font-size: var(--font-size--1);
         font-weight: 300;
     }
 }
--- a/src/components/Template/_Template.scss
+++ b/src/components/Template/_Template.scss
@@ -1,5 +1,5 @@
 $sidebar-width: 25%;
-$sidebar-width-s: calc(25% - calc(4 * #{var(--space-s)}));
+$sidebar-width-s: calc(25% - calc(4 * var(--space-s)));
 
 .main {
     @include wrapper;
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "@nypl/design-system-react-components",
-    "version": "0.18.4",
+    "version": "0.18.5",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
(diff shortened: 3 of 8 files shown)


src/components/Label/_Label.scss
```diff
@@ -7,7 +7,7 @@
     justify-content: space-between;
 
     &__required-helper {
-        font-size: var(--type-size--1);
+        font-size: var(--font-size--1);
```
EdwinGuzman, line 10: This seemed to be just a variable name that wasn't updated if/when `type` was changed to `font`.

src/components/Template/_Template.scss
```diff
@@ -1,5 +1,5 @@
 $sidebar-width: 25%;
-$sidebar-width-s: calc(25% - calc(4 * #{var(--space-s)}));
+$sidebar-width-s: calc(25% - calc(4 * var(--space-s)));
```
EdwinGuzman, line 2: The css variable doesn't need to be interpolated. I originally thought it was because this is doing `25% - (4 x 16px)` and the units were not the same. It doesn't seem to care and the UI looks the same in Storybook. But can this be double-checked, please?

helenvholmes:
> On a related not but maybe not for this PR, the Github Actions flow runs `npm lint` and then `npm test`. The test command, `npm test`, internally runs `npm lint` so it's running twice. Not sure if that's intended. I added the `stylelint` command into the `npm run test` command but it can be added into the `lint` command instead. Thoughts?

Sounds redundant to me! I have a bug ticket for one of our Github Actions (#449), so maybe we can shove removing the `npm lint` command in with that work

src/components/Template/_Template.scss
helenvholmes: You got it—this is cruft from I assume Sass needing to interpolate different values but the native browser calc _not_ needing to interpolate values

helenvholmes (approved):
It's beautiful, thanks Edwin

Many will probably disagree with me on alphabetizing CSS selectors but it definitely makes it way easier to a) not accidentally repeat and override yourself and b) easier to find what you need ¯\_(ツ)_/¯

EdwinGuzman:
Thanks. I sort of tried it with component props and or HTML attributes but I can't always get used to it. I'm fine with it in scss though.

helenvholmes:
> Thanks. I sort of tried it with component props and or HTML attributes but I can't always get used to it. I'm fine with it in scss though.

That's next level. We need editors to handle this sort of thing for us 😅

EdwinGuzman:
Agreed. It's very helpful, especially when finding out new things like the `calc` function and variable interpolation. It would not have been found otherwise.

PR #463, opened 2020-12-04
Using env files for the github release action.

### LINKED ISSUE #449: `set-env` in Github Actions erroring out

EdwinGuzman:
Fixes #449 

- I think this should fix the `set-env` issue. That's deprecated and environment files is the suggested alternative. I also updated the version of the changelog reminder action.

Not sure if this will work so it needs to be tested in production, unfortunately.

### Front End Review:

### DIFF
--- a/.github/workflows/rebase.yml
+++ b/.github/workflows/rebase.yml
@@ -7,7 +7,7 @@ jobs:
         steps:
             - uses: actions/checkout@master
             - name: Changelog Reminder
-              uses: peterjgrainger/action-changelog-reminder@v1.2.0
+              uses: peterjgrainger/action-changelog-reminder@v1.3.0
               with:
                   changelog_regex: "CHANGELOG.md"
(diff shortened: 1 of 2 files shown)


helenvholmes (approved):
Looks good, let's run it and see how it goes.

PR #476, opened 2021-01-04
Accordion input styles only apply to direct child

hokei:
Fixes NO-REF

- Accordion input styles used to apply to all inputs under accordion. This made it impossible to put, for example, a list of checkboxes inside an accordion element.  The style is now correctly scoped.  

### Front End Review:
- [ ] View [the example in Storybook]() <-- There is no Storybook example, but I'm happy to put one in if it is helpful!

### DIFF
--- a/src/components/Accordion/Accordion.tsx
+++ b/src/components/Accordion/Accordion.tsx
@@ -37,7 +37,11 @@ export default function Accordion(
             id={id}
             className={bem("accordion", modifiers, blockName, [className])}
         >
-            <input id={`accordion-${inputId}`} type="checkbox" />
+            <input
+                id={`accordion-${inputId}`}
+                className={bem("input", modifiers, "accordion")}
+                type="checkbox"
+            />
             <label
                 htmlFor={`accordion-${inputId}`}
                 className={bem("label", modifiers, "accordion")}
--- a/src/components/Accordion/_Accordion.scss
+++ b/src/components/Accordion/_Accordion.scss
@@ -33,7 +33,7 @@
         }
     }
 
-    input {
+    &__input {
         opacity: 0;
         position: absolute;
         z-index: -1;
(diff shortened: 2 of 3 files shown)


src/components/Accordion/_Accordion.scss
```diff
@@ -33,7 +33,7 @@
         }
     }
 
-    input {
+    > input {
```
helenvholmes: Can we instead use a BEM block here instead of styling an `input` directly?

hokei: yes, let's do that instead!

src/components/Accordion/Accordion.tsx
```diff
@@ -37,7 +37,11 @@ export default function Accordion(
             id={id}
             className={bem("accordion", modifiers, blockName, [className])}
         >
-            <input id={`accordion-${inputId}`} type="checkbox" />
+            <input
+                id={`accordion-${inputId}`}
```
helenvholmes, line 41: This might not be a good change for this PR, which is fine—but I dislike prepending the IDs with a string, I think it's a little unclear for consumers to have the ID they pass in a prop not be the ID-hook they'll need in other places. Do you have an opinion on this @hokei?

helenvholmes (approved):
Looks good to me!

src/components/Accordion/Accordion.tsx
EdwinGuzman: This pattern is used in other components. I did get confused at first when using some components because the output ID was not what I passed in. It makes sense for the association done with the label. You're right that if this is redone, it should be a new ticket/PR.

helenvholmes: I can file tickets for this work per component if @hokei agrees it's work worth doing.

hokei: I want to first know if this is a thing people will need.  

For most users, the fact that Accordion is also a checkbox is not a thing they need to care about, and if it's possible to have this "just work" without needing to know the internals then I say we should go for it.  But if someone finds that they absolutely need to access this we should change it then.

helenvholmes: Oh yikes, I'm seeing that I've gotten turned around—this is basically a "local" id. 

What I originally was questioning was whether external, consumer ids should ever be prepended in any way. I'd still argue not. If/when we have any of those left, let's remove any prepended string(s), but now looking at this properly this isn't one of those cases. Ignore me!

PR #477, opened 2021-01-04
Added onChange prop to input

hokei:
Fixes #xxx

- Adds `onChange` prop to input

### Front End Review:


src/components/Input/Input.tsx
```diff
@@ -56,6 +59,7 @@ const Input = React.forwardRef<HTMLInputElement, InputProps>((props, ref?) => {
     let inputProps = {
         id: id ? `input-${id}` : null,
         className: bem("input", modifiers, blockName, [className]),
+        defaultValue: defaultValue,
         type: type,
         value: value,
```
EdwinGuzman: There's already a `value` attribute in this object that will go into the `input` element. I haven't tested this yet but it'll probably complain about the controlled vs uncontrolled component type. I think if you're adding a `defaultValue` then you also need an `onChange` function similar to PR #413 but I'll test it.

EdwinGuzman: Okay so I was already using `defaultValue` under the `attributes` prop and I _don't_ need an `onChange` handler because I use uncontrolled components. It'll be great to test this out using a controlled flow and form.

EdwinGuzman:
This looks good but should be tested. Maybe Bill can do a quick test in his app.

hokei:
After conversations on Slack, I'm closing this in favor of a PR making Input a controlled component

hokei:
in Checkbox, there's a controlled and uncontrolled story.  I'm not sure that stories is the best place to show this functionality, when most of the stories demonstrate visual differences.  (Link is a notable exception).  In particular, the uncontrolled checkbox story just demonstrates passing in all the props.  Without a consuming app, it's not easy to see ref functionality. 

Is this already tested OK by having a unit test that checks for a passed ref, and another unit test that checks for onChange functionality?  

I think a Controlled/Non controlled example app would be the best way to ensure all of this works, but we don't have that now ... 

Thoughts? Tagging @nypl-wluisi in case you have thoughts on this as well

nypl-wluisi:
@hokei in the case of the checkbox, it seems worth it to me to show the uncontrolled/controlled variations in Storybook. The controlled version is wired up to Storybook's "Controls", so you can toggle the checked state of the component via the `checked` prop:

https://nypl.github.io/nypl-design-system/storybook-static/?path=/story/checkbox--controlled-checkbox

The story itself might be helpful to developers using the DS, as it provides a basic example of how to wire up state to a controlled checkbox:

https://github.com/NYPL/nypl-design-system/blob/2f47bcb56865fe47006dd79de1850a4fadc26647/src/components/Checkbox/Checkbox.stories.tsx#L41

We should have a more lengthy discussion about this, but I don't know if an example app to only show basic uncontrolled/controlled components would be worth it, if you can do a basic version of that in storybook with little effort? I haven't given this much thought, but I could see a separate example app that shows an example of using Redux or MobX (or some other 3rd party state management tool) to handle state on controlled components, as those packages are probs not something you'd want to demonstrate using in storybook?

hokei:
I feel like storybook makes sense to be controlled by default, the uncontrolled components are the things that can't easily be tested in storybook, right? Like, what does it mean to pass a ref and test it there?  

I guess it makes a lot of sense to do in checkbox because of the `checked` prop, but there aren't props that are in a controlled `input` that aren't in an uncontrolled one and vice versa (except ref).  It's more that an uncontrolled `input` doesn't need an `onChange` but it's necessary in a controlled one to work

Is this a distinction that makes sense to put in stories here?

helenvholmes:
> [...] I could see a separate example app that shows an example of using Redux or MobX (or some other 3rd party state management tool) to handle state on controlled components, as those packages are probs not something you'd want to demonstrate using in storybook?

I would love an example app to demonstrate things like this, but unfortunately at the moment we don't have the resourcing for something like this.

hokei:
Let's move the Storybook discussions to https://github.com/NYPL/nypl-design-system/issues/482 so as to not block this ticket.

EdwinGuzman:
I'll write this in the #482 ticket but maybe mdx can solve this issue. If we can write longer documentation in a story maybe that'll help. You can also hook up stories with context and redux (I'm not sure how though) so there might not be a need for a separate app.

src/components/Input/Input.test.tsx
```diff
@@ -49,6 +52,12 @@ describe("Renders Input", () => {
     container.simulate("click");
     expect(clickHander.callCount).to.equal(1);
   });
+
+  it("Changing the value calls the onChange handler", () => {
+    container.find("input").simulate("change", { target: { value: "Hello" } });
+
+    expect(changeHandler.callCount).to.equal(1);
```
EdwinGuzman, line 59: Wouldn't this be 5 since each character calls the onChange function?

EdwinGuzman (approved):
Have you tested this in DRB? I'm approving because this works for me as an uncontrolled component but do want to know if it works as intended for you.

src/components/Input/Input.test.tsx
hokei: No, it's a fun Enzyme thing.  When you call `simulate("change")`it doesn't actually put the value in the input (meaning that calling `input.value.text` or the equivalent doesn't do anything.  All it does is call the changeHandler.  If you want to update the value you have to set `wrapper.props(value)` directly

Anyway, it's another reason I dislike Enzyme.

hokei:
@EdwinGuzman This works as intended in DRB. Will merge.

PR #478, opened 2021-01-04
Added ESLint and associated modifications

### LINKED ISSUE #467: ESLint
### Description of issue:
- This repo is still using TSLint, which has been deprecated since 2019.  Let's move it to use ESLint
- This would require setting it up and running the linter, it would be a PR with a lot of changes but not a lot of work. 

- All

- [Metronome](XXXXXX)

- [Figma](XXXXXX)

hokei:
Fixes #467

- Removes SectionTitle because I didn't want to do the work in getting it fixed up if it's not supposed to be in Design System. 
- Adds ESLint and more strict Prettier settings
- Please don't bother reviewing every file, take a look at `.prettierrc` and `.eslintrc` - all the other files are to make them happy (either by inserting `any` or taking the suggested refactors) 

### Front End Review:
- [x] Make sure Storybook hasn't broken
- [x] Add some documentation on how to set up your IDE for this (if it is a thing you want in your IDE) https://github.com/NYPL/nypl-design-system/wiki/Setting-up-VSCode-for-Linting


.eslintrc.js
```diff
@@ -0,0 +1,91 @@
+module.exports = {
```
hokei, line 1: I copied over the config from circulation-patron-web, which I liked.  If you have one you like, copy that over and we can discuss!

.prettierrc
```diff
@@ -1,6 +1,18 @@
...
+    "jsxBracketSameLine": false,
+    "jsxSingleQuote": false,
+    "printWidth": 80,
+    "proseWrap": "preserve",
+    "quoteProps": "as-needed",
+    "requirePragma": false,
     "semi": true,
-    "singleQuote": false
+    "singleQuote": false,
+    "tabWidth": 4,
```
hokei: I copied this over from what it was before, but how do people feel about "tabWidth": 2

helenvholmes: I prefer it, honestly

CHANGELOG.md
```diff
@@ -318,7 +328,7 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 -   Adds `disabled` prop to `Input` and `Button` components
 -   Adds `disabled` styling to `Button.Primary`
 -   Adds `className?: string;` as additional props to all components
--   Adds `attributes?: {};` to `Input`, `Button`, `Select`, `Link`, and `Image`
+-   Adds `attributes?: { [key: string]: string }; ` to `Input`, `Button`, `Select`, `Link`, and `Image`
```
helenvholmes: Why is this change being inserted?

hokei: It's an ESLint convention to not use {} as a type: 
Screen Shot 2021-01-05 at 10 06 26 AM

We could turn it off, but using `{}` also negates some of the benefits of using typescript.  It's kind of like using `any`

helenvholmes (approved):
I do not have a favorite ESLint file. This looks great to me!

I have one question about an insertion in the CHANGELOG at a weird place I don't understand. Could be a weird merge, or maybe Github is causing me to see additions in weird places... it's happened before.

> SectionTitle removed

Hoorah!... why was this even still in there? Is there any CSS we want to preserve from that lockup?

`// eslint-disable-next-line id-blacklist`

I wasn't able to find a quick-and-easy explanation for these, does the editor recommend when they should be added?

.prettierrc
EdwinGuzman: I prefer 2 as well.

package.json
```diff
@@ -16,15 +16,16 @@
...
+        "test": "npm run lint && npm run test:ci",
         "test:ci": "npm run build-ts && cp src/styles.scss lib && mocha --require lib/__tests__/setupEnzyme.js --recursive lib/**/*.test.js",
         "format": "prettier --write 'src/**/*.{ts, tsx, js,jsx,json}'",
         "pretty-quick": "pretty-quick",
         "prepare": "npm run dist",
         "build-ts": "rm -rf lib && tsc",
         "dist": "rm -rf dist && webpack --mode production --progress --display-modules --config webpack.config.js",
         "stylelint": "stylelint '**/*.scss'",
-        "lint": "tslint -c tslint.json src/__tests__/components/*.tsx src/components/**/**/*.tsx src/components/**/**/**/*.tsx src/utils/*.tsx",
+        "lintts:fix": "eslint --fix --ext=jsx,ts,tsx src",
```
EdwinGuzman: Is this a command to run manually?

src/components/Breadcrumbs/Breadcrumbs.stories.tsx
```diff
@@ -47,8 +47,8 @@ export const breadcrumbs = () => (
             : (showLongItems = false)}
         <Heading level={1}>Breadcrumbs</Heading>
         <p>
-            Breadcrumbs background color matches the secondary section color of
-            the website section they're placed into. See more about brand colors{" "}
+            {"Breadcrumbs background color matches the secondary section color of " +
+                "the website section they're placed into. See more about brand colors "}
```
EdwinGuzman: This bothers me because I won't remember to write text like this just for prettier. But if it updates this on its own, I'm fine with it.

src/components/Icons/Icon.tsx
```diff
@@ -75,13 +76,13 @@ export default function Icon(props: React.PropsWithChildren<IconProps>) {
         children,
     } = props;
 
-    let icon_base_class = "icon";
+    const icon_base_class = "icon";
```
EdwinGuzman: The change form underscore to camelcase was done in other files, why not this one?

src/components/Image/Image.tsx
```diff
@@ -53,7 +52,12 @@ export default function Image(props: React.ComponentProps<"img"> & ImageProps) {
         <>
             {imageCaption || imageCredit ? (
                 <figure className={bem("figure")}>
-                    <img {...imageProps} {...imgHTMLProps} {...attributes} />
+                    <img
+                        alt={imageProps.alt}
```
EdwinGuzman: Did an a11y plugin complain about this?

package.json
hokei: You can run it with `npm run lintts:fix` ... I have never been able to get `npm run lint --fix` to work with ESLint but maybe I've just not set it up properly?

src/components/Breadcrumbs/Breadcrumbs.stories.tsx
hokei: It ... doesn't.  
Honestly, I kinda doing all that editing too, I'm happy to disable this and change these guys back.

src/components/Icons/Icon.tsx
hokei: Oh because I put `/* eslint-disable camelcase */` at the top of this file to deal with the fact that a lot of the svg filenames were in snake case.  I wish I could just disable eslint for that block ...

src/components/Image/Image.tsx
hokei: yes

src/components/Breadcrumbs/Breadcrumbs.stories.tsx
helenvholmes: I will definitely have to google this each and every time if the editor doesn't do it automatically for us 😟

package.json
EdwinGuzman: Yeah, I meant more of a, will it be called in other scripts or other processes such as Github Actions?

src/components/Breadcrumbs/Breadcrumbs.stories.tsx
EdwinGuzman: The IDE might be able to point out the issue and also give a recommendation, but not sure if that's true for every case.

src/components/Icons/Icon.tsx
EdwinGuzman: Ah, makes sense. Maybe just update this variable name but leave the SVG filenames in snake case, since this variable is camelcase in other files.

package.json
hokei: `npm run lint` is run in `npm run test`, so it should be running in everything?  I'll double check tho.  
Is it general practice to run --fix in CI?  Or is it normal to let it fail and have the user go back and double check it?

src/components/Breadcrumbs/Breadcrumbs.stories.tsx
hokei: The IDE does give a recommendation, but it's extra work that's frustrating I don't think it helps us much in the clean code department.

hokei:
@scooooooooby 
> // eslint-disable-next-line id-blacklist


By convention, generally we don't want to name our variables any number of reserved words, like `boolean` or `number`.  The blacklisted IDs are found here:  https://github.com/NYPL/nypl-design-system/pull/478/files#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5R71

I think we should keep them, it's only `InputType.number` that really needs to be disabled - like, if you're gonna call your variable `number`, you better be really sure it's absolutely got to be called that.  The other place it's disabled for is storybook knobs, something that we should work towards getting rid of anyway.

CHANGELOG.md
helenvholmes: This is super weird, but this isn't where I intended to leave this comment... Thanks for the explanation, though! I think Github is showing changes in odd places for me.

src/components/Breadcrumbs/Breadcrumbs.stories.tsx
hokei: It's muuuch too troublesome to go back and remove these, but I have disabled it in the ESLint config file and you can see that it works by the fact that this line no longer errors.

hokei:
PS I'm glad y'all got a change to review it before because with the tab spacing change, this PR is now completely unreviewable.

package.json
EdwinGuzman: I'm not sure. We haven't had a `:fix` script in our process before. It makes sense to run that when we commit but not in CI since we want to see the changes in PRs and in CI it doesn't matter or the code shouldn't suddenly change (if it makes any "fixes" we can't see).

EdwinGuzman (approved):

helenvholmes (approved):
I don't understand Edwin's discussion about the `--fix` script, but this looks great to me. Thanks Crystal!

PR #483, opened 2021-01-06
Reworks Select to not use internal state

### LINKED ISSUE #346: Select test for console warning
### Component in question:
- `Select`, `Select.test.tsx`

### Description of issue:
- L263
- Select should have a test for the warning that is thrown when there are more than 7 options.

### LINKED ISSUE #451: Missing Select Test
### Description of issue:
- There should be a test for `Select` for the warning it throws on this line: https://github.com/NYPL/nypl-design-system/blob/development/src/components/Select/Select.test.tsx#L244

- `Select`

### LINKED ISSUE #472: 'Select' causing "Invalid hook call" error
### Description of issue:
- When the 'Select' component is used in an application, this error is raised:

> Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
> 1. You might have mismatching versions of React and the renderer (such as React DOM)
> 2. You might be breaking the Rules of Hooks
> 3. You might have more than one copy of React in the same app
> See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.

- [Select](https://nypl.github.io/nypl-design-system/storybook-static/?path=/story/select--select-with-optional-label)

hokei:
Fixes #472 
Fixes #451 
Fixes #346 

- Select is reworked to not use internal state

### Front End Review:
- [ ] View [the example in Storybook](https://deploy-preview-483--stoic-murdock-c7f044.netlify.app/?path=/story/select--select-with-optional-label)


hokei:
@EdwinGuzman how should I test that this hasn't broken the uncontrolled version?

CHANGELOG.md
```diff
@@ -8,8 +8,9 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 ## [Unreleased]
 
-### Changes
+### Changes:
```
helenvholmes: We don't use semicolons anywhere else, was this an ESLint addition?

hokei: no, it was my fat fingers.

helenvholmes (commented):
I can't actually do this through the UI, but I'm going to recommend that I step out as a reviewer on this one—because I'm not an applications developer building applications that need state management (nor have I ever built an application with statement management), I make for a poor reviewer on these controlled/uncontrolled PRs. Sorry 😥

EdwinGuzman:
> @EdwinGuzman how should I test that this hasn't broken the uncontrolled version?

I'm not sure it's being used by anyone as the uncontrolled version.

hokei:
@EdwinGuzman Ah ok!  Then I'll just make sure it's working ok the controlled version and call it good for now.

EdwinGuzman:
This renders without the hook issue so that's fixed. I haven't used this component before so I can't really test in a real environment. It'll be good to have the knobs or controls work for selecting a new option.

In Serena's original example, she had:

```
<Select isRequired={true} name="field">
  <option>All Fields</option>
</Select>
```

This does render an error and I noticed it only happens with _one_ `option` child. Maybe the `React.Children.map` should be updated or we somehow enforce that more than one option is used (which makes sense because you wouldn't use a `Select` component with only one option...).

Screen Shot 2021-01-07 at 11 29 52 AM

hokei:
Oh yeah that's a great idea.  I think it should not fail to render, but it should throw a warning -- It shouldn't fail the entire page because I can imagine people populating the select based on the outcome of an api or db call, and perhaps something failing on that end. I'll definitely look into it a bit more.

siberry:
This is working for me on `discovery-front-end`. Thanks, Crystal! I agree with you and Edwin that we should handle the case of one option better. A warning sounds like the right idea, because of the case you mentioned. Otherwise, would approve. Code looks good. Ran the build as well.

src/components/Select/Select.test.tsx
```diff
@@ -219,10 +218,11 @@ describe("Select", () => {
     expect(wrapper.find("select").props().value).to.equal("value1");
 
     wrapper.find("select").simulate("change", { target: { value: "value2" } });
-    expect(wrapper.find("select").props().value).to.equal("value2");
+    console.log("wrapper html", wrapper.html());
```
siberry: Guessing this should be removed

EdwinGuzman (approved):

Issue #240, opened 2020-06-22
CSS Scoping


kristojorg:
I have imported `styles.scss` into my app, and now all of my components are affected by the global styles within. I realize that in some sense this might be desirable, but it can also be quite disruptive. I'm curious what your thoughts are here. One option is to simply stop relying on global selectors like `a { /** styles */ }`. 

What do you all think?

helenvholmes:
I'd definitely like to namespace all of our styles. Ultimately I'd like to do two things:

1. Namespace everything with a prefix on compile, something like `nds-` in front of every className.
2. Add a hash at the end of all the styles to make conflicts on existing names even less likely, again on compile (probably with something like PostCSS).

Adding the prefix will help people understand that the className is coming from this library. The hash will help with any place where the className is just too common (I'm using `nds-` as the example proposal here, but even if we, say, did `nypl-` it's totally possible that a bunch of the apps here would have conflicts in styles for something like `.nypl-card`.)

@kristojorg thoughts? This is as far as I've gotten thinking through this.

kristojorg:
Yea I think both of those are a good idea. You can probably accomplish both at build time and just avoid setting global css. You could provide people a class they can put at the top level from where they want the NYPL styles to apply like `.nds-root`, and then put your design-system global styles in there.

kristojorg:
You might look at something like `css-modules` for style encapsulation. I think in general it would also be good to move towards styles separated per-component. So each component would have a `styles.scss` file next to it in the file tree which would get imported in the component's `index.tsx`, and with `css-modules` would be scoped automatically to only apply to that component. It would also means that consumers would only be importing/loading the css that is relevant for the components they're using.

helenvholmes:
Yeah, a lot of these global styles are holdover from another project that was using a framework that set up those styles. I also really want to move toward style encapsulation per component—the work we did in "Storybook Cleanup" to move everything into scoped folders was in support of that (before it was challenging to even know where you were supposed to put new CSS). 

Glad to have my thoughts validated by another person! I am going to have to put this in the Backlog for now but since it's going to continue affecting multiple projects using this stuff I'll see if we can't get this put into the next sprint.

helenvholmes:
@EdwinGuzman shared this plugin on #489: https://github.com/dbtedman/postcss-prefixwrap

helenvholmes:
I've been working on this ticket this week and want to give an update as it's kind of a big ticket. Here are my hopes:

[...]

PR #489, opened 2021-01-11
Type styles scoped

### LINKED ISSUE #456: Type styles not scoped
### Description of issue:
- While trying to use Design System for the new Research Catalog Account page, I was running into conflicts with `design-toolkit` for an `h1`. The styling from DS is applied using an `h1` selector. It is harder to tell what exactly the DS heading should look like.

- Example: [Heading](https://nypl.github.io/nypl-design-system/storybook-static/?path=/docs/heading--all-headings)

helenvholmes:
Fixes #456 

- Scopes any element selectors styling to the `.nypl-ds` class

### Front End Review:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 ### Breaking Changes
 
 - Removed `SectionTitle`
+- Scopes all element selects to the `.nypl-ds` class
 
 ## 0.19.1
 
--- a/src/components/Heading/_Heading.scss
+++ b/src/components/Heading/_Heading.scss
@@ -89,25 +89,27 @@
   }
 }
 
-h1 {
-  @include heading-xl;
-}
+.nypl-ds {
+  h1 {
+    @include heading-xl;
+  }
 
-h2 {
-  @include heading-large;
-}
+  h2 {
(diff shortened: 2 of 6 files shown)


siberry (approved):
LGTM. Any thoughts about the how this affects using the CDN as you brought up before?

EdwinGuzman (approved):
Looks good. You mentioned using PostCSS for post-processing. If that's used, [this plugin](https://github.com/dbtedman/postcss-prefixwrap) can be used to do what this PR is doing so you don't miss a rule doing it manually.

helenvholmes:
> LGTM. Any thoughts about the how this affects using the CDN as you brought up before?

The CDN isn't super user-friendly, and will continue to not be so. It does mean that if anyone is currently styling a particular element within their app using the DS selectors, those have now changed slightly (hence the breaking change).

Which isn't great, but the name of the game with these CSS changes.

> Looks good. You mentioned using PostCSS for post-processing. If that's used, [this plugin](https://github.com/dbtedman/postcss-prefixwrap) can be used to do what this PR is doing so you don't miss a rule doing it manually.

I have a few ideas written up for universal scoping in #240. I had envisioned prefixing all of our classes in some way rather than scoping them, but now that I'm confronted with the choice I'd like to think through the pros/cons. I'll add this info over there.

PR #492, opened 2021-01-13
Dropdown component

### LINKED ISSUE #412: Dropdown
### Description of issue:
- [Designs](https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Main?node-id=17519%3A64445)

- New component

### Metronome Documentation
- [SCC Support Components, expand `Dropdown`](https://airtable.com/shrnoS9AtWEPcFJFo)

### Additional Inclusive Design Article
[Heydon Pickering](https://inclusive-components.design/menus-menu-buttons/)

siberry:
Fixes #412

- Adds the dropdown component to Design System

### Front End Review:
- [ ] View [the example in Storybook](https://deploy-preview-492--stoic-murdock-c7f044.netlify.app/?path=/story/dropdown--example-dropdown)

### DIFF
--- a/src/components/Dropdown/Dropdown.stories.tsx
+++ b/src/components/Dropdown/Dropdown.stories.tsx
@@ -1,12 +1,7 @@
 import * as React from "react";
 import { Meta } from "@storybook/react/types-6-0";
-import range from "lodash/range";
 import Dropdown from "./Dropdown";
 import { withDesign } from "storybook-addon-designs";
-
-import List from "../List/List";
-import { ListTypes } from "../List/ListTypes";
-import Modal from "../Modal/Modal";
 
 export default {
   title: "Dropdown",
--- a/src/components/Dropdown/Dropdown.stories.tsx
+++ b/src/components/Dropdown/Dropdown.stories.tsx
@@ -4,7 +4,9 @@ import { Meta } from "@storybook/react/types-6-0";
 import Dropdown from "./Dropdown";
 import { withDesign } from "storybook-addon-designs";
 import { ExampleCard } from "../Card/Card.stories";
+import Checkbox from "./../Checkbox/Checkbox";
 
+// Set defaults for the Dropdown stories.
 export default {
   title: "Dropdown",
   component: Dropdown,
@@ -34,12 +36,14 @@ ExampleDropdown.args = {
       <fieldset className={bem("fieldset", this.modifiers, "dropdown")}>
         {options.map(option => (
           <div key={`option-${option.name}`}>
-            <input
-              type="checkbox"
-              id={option.name}
-              className={bem("checkbox", this.modifiers, "dropdown")}
+            <Checkbox
+              name={option.name}
+              checkboxId={option.name}
+              labelOptions={{
+                id: "label",
+                labelContent: <>{option.name}</>,
+              }}
             />
-            <label htmlFor={option.name}>{option.name}</label>
           </div>
         ))}
       </fieldset>
--- a/src/components/Dropdown/Dropdown.stories.tsx
+++ b/src/components/Dropdown/Dropdown.stories.tsx
@@ -4,6 +4,10 @@ import range from "lodash/range";
 import Dropdown from "./Dropdown";
 import { withDesign } from "storybook-addon-designs";
 
+import List from "../List/List";
+import { ListTypes } from "../List/ListTypes";
+import Modal from "../Modal/Modal";
+
 export default {
   title: "Dropdown",
   component: Dropdown,
@@ -16,49 +20,38 @@ const DropdownTemplate = ({ ...args }) => (
   <Dropdown {...args} inputId={`${args.inputId}`} />
 );
 
-export const SingleDropdown = DropdownTemplate.bind({});
+export const ExampleDropdown = DropdownTemplate.bind({});
+
+const options = [{ name: "Cats" }, { name: "Dogs" }, { name: "Rabbits" }];
 
 // The `args` allow these props to be updated in the UI through the
 // "Controls" tab.
-SingleDropdown.args = {
+ExampleDropdown.args = {
+  modifiers: ["fieldset"],
   dropdownLabel: "Click to expand",
   inputId: "dropdownBtn",
-  count: 1,
-  options: [{ name: "Cats" }, { name: "Dogs" }, { name: "Rabbits" }],
+  children: (
+    <form>
+      <fieldset>
+        {options.map(option => (
+          <div key={`option-${option.name}`}>
+            <input type="checkbox" name={option.name} />
+            <label htmlFor={option.name}>{option.name}</label>
+          </div>
+        ))}
+      </fieldset>
+    </form>
+  ),
 };
 
-SingleDropdown.argTypes = {
+ExampleDropdown.argTypes = {
   blockName: { table: { disable: true } },
   className: { table: { disable: true } },
   id: { table: { disable: true } },
   modifiers: { table: { disable: true } },
 };
 
-SingleDropdown.parameters = {
-  design: {
-    type: "figma",
-    url:
-      "https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Main?node-id=17519%3A64445",
-  },
-};
-
-// This is one specific story where it will render a group of Dropdowns.
-export const DropdownSet = DropdownTemplate.bind({});
-DropdownSet.args = {
-  dropdownLabel: "Dropdown Set",
-  modifiers: ["set"],
-  inputId: "dropdownBtn",
-  count: 3,
-  options: [{ name: "Cats" }, { name: "Dogs" }, { name: "Rabbits" }],
-};
-
-DropdownSet.argTypes = {
-  blockName: { table: { disable: true } },
-  className: { table: { disable: true } },
-  id: { table: { disable: true } },
-};
-
-DropdownSet.parameters = {
+ExampleDropdown.parameters = {
   design: {
     type: "figma",
     url:
--- a/src/components/Dropdown/Dropdown.stories.tsx
+++ b/src/components/Dropdown/Dropdown.stories.tsx
@@ -30,7 +30,7 @@ ExampleDropdown.args = {
       <fieldset>
         {options.map(option => (
           <div key={`option-${option.name}`}>
-            <input type="checkbox" name={option.name} />
+            <input type="checkbox" id={option.name} />
             <label htmlFor={option.name}>{option.name}</label>
           </div>
         ))}
--- a/src/components/Dropdown/Dropdown.stories.tsx
+++ b/src/components/Dropdown/Dropdown.stories.tsx
@@ -4,11 +4,6 @@ import range from "lodash/range";
 import Dropdown from "./Dropdown";
 import { withDesign } from "storybook-addon-designs";
 
-import List from "../List/List";
-import { ListTypes } from "../List/ListTypes";
-import { list as ListStory } from "../List/List.stories";
-import Link from "../Link/Link";
-
 export default {
   title: "Dropdown",
   component: Dropdown,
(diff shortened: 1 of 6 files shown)


nypl-wluisi:
@siberry This looks great thus far. A few general questions about the component.

For the Dropdown, Location Finder has "Clear" and "Save" buttons. What would be the best way to close the dropdown when either of these buttons is pressed? These buttons aren't part of the component, and would be added by the consuming app, but we'd need a way to open/close the dropdown either via state to mark the parent checkbox as not checked or something else?

Looks like this on location finder:

lf-filters

If you click outside of an open Dropdown, should it close the dropdown? Should this functionality be part of the component in the DS?

@scooooooooby @EdwinGuzman @hokei

helenvholmes:
> If you click outside of an open Dropdown, should it close the dropdown? Should this functionality be part of the component in the DS?

This makes sense to me.

src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+  component: Dropdown,
+  decorators: [withDesign],
+} as Meta;
+
+// Set up the reusable template to create multiple stories for the
+// Dropdown component.
+const DropdownTemplate = ({ ...args }) => (
+  <>
+    <Dropdown {...args} inputId={`${args.inputId}`} />
+    <ExampleCard />
```
helenvholmes, line 21: I think this can be removed, along with its import on L5.

siberry:
> > If you click outside of an open Dropdown, should it close the dropdown? Should this functionality be part of the component in the DS?
> 
> This makes sense to me.

SCC uses a package called `focus-trap` to implement this. Should we use that in Design System? @scooooooooby

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,61 @@
...
+      <input
+        id={`dropdown-${inputId}`}
+        className={bem("input", modifiers, "dropdown")}
+        type="checkbox"
+      />
+      <label
+        htmlFor={`dropdown-${inputId}`}
+        className={bem("label", modifiers, "dropdown")}
+      >
+        {dropdownLabel}
```
helenvholmes: I think this line might need its own wrapper—perhaps what's currently `&__label` becomes `&__label-group` and this is wrapped in `&__label`. That way we can add some spacing between the text and the plus icon.

src/components/Dropdown/_Dropdown.scss
```diff
@@ -0,0 +1,80 @@
...
+
+  &__label {
+    @include button--outline;
+
+    align-items: center;
+    border: 1px solid var(--ui-gray-medium);
+    cursor: pointer;
+    font-weight: 300;
+    justify-content: space-between;
+    width: 148px;
```
helenvholmes: Please remove the defined width—we'll want this to explain for whatever string we populate the dropdown with

src/components/Dropdown/_Dropdown.scss
```diff
@@ -0,0 +1,80 @@
...
+  display: inline-block;
+
+  &__content {
+    background-color: var(--ui-white);
+    border: 1px solid var(--ui-gray-medium);
+    display: none;
+    padding: var(--space-xxs) var(--space-s) var(--space-xs);
+    position: absolute;
+    transition: all 0.35s;
+    width: 259px;
```
helenvholmes: We should remove this fixed width as well

src/components/Dropdown/_Dropdown.scss
```diff
@@ -0,0 +1,80 @@
+.dropdown {
+  color: inherit;
+  display: inline-block;
+
+  &__content {
+    background-color: var(--ui-white);
```
helenvholmes, line 6: This should be inherited naturally, no? Can we remove this?

src/components/Dropdown/_Dropdown.scss
```diff
@@ -0,0 +1,80 @@
...
+    background-color: var(--ui-gray-light);
+    border-bottom: 0;
+    border-bottom-left-radius: 0;
+    border-bottom-right-radius: 0;
+    // compensate for absence of border-bottom
+    margin-bottom: 1px;
+  }
+
+  &--fieldset {
+    fieldset {
```
helenvholmes: Let's give each of these selectors BEM classnames instead of styling the elements themselves.

src/components/Dropdown/_Dropdown.scss
```diff
@@ -0,0 +1,80 @@
...
+      + .dropdown__label .icon--minus {
+        display: inline;
+      }
+
+      + .dropdown__label .icon--plus {
+        display: none;
+      }
+
+      ~ .dropdown__content {
+        background-color: #ffffff;
```
helenvholmes: Same question as L6. If it's needed, though, let's use the `var(--` associated with white

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,61 @@
...
+export interface DropdownProps {
+  /** Inner label on the button that opens the dropdown */
+  dropdownLabel?: React.ReactNode;
+  /** BlockName for use with BEM. See how to work with blockNames and BEM here: http://getbem.com/introduction/ */
+  blockName?: string;
+  /** className you can add in addition to 'input' */
+  className?: string;
+  /** ID that other components can cross reference for accessibility purposes */
+  id?: string;
+  /** dropdownLabel's input ID */
```
helenvholmes: Since we already have the public `id`, so we need the additional `inputId` to be public?

helenvholmes (changes requested):
This is looking great!

Bill's attached designs are useful for seeing how the designers want to use this component in context. A question I have is, if a dropdown is close to the right edge of the screen, how does the content area adjust to stay in the screen area without causing a horizontal scroll?

helenvholmes:
> > > If you click outside of an open Dropdown, should it close the dropdown? Should this functionality be part of the component in the DS?
> > 
> > 
> > This makes sense to me.
> 
> SCC uses a package called `focus-trap` to implement this. Should we use that in Design System? @scooooooooby

I'm agnostic on the approach, but I'd check in with @EdwinGuzman (who's implemented some `Modal` adjustments in LCA for desktop that this library pull in, I'll file a ticket) since he's done the thinking already. 

EDIT: Okay, filed #497 for this.

EdwinGuzman:
I haven't used the focus-trap library in the `Modal` component. I'm not using it as a traditional modal but more like a popup that covers the entire page just to display a loading indicator and then closes.

I'm not sure if this is still valid but it's recommended for users to stay focused into a widget that they open. So if a patron tabs inside the dropdown, they should stay there until they close the dropdown or the dropdown closes as a side effect. So maybe it is a good idea to try to implement `focus-trap`.

> If you click outside of an open Dropdown, should it close the dropdown? Should this functionality be part of the component in the DS?

Yes.

> For the Dropdown, Location Finder has "Clear" and "Save" buttons. What would be the best way to close the dropdown when either of these buttons is pressed? 

This seems more like an app pattern we should all agree on. I'm a bit torn because I think the dropdown should close when you click either "Clear" or "Save" but does the user expect that behavior to happen if there's no explicit "Close" button or label?

EdwinGuzman:
The code for the `Dropdown` component is pretty much the same as the code for the `Accordion` component. I think we briefly talked about this but one difference between the components is that the `Dropdown` component overlaps content whereas the `Accordion` pushes content down. There may be some benefit to somehow combining the components and adding props to change the styling and behavior - but I forgot where we left off on this. This PR shouldn't be thrown away or redone but if there's an iteration of this in the future, it might be good to come back to this.

nypl-wluisi:
> > For the Dropdown, Location Finder has "Clear" and "Save" buttons. What would be the best way to close the dropdown when either of these buttons is pressed?
> 
> This seems more like an app pattern we should all agree on. I'm a bit torn because I think the dropdown should close when you click either "Clear" or "Save" but does the user expect that behavior to happen if there's no explicit "Close" button or label?

Yeah we'll have to discuss that. I've been playing around with a "local" version of Dropdown component today in the location finder app. As long as the dropdown can be used as a controlled component (`checked` and `onChange` props) I'm able to close a dropdown when another opens, or close a dropdown when I click "cancel" or "save", etc. Need to experiment more tho.

nypl-wluisi:
> The code for the `Dropdown` component is pretty much the same as the code for the `Accordion` component. I think we briefly talked about this but one difference between the components is that the `Dropdown` component overlaps content whereas the `Accordion` pushes content down. There may be some benefit to somehow combining the components and adding props to change the styling and behavior - but I forgot where we left off on this. This PR shouldn't be thrown away or redone but if there's an iteration of this in the future, it might be good to come back to this.

I've been looking at how the Carbon DS does this and they have separate components for `Accordion` and `Dropdown`. They're using `downshift` for the Dropdown, but not entirely sure why:

Storybook:
https://react.carbondesignsystem.com/?path=/story/dropdown--default

Github Code:
https://github.com/carbon-design-system/carbon-components-react/blob/master/src/components/Dropdown/Dropdown.js

The accordion they have is more basic: https://github.com/carbon-design-system/carbon-components-react/blob/master/src/components/Accordion/Accordion.js

src/components/Dropdown/Dropdown.stories.tsx
siberry: @hokei suggested having some content below to show that it opens on top of it, rather than pushing content down, like an accordion. Thoughts?

src/components/Dropdown/_Dropdown.scss
siberry: It's see through for some reason. But the line you commented on below is redundant. Will remove that.

siberry (commented):
Thanks for the feedback so far! Really helpful. Resolved most of Helen's initial comments. 

@wluisi are the local changes you are making in Design System or all in your app? If the former, feel free to open a PR to merge into this one.

helenvholmes:
> The code for the `Dropdown` component is pretty much the same as the code for the `Accordion` component. I think we briefly talked about this but one difference between the components is that the `Dropdown` component overlaps content whereas the `Accordion` pushes content down. There may be some benefit to somehow combining the components and adding props to change the styling and behavior - but I forgot where we left off on this. This PR shouldn't be thrown away or redone but if there's an iteration of this in the future, it might be good to come back to this.

I can provide context here. I recommended we separate them out. A lot of the code feels shared, and feels like you could concatenate them together, but where they're used, what they look like, how they act, is different. Keeping them separate allows us to diverge these components more visually in the future. I will say that the current design iteration of `Dropdown` feels young. I'm trying to build in room for it to change (because I think it will).

> > For the Dropdown, Location Finder has "Clear" and "Save" buttons. What would be the best way to close the dropdown when either of these buttons is pressed?
> 
> This seems more like an app pattern we should all agree on. I'm a bit torn because I think the dropdown should close when you click either "Clear" or "Save" but does the user expect that behavior to happen if there's no explicit "Close" button or label?

The designers have built it this way. Kathy did user testing on this and it sounded like it was successful—I say we not overthink this and we adjust in the future if/when an application finds that something is confusing.

> I've been looking at how the Carbon DS does this and they have separate components for `Accordion` and `Dropdown`. They're using `downshift` for the Dropdown, but not entirely sure why:

This cemented my reasoning above, @EdwinGuzman—felt like the right move.

siberry:
> A question I have is, if a dropdown is close to the right edge of the screen, how does the content area adjust to stay in the screen area without causing a horizontal scroll?

I was wondering about this, too. Is this a UX question? Or should we figure out a way to have it float right if it's going to hit the right edge?

src/components/Dropdown/Dropdown.stories.tsx
helenvholmes: Oh, that makes perfect sense, I think I'm just not seeing it in the story for some reason... are you seeing it locally?

src/components/Dropdown/_Dropdown.scss
helenvholmes: When we have the content underneath we should be able to test this more carefully—it's only a line so it's not hurting anybody, it's just odd that we're seeing different things

helenvholmes:
> > A question I have is, if a dropdown is close to the right edge of the screen, how does the content area adjust to stay in the screen area without causing a horizontal scroll?
> 
> I was wondering about this, too. Is this a UX question? Or should we figure out a way to have it float right if it's going to hit the right edge?

I think we should add a story for it. Honestly the box should just move to accommodate, I'm not sure UX would tell us anything different...

siberry:
Should the Dropdown component come with the "Clear" and "Save" buttons and a list of options with checkboxes out of the box? Or should we stick with the implementation here in which it render whatever is passed as child components? @hokei suggested the latter.

helenvholmes:
> Should the Dropdown component come with the "Clear" and "Save" buttons and a list of options with checkboxes out of the box? Or should we stick with the implementation here in which it render whatever is passed as child components? @hokei suggested the latter.

Definitely what @hokei suggested. We want to keep these components super flexible. While Bill's app needs cancel/save buttons, we don't know what DRB will need, or what SCC will need. Keeping them empty with devs able to pass in what they need for their application is an important part of the design for each component.

siberry:
> I'm not sure if this is still valid but it's recommended for users to stay focused into a widget that they open. So if a patron tabs inside the dropdown, they should stay there until they close the dropdown or the dropdown closes as a side effect. So maybe it is a good idea to try to implement `focus-trap`.

The filter/dropdowns on SCC behave this way. 

Currently, our "Clear" button behaves more like a "Cancel" or "Close" button (Sean's decision). But multiple people have singled this out as a bug and think the button should remove the selected filters.

EdwinGuzman:
> This cemented my reasoning above, @EdwinGuzman—felt like the right move.

Yeah, this makes sense. Thanks. As an aside, it looks like Carbon's dropdown forces you to render a list of options to select from, so you can't add custom children nodes.

> Currently, our "Clear" button behaves more like a "Cancel" or "Close" button (Sean's decision). But multiple people have singled this out as a bug and think the button should remove the selected filters.

This now makes sense to me. When I first tried the dropdown on the SCC link you shared regarding the Accordion issue you had, I clicked on the "Clear" button but it didn't remove the selected filters. I thought it was a bug because I expected it to remove the selected filters rather than close the dropdown.

src/components/Dropdown/Dropdown.stories.tsx
helenvholmes: Nvrmnd—seeing it now!

src/components/Dropdown/Dropdown.test.tsx
```diff
@@ -0,0 +1,21 @@
...
+    );
+  });
+
+  it("Renders an input checkbox and label", () => {
+    expect(container.find("input").exists()).to.equal(true);
+    expect(container.find(".dropdown__label").text()).to.contain(
+      "Click to expand"
+    );
+  });
+});
```
helenvholmes, line 21: After the focustrap work, let's add in a test for it closing when you click off

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <input
+        id={`dropdown-${id}`}
+        className={bem("input", modifiers, "dropdown")}
+        type="checkbox"
+      />
+      <label
+        htmlFor={`dropdown-${id}`}
+        className={bem("label-group", modifiers, "dropdown")}
+      >
+        <span className={bem("label", modifiers, "dropdown")}>
```
helenvholmes, line 64: `span`s are inline, so you won't be able to apply a margin to this in the way you're expecting to be able to. My recommendation would be to make the label group a flex group with this as a div so that

1. have the label contexts float next to each other with the new div
2. apply the margin to the label text
3. align-content so that the icons center in relation to the text

I think that will get us closer to the designs we're seeing in [the specs](https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Main?node-id=17519%3A64445).

helenvholmes (changes requested):
Some suggestions for how I would make the label contents match the designs, we're getting close

bigfishdesign13:
@wluisi 
The last activity for this PR was on Feb 1.  What are the next steps?

nypl-wluisi:
> @wluisi
> The last activity for this PR was on Feb 1. What are the next steps?

@bigfishdesign13 For now this is on hold. I should be finished w/ the location finder v2 next week, and then I'll have more time to revisit this. I think we might want to reconsider the approach, I can show you some of the issues we have with making this a reusable DS component at our next 1:1 on 3/16

bigfishdesign13:
We're moving towawrd using a Multiselect component in place of the Dropdown.  This code is no longer relevant, so I am closing out this PR.

PR #493, opened 2021-01-13
First stab at css modules with PostCSS

### LINKED ISSUE #240: CSS Scoping

helenvholmes:
Fixes #240 

- [x] Scopes component CSS
- [ ] Sets up custom-media imports
- [ ] Updates `.stylelintrc.json` for new PostCSS syntax
- [ ] Removes all SCSS
- [ ] Nests CSS in the .nypl-ds class


bigfishdesign13:
We will be addressing CSS scpoing with other methods.

PR #507, opened 2021-01-29
[UX-251] Add Windows support to node build scripts

kaplanmichael:
…on both Mac and PC platform

Fixes [https://jira.nypl.org/browse/UX-251]

- This PR fixes an error that Windows/PC users will see when attempting to run `npm install` , `npm build-td` and `npm dist`. 

### Front End Review:


EdwinGuzman (approved):
Looks okay to me. Why was it installed in the repo before if it wasn't used?

nypl-wluisi (commented):
Looks good to me. Do we have a way of testing this on Windows/PC ?

kaplanmichael:
> Looks good to me. Do we have a way of testing this on Windows/PC ?

Bill, I actually do my development on a PC (rare, I know!) and the fix seems to work OK.

kaplanmichael:
> Looks okay to me. Why was it installed in the repo before if it wasn't used?

I was wondering the same thing. I initially assumed that it wasn't in the bundle, but when I created this PR and saw the diff, I was like, "Oh weird; there was an older version in there already!"

nypl-wluisi (approved):

PR #508, opened 2021-02-01
Dropdown Component Tweaks

nypl-wluisi:
Fixes #xxx

- Merges latest from `development` branch
- Allows `Dropdown` component to be used as a controlled component or uncontrolled component.

### Front End Review:


src/components/Dropdown/Dropdown.tsx
```diff
@@ -14,6 +17,12 @@ export interface DropdownProps {
   id?: string;
   /** Modifiers array for use with BEM. See how to work with modifiers and BEM here: http://getbem.com/introduction/ */
   modifiers?: string[];
+  /* For a controlled component only. If checked is not defined, the state of the uncontrolled native `input` component will be used. */
+  checked?: boolean;
```
siberry, line 21: could we call this `open`? Maybe a discussion for the main PR to get others' opinions?

siberry (approved):

src/components/Dropdown/Dropdown.tsx
nypl-wluisi: if we're going to use custom prop names that differ from the standard react "checkbox" props names then we should discuss that with the others. in general, i like using the standard prop names, but in this case, i could see using custom names, since the actual component is a "Dropdown" and not a "Checkbox", but we'd have to come up with something consistent for these props: `checked`, `defaultChecked`, and `onChange`

PR #512, opened 2021-02-10
DSD-101 mixin export

hokei:
Fixes https://jira.nypl.org/browse/DSD-101

- Compiles resources into an scss file that can be used by consuming projects

### Front End Review:
- [x] Linked to DRB and verified that it was working


gulpfile.js
```diff
@@ -0,0 +1,13 @@
+"use strict";
+
+var gulp = require("gulp");
```
kaplanmichael, line 3: Looks good.  Just wondering why you chose to write it in gulp rather than adding to the webpack config?

kaplanmichael (changes requested):

gulpfile.js
hokei: I tried looking for a way to do it through the webpack config but wasn't able to - I think webpack isn't set up to export two copies of the same file, I was only able to figure out how to compile an scss file to css or to export as is, but not to do both with the same file.  But, I'd love to know a way to do it if you've found one -- I also would prefer to do it through webpack if possible.

kaplanmichael: Ah, yeah. It's way more complicated to get webpack to do that than it should be!  I'm going to approve your approach but also create a ticket to convert this to webpack ; just in case anybody on the team has a specific knack for webpack configuration (and/or the time to work on it).

kaplanmichael (approved):

EdwinGuzman:
I reviewed this last night and tried a couple of webpack configurations but I don't think it's really possible. The most promising config looks like this in `webpack.config.js`:

```
module.exports = {
  ...,
  entry: {
    app: ["./src/index.ts"],
    resources: ["./src/resources.scss"],
  },
  output: {
    path: path.resolve(__dirname, "./dist"),
    publicPath: "/",
    filename: "[name].min.js",
    library: "DesignSystemReactComponents",
    libraryTarget: "umd",
    // This allows a server-side app to import this package and not crash.
    globalObject: "this",
  },
  plugins: [
    new MiniCssExtractPlugin({
      filename: "[name].css",
      chunkFilename: "[id].css",
    }),
  ],
 ....
```

This outputs `/dist/app.css` as before but also a new `/dist/main.css` which, ideally, contains only the mixins and variables defined in `resources.scss`. It gets close because it outputs:
```
:root{--breakpoint-small: 320px;--breakpoint-medium: 600px;--breakpoint-large: 960px;--breakpoint-xl: 1280px}.container{padding:var(--space-s)}
```
It adds the css variables but it's missing the actual mixins and scss variables.

Webpack is a JS bundler so it makes sense that it's hard to compile specific scss files using it.

I'm not sure adding a ticket for getting this to work in webpack is worth it and I'm okay with the gulp configuration but I would add a note for this in the README for documentation on how it works and how it should be imported if a dev wants to use it in their app.

PR #513, opened 2021-02-10
Link with class button has white text color

hokei:
Fixes https://jira.nypl.org/browse/DSD-102

- Links with class `button` were showing up with blue text, which made them hard to see on the blue button.  This changes the text color to be white. 

### Front End Review:


kaplanmichael (approved):
LGTM

PR #519, opened 2021-02-19
Adding padding to every story as a decorator to display the focus ring.

EdwinGuzman:
Fixes [DSD-30](https://jira.nypl.org/browse/DSD-30)

- Adds a global decorator to Storybook stories so that the focus ring displays correctly.

### Front End Review:
- [ ] View [first screenshot specified in JIRA ticket in Storybook](https://deploy-preview-519--stoic-murdock-c7f044.netlify.app/?path=/story/link--link) and [second screenshot](https://deploy-preview-519--stoic-murdock-c7f044.netlify.app/?path=/story/checkbox--uncontrolled-checkbox)

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -1,6 +1,7 @@
 // Typically, this file isn't created for a Storybook service, but we want it
 // to be able to add the custom styles.
 import "!style-loader!css-loader!sass-loader!import-glob-loader!../src/styles.scss";
+import React from "react";
 // We also want to add MDX-style documentation here:
 import { addParameters } from "@storybook/react";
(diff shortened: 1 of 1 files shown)


hokei (approved):
yay!  
Accordion focus has always not worked, right?  
Also, modal - the button doesn't seem to be focus-trapped.  But I think these are issues from before, so we should make new tickets for them if they don't already exist

EdwinGuzman:
Yeah, the `Accordion` focus didn't work before but there's a PR for that.
I'm not sure about the `Modal`. There are a few modal-related tickets on the JIRA board but I'm not sure if it has the focus-trap issue.

PR #521, opened 2021-02-22
Add `attributes` to `SearchBar` form

siberry:
Fixes JIRA ticket [DSD-112](https://jira.nypl.org/browse/DSD-112)

- The `SearchBar` accepts an `attributes` prop, but it is currently not being implemented. Seems like a small oversight. I need to use it for SCC to add `method` and `action` attributes, which are on existing SCC search bar.

### Front End Review:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 Currently, this repo is in Prerelease. When it is released, this project will adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 ========
 
+## 0.20.3
+
+### Changes
+
+- Add the `attributes` property to the `SearchBar` `<form>`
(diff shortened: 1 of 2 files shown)


EdwinGuzman:
Can you create a bug ticket in the DSD board and link it in the description? Just so your contribution is also on the JIRA board.

This is fine but `method` and `action` are actual `form` element attributes that I would prefer to be directly passed rather than in an `attributes` object. How do others feel about this? This is not a blocker for this PR.

hokei:
I agree with Edwin - `attributes` is a fallback we're using for people who need stuff quickly, we can totally keep it but if there are specific form props that you need, let's put them here explicitly

siberry:
Added `method` and `action` to `SearchBar` as well! Thanks for the suggestions. Also ticket here: https://jira.nypl.org/browse/DSD-112

kaplanmichael (approved):
Good to go! 😄

PR #522, opened 2021-02-23
Regenerating package lock

hokei:
Fixes #xxx

- Regenerates package lock because the old one seems like it was corrupted. 

### Front End Review:


EdwinGuzman (approved):

EdwinGuzman (changes requested):
Prematurely approved, let's fix the tests first.

PR #523, opened 2021-02-25
RENO-2068 - Design System: StatusBadge Component Updates

knezmilos:
Fixes #2068

- Removes statusBadgeText prop to allow text to be inside component

### Front End Review:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 Currently, this repo is in Prerelease. When it is released, this project will adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 ========
 
+## 0.20.3
+
+### Changes
+
+- Removes `statusBadgeText` prop in StatusBadge to allow text to be inside component
+
 ## 0.20.2
 
 ### Changes
--- a/src/components/StatusBadge/StatusBadge.stories.tsx
+++ b/src/components/StatusBadge/StatusBadge.stories.tsx
@@ -1,8 +1,8 @@
 import * as React from "react";
 
-import {Story} from "@storybook/react/types-6-0";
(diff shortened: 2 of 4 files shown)


EdwinGuzman:
This looks good. One minor comment that doesn't need to hold this PR, if no children are passed, the component will render an empty colored bar. That's probably not what anyone wants, right? This accidental rendering won't occured with the `if (!props.children) ...` check that was deleted.

Screen Shot 2021-02-25 at 3 25 51 PM

nypl-wluisi:
Yeah we should add this back, right? You shouldn't use this component w/o passing text inside it.

```
if (!props.children) {
  console.warn("Status Badge has no children");
}
```

@knezmilos @EdwinGuzman

knezmilos:
The intention was to always pass text to it, but the check can be added back to be sure there's always text.  Please confirm.
@EdwinGuzman

EdwinGuzman:
I think it'll be good to add back. What do you think @hokei @kaplanmichael ?

kaplanmichael:
> I think it'll be good to add back. What do you think @hokei @kaplanmichael ?

Do we have a history in the DS of doing things like this to prevent undesirable output?  If so, then I agree we should handle this in a manner consistent with what we've done elsewhere.  If not, we should probably discuss as a team to what extent we want to "save DS users from themselves"...

hokei:
In the past we've warned, and I think we can do that here - this component doesn't assume that people will be passing text from, say, an api response, but there's nothing to say it _can't_ 
In terms of "saving users from themselves" the only thing we've really settled on is that DS fail to a fallback rather than crashing the app that it's in.

EdwinGuzman:
@knezmilos let's add that snippet back in. It won't crash the app but warn devs that something is wrong.

Since this is a component update, can you also update the changelog with a new version and changes?

knezmilos:
@EdwinGuzman both changes are done

EdwinGuzman (approved):
Looks good but there are some merge conflicts.

kaplanmichael (approved):

PR #530, opened 2021-03-08
Makes list bullet grey

hokei:
Fixes #DSD-117

- Makes the default list bullet grey. 
- I did a bit of research into CSS pseudo content.  The consensus seems that it's ok in anything newer than IE11, as long as the content is decorative, which it is here. 

### Front End Review:


kaplanmichael (changes requested):
@hokei the bullets look great--the only issue is when you're looking at an _ordered_ list.  In that case, the `list-style-none` removes the numbering.  We should probably also clarify whether the numbering of the ordered list should be gray as well...

hokei:
Good catch - the mock says that the ordered list should be black, so I will go back and make those changes.

kaplanmichael (changes requested):
I'm seeing an extra type field in the controls.  Is this intentional?   
image

kaplanmichael:
@hokei Would you mind please also setting it up so that the Figma design is visible for both the ordered and unordered list stories (since the one Figma file has designs for both of those types of lists).

hokei:
@kaplanmichael I removed the ordered list story because we can get there by using the control.  Made the other change as well.

kaplanmichael (approved):
Looks great, thanks!

PR #531, opened 2021-03-08
Adds fixed-height example and defaultOpen to accordion

hokei:
Fixes #DSD-113

- Adds defaultOpen prop and fixedHeight modifier 
- This PR is to enable the implementation of filters in DRB. 
Screen Shot 2021-03-08 at 4 08 12 PM

### Front End Review:


kaplanmichael:
@hokei we might want to try using `overflow-y: auto` so that the user will only see a scrollbar when the content actually overflows: 
image

src/components/Accordion/Accordion.stories.mdx
```diff
@@ -60,3 +60,12 @@ designs. For example, when building out FAQ-like pages, pass
 <Preview withToolbar>
   <Story story={stories.AccordionAsFAQSet} />
 </Preview>
+
+## Scrollable example
+
+Use `modifiers=["fixed-height"]` when the Accordion must conform to the height of its parent.
+Any content underneath the accordion will continue to be pushed down.
+
```
kaplanmichael, line 75: Please update with the following for clarity:

"Use `modifiers=["fixed-height"]` when you'd like the accordion height to match that of its parent container.  In this example, we wrap the accordion in a container whose height is set to 300px.  If the contents within the accordion are taller than the height of the parent container, a scrollbar will appear. 

As in the other examples, any page content underneath the accordion will continue to be pushed down."

src/components/Accordion/Accordion.stories.tsx
```diff
@@ -45,3 +46,21 @@ AccordionAsFAQSet.args = {
   children: faqContent,
   count: 3,
 };
+
+// One accordion with scrollable list
+export const AccordionScroll = () => (
+  <>
```
kaplanmichael, line 52: The accordion looks a bit funky when it is focused/active: 
image

src/components/Accordion/Accordion.stories.tsx
```diff
@@ -45,3 +46,21 @@ AccordionAsFAQSet.args = {
...
+      <Accordion
+        accordionLabel={"Content"}
+        modifiers={["fixed-height"]}
+        defaultOpen={true}
+      >
+        <ListStory {...ListStory.args}></ListStory>
+      </Accordion>
+    </div>
+    <div>Example content underneath</div>
+  </>
```
kaplanmichael, line 63: Although I much prefer the use of fragment tags over meaningless DIVs, we might want to use a DIV for the time being otherwise storybook will put `<React.Fragment>` tags in your component's "View Code" area, which may be undesirable:   
image

kaplanmichael (changes requested):
Looks good, just a few quick things.

src/components/Accordion/Accordion.stories.tsx
hokei: I fixed everything else, but I'm having trouble making the code block show up in the docs at all so I'll hold off on fixing this until I can get that to show up

hokei:
@kaplanmichael The code block looks like this: 
Screen Shot 2021-03-12 at 5 13 48 PM

It still has that bind error, but I think that should be in a different ticket.  
Also, I would love to figure out a way to separate the part of the story that's about generating the component and the part of the story that's wrapper (eg: example content, etc).  I couldn't quite figure it out, but I bet it's possible.  I also think it could be in a different ticket though, so as not to block work that's dependent on this accordion functionality.

src/components/Accordion/Accordion.stories.tsx
kaplanmichael: @hokei Still seeing this issue.

kaplanmichael:
> @kaplanmichael The code block looks like this:
> Screen Shot 2021-03-12 at 5 13 48 PM
> 
> It still has that bind error, but I think that should be in a different ticket.
> Also, I would love to figure out a way to separate the part of the story that's about generating the component and the part of the story that's wrapper (eg: example content, etc). I couldn't quite figure it out, but I bet it's possible. I also think it could be in a different ticket though, so as not to block work that's dependent on this accordion functionality.

This issue has now been moved to its own user story.

kaplanmichael (approved):
LGTM

PR #532, opened 2021-03-08
Searchbar MDX Story update

EdwinGuzman:
Fixes [DSD-121](https://jira.nypl.org/browse/DSD-121)

- This is the PR for the `SearchBar` MDX story update which I partially did through the live-coding session. The first commit is from the meetings and the second commit is me finalizing and taking a closer look at why some components did not render correctly. It turns out that `SearchBar` only expects `Input` and `Button` (`Select` is optional) components so the `Label` and `HelperErrorText` components broke the styling when they were passed as children. 
- There's a small refactor in the `Autosuggest` story to allow rendering the library search example with different children. This is needed to render a single `Input` when it's used in the `SearchBar` example.

Adding everyone who was at the presentation in case you're curious about the final result.

### Front End Review:
- [ ] View [the Basic example in Storybook](https://deploy-preview-532--stoic-murdock-c7f044.netlify.app/?path=/story/searchbar--basic)
- [ ] View [the Search Autocomplete example in Storybook](https://deploy-preview-532--stoic-murdock-c7f044.netlify.app/?path=/story/searchbar--search-autocomplete)
- [ ] View [the Optional Pattern example in Storybook](https://deploy-preview-532--stoic-murdock-c7f044.netlify.app/?path=/story/searchbar--optional-pattern)


hokei:
Instead of having the optional select be a third story, is it possible to have it as one of the controls?

EdwinGuzman:
Yes, it is. I'll make that change but it won't reflect the Figma file as closely.

siberry:
I did a little googling about the `MDXCreateElement` issue. Here is a suggest approach, defining the story in Component Story Format and then importing it into the MDX: https://github.com/storybookjs/storybook/issues/11542#issuecomment-773617141

How do you feel about this?

EdwinGuzman:
That's useful, @siberry . I tried and am using that pattern for one story and I think what many are doing is separating the complicated stories with the documentation -- the CSF format goes in the `.stories.tsx` file and the more straight-forward documentation goes in the `.stories.mdx` file. It does make the mdx file more readable and I'm not opposed to this if it helps out rendering elements correctly.

src/components/SearchBar/SearchBar.tsx
```diff
@@ -30,6 +30,8 @@ export interface SearchBarProps {
   onSubmit: (event: React.FormEvent) => void;
```
kaplanmichael, line 30: While we are working on this component we might want to fix this also:  
image
This will always occur with adjacent focusable elements. My recommendation would be to add something like `z-index:999` to the focus state--we don't care so much if the focus outline overlaps the adjacent element, so long as it doesn't drop behind it.

kaplanmichael: One more thing:  Could we please also add `pointer-events: none;` to the button's disabled state styles?
image

kaplanmichael: For whatever reason, the button shrinks slightly when you are looking at the autocomplete example:  

image

kaplanmichael: According to the Figma design, the helper text should line up horizontally with the input element:
image

Whereas currently, it aligns to the left:
image

src/components/SearchBar/SearchBar.stories.mdx
```diff
@@ -0,0 +1,293 @@
...
+    args={{
+      showHelperText: true,
+      showSelect: true,
+      isErrorState: false,
+      disabledButton: false,
+    }}
+  >
+    {args => (
+      <>
+        <SearchBar {...args}>
```
kaplanmichael, line 125: As I was mentioning in the Slack channel, we might want to switch to DIV tags instead of fragment tags only until Storybook stops rendering fragments in the source code areas.  The idea is that the code there should be entire copy/paste-able.

kaplanmichael (changes requested):
Great job with the story content; it is very readable and clear!  Just a few things related to the component itself and you should be good to go.

EdwinGuzman:
My intention for this PR was to update the story but not necessarily update the component or make any big changes. So the suggestions kind of worry me because I'm not sure if they're good with everyone and if it might affect other components:

> My recommendation would be to add something like `z-index:999` to the focus state

I'll try this but check if it affects any other components in a bad way.

> Could we please also add pointer-events: none; to the button's disabled state styles?

This seems sensible.

> For whatever reason, the button shrinks slightly when you are looking at the autocomplete example:

I think this can be a separate ticket for the autocomplete component. This issue came up when trying to compose the component with other components so it's more of a bug I rather work on separately.

> According to the Figma design, the helper text should line up horizontally with the input element:

This is also an issue with how the components are composed together and probably hasn't come up before. This is a bigger change to align this correctly and I'd also like this to be a separate ticket since it can potentially make bigger changes to the layout and styling.

What do you think of breaking the last two points into separate tickets?

EdwinGuzman:
@kaplanmichael I updated the first three points. The third wasn't as big as I thought so I included it in this PR. I still left

> According to the Figma design, the helper text should line up horizontally with the input element:

for another potential ticket.

bigfishdesign13:
@EdwinGuzman 
Yes, let's create a new ticket for that last issue.  This might be a good place for me to get involved with creating tickets.

kaplanmichael:
> @EdwinGuzman
> Yes, let's create a new ticket for that last issue. This might be a good place for me to get involved with creating tickets.

That would be great, @bigfishdesign13 , thanks!  Let me know if you have any questions.

kaplanmichael (approved):
LGTM

PR #533, opened 2021-03-15
DSD-125 pagination refactor

hokei:
Fixes #533 

- Adds `getPageHref` to `Pagination`, allowing the component to generate hrefs to inject into the anchor elements, rather than always having `href="#"`

### Front End Review:


src/components/Pagination/Pagination.stories.mdx
```diff
@@ -0,0 +1,44 @@
...
+      url:
+        "https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Master?node-id=17226%3A932",
+    },
+  }}
+/>
+
+# Pagination
+
+<Description of={Pagination} />
+Pagination navigates between the pages of a multi-page application. It is commonly
```
hokei, line 27: The paragraph under "description" is not showing up in the storybook preview - anyone know why this is happening?

src/components/Pagination/Pagination.stories.mdx
```diff
@@ -0,0 +1,44 @@
...
+in search results. Use controls in the Basic example to explore this component with
+very many or very few pages.
+
+Pagination uses anchor tags because it is navigating between URLs.
+
+<Preview>
+  <Story story={stories.paginationUrlHref} />
```
hokei, line 34: This is linking back to the `canvas` page.  I could make different stories for `canvas` and `docs` to make this link to the `doc` page - would this be helpful or more confusing?

src/components/Pagination/Pagination.stories.tsx
```diff
@@ -3,15 +3,58 @@ import { Story } from "@storybook/react/types-6-0";
...
 };
 
-const PaginationTemplate: Story<PaginationProps> = args => {
-  const [page, setPage] = useState(7);
+const PaginationPageHrefTemplate: Story<PaginationProps> = args => {
+  const urlParams = new URLSearchParams(document.location.search);
+  const pageNumber = Number(urlParams.get("page"));
+  const location = window.location;
+
+  const getPageHref = page => {
```
hokei, line 29: Honestly, this is the most important part of the story re: teaching other people how to use the component.  Should I add this to a separate code block in `docs` since it doesn't show up in the "show code" block?

src/components/Pagination/Pagination.tsx
```diff
@@ -14,59 +14,69 @@ export interface PaginationProps {
...
 
-  previousPage = evt => {
-    const { currentPage } = this.props;
-    evt.preventDefault ? evt.preventDefault() : (evt.returnValue = false);
-    if (currentPage > 1) {
-      this.selectPage(evt, currentPage - 1);
-    }
-  };
+  const changeUrls =
+    typeof getPageHref !== "undefined" && typeof getPageHref === "function";
```
kaplanmichael, line 41: We might want to beef up the URL validation a bit since right now, it is possible to break the pagination if you navigate to something like this:  
`http://localhost:9001/?path=/story/pagination--pagination-url-href&page=x`
or 
`http://localhost:9001/?path=/story/pagination--pagination-url-href&page=-15`


image

src/components/Pagination/Pagination.tsx
```diff
@@ -14,59 +14,69 @@ export interface PaginationProps {
   pageCount: number;
   /** The current page selected. */
   currentPage: number;
+  /** A method that returns a Link component given the target page */
+  getPageHref?: (pageNumber: number) => string;
   /** The method to callback when an item is selected. Passes the selected page to the consuming app as an argument. */
-  onPageChange?: (selected) => void;
+  onPageChange?: (selected: number) => void;
 }
```
kaplanmichael, line 21: As a default we might want to have the component behave as though the user is on page 1 unless a valid page number is passed in the querystring.  Right now, without that parameter, no page is highlighted, but that isn't a valid use case since the user will always be _on_ a page whenever pagination controls are displayed.

src/components/Pagination/Pagination.tsx
```diff
@@ -80,17 +90,15 @@ export default class Pagination extends React.Component<PaginationProps> {
         <Link
           attributes={{ ...pageAttributes }}
           className={bem("link", modifiers, "pagination", [pageClass])}
-          href="#"
+          href={changeUrls ? getPageHref(item) : "#"}
```
kaplanmichael: Could we please try setting `pointer-events: none` on the highlighted page?  That will help to reinforce that that's the page you are on.  Right now, with it just being a different color, it can be mistaken at a glance for `a:visited {color: black;}` . Doing this will also save a trip to the router [if the user happened to click on the page he/she was already on].

kaplanmichael (changes requested):
@hokei this is looking good; just a few small things.  I didn't review the story part of this PR yet since I'll have to play around with Storybook a bit and see if I can solve those issues you've encountered.  If that turns out to be too time consuming, I can spin that out into a separate ticket as well like we did with the accordion.

src/components/Pagination/Pagination.tsx
hokei: Happy to validate getPage() to show that it should be done in the passed-in onClick.  That will prevent the page=-15 (or page=500) error - but won't prevent page="asdfaw;elkjfa;sdf" error.  But I think that gets into Storybook routing functionality, and pretty no consumer app uses Storybook as its main routing engine, so an example of this is unlikely to be helpful. 

What I could do instead is perhaps suggest in documentation that when users try to navigate to a page that doesn't exist, it returns to page 1, does that work better?

src/components/Pagination/_Pagination.scss
```diff
@@ -28,6 +28,7 @@
```
kaplanmichael, line 29: The figma file specs out a space of 8px between each number link, but you've got 8px padding on the left _and_ right of each link.  Consequently, it is causing a space of _16px_ between each number.

src/components/Pagination/Pagination.stories.tsx
```diff
@@ -3,17 +3,67 @@ import { Story } from "@storybook/react/types-6-0";
...
+const PaginationPageHrefTemplate: Story<PaginationProps> = args => {
+  const urlParams = new URLSearchParams(document.location.search);
+  const pageParam = urlParams.get("page");
+  const pageNumber =
+    pageParam &&
+    Number(pageParam) &&
+    Number(pageParam) > 0 &&
+    Number(pageParam) < args.pageCount
+      ? Number(pageParam)
+      : 1;
```
kaplanmichael: This is good though you might consider moving this into its own method--something like `const pageNumber = someFunction(pageParam);`

kaplanmichael (changes requested):
Two quick things, and then I will actually spin the docs -related issues off into a separate story so as not to hold up any project-related work.

src/components/Pagination/Pagination.stories.mdx
kaplanmichael: Per discussion, this issue will be resolved under a separate documentation-specific ticket.

src/components/Pagination/Pagination.stories.mdx
kaplanmichael: Per discussion, this issue will be resolved under a separate documentation-specific ticket.

src/components/Pagination/Pagination.stories.tsx
kaplanmichael: Per discussion, this issue will be resolved under a separate documentation-specific ticket.

kaplanmichael (approved):
LGTM

PR #534, opened 2021-03-19
DSD-57 Definition List

kaplanmichael:
Fixes https://jira.nypl.org/browse/DSD-57

- Adds a Definition List component

### Front End Review:


src/components/List/List.stories.tsx
```diff
@@ -55,6 +67,31 @@ list.parameters = {
   },
 };
 
+const DefinitionListTemplate = ({ items, ...args }) => (
+  <List type={ListTypes.Definition} {...args}>
+    {items.map((item, i) => (
+      <React.Fragment key={i}>
```
EdwinGuzman: This renders something like 

```
<React.Fragment>
  <dt></dt><dd><dd>
</React.Fragment>
<React.Fragment>
  <dt></dt><dd><dd>
</React.Fragment>
...
```

You can get away from the fragment by returning an array. So you can return `[<dt>...</dt>, <dd>...</dd>]` and it'll render:

```
  <dt></dt><dd><dd>
  <dt></dt><dd><dd>
  ...
```

src/components/List/List.stories.tsx
```diff
@@ -109,7 +146,7 @@ export const cardList = () => (
       type={select("List Type", ListTypes, ListTypes.Unordered)}
       modifiers={boolean("List Styling", false) ? null : ["no-list-styling"]}
     >
-      {cards}
+      <p>test</p> {cards}
```
EdwinGuzman: This renders fine without the test `p` tag.

src/components/List/List.tsx
```diff
@@ -20,26 +21,63 @@ export default function List(props: React.PropsWithChildren<ListProps>) {
 
   const baseClass = "list";
 
-  let listTag;
+  let listTag,
+    errorText = "";
```
EdwinGuzman: Nitpick: each variable should be declared on it's own line without the comma separator.

src/components/List/List.tsx
```diff
@@ -20,26 +21,63 @@ export default function List(props: React.PropsWithChildren<ListProps>) {
...
-      </ul>
-    );
-  } else if (type === ListTypes.Ordered) {
-    listTag = (
-      <ol id={id} className={bem(baseClass, modifiers, blockName, [className])}>
-        {children}
-      </ol>
-    );
+  switch (type) {
+    case "ol":
```
EdwinGuzman: The `ListTypes.Ordered` etc enum variables are better for these cases.

src/components/List/List.tsx
```diff
@@ -20,26 +21,63 @@ export default function List(props: React.PropsWithChildren<ListProps>) {
...
+          child.type !== "dt" &&
+          child.type !== "dd" &&
+          child.type !== React.Fragment
+        ) {
+          throw new Error(errorText);
+        }
+      });
+      listTag = (
+        <dl id={id} className={bem("definition-list", modifiers, baseClass)}>
+          <Heading id="heading1" level={3}>
```
EdwinGuzman: I'm not sure this is semantically correct but check with Zach. I'm guessing the DL in figma has this heading in place already? Do all DLs need this? Because if not, it can be annoying to always render `Details` and only `Details` (if it's not a variable) -- but not a blocker for this PR.

src/components/List/List.stories.tsx
kaplanmichael: Oops; I should have deleted that before pushing the code..  Fixed it now.

src/components/List/List.tsx
kaplanmichael: Fixed.

src/components/List/List.stories.tsx
kaplanmichael: Nice!!!  That _did_ work!  Very cool.

src/components/List/List.tsx
kaplanmichael: Why?  Both are valid.  Our linter didn't indicate this as being an issue.  Is this some kind of best practice set up for the Design System specifically?

src/components/List/List.tsx
kaplanmichael: This will become an optional prop, I just haven't finished this part yet.

src/components/List/List.tsx
EdwinGuzman: Declaring each variable on its own line is the syntax we've been following. I know the linter doesn't complain but just for consistency's sake.

kaplanmichael: OK, I updated it.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 ## Prerelease
 
+- Add Definition List component.
```
hokei: nit: in the past we've added the [type of change](https://keepachangelog.com/en/1.0.0/#how) in the changelog

src/components/List/List.stories.tsx
```diff
@@ -32,19 +32,35 @@ const itemGroups = [
...
-const ListTemplate = ({ listType, items, ...args }) => (
-  <List type={listType} {...args}>
-    {items.map((item, i) => (
-      <li key={i}>{...item}</li>
-    ))}
+const definitions = [
+  {
+    term: "Summary",
+    definition:
+      "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. ",
```
hokei, line 39: Can we get an example of a definition with `<em>` or links?  Or is that impossible in definitions?

src/components/List/List.tsx
```diff
@@ -11,35 +12,87 @@ export interface ListProps {
...
+          throw new Error(errorText);
+        }
+      });
+      listTag = (
+        <section
+          id={id}
+          className={bem("definition-list", modifiers, baseClass)}
+        >
+          {title && (
+            <Heading id="heading1" level={3}>
```
hokei: can we better ensure uniqueness of this heading id, maybe by using the id of the list?

hokei: will DL titles always be pure strings?  (eg: will they never have icons or parts of it be `<em>`)?)

src/components/List/List.stories.tsx
```diff
@@ -32,19 +32,35 @@ const itemGroups = [
   "Villagers",
 ];
 
-const ListTemplate = ({ listType, items, ...args }) => (
-  <List type={listType} {...args}>
-    {items.map((item, i) => (
-      <li key={i}>{...item}</li>
-    ))}
+const definitions = [
```
hokei, line 35: One of the most common use cases for DL is a list of links - it would be super helpful to have that as an example 
Screen Shot 2021-04-02 at 2 47 37 PM

src/components/List/_List.scss
```diff
@@ -10,6 +10,53 @@
...
+      grid-template-columns: 100%;
+      grid-template-rows: 1fr;
+      margin-bottom: 0;
+
+      dt {
+        border-top: 1px solid var(--ui-gray-light);
+        font-size: var(--font-size-0);
+        font-weight: bold;
+        padding-bottom: 0;
+        padding-right: var(--space-m);
```
hokei: Why is padding preferred over margin?  `space-stack` [uses margin](https://github.com/NYPL/nypl-design-system/blob/development/src/styles/03-space/_space-stack.scss) and a lot of things use space-stack, I'm wondering, if padding is a better thing to use, if it might be good to make similar changes to space-stack

src/components/List/_List.scss
```diff
@@ -10,6 +10,53 @@
...
+    border-bottom: 1px solid var(--ui-gray-light);
+
+    h3 {
+      border-top: 3px solid var(--section-research-primary);
+      padding-top: var(--space-xs);
+    }
+
+    dl {
+      display: grid;
+      gap: 0px;
```
hokei: I'm super new to grid but wanted to know what's the purpose of setting gap here?  It looks like it's not supported in safari but I don't see a visual difference between how it looks in safari and on ff?

hokei (changes requested):
Overall looks good! a bunch of little nitty things.  

I don't see a Figma for DL linked, but when the screen grows, is it expected that the gap between the term and definition grows? Intuitively I'd expect the definition to grow as the screen grows, but the term to stay the same size.  

On a wide screen the gap between the term and the definition can get pretty wide

Screen Shot 2021-04-02 at 2 58 36 PM

CHANGELOG.md
kaplanmichael: Done!

src/components/List/List.tsx
kaplanmichael: > can we better ensure uniqueness of this heading id, maybe by using the id of the list?

Good catch!  Fixed now.

kaplanmichael: > will DL titles always be pure strings? (eg: will they never have icons or parts of it be `<em>`)?)

Yes they will only be strings.  If there becomes a need to support other markup, that will need to be specified from the UX/Design team, added to the Figma, etc...

src/components/List/_List.scss
kaplanmichael: It is supported--I think you might have been looking at the _flexbox_ gap property.  You are right though that it appears not to be necessary here since 0px is the default setting for the grid gap property.  I went ahead and removed it.

src/components/List/_List.scss
kaplanmichael: Tbh, I wasn't aware of the space-stack mixin until you mentioned it!  In this case, I used padding because I needed the space to appear on the inside of the box model otherwise the gray border would have separated.  On the bottom, I used padding for consistency, but if it is preferable to use the mixin, I could switch over to that.  I think @bigfishdesign13 has some thoughts on the mixins that we can talk about in an upcoming meeting as well.

src/components/List/List.stories.tsx
kaplanmichael: Sure, no problem. I will add.

kaplanmichael:
> Overall looks good! a bunch of little nitty things.
> 
> I don't see a Figma for DL linked, but when the screen grows, is it expected that the gap between the term and definition grows? Intuitively I'd expect the definition to grow as the screen grows, but the term to stay the same size.
> 
> On a wide screen the gap between the term and the definition can get pretty wide
> 
> Screen Shot 2021-04-02 at 2 58 36 PM

@bigfishdesign13 Could we get a UX check on how they want to handle this?  Right now it is fluid but see Crystal's point above.

src/components/List/_List.scss
bigfishdesign13: Padding vs. Margin depends on the situation.  As Michael noted, for his case padding was necessary to accomplish the look that he needed.  But maybe we need to expand the spacing options because of problems like what Michael was facing.

src/components/List/List.stories.tsx
```diff
@@ -113,3 +159,53 @@ export const cardList = () => (
     </List>
   </>
 );
+/* eslint-disable jsx-a11y/anchor-is-valid */
+export const listOfLinks = () => (
+  <List title="Details" type={ListTypes.Definition}>
+    <dt>Authors</dt>
+    <dd>
+      <a href="#">Chirwa, Ephraim Wadonda, author</a>
+      <br />
```
hokei, line 184: does this mean that the correct way to implement a definition list within design system is to use brs between every link?  Is it against standards to do 
```
  <dt>Authors</dt>
    <dd>
      <li><a href="#">Chirwa, Ephraim Wadonda, author</a></li>
     <li><a href="#">Dorward, Andrew, author</a></li>
```

Or, is it required that `dl` should only be used for definitions the way that tables shouldn't ever be used for anything except tabular data, and therefore, is this not the correct component for this use case?

src/components/List/List.stories.tsx
```diff
@@ -113,3 +159,53 @@ export const cardList = () => (
...
+      <a href="#">Agricultural subsidies.</a>
+      <br />
+      <a href="#">BUSINESS &amp; ECONOMICS -- Industries</a>
+      <br />
+    </dd>
+  </List>
+);
+/* eslint-enable jsx-a11y/anchor-is-valid */
+
+listOfLinks.storyName = "List of Links";
```
hokei: Nit: I'd change this to "definition list of links"

src/components/List/List.stories.tsx
EdwinGuzman: The `<br />` doesn't seem right and maybe can be fixed with CSS. Your example is somewhat correct, it just needs an `<ul>`:

```
<dt>Authors</dt>
<dd>
  <u>
    <li><a href="#">Chirwa, Ephraim Wadonda, author</a></li>
    <li><a href="#">Dorward, Andrew, author</a></li>
  </ul>
</dd>
```

kaplanmichael: You can basically put any markup you want inside the DD.  I wasn't advocating BRs by including them in the example.   The UL suggestion is a good one, or you could also use CSS to set the links to `display: block` and use margin to hone the spacing between them.

src/components/List/List.stories.tsx
kaplanmichael: Done!

kaplanmichael:
> Overall looks good! a bunch of little nitty things.
> 
> I don't see a Figma for DL linked, but when the screen grows, is it expected that the gap between the term and definition grows? Intuitively I'd expect the definition to grow as the screen grows, but the term to stay the same size.
> 
> On a wide screen the gap between the term and the definition can get pretty wide
> 
> Screen Shot 2021-04-02 at 2 58 36 PM

Fixed!

src/components/List/_List.scss
```diff
@@ -10,6 +10,52 @@
...
+        padding-top: var(--space-xs);
+
+        @include breakpoint($breakpoint-medium) {
+          border-top: 1px solid var(--ui-gray-light);
+        }
+      }
+
+      @include breakpoint($breakpoint-medium) {
+        grid-template-columns: max(225px) 1fr;
+        margin-bottom: var(--space-s);
```
bigfishdesign13: Looking at the Figma, it would be good to include something to set a min height for the rows.  I believe `grid-auto-rows` could be used for this.  Adding this, however, may also impact spacing that has already been set in other margin/padding vlaues.

bigfishdesign13:
@hokei I am picking up this ticket to get it completed.  If you have some time, could you take care of the final review?  Thanks.

src/components/List/_List.scss
bigfishdesign13: Not going to do this now.  Will address later, if need be.

hokei (approved):

PR #537, opened 2021-03-24
DSD-156 Template Flex-wrap and Reflow match

hokei:
Fixes #DSD-156

- Template used to have an awkward part where the components flexed downward but the wide-width styles were not applying.  This has been fixed. 
- A lot of the math in the Template component seems to come from the fact that we are using Flexbox when we really should be using Grid. 

### Front End Review:

### DIFF
--- a/src/components/Template/_Template.scss
+++ b/src/components/Template/_Template.scss
@@ -1,6 +1,10 @@
 $sidebar-width: 25%;
 $sidebar-width-s: calc(25% - calc(4 * var(--space-s)));
 
+// Adding an extra margin to the breakpoints allows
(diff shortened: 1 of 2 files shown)


src/components/Template/_Template.scss
```diff
@@ -1,6 +1,10 @@
 $sidebar-width: 25%;
 $sidebar-width-s: calc(25% - calc(4 * var(--space-s)));
```
kaplanmichael: I can't even...!

kaplanmichael (approved):

bigfishdesign13:
The CSS changes look good.

PR #538, opened 2021-03-25
[DSD-161] Modify lint and test scripts so that function cross-platform.

kaplanmichael:
Fixes https://jira.nypl.org/browse/DSD-161

### Front End Review:
- [ ] At your console prompt, try running both `npm lint` and `npm test`.  They should both complete without any _script-specific_ errors (they might encounter legitimate lint issues and/or failing code tests which is fine--we just want to make sure that each of the commands runs properly)


EdwinGuzman (approved):
Works well for me locally.

Typically, dev changes that are not component changes don't need to be written in the `CHANGELOG` file, though this rule may have been updated.

kaplanmichael:
> Works well for me locally.
> 
> Typically, dev changes that are not component changes don't need to be written in the `CHANGELOG` file, though this rule may have been updated.

Thanks for the tip. I'll avoid that going forward.

PR #541, opened 2021-03-26
DSD-165 added new icons and enabled color selection

bigfishdesign13:
Fixes DSD-165
https://jira.nypl.org/browse/DSD-165

- added three new icons: Error (solid and outlined), Speaker Notes
- enabled icon color selection

### Front End Review:


src/components/Icons/Icon.tsx
```diff
@@ -69,6 +82,7 @@ export default function Icon(props: React.PropsWithChildren<IconProps>) {
     className,
     desc,
     iconRotation,
+    color = "ui-black",
```
EdwinGuzman: Use `IconColors.ui_black` instead.

EdwinGuzman: Oiy, sorry. I got an email you added me as a reviewer but now I'm actually not seeing that message on this page. I'll refrain from commenting!

bigfishdesign13: @EdwinGuzman 
I didn't explicitly add anyone as a reviewer, so I'm not sure how that happened.  Michael was going to take a look at this one, but, by all means, if you have time, go for it.

src/components/Icons/Icon.tsx
```diff
@@ -82,6 +96,10 @@ export default function Icon(props: React.PropsWithChildren<IconProps>) {
     modifiers.push(iconRotation);
   }
 
+  if (color) {
+    modifiers.push(color);
+  }
+
   const iconProps = {
```
kaplanmichael, line 70: One thing you have to watch out for here is that every time it re-renders, it's going to push that value to the modifiers string and eventually you will end up with something like this:
image

kaplanmichael (changes requested):
This bug is blocking me from testing other aspects so let's see about fixing this first, and I can review again once that's done.

src/components/Icons/Icon.tsx
bigfishdesign13: I made some updates to the stories.tsx file to address this issue.

src/components/Icons/Icon.stories.tsx
```diff
@@ -79,14 +120,16 @@ Logos.parameters = {
...
 
-export const customIcon = () => (
-  <>
+// CUSTOM ICONS *****************************
+
+const customIconTemplate = args => (
+  <div>
     <p>
       {
         "If you'd like to pass a custom svg to your application, you can pass it as children to Icon:"
```
kaplanmichael: This text shouldn't be part of the custom icon's code, otherwise it shows up when the user clicks "Show Code":  
image

src/components/Icons/Icon.tsx
```diff
@@ -44,11 +56,15 @@ export interface IconProps {
   /** className that appears in addition to "icon" */
   className?: string;
   /** Decorative icons are skipped by screenreaders */
-  decorative: boolean;
+  decorative?: boolean;
   /** Desc prop added to the <svg> element */
```
kaplanmichael: place "<svg>" inside back ticks so that it doesn't  break the args table:  
image

src/components/Icons/Icon.tsx
```diff
@@ -44,11 +56,15 @@ export interface IconProps {
...
   desc?: boolean;
```
kaplanmichael: Could we add a bit more detail about what the _desc_ prop actually does?

src/components/Icons/Icon.tsx
```diff
@@ -63,27 +79,39 @@ export interface IconProps {
...
+
+  if (size) {
+    iconModifiers.push(size);
   }
 
   const iconProps = {
-    className: bem(baseClass, modifiers, blockName, [className]),
+    className: bem(baseClass, iconModifiers, blockName, [className]),
     role: decorative ? "img" : role,
     "aria-hidden": decorative,
```
kaplanmichael, line 75: Would be great to get a bit more detail in the args table about what the _role_ and _title_ props do.  For example, with role we can say something like "Applies an `aria-role` parameter to the icon for accessibility purposes." Would be great to do similar for the _title_ prop.

src/components/Icons/IconTypes.tsx
```diff
@@ -1,9 +1,9 @@
...
 export enum IconRotationTypes {
-  rotate180 = "rotate-180",
-  rotate270 = "rotate-270",
   rotate0 = "rotate-0",
   rotate90 = "rotate-90",
+  rotate180 = "rotate-180",
+  rotate270 = "rotate-270",
 }
 
 export enum IconNames {
```
kaplanmichael, line 9: Just wondering if this enum is necessary given that we already have the allSvgs object declared in icon.tsx.  It might be necessary--it just flagged my attention so I though it worth mentioning.

kaplanmichael (changes requested):
Looking good!  Just a few things:

src/components/Icons/Icon.stories.tsx
bigfishdesign13: Agreed.  And fixed.  I switched over to `.mdx` and I moved this info into the description for this story.

src/components/Icons/Icon.tsx
bigfishdesign13: Good catch.  Fixed.

src/components/Icons/Icon.tsx
bigfishdesign13: As I look at this more closely, this prop is not being used correctly, so this portion will need to be refactored.

src/components/Icons/Icon.tsx
bigfishdesign13: Similar to `desc`, this prop is being used different from how I would think it should be.  I will work on this as well.

src/components/Icons/IconTypes.tsx
bigfishdesign13: @kaplanmichael 
I think the names were used in one of the stories - a story that I am now omitting in the `.mdx` version of the stories.  I am omitting the story becaseu I cann't figure out a way to redo what was being done in the `.tsx` version.  Do you think it is necessary to include a story that shows all available icons at one time?

src/components/Icons/Icon.tsx
kaplanmichael: Done!

src/__tests__/setupEnzyme.ts
```diff
@@ -23,7 +23,9 @@ global.navigator = {
 const noop = () => {};
 
 // Ignore SVG
-require.extensions[".svg"] = noop;
+require.extensions[".svg"] = obj => {
```
EdwinGuzman, line 26: This was what was needed to make tests _run_ first before they could pass.

EdwinGuzman:
@bigfishdesign13 The tests should run now.

In the last commit I made these changes:

- Updated mocha test runner `.extensions` so it can ignore .svg files.
- Cleaned up and returned the actual SVG react component from the SVG file import.
- Updated how export and import was done in `IconSvgs.tsx`. It's not a big deal because the previous one would work but wanted to make it clearer what was being exported. Now, the default from each svg file is being exported rather than the "ReactComponent" or the "default" export. Export/import can be tricky.
- Updated the `tsconfig.json` file because there were unnecessary files and directories in the `include` attribute. (I initially thought this was what was the problem for including the svg files).
- Updated the webpack config but I should revisit that later.

Now I need to know what's left to do because

- the ticket says to include the `Alert` and `Speaker Notes` svgs but I only see Speaker Notes here. In the changelog I see there it explicitly says it also includes "Error solid" and "Error outlined", but there are so many new SVGs that were added that are not in the changelog. I'm assuming the changelog just hasn't been updated. I can update it but just want to confirm that all the svgs in the "icons/svg/new" directory should be included. If so, they should just go in "icons/svg" without the "/new" subfolder (unless if there's something I don't know and we need this).
- The `Icon.stories.mdx` file should be updated and I can do that. I think it'll be beneficial to include all svgs as a showcase along with mixes of new props in place.


Is there anything else I'm missing before I continue?

EdwinGuzman:
@bigfishdesign13  Alright, updated the mdx storybook page and fixed the directory of the new svg files. Almost all logo svgs are way too big and there's no `size` prop to display it correctly. This screenshots shows how they're cut off:
Screen Shot 2021-06-30 at 5 22 39 PM

I noted that this is a work in progress. Either we remove these svgs or we merge this in and make a ticket to fix this issue later. What do you think? Easier to see in the deploy - https://deploy-preview-541--stoic-murdock-c7f044.netlify.app/

src/components/Icons/_Icons.scss
```diff
@@ -28,24 +30,84 @@
     margin-left: var(--space-xxs);
   }
 
+  // SIZE
```
bigfishdesign13, line 33: This just came in from UX.  The values in the size classes need to be changed to the following:
- small: `0.875rem` (14px)
- medium: `1.125rem` (18px)
- large: `--space-m` 1.5rem (24 px)
- xlarge: `--space-l` 2rem 
- xxlarge: `--space-xl` 3rem 
- xxxlarge: `--space-xxl` 4rem 

Continue ot use the CSS vars as much as possible and just use the straight `rem` values for `small` and `medium`.

EdwinGuzman: Gotch. Any explanation for this update?

bigfishdesign13: I am writing an Iconography section for the Style Guide and I am working with Kathy to resolve descrepencies between the Figma files and how the component was executed.

EdwinGuzman:
I updated the spacing variables. @hokei @wluisi can you both take a look at this PR? It's a blocker for other work.

hokei:
~~Just for prettiness, can there be a bit of a padding around the icons in the storybook?~~

Screen Shot 2021-07-06 at 8 55 54 AM

src/components/Icons/Icon.stories.mdx
```diff
@@ -0,0 +1,276 @@
...
+      },
+    }}
+  >
+    {args => <Icon {...args} />}
+  </Story>
+</Preview>
+
+<ArgsTable story="Icon Props Control" />
+
+export const iconRow = (icon, opts = {}) => {
```
hokei, line 88: ooooh! I totally thought the icons were literally black boxes haha 
Can we use a not-black colour (ui-gray-xdark) attacht he background color to the `<li>` element rather than the icon? I think this might give a better sense of what's going on.  

Screen Shot 2021-07-06 at 9 04 09 AM

Not really functionality related though, might belong on whatever ticket that's for fixing the logo size display issue Edwin mentioned above.

hokei (approved):
Assuming there's another ticket to fix the logos-are-too-big-no-`size`-prop issue, this looks good to go

src/components/Icons/Icon.stories.mdx
EdwinGuzman: Yeah, I pointed that above that all the library-svg logos are really big in size. We'll have to create a ticket to discuss including them and then fixing the sizes because right now the largest css modifier is not big enough to display the whole svg.

PR #542, opened 2021-04-05
DSD-174 updated Brand and Utility colors to match Figma main file

bigfishdesign13:
Fixes #DSD-94 and #DSD-174

- updated Brand colors based on DSD-174
- updated Utility colors based on DSD-94

### Front End Review:


CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 ## Prerelease
 
+- Changes `Brand` and `Utility` colors to match Figma main file
```
hokei: The removal of `ui-success` and `ui-error` should be highlighted under a "Breaking Changes" section so that everyone who's upgrading knows to make those changes.

bigfishdesign13: @hokei 
`ui-error` is still in there.  It just got bumped down in the list of colors.

As for `ui-success`, I searched the DS codebase and noticed that var is not used in any other files.  That being the case, should the removal of that CSS var still be listed under Breaking Changes?

src/styles/01-colors/_colors-brand.css
```diff
@@ -1,23 +1,27 @@
...
   /* Books and More is used for the Catalog, Staff Picks, Recommendations, New Arrivals, and E-Book Central. */
-  --section-books-and-more-primary: #d0343a;
-  --section-books-and-more-secondary: #bb1d12;
+  --section-books-and-more-primary: var(--brand-primary);
+  --section-books-and-more-secondary: var(--brand-secondary);
 
-  /* Research is used for the Research Catalog and SCC. */
-  --section-research-primary: #377f8b;
-  --section-research-secondary: #02626a;
+  /* Research is used for the Research Catalog and SCC. - UPDATED TO MATCH FIGMA */
```
EdwinGuzman: Can you add a link here to the figma colors documentation? At the top of the file may be best.

bigfishdesign13: @EdwinGuzman 
This has been addressed.  Thanks for the good suggestion.  I added reference links to the top of both CSS files.

src/styles/01-colors/_colors-utility.css
```diff
@@ -1,3 +1,10 @@
+/**
+ * Utility Colors
+ *
+ * Color Reference in Figma Main File
+ * https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Main?node-id=14989%3A37
+ *
+ */
```
kaplanmichael: Great idea to have Figma link here!

src/styles/01-colors/_colors-utility.css
```diff
@@ -13,9 +20,17 @@
...
+  --ui-success-secondary: #095212;
+  --ui-warning-primary: #f0974e;
+  --ui-warning-secondary: #ec7b1f;
+  --ui-status-primary: #f9e08e;
+  --ui-status-secondary: #fbe7e1;
+  --ui-error: #97272c;
+  --ui-disabled-primary: #e0e0e0;
+  --ui-disabled-secondary: #fafafa;
+  --ui-test: #ff6347;
 }
```
kaplanmichael: Consider alphabetizing the variables within each category  --  we actually have a linter setting that recommends doing that.  Let me know if you'd like to set that up in your IDE if you haven't already.

src/styles/01-colors/_colors-brand.css
```diff
@@ -1,23 +1,33 @@
...
-  /** What's On is used for Exhibitions & Events. */
-  --section-whats-on-primary: #8a1a9c;
-  --section-whats-on-secondary: #7c1097;
+  /* Brand fills are the NYPL primary and secondary colors */
+  --brand-primary: #c60917;
+  --brand-secondary: #760000;
+
+  /* What's On is used for Exhibitions & Events. */
+  --section-whats-on-primary: #242424;
+  --section-whats-on-secondary: #000000;
```
kaplanmichael: One thing we should think about is whether we want to try giving the hex values some color name (example "fucshia") which describes the color rather than something like "--section-whats-on-primary".  This would be helpful when designing components based off of Figma designs that use certain colors but don't necessarily belong to, for example, the "What's On" section.  This approach would be similar to what you are already doing for `--brand-primary ` and `--brand secondary`.  For example:  ```--section-whats-on-primary: var(--cl-fucshia);```

CHANGELOG.md
kaplanmichael: Agree with @hokei here.

kaplanmichael (changes requested):

src/styles/01-colors/_colors-utility.css
bigfishdesign13: This is the order that the colors appear in the Figma.  And I honestly don't know which way to go with this.  My visual side, in order to make a quick reference, wants the order to match the order in the Figma file.  But my logic side understands the benefits of alphabetizing.

bigfishdesign13: @kaplanmichael 
OK, I alphebetized the vars where it made sense.  I did not touch the ordering of the grayscale vars, as they are curerently ordered based on the color values from dark to light.

src/styles/01-colors/_colors-utility.css
bigfishdesign13: Edwin's gets the credit for that.

src/styles/01-colors/_colors-brand.css
bigfishdesign13: @kaplanmichael 
Ideally, the names of the CSS vars match the color names used in the Figma file.  Which I believe they do, at this point.  I am hoping to start talking more seriously about global NYPL design tokens soon and this discussion will definitley be part of that.

CHANGELOG.md
bigfishdesign13: OK, I made some changes to the wording in the CHNAGELOG.

hokei: >  I searched the DS codebase and noticed that var is not used in any other files

Just to clarify in case people are searching in the future -- consuming apps can use these CSS variables directly, so any changes here should come with instructions on what to use instead.

hokei (approved):

CHANGELOG.md
bigfishdesign13: @hokei 
Ah, yes, in that light, this makes much more sense.  Thanks for the catch andn suggestion.

bigfishdesign13:
@kaplanmichael 
I think you need to mark your change requests as completed before this one can move forward.

kaplanmichael (approved):
LGTM

PR #543, opened 2021-04-06
Dsd 201 radio buttons

kaplanmichael:
Fixes https://jira.nypl.org/browse/DSD-201

- Moves radio button from Input component into its own new "Radio" component.
- Updates automated tests

### Front End Review:


src/components/Input/Input.stories.tsx
```diff
@@ -172,52 +172,3 @@ export const inputAttributes = () => (
     </HelperErrorText>
   </>
 );
-
-export const RadioInput = () => {
-  const [stateValue, setStateValue] = useState("");
```
EdwinGuzman, line 177: Minor: since this was removed `useState` can be removed from the import. Github gives a nice message about it:

Screen Shot 2021-04-09 at 11 41 09 AM

src/components/Radio/Radio.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+    container.simulate("click");
+    expect(clickHander.callCount).to.equal(1);
+  });
+
+  it("Changing the value calls the onChange handler", () => {
+    container.find("input").simulate("change", { target: { value: "Hello" } });
+    expect(changeHandler.callCount).to.equal(1);
+  });
+
+  it("Renders with label", () => {
```
EdwinGuzman, line 57: Can you add to this test that the `htmlFor` (`htmlfor` in case the react version doesn't get picked up by Enzyme) is the same as the `id` that is passed as a prop to `Radio`? Having a label is great, but unless it's associated to the `input` element then it's not useful.

src/components/Radio/Radio.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+  it("Renders HTML attributes passed through the `attributes` prop", () => {
+    const onChangeSpy = stub();
+    const onBlurSpy = stub();
+    container = Enzyme.mount(
+      <Radio
+        id="inputID-attributes"
+        attributes={{
+          onChange: onChangeSpy,
+          onBlur: onBlurSpy,
+          maxLength: 10,
```
EdwinGuzman: This is a bit weird to pass because it doesn't do anything to a radio type input.

src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,66 @@
+import * as React from "react";
+import bem from "../../utils/bem";
+import Label from "../Label/Label";
+
+export interface InputProps {
+  /** An optional text label for the radio button */
+  labelText?: string;
+  /** Additional attributes to pass to the <input> tag */
```
EdwinGuzman: Wrap `<input>` in backticks. It renders weirdly without in storybook:

Screen Shot 2021-04-09 at 12 06 15 PM

src/components/Input/Input.stories.tsx
EdwinGuzman: Looks like the Github Action test failed because of this linting issue as well.

src/components/Radio/Radio.stories.mdx
```diff
@@ -0,0 +1,43 @@
...
+  ArgsTable,
+  Preview,
+  Description,
+} from "@storybook/addon-docs/blocks";
+import { withDesign } from "storybook-addon-designs";
+
+import Radio from "./Radio";
+
+<Meta
+  title="Form Elements/Radio"
```
EdwinGuzman, line 14: Not sure if this was intentional or settled in a different conversation, but this makes this component render in a different subtitle than the rest of the components. This is a nice change!

Screen Shot 2021-04-09 at 12 08 42 PM

EdwinGuzman:
Looking good so far. There's an accessibility error for the main story:
Screen Shot 2021-04-09 at 12 10 09 PM

The `Radio` story is just missing an explicit `id` prop, similar to how you added `labelText: "Test Label"`. This is an easy miss and eventually, I think this could use a larger discussion -- the `id` prop is optional but breaks this accessibility check. Perhaps other components behave in a similar way and shouldn't  but this isn't necessary for this PR. Just adding the `id` to begin with and making a note of it in the story description is enough for now.

Are you going to add other stories for the error, disabled, checked, etc states or is that a separate PR?

src/components/Radio/Radio.stories.mdx
bigfishdesign13: @EdwinGuzman 
This was intentional.  We will be moving the components into logic groups within Storybook and Form Elements is the first group we are addressing.

src/components/Input/Input.stories.tsx
kaplanmichael: Fixed!  Thanks for catching that!

src/components/Radio/Radio.tsx
kaplanmichael: Fixed!

src/components/Radio/Radio.stories.mdx
kaplanmichael: @EdwinGuzman Thanks!  Yeah, this grew out of a discussion with @bigfishdesign13 about how to better organize the various components.  I'm glad you like it; that means we're on the right track!

src/components/Radio/Radio.test.tsx
kaplanmichael: Fixed!

kaplanmichael:
> Looking good so far. There's an accessibility error for the main story:
> Screen Shot 2021-04-09 at 12 10 09 PM
> 
> The `Radio` story is just missing an explicit `id` prop, similar to how you added `labelText: "Test Label"`. This is an easy miss and eventually, I think this could use a larger discussion -- the `id` prop is optional but breaks this accessibility check. Perhaps other components behave in a similar way and shouldn't but this isn't necessary for this PR. Just adding the `id` to begin with and making a note of it in the story description is enough for now.
> 
> Are you going to add other stories for the error, disabled, checked, etc states or is that a separate PR?

@bigfishdesign13 If its absence causes an accessibility violation, should the `id` prop instead be mandatory?

src/components/Radio/Radio.test.tsx
kaplanmichael: Done!

src/components/Radio/Radio.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+  });
+
+  it("Changing the value calls the onChange handler", () => {
+    container.find("input").simulate("change", { target: { value: "Hello" } });
+    expect(changeHandler.callCount).to.equal(1);
+  });
+
+  it("Renders with label", () => {
+    container = Enzyme.mount(
+      <>
```
EdwinGuzman: Nit: you don't need the fragment element wrapper.

src/components/Radio/Radio.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+        id="inputID"
+        attributes={{ onClick: clickHander }}
+        onChange={changeHandler}
+        labelText="Test Label"
+        showLabel={false}
+      ></Radio>
+    );
+  });
+
+  it("Renders with appropriate 'aria-label' attribute and value when 'showLabel' prop is set to false", () => {
```
EdwinGuzman, line 30: 👍 
Can you add a check to an existing test to show that when `showLabel` is set to `true`, that the `aria-label` attribute does not get rendered?

EdwinGuzman:
The update looks good! Just two minor comments.

Re: making the `id` prop mandatory, that's probably the safest way to go so developers always set one. Nbd but the `Radio` components in the `<Canvas>` should also have an id prop.

src/components/Radio/Radio.test.tsx
kaplanmichael: Good catch; fixed!

src/components/Radio/Radio.test.tsx
kaplanmichael: Done!

EdwinGuzman (approved):
Looks good to me but wait for one more review.

src/components/Radio/Radio.test.tsx
kaplanmichael: Fixed!

EdwinGuzman (approved):
Approving latest update.

src/components/Radio/Radio.stories.mdx
bigfishdesign13: @kaplanmichael You'll eventually want to implement the auto-categorization method from DSD-220.

src/components/Radio/_Radio.scss
```diff
@@ -0,0 +1,67 @@
+.input {
```
bigfishdesign13, line 1: We never talked directly about the philosophy behind the root class name for this component.  I feel this class name should be unique to this componet and style inheritance should come from the base styles.  Not from other components.  We should have a call about this.

src/components/Input/Input.test.tsx
```diff
@@ -215,36 +215,6 @@ describe("Forwarding refs", () => {
   });
 });
 
-describe("Radio buttons", () => {
```
bigfishdesign13, line 218: The Input component will eventually be deprecated, so I am not sure these changes are necessary.

bigfishdesign13 (dismissed):

src/components/Radio/Radio.test.tsx
kaplanmichael: Done!

src/components/Radio/_Radio.scss
kaplanmichael: Per our conversation on Monday, 4/19, we will move these reorganization changes to a new JIRA story.

src/components/Input/Input.test.tsx
kaplanmichael: @bigfishdesign13 You're right. I did this so that I could run the test suite locally and not have it error.  Since the changes were already done a while back, I would recommend keeping as-is for now.  Thoughts?

PR #544, opened 2021-04-07
DSD-189 initial setup for text input

bigfishdesign13:
Fixes #DSD-189
https://jira.nypl.org/browse/DSD-189

-Adds TextInput component

### Front End Review:


src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,160 @@
...
+        <textarea
+          id={id}
+          className={bem("inputfield", modifiers, "", [className])}
+          aria-required={required}
+          aria-hidden={type === TextInputTypes.hidden}
+          disabled={disabled}
+          required={required}
+          placeholder={placeholder}
+          // onChange={onChange}
+          // ref={ref}
```
kaplanmichael: Curious why these are commented-out?

src/components/TextInput/_TextInput.scss
```diff
@@ -0,0 +1,75 @@
+$break-input-mobile: 400px;
+
+.textinput {
+  .inputfield {
+    @include space-inset-xs;
+    @include space-stack-xs;
+
```
kaplanmichael: We should think about whether these belong in the component or if this should instead be determined by the parent container.

src/components/TextInput/_TextInput.scss
```diff
@@ -0,0 +1,75 @@
...
+.input-group {
+  @include breakpoint($break-input-mobile) {
+    display: flex;
+    justify-content: space-between;
+
+    > *:not(:last-child) {
+      @include space-inline-m;
+    }
+  }
+}
```
kaplanmichael, line 61: Will there be an input group component in the same way we are planning to do a checkbox and radio group? If so, these styles would probably belong in the group component instead.

src/components/TextInput/TextInputTypes.tsx
```diff
@@ -0,0 +1,38 @@
...
+  hidden = "",
+  // eslint-disable-next-line id-blacklist
+  number = "",
+  password = "",
+  text = "",
+  // eslint-disable-next-line prettier/prettier
+  tel = "([0-9]{3}) [0-9]{3}-[0-9]{4}",
+  textarea = "textarea",
+  url = "https?://.+",
+}
```
kaplanmichael, line 38: Do these get used anywhere?  On the subject, I wonder if we should offer the user the choice of enabling HTML5 field validation or not.  If so, we could also allow them to enter a custom pattern or select from the defaults we have in the enum above.

src/components/TextInput/TextInput.test.tsx
```diff
@@ -0,0 +1,292 @@
...
+});
+
+describe("TextInput Group", () => {
+  let container;
+  before(() => {
+    container = Enzyme.mount(
+      <fieldset>
+        <legend>Input Group Label</legend>
+        <>
+          <>
```
kaplanmichael: Only the outer fragment is needed.  The child components don't need it. For example:
```<><TextInput></TextInput><TextInput></TextInput></>```

File: src/components/TextInput/TextInput.stories.mdx [hunk 6]
```diff
@@ -0,0 +1,149 @@
+import React from "react";
+
+import TextInput from "./TextInput";
+import { TextInputTypes } from "./TextInputTypes";
+import Button from "../Button/Button";
+import { ButtonTypes } from "../Button/ButtonTypes";
+import HelperErrorText from "../HelperErrorText/HelperErrorText";
+// eslint-disable-next-line id-blacklist
+import { text, boolean, select, number } from "@storybook/addon-knobs";
```
kaplanmichael: knobs are deprecated in favor of controls.

src/components/TextInput/TextInput.stories.mdx
```diff
@@ -0,0 +1,149 @@
...
+import { action } from "@storybook/addon-actions";
+
+export default {
+  title: "Form Elements/Text Input",
+  component: TextInput,
+};
+
+<Meta
+  title="Form Elements/Text Input"
+  component={TextInput}
```
kaplanmichael, line 18: This can use the new logic you created in the recently-merged categorization PR.

src/components/TextInput/TextInput.stories.mdx
```diff
@@ -0,0 +1,149 @@
...
+import { ButtonTypes } from "../Button/ButtonTypes";
+import HelperErrorText from "../HelperErrorText/HelperErrorText";
+// eslint-disable-next-line id-blacklist
+import { text, boolean, select, number } from "@storybook/addon-knobs";
+import { action } from "@storybook/addon-actions";
+
+export default {
+  title: "Form Elements/Text Input",
+  component: TextInput,
+};
```
kaplanmichael: This export might not be necessary in MDX--I believe that is all handled by the Meta below.  Would recommend having a quick look at radio component to compare.

src/components/TextInput/TextInput.stories.tsx
```diff
@@ -0,0 +1,133 @@
+import React from "react";
```
kaplanmichael: Is this file still being used for anything?  Most of the time when there's an MDX file in a component folder, it is being used as the component's sole story source.  In a few cases, some of the devs have their MDX files refer to the old TSX, but that doesn't seem to be what you're doing here.

src/components/TextInput/TextInput.stories.mdx
```diff
@@ -0,0 +1,149 @@
...
+        "https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Master?node-id=11895%3A547",
+    },
+  }}
+/>
+
+# TextInput
+
+<Description of={TextInput} />
+
+export const input = () => (
```
kaplanmichael: Using the new MDX format for your stories should allow the user to be able to modify the properties in real-time using the args table.  Recommend having a quick look at Radio component for comparison.

kaplanmichael (changes requested):
Looking good.  Most of the questions I had were about there being both a .stories.tsx and an .stories.mdx file, and that you might be using some tsx formatting in the MDX which is preventing it from taking advantage of all of the MDX benefits and features.

src/components/TextInput/TextInput.tsx
bigfishdesign13: @kaplanmichael 
I commented out these two lines of code because they were throwing an error during development, but, at the time, I was not sure why that was the case.  I then decided to ignore those two lines and moved on.  However, I now see that the errors come up are due to what is being declared on line 41.  The forwardRef is referencing `HTMLInputElement`, but the element where I commented the code is `HTMLTextAreaElement`.  How can I assign that dynamically?

src/components/TextInput/_TextInput.scss
bigfishdesign13: @kaplanmichael 
`space-inset-xs` is the padding within the input field itself, which makes it part of the NYPL styling of the input field.  But, based on what you are suggesting, I will rethink how the other spacing is handled.

src/components/TextInput/_TextInput.scss
bigfishdesign13: @kaplanmichael 
The `Input Group` story was part of the original Input component and I simply kept it along for the ride as I was working on this component.  I actually don't think it is necessary, so I will remove it.  And with that, I do not think a `TextInputGroup` component is necessary.  Do you?

src/components/TextInput/TextInputTypes.tsx
bigfishdesign13: @kaplanmichael 
Those are leftover from my development process.  I was planning on using them, but React was stripping out some of the formatting in the string.  At the momenet, I don't recall the exact problem, but it was related to the backslash character.

HTML5 validation is activated automatically based on the type attribute.  The `pattern` value enhances that validation.

Right now, `pattern` can be passed in as part of the attributes.  That is currently being done on the Input Attribiutes story.

src/components/TextInput/TextInput.test.tsx
bigfishdesign13: @kaplanmichael 
Agreed.  Fixed.

src/components/TextInput/TextInput.stories.tsx
bigfishdesign13: @kaplanmichael 
I took what was already in place with the Input component (.tsx)  and ran with that.  I knew that `.mdx` was the better way, but I balked and simply carried out what had already been started.  The `.mdx` file include here was my first attempt with that formatting, but I didn't get very far.  The `.tsx` file is fully functional for the story, but I would like to convert it to `.mdx`.  What do you think?  Do it now or come back to it after this component has been released?  I'm thinking to do it now.

src/components/TextInput/TextInput.stories.mdx
bigfishdesign13: @kaplanmichael 
I will start working on converting the `.tsx` to `.mdx`.

src/components/TextInput/_TextInput.scss
kaplanmichael: Yeah, probably not.  Most people would use some kind of grid or other layout tool to organize their text inputs.  With checkboxes and radios, a component that groups them and encapsulates their logic is probably more helpful.

src/components/TextInput/TextInput.test.tsx
kaplanmichael: Looking more closely, I see that you actually don't need the fragment tags at all here since you already have a valid wrapping element (fieldset in this case).

src/components/TextInput/TextInput.stories.mdx
kaplanmichael: Looks good now.

src/components/TextInput/TextInput.stories.mdx
kaplanmichael: To clarify, I was referring to `getCategory("__KEY__")`

src/components/TextInput/TextInput.stories.tsx
kaplanmichael: Sorry I missed this question you had asked me, but I do see you've moved forward with MDX and it looks great!

src/components/TextInput/TextInputTypes.tsx
kaplanmichael: How would a user disable HTML5 validation if they don't want to (or can't ) use it in their project?  For example, if their system is set up to do server-side validation, etc...

src/components/TextInput/_TextInput.scss
kaplanmichael: Ah, OK, yes if the space is internal to the component (or field, in this case) then it does make sense. My bad.

kaplanmichael:
The placeholder text's style and color don't match the Figma designs.

src/components/TextInput/TextInput.tsx
kaplanmichael: I see that you've commented out the text area which is fine for now.  Happy to help you troubleshoot that--even if we spin that off into another ticket.  One thing though:  I recommend you also comment textarea out of the textinputtypes file as well, otherwise it will still be selectable in the dropdown: 
image

bigfishdesign13: Good catch.  I commented out `textarea`.

bigfishdesign13: I also created a new Jira ticket so we can address `textarea` in the future.
https://jira.nypl.org/browse/DSD-244

src/components/TextInput/TextInput.test.tsx
bigfishdesign13: Again, good suggestion.  This has been fixed.

src/components/TextInput/TextInputTypes.tsx
bigfishdesign13: This can be done by adding `novalidate` for a `<form>` tag, so that would be in the hands of the developer when building the full form.

src/components/TextInput/TextInput.stories.mdx
bigfishdesign13: Got it.  And done.

bigfishdesign13:
@kaplanmichael 
I've addressed all the issues that you noted in your review.  Where do we go from here?

kaplanmichael (approved):
G2G

PR #545, opened 2021-04-13
DSD-220 categorizing form components

bigfishdesign13:
Fixes #DSD-220
https://jira.nypl.org/browse/DSD-220

- Adds `Form Elements` heading in Storybook components list
- Changes `Button`, `Checkbox`, `Input`, `Label`, `SearchBar` and `Select` to sit under the new `Form Elements` heading

### Front End Review:


EdwinGuzman:
This looks fine. I'm not sure if you need the changelog update since the updates only affect Storybook and not the components themselves.

Eventually, if/when more components are grouped together and not for this PR, it'd be good to have a mapping of the categories titles so you don't have to hard code the title. I'm thinking of something like 
```
// src/docs/categories.tsx
export default {
  formElements: "Form Elements/",
  ...
}
```
and then import it into a story and add it to the title ```title: `${formElements}Checkbox`",```. It's a bit over-engineered since right now there's only one category, but it may help reduce copying/pasting issues and it's easy to see all the categories at a glance.

kaplanmichael (commented):
I agree with @EdwinGuzman that maybe the changelog should only feature changes related to the components themselves and not how NYPL chooses to present them to the world (Storybook, in this case).  That information wouldn't be relevant to users outside NYPL who opt to use the Design System.

bigfishdesign13:
@EdwinGuzman and @kaplanmichael 
I made some changes to enable dynamic category assignment and ease category maintenance.
`src/utils/componentCategories.ts`

I also removed my entry in the changelog.

EdwinGuzman (approved):
The update looks good!

kaplanmichael (approved):
LGTM

PR #547, opened 2021-04-23
Dsd 231 checkbox

kaplanmichael:
Fixes https://jira.nypl.org/browse/DSD-231

-Adds a new checkbox component which matches the approved Figma design.
- Refactors the color variables so that there are both CSS variables (custom properties) _and_ SASS variables, which were needed for some preprocessing operations.
- Removes linter plugin which is no longer necessary since we are now creating the CSS vars dynamically.

### Front End Review:


src/styles/01-colors/_colors-utility.scss
```diff
@@ -0,0 +1,66 @@
...
+  }
+}
+
+//generates a collection of CSS vars
+@mixin css-vars {
+  @each $name, $color in $colors {
+    --ui-#{$name}: #{$color};
+  }
+}
+
```
EdwinGuzman: Looks like you can have a utility scss file for these mixins so you don't have to declare them again in this file. Also, it'd be good to have this in a separate PR next time, just easier to see all the related changes

src/index.ts
```diff
@@ -5,7 +5,7 @@ export { default as Breadcrumbs } from "./components/Breadcrumbs/Breadcrumbs";
 export { default as Button } from "./components/Button/Button";
 export { ButtonTypes } from "./components/Button/ButtonTypes";
 export { default as Card } from "./components/Card/Card";
-export { default as Checkbox } from "./components/Checkbox/Checkbox";
+export { default as CheckboxOld } from "./components/Checkbox-old/CheckboxOld";
```
EdwinGuzman: By doing this you're not exporting the "new" `Checkbox` component, just the `CheckboxOld` component which, if no one should use, should not be exported.

EdwinGuzman:
Is there a reason why the "older" component version is being kept around? I'm not sure if the "deprecated" label in storybook or the appended "old" in the component name really work. The `Checkbox` component went through some updates but the changes are not vastly different in what it renders or how its styles that there should be an "old" version of it. If the new designs are approved, the older version shouldn't still be allowed to be used.

The LCA, for example, implements the `Checkbox` component. When I update to the version of DS that includes this change, I would expect the styling to be updated and that I need to include new props. But I would not switch to the `CheckboxOld` component if it goes against what Figma has approved for styling. 

All of this to say that it seems safe to remove the "old" files.

EdwinGuzman:
Also, I'm not sure if it's this PR but something is wrong with the `Hero` component in Storybook -- the homepage link in this PR throws an error: https://deploy-preview-547--stoic-murdock-c7f044.netlify.app/

src/components/Checkbox/Checkbox.tsx
```diff
@@ -1,69 +1,77 @@
...
+      name,
       onChange,
+      showLabel,
+      value,
     } = props;
 
-    const baseClass = "checkbox";
+    const attributes = props.attributes || {};
+    const modifiers = props.modifiers ? props.modifiers : [];
+    const checkboxID = id || generateUUID();
```
EdwinGuzman, line 53: Why not set these defaults in the destruction assignment?

kaplanmichael:
> Is there a reason why the "older" component version is being kept around? I'm not sure if the "deprecated" label in storybook or the appended "old" in the component name really work. The `Checkbox` component went through some updates but the changes are not vastly different in what it renders or how its styles that there should be an "old" version of it. If the new designs are approved, the older version shouldn't still be allowed to be used.
> 
> The LCA, for example, implements the `Checkbox` component. When I update to the version of DS that includes this change, I would expect the styling to be updated and that I need to include new props. But I would not switch to the `CheckboxOld` component if it goes against what Figma has approved for styling.
> 
> All of this to say that it seems safe to remove the "old" files.

@EdwinGuzman thanks for your thoughts on this! @bigfishdesign13 is still mulling how we might deprecate components so I created the "deprecated" section more as a POC of one way we could do it, but that's not set in stone!  We should probably do a team discussion on this.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -1,69 +1,77 @@
...
+    }
     return (
-      <div className={bem(baseClass, modifiers, blockName, [className])}>
+      <>
         <input
-          id={checkboxId}
-          name={name}
-          className={bem("input", [], baseClass)}
-          type="checkbox"
-          {...(checked !== undefined
```
nypl-wluisi, line 53: @kaplanmichael Did you intentionally remove this? The whole point of this conditional is to allow the component to be used a controlled component or uncontrolled component.

If you use pass a `checked` prop, then you're doing a "controlled" component, and it gives you `checked` prop as well as `aria-checked` and `onChange` props. If you don't pass a `checked` prop, then you're doing an "uncontrolled" component, and it just gives you a `defaultChecked` prop.

src/styles/01-colors/_colors-utility.scss
kaplanmichael: Done!

src/components/Checkbox/Checkbox.tsx
kaplanmichael: Done!  (Note:  the id stays where it is because I also want the case of "" to generate a UUID. If I define that in the destructure, it only cares that the value is truthy and so it allows "" to pass.)

src/components/Checkbox/Checkbox.tsx
kaplanmichael: You only need `aria-checked` if you are giving something other than a checkbox the the role of a checkbox (`role=checkbox`).  Also the user should be able to attach an onChange regardless of whether it is controlled or not.
For `defaultChecked`, you get the ability to query that regardless of whether it is passed as a prop or not (again, because we're using a native checkbox element and not some other element given the role of a checkbox).

kaplanmichael:
> Also, I'm not sure if it's this PR but something is wrong with the `Hero` component in Storybook -- the homepage link in this PR throws an error: https://deploy-preview-547--stoic-murdock-c7f044.netlify.app/

NB: @bigfishdesign13

src/components/Checkbox/Checkbox.tsx
kaplanmichael: @nypl-wluisi this should be good now.  Please let me know.

src/components/Checkbox/_Checkbox.scss
```diff
@@ -1,8 +1,70 @@
...
+        border-color: $error;
+
+        &:before {
+          background: url('data:image/svg+xml;utf8,<svg viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><path fill="#{url-friendly-colour($checkmark-color-error)}" d="M8.795 15.875l-4.17-4.17-1.42 1.41 5.59 5.59 12-12-1.41-1.41-10.59 10.58z"/></svg>')
+            50% 50% no-repeat;
+        }
+      }
+    }
+
+    &:disabled {
```
bigfishdesign13, line 64: There are some descrepencies in the Figma file regarding the styling of various states of Checkbox and Radio.  I'd like to get that ironed out before moving forward.  I'll keep you posted.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -1,69 +1,86 @@
...
-      labelOptions,
+      attributes = {},
       checked,
+      className,
       defaultChecked,
+      disabled,
+      errored,
+      id,
+      labelText,
+      modifiers,
```
bigfishdesign13: `modifers` is handled differently in `Radio`.  Can we mirror the streucture used there, where it is pulled out of this props list and defined in the same area where `checkboxID` is defined below?

`const modifiers = props.modifiers ? props.modifiers : [];`

bigfishdesign13: I am aslo finding that there is an error showing in Storybook when the errored prop is toggled.

src/components/Checkbox/Checkbox.stories.mdx
```diff
@@ -0,0 +1,82 @@
...
+  ArgsTable,
+  Preview,
+  Description,
+} from "@storybook/addon-docs/blocks";
+import { withDesign } from "storybook-addon-designs";
+
+import Checkbox from "./Checkbox";
+
+<Meta
+  title="Form Elements/Checkbox"
```
bigfishdesign13: Make sure to utilize `getCategory("Checkbox")` here.

bigfishdesign13 (changes requested):

bigfishdesign13:
> Also, I'm not sure if it's this PR but something is wrong with the `Hero` component in Storybook -- the homepage link in this PR throws an error: https://deploy-preview-547--stoic-murdock-c7f044.netlify.app/

@kaplanmichael 
I am not seeing this issue in any other builds.  What about your local build?

bigfishdesign13:
@kaplanmichael 
I am also noticing that the Docs tab in Storybook is throwing an error for `Checkbox`.  Are you seeing the same?

bigfishdesign13:
> Is there a reason why the "older" component version is being kept around? I'm not sure if the "deprecated" label in storybook or the appended "old" in the component name really work. The `Checkbox` component went through some updates but the changes are not vastly different in what it renders or how its styles that there should be an "old" version of it. If the new designs are approved, the older version shouldn't still be allowed to be used.
> 
> The LCA, for example, implements the `Checkbox` component. When I update to the version of DS that includes this change, I would expect the styling to be updated and that I need to include new props. But I would not switch to the `CheckboxOld` component if it goes against what Figma has approved for styling.
> 
> All of this to say that it seems safe to remove the "old" files.

@EdwinGuzman and @kaplanmichael 
I think Edwin is spot on with this.  Especially so, considering we are still in a pre-release/beta mode.  I will send out a note to the #design-system Slack channel to annopunce that this change is coming.  I will also make note of this change once the release is completed.

src/components/Checkbox/Checkbox.tsx
kaplanmichael: Fixed!

src/components/Checkbox/Checkbox.stories.mdx
kaplanmichael: Fixed!

kaplanmichael:
> > Also, I'm not sure if it's this PR but something is wrong with the `Hero` component in Storybook -- the homepage link in this PR throws an error: https://deploy-preview-547--stoic-murdock-c7f044.netlify.app/
> 
> @kaplanmichael
> I am not seeing this issue in any other builds. What about your local build?

I don't have any problems locally.

bigfishdesign13 (approved):
Looks good.

PR #548, opened 2021-04-30
DSD-152: Notification

siberry:
- Adds the "Notification" component to Design System

This is not complete. Just want to check in that this is going in the right direction. I have one week left at the library. Maybe I can finish this? If not, want to leave it as easy to wrap up as possible. To come:
- Add in correct icons from PR #541
- Test coverage

### Front End Review:
- [ ] View [the example in Storybook](https://deploy-preview-548--stoic-murdock-c7f044.netlify.app/?path=/story/notification--default)


src/index.ts
```diff
@@ -26,6 +26,11 @@ export { LinkTypes } from "./components/Link/LinkTypes";
 export { default as List } from "./components/List/List";
 export { ListTypes } from "./components/List/ListTypes";
 export { default as Modal } from "./components/Modal/Modal";
+// export {
```
EdwinGuzman: Is this ready to be exported or why is this commented out?

bigfishdesign13: Yes, an oversight.  Fixed.

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,201 @@
...
+  /** Optional prop to control the margin around the `Notification` component */
+  noMargin?: boolean;
+  /** Optional prop to control the coloring of notification text and the visibility of an applicable icon */
+  notificationType?: NotificationTypes;
+  /** Modifiers array for use with BEM. See how to work with modifiers and BEM here: http://getbem.com/introduction/ */
+  modifiers?: string[];
+}
+
+// NotificationHeading child-component
+export function NotificationHeading({ ...props }) {
```
EdwinGuzman: `NotificationHeading` and `NotificationContent` also have their own file declarations so these should be cleaned up. Both declarations are different from the file components.

Also, it's not necessary to do `NotificationHeading({ ...props })` and `NotificationHeading(props)` is the same.

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,201 @@
...
+    className,
+    dismissible = false,
+    noMargin = false,
+    icon,
+    id,
+    modifiers = [],
+    notificationType = "standard",
+  } = props;
+
+  const [isOpen, setIsOpen] = useState(true);
```
EdwinGuzman, line 71: Ah, I thought we were trying not to have components with their own internal states. Did you get a chance to try this branch in a working app? I've encountered problems with components with internal state before so I want to make sure that's not a problem anymore.

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,201 @@
...
+  const notificationModifiers = modifiers ? modifiers : [];
+
+  notificationModifiers.push(notificationType);
+
+  if (centered) notificationModifiers.push("centered");
+
+  if (noMargin) notificationModifiers.push("no-margin");
+
+  const iconElement = () => {
+    if (icon) return icon;
```
EdwinGuzman, line 90: This is fine and I would not change this, but in general, I think using global scope props in functions can be buggy. Sometimes it's better to pass them all in as arguments. Don't update this though as it can be nitpicky.

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,201 @@
...
+            modifiers={["large"]}
+          />
+        );
+      default:
+        return null;
+    }
+  };
+
+  let childHeading,
+    childContent,
```
EdwinGuzman: The linter doesn't throw an error but create a new declaration for each variable. So,
```
let childheading;
let childContent;
...
```

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,201 @@
...
+      "Error: Only one NotificationContent child component may be passed to Notification.";
+  }
+
+  if (
+    headingCount &&
+    (icon ||
+      ((notificationType === NotificationTypes.Announcement ||
+        notificationType === NotificationTypes.Warning) &&
+        !centered))
+  )
```
EdwinGuzman: Nit: for multiline `if` statements, add the curly braces around line 164.

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,201 @@
...
+    modifiers.push("align-text");
```
EdwinGuzman: I think you meant `notificationModifiers`, right?

src/components/Notification/Notification.stories.mdx
```diff
@@ -0,0 +1,180 @@
...
+      purus sit amet fermentum. Maecenas faucibus mollis interdum. Morbi leo
+      risus, porta ac consectetur ac, vestibulum at eros. Cum sociis natoque
+      penatibus et magnis dis parturient montes, nascetur ridiculus mus. Vivamus
+      sagittis lacus vel augue laoreet rutrum faucibus dolor auctor.
+    </NotificationContent>
+  </Notification>
+</Canvas>
+
+<Canvas>
+  <Notification notificationType={NotificationTypes.Announcement}>
```
EdwinGuzman, line 90: I think the `announcement` and `warning` types would be good to have as standalone stories under the `Notification` category. Or maybe a `Variants` story with these two under it.

EdwinGuzman: Perhaps also a separate story with the closing button on.

src/components/Notification/Notification.stories.mdx
```diff
@@ -0,0 +1,180 @@
...
+    <NotificationContent>
+      This is a Notification with a custom icon. Cras mattis consectetur purus
+      sit amet fermentum. Maecenas faucibus mollis interdum. Morbi leo risus,
+      porta ac consectetur ac, vestibulum at eros. Cum sociis natoque penatibus
+      et magnis dis parturient montes, nascetur ridiculus mus.
+    </NotificationContent>
+  </Notification>
+</Canvas>
+
+### Errors
```
EdwinGuzman: Would "Development Errors" be a better title? This would only really happen while actively working the DS component and is not a variant.

EdwinGuzman (changes requested):
Looking good so far but have the question about whether this component with internal state was testing in a consuming app and the extra components (and files) need to be removed or cleaned up.

src/components/Notification/Notification.tsx
bigfishdesign13: Ah, an oversight.  I started with the separate files, but eventually moved everything within the main `Notification.tsx` file.  Do you have a preference (internal definitions or separate files)?  Pros and cons of the two methods?

src/components/Notification/Notification.tsx
bigfishdesign13: This is something that was included before I picked up the development.  I will pull `Notification` into the test app and see what happens.

src/components/Notification/Notification.tsx
bigfishdesign13: Got it.  Thanks.

src/components/Notification/Notification.tsx
bigfishdesign13: Done.

src/components/Notification/Notification.tsx
bigfishdesign13: Done.

src/components/Notification/Notification.tsx
bigfishdesign13: You're right.  I did mean `notificationModifiers`.  Odd that it was working properly with the wrong var name in there.  Fixed now.

src/components/Notification/Notification.stories.mdx
bigfishdesign13: When I built up the Story page, I imagined users relying on the `Docs` tab and adjusting props in the `ArgsTable`.  With that in mind, I felt that multiple stories would not be necessary.  The `Variants` section was more of a quick visual reference for what could be done with the component.  I do, however, now feel that separate stories would be good for `Cusotm Icon` and `Development Errors`.

src/components/Notification/Notification.stories.mdx
bigfishdesign13: Good suggesttion.  Done.

EdwinGuzman (approved):
I'm going to approve this to move forward with this update because it's been open for a long time. We talked about not implementing hooks or using state in components so that users handle that themselves. I think it'd be good to talk about our decision on this again since we will soon integrate Chakra and that library uses hooks and internal state in its own components.

For now, this is fine but let's keep in mind that we will need to revisit a previously agreed convention.

PR #550, opened 2021-05-03
DSD-180: StatusBadge Component Set Default Width

nypl-wluisi:
Jira ticket: https://jira.nypl.org/browse/DSD-180

- Sets a default width of `fit-content` to StatusBadge component and adds more padding inside to match Figma. Previously, this component would just expand full width, and require custom styles. But in all cases in Figma, the component's width should match the width of the inner text.

### Front End Review:
- [ ] View [the example in Storybook](https://deploy-preview-550--stoic-murdock-c7f044.netlify.app/?path=/story/status-badge--status-badge)


src/components/StatusBadge/_StatusBadge.scss
```diff
@@ -4,8 +4,9 @@
   display: block;
   font-size: var(--font-size--1);
   font-style: italic;
-  padding: var(--space-xxs);
+  padding: var(--space-xxs) var(--space-xs);
   text-align: center;
+  width: fit-content;
```
kaplanmichael, line 9: A width of `fit-content` might not work properly on Firefox unless you use a vendor prefix.  Could you please verify that this looks good on FF (I'd do it myself but I am still trying to get a license for BrowserStack )

CC @bigfishdesign13

kaplanmichael (changes requested):

src/components/StatusBadge/_StatusBadge.scss
kaplanmichael: @wluisi Yeah, it looks like you'll probably need the browser prefix. Here's what it loox like on FF:
image

nypl-wluisi: @kaplanmichael Ok I can add that to this PR. For some reason I thought the DS was already using something like this? https://www.npmjs.com/package/autoprefixer -- for now, I'm fine with adding the prefixes manually to get this PR finished, but do you think we should create a ticket to add autoprefixing ?

nypl-wluisi:
@kaplanmichael I added the prefix for moz FF. Also, this is pre-existing, but we're setting the border radius with this:
```
border-radius: 4px;
```

Should this use a css var instead of hardcoding a specific px value?

src/components/StatusBadge/_StatusBadge.scss
kaplanmichael: That's funny--I had thought that we had an auto-prefixer as well.  I'll mention that to Marty and see if we can make that happen.

kaplanmichael:
> @kaplanmichael I added the prefix for moz FF. Also, this is pre-existing, but we're setting the border radius with this:
> 
> ```
> border-radius: 4px;
> ```
> 
> Should this use a css var instead of hardcoding a specific px value?

Yes, @wluisi if you wouldn't mind using ` --space-xxs`, that would be great!  Thanks!

kaplanmichael (approved):
LGTM

PR #552, opened 2021-05-04
Adding documentation on layout bidirectionality

EdwinGuzman:
Fixes [DSD-249](https://jira.nypl.org/browse/DSD-249)

- Just documentation on webpage directionality as a precursor to making any updates on components that will be needed for internationalization. For many left to right (LTR) languages such as English, no updates need to be made. For right to left (RTL) languages such as Arabic, updates need to be made to the app to support the RTL languages.
- The NYPL DS _does not_ need to worry about application-specific implementation but I wrote background on the why and how to implement.
- What the NYPL DS _does_ need to worry about is displaying components appropriately when the layout is mirror and when text is aligned right.

There's an example in the [documentation page](https://deploy-preview-552--stoic-murdock-c7f044.netlify.app/?path=/story/bidirectionality--page#component-updates) I wrote in Storybook but basically, using rules like `padding-left`/`padding-right` or `margin-left`/`margin-right` works well now using LTR languages, but in RTL languages the UI doesn't look right. In the `Button` example, when the browser switches the placement of the icon and the text inside a button there is no space between the icon and the text. The `margin-right` rule in the icon doesn't work when the text is on the left side.

Screen Shot 2021-05-04 at 2 17 15 PM

If this PR sounds right, the next step to this approach would be making an audit and seeing how we can update margin specific rules to be more universal to either side of an element.

### Front End PR Review:
- [ ] View the [Bidirectionality example in Storybook](https://deploy-preview-552--stoic-murdock-c7f044.netlify.app/?path=/story/bidirectionality--page)


src/components/Template/Bidirectionality.stories.mdx
```diff
@@ -0,0 +1,248 @@
...
+import SearchBar from "../SearchBar/SearchBar";
+import Input from "../Input/Input";
+import { InputTypes } from "../Input/InputTypes";
+import Icon from "../Icons/Icon";
+import { IconNames } from "../Icons/IconTypes";
+import Button from "../Button/Button";
+import { ButtonTypes } from "../Button/ButtonTypes";
+import HelperErrorText from "../HelperErrorText/HelperErrorText";
+
+<Meta title="Bidirectionality" />
```
bigfishdesign13: There is something odd happening with the alphabetical sort order in the Storybook left menu.  Out of curiosity, I changed the title to `Template Bidirectionality` and the link jumped all the way to the bottom of the Stories list.  Are you seeing the same in your local build?

bigfishdesign13 (changes requested):
This provides a lot of excellent info as a precurosr to making changes to better support RTL languages.  We certainly will need to have some further discussions to map out the best way to handle this with DS.  The only issue I am seeing for this PR is in regard to the sorting in the Storybook left menu.

src/components/Template/Bidirectionality.stories.mdx
EdwinGuzman: Yes, that happens with the live-reload. If you refresh the page, it'll sort correctly. I'll change the title to `Template Bidirectionality`.

EdwinGuzman:
Updated. Stories are not sorted alphabetically. They are sorted by the order that they get read in the repo structure. It just so happens that all the components are sorted alphabetically (with the exception of the "form elements" section and the "introduction" page).

It's possible to [sort alphabetically](https://storybook.js.org/docs/react/writing-stories/naming-components-and-hierarchy#sorting-stories) but it messes up with the "introduction" page which we want to show first in the list. If you want, we can dig more into this but it should be a separate ticket.

src/components/Template/Bidirectionality.stories.mdx
bigfishdesign13: Got it.  Looks good now.

bigfishdesign13 (approved):
Looks good.

bigfishdesign13:
> Updated. Stories are not sorted alphabetically. They are sorted by the order that they get read in the repo structure. It just so happens that all the components are sorted alphabetically (with the exception of the "form elements" section and the "introduction" page).
> 
> It's possible to [sort alphabetically](https://storybook.js.org/docs/react/writing-stories/naming-components-and-hierarchy#sorting-stories) but it messes up with the "introduction" page which we want to show first in the list. If you want, we can dig more into this but it should be a separate ticket.

Just saw this comment.  And understood.  We will eventually move all the components into category groupings (ex. Form Elements), which will resolve some of the sorting issues.  But, as it is right now, this looks good with the `Template Bidirectionality` title.

PR #554, opened 2021-05-08
DSD-243: Add Tertiary Hero component

knezmilos:
Jira ticket: https://jira.nypl.org/browse/DSD-243

- Adds Tertiary Hero component.  Background color is defaulted to "--ui-gray-xdark" but there is a selection of colors same as in the Hero Secondary component.

### Front End Review:
- [ ] View [the example in Storybook](http://localhost:9000/?path=/story/hero--hero-tertiary)


nypl-wluisi (changes requested):
@knezmilos This looks pretty good so far. The padding around the text here seems different than what's in the Figma design for "Online Resources":

image-2021-04-28-11-48-58-038

Also, in terms of Storybook, I think an example with short description text would be good, but also an example with longer text that will stretch 2 lines?  Does this sound like a good idea: @bigfishdesign13 @kaplanmichael

src/components/Hero/_Hero.scss
```diff
@@ -67,6 +67,39 @@
   }
 }
 
+.hero--tertiary {
```
bigfishdesign13: For consitency, move the CSS for `.hero--tertiary` below the CSS for `.hero--secondary`.

src/components/Hero/_Hero.scss
```diff
@@ -67,6 +67,39 @@
...
+  .hero__content {
+    @include wrapper;
+
+    color: var(--ui-white);
+    display: flex;
+    flex-flow: column nowrap;
+    padding: var(--space-m) var(--space);
+
+    .hero__heading {
+      @include space-stack-none;
```
bigfishdesign13: Based on the Figma, the spacing between the header and the body text should be increased to `space-stack-s`.

src/components/Hero/_Hero.scss
```diff
@@ -67,6 +67,39 @@
...
+    background-color: var(--section-locations-primary);
+  }
+
+  .nypl--research & {
+    background-color: var(--section-research-primary);
+  }
+
+  .nypl--whats-on & {
+    background-color: var(--section-whats-on-primary);
+  }
```
bigfishdesign13: Let's keep this color organization as is for now, but this should eventually be replaced with a more comprehensive solution for applying background colors.

src/components/Hero/_Hero.scss
```diff
@@ -67,6 +67,39 @@
...
+
+  .hero__content {
+    @include wrapper;
+
+    color: var(--ui-white);
+    display: flex;
+    flex-flow: column nowrap;
+    padding: var(--space-m) var(--space);
```
bigfishdesign13: The padding for the top/bottom should be bumped up to `--space-l` (32px).

bigfishdesign13 (changes requested):
Milos, please see my comments regarding the CSS.

bigfishdesign13:
> @knezmilos This looks pretty good so far. The padding around the text here seems different than what's in the Figma design for "Online Resources":
> 
> image-2021-04-28-11-48-58-038
> 
> Also, in terms of Storybook, I think an example with short description text would be good, but also an example with longer text that will stretch 2 lines? Does this sound like a good idea: @bigfishdesign13 @kaplanmichael

Yes, I agree.  An example with longer body text (2-3 lines at full width) would be nice to see.

knezmilos:
Sure, I'll work on the changes.  Quick note, having .hero--tertiary below the CSS for .hero--secondary gave an error when the scss was compiling, but I'll work on resolving that.
Thanks

knezmilos:
Adjustments are made based on above comments.  As far as longer body text, I added the text from Figma, it goes across 2 lines.
Also, spoke briefly with Marty today about the background color input, it was decided to use the way Primary and Hero5050 components use the background color input and to work on a better option for passing color for all components later.

bigfishdesign13 (approved):

src/components/Hero/_Hero.scss
```diff
@@ -165,3 +165,16 @@
...
+  background-color: var(--ui-gray-xdark);
+
+  .hero__content {
+    @include wrapper;
+
+    color: var(--ui-white);
+    display: flex;
+    flex-flow: column nowrap;
+    padding: var(--space-l) var(--space);
+  }
```
bigfishdesign13, line 192: What happened to `.hero_heading`?  The spacing for that element still needs to be set to `space-stack-s` (16px or 1rem).

src/components/Hero/_Hero.scss
```diff
@@ -165,3 +165,16 @@
     }
   }
 }
+
+.hero--tertiary {
+  background-color: var(--ui-gray-xdark);
```
bigfishdesign13, line 170: Are the background-color variants being picked up by the classes in `.hero--secondary`?

bigfishdesign13 (changes requested):
It looks like you were able to remove some of the classes when you moved the `tertiary` code to the end of the file.  However, the `.hero_heading` class should be added back in.  Please see my inline comments.

src/components/Hero/_Hero.scss
knezmilos: `.hero_heading` was removed because running `npm run test`, it returns this error:
"180:5  ✖  Expected selector ".hero--tertiary .hero__content .hero__heading" to come before selector ".nypl--books-and-more  .hero--secondary .hero__content .hero__heading" - no-descending-specificity"
That wasn't an issue when `.hero--tertiary` was above `.hero--secondary`.  Would you happen to know how to resolve that?

src/components/Hero/_Hero.scss
knezmilos: The color variants will be added through the `backgroundColor` parameter, same as the Hero Primary or Hero 5050 would pass the background color.  
I suppose background color can be added the same way as Hero Secondary, but I was running into the lint issue described in the previous comment.

src/components/Hero/_Hero.scss
bigfishdesign13: Hmm, I am not seeing any errors when I run `npm run test`.  Maybe kill the `node_modules` folder and run `npm install` before running the tests?  If the error persists, then just move the `tertiary` block back above the `secondary` block.

src/components/Hero/_Hero.scss
bigfishdesign13: OK, that works.

bigfishdesign13 (approved):

src/components/Hero/_Hero.scss
```diff
@@ -165,3 +165,22 @@
...
+    @include wrapper;
+
+    color: var(--ui-white);
+    display: flex;
+    flex-flow: column nowrap;
+    padding: var(--space-l) var(--space);
+
+    /* stylelint-disable */
+    .hero__heading {
+      @include space-stack-s;
```
bigfishdesign13, line 182: Sorry, missed this the first time around.  We need to add some logic/CSS that sets bottom-margin to `space-stack-none` when the `subHeaderText` is not shown.  Either that or wrap the `subHeaderText` in a `<p>` and add top-margin to that element.

bigfishdesign13: To keep in alignment with how our CSS is structured for the DS, the first option makes the most sense.

bigfishdesign13 (changes requested):

src/components/Hero/_Hero.scss
knezmilos: I've made some adjustments.  `subHeaderText` needs to be wrapped in a `<p>` tag and with that only bottom margin is set.  Please review

bigfishdesign13: Looks good.

bigfishdesign13 (approved):

bigfishdesign13 (approved):
Looks good to me.

knezmilos:
@nypl-wluisi, this is approved to go get merged, can you approve as well so I can merge.

nypl-wluisi (approved):
@knezmilos Looks good!

PR #559, opened 2021-05-20
[DSD-201] Fix inconsistent behavior with checked/unchecked in Radio and Checkbox components

kaplanmichael:
Remove 'checked' prop from args table to avoid issue where storybook would fail to toggle between controlled/uncontrolled properly.


bigfishdesign13 (approved):
Looks good!

PR #568, opened 2021-06-03
DSD-289: heading mdx story

bigfishdesign13:
Fixes DSD-289

- Changes `Heading` story from `.tsx` to `.mdx`

### Front End Review:


hokei (changes requested):
Kind of same as all the other PRs:  Show the "basic" heading many times with different args in the "all headings" page, instead of creating a separate "all headings" story. Otherwise, looks great!

src/components/Heading/Heading.stories.mdx
```diff
@@ -0,0 +1,92 @@
...
+  >
+    {args => <Heading {...args} />}
+  </Story>
+</Preview>
+
+<ArgsTable story="Basic" />
+
+## All Heading Sizes
+
+<Preview>
```
hokei, line 54: If you nest this all inside a div (like 

```
<Preview>
  <div>
    <Heading id="heading1" level={1} text="Heading 1" />
    <Heading id="heading2" level={2} text="Heading 2" />
    <Heading id="heading3" level={3} text="Heading 3" />
    <Heading id="heading4" level={4} text="Heading 4" />
    <Heading id="heading5" level={5} text="Heading 5" />
    <Heading id="heading6" level={6} text="Heading 6" />
  </div>
</Preview>
```

It will all appear on different lines.  

In the future also, you can make storybook-specific styles to overwrite storybook stuff.  
Like, if you went 
<Preview>
  <div className="sb-only-class"> 
     <Component />
  </div>
</Preview>


We should figure out how to not export these classes, but I think the only sb class that's left is in `card` and it's not actually being used.

bigfishdesign13: Thanks for the suggestion.  Works great.

bigfishdesign13: I made a note to myself to take a look at the Storybook-specific styles at some point.

hokei (approved):

PR #569, opened 2021-06-04
DSD-295: horizontal rule

bigfishdesign13:
Fixes DSD-295

- Adds HorizontalRule component

### Front End Review:


hokei (changes requested):
There's no need to have both a "basic" and "various configurations" story.  Just show "basic" with the configurations in the docs tab.  The user can use controls to preview their desired configurations themselves.  

(UNLESS these are the only allowed configurations.  As of now, I'd rather give the users freedom over their configurations)

src/components/HorizontalRule/HorizontalRule.tsx
```diff
@@ -0,0 +1,44 @@
...
+  modifiers?: string[];
+  /** Optional width value.  This value should be entered with the same formatting as a CSS width attribute (ex. `50%`, `640px`, `20rem`).  If omitted, the horizobntal rule will have a default width of 100%. */
+  width?: string;
+}
+
+export default function HorizontalRule(
+  props: React.ComponentProps<"hr"> & HorizontalRuleProps
+) {
+  const { align, className } = props;
+  const attributes = props.attributes || {};
```
hokei: If you change L23 to `  const { align, className, modifiers, width, height } = props;`
you can avoid invoking `props` every time.

src/components/HorizontalRule/HorizontalRule.tsx
```diff
@@ -0,0 +1,42 @@
...
+import * as React from "react";
+import bem from "../../utils/bem";
+
+export interface HorizontalRuleProps {
+  /** Optional alignment value to align the horizontal rule to one side or the other when the width is less than 100%. If omitted, the horizobntal rule will have a default center alignment. */
+  align?: "left" | "right";
+  /** Additional attributes passed to the horizontal rule */
+  attributes?: { [key: string]: any };
+  /** ClassName you can add in addition to `horizontal-rule` */
+  className?: string;
```
hokei, line 11: If one wants to add more than a single additional class, do they just pass all in a string?

src/components/HorizontalRule/HorizontalRule.tsx
hokei: I rewrote this class to be 

```

export default function HorizontalRule(
  props: React.ComponentProps<"hr"> & HorizontalRuleProps
) {
  const {
    align,
    attributes = {},
    className,
    width = "auto",
    height = "2px",
    modifiers = [],
  } = props;

  attributes.style = {
    width: width,
    height: height,
  };

  if (align) {
    modifiers.push(align);
  }

  return (
    <hr
      className={bem("horizontal-rule", modifiers, "", [className])}
      {...attributes}
    />
  );
}
```

and it seemed to work ok?  Let me know if this doesn't work.

hokei (changes requested):
forgot to hit "request changes"

src/components/HorizontalRule/HorizontalRule.tsx
bigfishdesign13: Works great.

src/components/HorizontalRule/HorizontalRule.tsx
bigfishdesign13: Yes.  

`className="classone classtwo"`

.storybook/manager-head.html
```diff
@@ -0,0 +1,5 @@
+<style>
```
hokei, line 1: Is this being used at all?  If not, remove until we actually need it

hokei (approved):
Some nits, but they're not code issues so I'm gonna pre-emptively approve

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 ### Changes
 
+- Adds `HorizontalRule` component
```
hokei, line 18: You want this to be in a separate section, like 
```
### Adds 
- Adds `HorizontalRule` component

### Changes 
- Changes `Breadcrumbs` story from `.tsx` to `.mdx`

bigfishdesign13: Fixed.

.storybook/manager-head.html
bigfishdesign13: Removed.

PR #578, opened 2021-06-23
DSD-322: form component

bigfishdesign13:
Fixes [DSD-322](https://jira.nypl.org/browse/DSD-322)

- Adds `Form` component for formatting forms and their child input fields

### Front End Review:


bigfishdesign13:
@nypl-wluisi, @hokei, @EdwinGuzman 
I have been thinking a lot about how to help developers with formatting a new page.  With that in mind, I put together a `Form` component to control the spacing of input fields within a `<form>`.  Please take a look and give me your thoughts on what else might be needed in a component like this.  Right now, it is simply a component for formatting the UI.  But should it be more than that?

src/components/Form/Form.stories.mdx
```diff
@@ -0,0 +1,130 @@
...
+        "https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Master?node-id=11895%3A547",
+    },
+  }}
+/>
+
+# Form
+
+<Description of={Form} />
+
+The `Form` component renders a standard `<form>` element and should be used to handle spacing for child input fields. Use the `FormRow` and `FormField` child-components to arrange your input fields accordingly.
```
EdwinGuzman: Can you say, "should be used to handle layout and spacing ..."? It'll be good to note that the combination of these components helps with the overall layout and not just spacing between form inputs.

Maybe it'll be clearer to say that `FormRow` and `FormField` are components that should be used as children only in the `Form` component, rather than saying they are "child-components".

src/components/Form/Form.test.tsx
```diff
@@ -0,0 +1,51 @@
+// import { expect } from "chai";
```
EdwinGuzman: This should be updated eventually. It may be hard to write tests for layout-only components... but we can think of a way to test them.

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,45 @@
...
+// FormField child-component
+export function FormField({ ...props }) {
+  const { children, className } = props;
+  return (
+    <div className={bem("form-field", [], "", [className])}>{children}</div>
+  );
+}
+
+export default function Form(props: React.ComponentProps<"form"> & FormProps) {
+  const { attributes = {}, children, className, id, modifiers = [] } = props;
```
EdwinGuzman: We'll have to discuss this later but in case someone forgets to use the `attributes` prop and adds props directly to the component, this would need a `...rest` variable to catch those props.

EdwinGuzman:
> But should it be more than that? 

I think it should have `method` and `action` explicit props since those are app-behavior attributes on the `form` that are commonly used.

I personally very much like the idea of composable components that can be used for layouts so this is a great step. Chakra UI uses a similar approach, right?

The storybook doc page is very clean but I think it's missing a code snippet example so it's easier for devs to see the component hierarchy and the parent-child relationships (multiple examples to show the explanation that was given).

bigfishdesign13:
@hokei and @wluisi 
Just a reminder to take a look at and comment on the PR for Form.  Thanks.

src/components/Form/Form.stories.mdx
```diff
@@ -0,0 +1,130 @@
...
+
+Multiple `FormField` components can be placed in a parent `FormRow` component. However, a `FormField` component does not have to placed within a `FormRow` component. `FormField` components can stand alone within `Form`.
```
hokei: When should `FormField` components not be in `FormRow` components? If there's only one row?

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,45 @@
+// HorizontalRule
```
hokei: ?

src/components/Form/Form.stories.mdx
hokei: I'm trying to figure out if there's a way to make these divs not so deeply nested.  
For the `<textInput>`, which is a wrapper around the label and input, there's really no need for both a `<FormField>` and a `<TextInput>`  It makes more sense to just have a <FormRow> that handles box spacing and stuff.  

But I do see why `<FormField>` is necssary in the checkbox and radio groups - I wonder if it makes more sense to make `FormRow` required, and `FormField` optional.

hokei:
Perhaps another ticket:  It looks like the field validation is happening natively here: 
Screen Shot 2021-06-28 at 4 38 45 PM

Whereas in other places, it uses HelperErrorText (https://deploy-preview-578--stoic-murdock-c7f044.netlify.app/?path=/story/form-elements-textinput--text-input&args=errored:true) 

Which is the way DS reccommends?

EdwinGuzman:
@hokei that's interesting. What browser are you using? That popup doesn't appear for me. In general, native validation should appear before the DS validation updates. I think this is because when you try to tab or when you focus away from an input element, the browser stops you from doing so. That then blocks the DS from updating and showing you its own error. I could be way wrong but from doing form validation that's what I noticed.

hokei:
@EdwinGuzman I was on Firefox, and got that error by hitting "submit" without entering anything.  It weirdly doesn't seem to happen when I do this on `searchbar` which I thought also uses the `<form>` tag?

EdwinGuzman:
Yeah, I see the same in Chrome when I hit submit. I thought you were just tabbing through. This doesn't happen in `SearchBar` because, in this example, there are `required` attributes in the `TextInputs` components. Since that's the native attribute, you'll get the native validation popup.

hokei:
aaaaaah so it's `searchbar` that should be using the native validation.  Makes sense, thank you

src/components/Form/Form.test.tsx
bigfishdesign13: @EdwinGuzman Should I just remove this `test` file?  Or should I leave it in there with nothing in it?

src/components/Form/Form.tsx
bigfishdesign13: @EdwinGuzman I will create a new ticket for this functionality.  For now, I will release this as a basic UI component.

src/components/Form/Form.stories.mdx
bigfishdesign13: I adjusted the language to clarify how the components should be used.

src/components/Form/Form.tsx
bigfishdesign13: I changed my mind and added props for `action` and `method`.

src/components/Form/Form.stories.mdx
bigfishdesign13: I understand what you are saying about div wrappers around the form input components.  We are not there yet, but we can get there.  For now, however, we will continue to use `FormField` and `FormRow`.  I rewrote the description trying to clarify how the child components should be used.

src/components/Form/Form.tsx
bigfishdesign13: No clue.  Removed it.

src/components/Form/Form.test.tsx
EdwinGuzman: Just remove the entire file.

bigfishdesign13: I was able to write some tests for this.  The tests are fine, but I think they could be better.  For example, I was not able to get Enzyme's `parent()` method to work the way I wanted.

src/components/Form/Form.test.tsx
```diff
@@ -0,0 +1,61 @@
+import { expect } from "chai";
+import * as Enzyme from "enzyme";
+import * as React from "react";
+
+import Form, { FormRow, FormField } from "./Form";
+import TextInput from "../TextInput/TextInput";
+
+describe("Form", () => {
+  //   let wrapper: Enzyme.ShallowWrapper<any, any>;
```
EdwinGuzman: You can remove this.

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,62 @@
...
+  /** Optional ID that other components can cross reference */
+  id?: string;
+  /** Optional form `method` attribute */
+  method?: "get" | "post";
+  /** Optional modifiers array for use with BEM. See how to work with modifiers and BEM here: http://getbem.com/introduction/ */
+  modifiers?: string[];
+}
+
+// FormRow child-component
+export function FormRow({ ...props }) {
```
EdwinGuzman: For this line and line 27, you can replace this with `(props) {` since `{ ...props }` is the same as `props`.

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,62 @@
...
+    action,
+    attributes = {},
+    children,
+    className,
+    id,
+    method,
+    modifiers = [],
+  } = props;
+
+  const formID = id || generateUUID();
```
EdwinGuzman: You can also make this the default `id` value by setting `id = generateUUID(),` on line 40.

EdwinGuzman (commented):
Some minor comments but the update looks good. Mentioning @hokei and @wluisi in case they have more feedback.

hokei:
it looks good to me!

EdwinGuzman (approved):

src/components/Form/Form.tsx
bigfishdesign13: Done.

src/components/Form/Form.tsx
bigfishdesign13: Good suggestion.  Thanks.  Done.

PR #579, opened 2021-06-23
DSD-325: add Style Guide category

bigfishdesign13:
Fixes [DSD-325](https://jira.nypl.org/browse/DSD-325)

- adds `Style Guide` category to Storybook sidebar

### Front End Review:

### DIFF
--- a/src/utils/componentCategories.ts
+++ b/src/utils/componentCategories.ts
@@ -22,7 +22,7 @@ const categories = {
   },
   deprecated: {
     title: "Deprecated",
-    components: ["CheckboxOld"],
+    components: [],
   },
   feedback: {
     title: "Feedback",
--- a/src/utils/componentCategories.ts
+++ b/src/utils/componentCategories.ts
@@ -20,6 +20,10 @@ const categories = {
       "Promo",
     ],
   },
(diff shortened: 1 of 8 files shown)


src/utils/componentCategories.ts
```diff
@@ -20,6 +20,10 @@ const categories = {
       "Promo",
     ],
   },
+  deprecated: {
+    title: "Deprecated",
+    components: ["CheckboxOld"],
```
EdwinGuzman: This isn't being displayed. Is it suppose to or is this meant to just be a declaration?

bigfishdesign13: Michael inserted that when he was reworking `Checkbox`.  We opted to not deprecate the original component, so `CheckboxOld` was never actually created.  However, I think we will eventually want to have a `Deprecated` category, so I left it in there.

EdwinGuzman (approved):
I like this section in Storybook.

src/utils/componentCategories.ts
bigfishdesign13: You already approved this PR, but I removed `CheckboxOld` from the deprecated section.

PR #580, opened 2021-06-23
DSD-326: heading display sizes

bigfishdesign13:
Fixes [DSD-326](https://jira.nypl.org/browse/DSD-326)

- Adds `displaySize` prop to `Heading` for overriding default heading styles

### Front End Review:

### DIFF
--- a/src/components/Heading/Heading.stories.mdx
+++ b/src/components/Heading/Heading.stories.mdx
@@ -32,7 +32,7 @@ import Heading from "./Heading";
 
 <Description of={Heading} />
 
-A `Heading` component will render a standard HTML `<h>` tag (1-6). The heading's text can be passed in through a `text` prop or as a child. Default styles for semantic elements can be overwriten using the `displaySize` prop.
+A `Heading` component will render a standard HTML `<h>` tag (1-6). The heading's text can be passed in through a `text` prop or as a child. Default styles for semantic elements can be overwritten using the `displaySize` prop.
 
 <Preview withToolbar>
   <Story
--- a/src/components/Heading/Heading.stories.mdx
+++ b/src/components/Heading/Heading.stories.mdx
@@ -9,6 +9,7 @@ import {
 import { withDesign } from "storybook-addon-designs";
 
 import Heading from "./Heading";
+import { DisplaySizes } from "./HeadingDisplaySizes";
 
 <Meta
   title="Heading"
@@ -36,7 +37,7 @@ A `Heading` component will render a standard HTML `<h>` tag (1-6). The heading's
 
 <Preview withToolbar>
   <Story
-    name="Basic"
+    name="Heading Props"
     args={{
       id: "heading1",
       level: 1,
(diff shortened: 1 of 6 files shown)


src/components/Heading/Heading.tsx
```diff
@@ -35,6 +39,10 @@ export default function Heading(props: React.PropsWithChildren<HeadingProps>) {
 
   const baseClass = "heading";
 
+  if (displaySize) {
```
EdwinGuzman, line 42: There should also be a check to make sure that `displaySize` is one of the appropriate `DisplaySizes` options. Right now, you can add any string and the component won't complain about it (and it also doesn't affect the size).

src/components/Heading/Heading.tsx
```diff
@@ -1,30 +1,34 @@
...
-  /** Number 1-6, creating the <h*> tag */
+  /** Number 1-6; used to create the `<h*>` tag */
   level: number;
   /** Modifiers array for use with BEM. See how to work with modifiers and BEM here: http://getbem.com/introduction/ */
   modifiers?: string[];
-  /** Inner text of the <h*> element */
+  /** Inner text of the `<h*>` element */
   text?: string;
-  /** URL that header points to */
+  /** Optional URL that header points to */
```
EdwinGuzman: Can you add that adding the `url` prop will create and anchor element child.

EdwinGuzman (changes requested):
Can you add a story section to display all of the `DisplaySizes` options?

Not necessary for this PR but eventually the documentation for the `Headings with Links` section should be clearer that the `url` prop will create an anchor element for you when it renders.

src/components/Heading/Heading.tsx
bigfishdesign13: Done.

src/components/Heading/Heading.tsx
bigfishdesign13: I will create a ticket to add this functionality in the future.

bigfishdesign13:
> Can you add a story section to display all of the `DisplaySizes` options?

Done.

> Not necessary for this PR but eventually the documentation for the `Headings with Links` section should be clearer that the `url` prop will create an anchor element for you when it renders.

I added some more details to the `url` prop and to the `Headings with Links` story.

src/components/Heading/Heading.stories.mdx
```diff
@@ -62,6 +65,39 @@ A `Heading` component will render a standard HTML `<h>` tag (1-6). The heading's
...
 </Preview>
 
+## DisplaySize Styles
+
+<Story name="DisplaySize Styles" />
+
+<Preview>
+  <div>
+    <Heading
+      id="heading1"
```
EdwinGuzman: The page looks good. If you include `<Story>` inside of `<Preview>`, you'll get this display in both the "Canvas" page and the "Docs" page. Right now, the initial "Canvas" page is blank when you click on it. Something like:
```
<Preview>
  <Story name="DisplaySize Styles" />
    <div>
      <Heading
```

bigfishdesign13: Ha!  I wondered why things were not showing unnder the Canvas tab.  Now I know.  I made the change.

EdwinGuzman (approved):

PR #583, opened 2021-06-25
DSD-329: typography style guide

bigfishdesign13:
Fixes [DSD-329](https://jira.nypl.org/browse/DSD-329)

- Updates content in Style Guide `Typography`

### Front End Review:


bigfishdesign13:
@EdwinGuzman, @hokei, @nypl-wluisi:
As you look at this, imagine you are a developer who needs help defining text styles.  What details would you want to know?

EdwinGuzman (commented):
Is the intention for this to be purely a static documentation page or also have dynamic stories? By adding `<Story name="..." />`, it's creating the hierarchy of stories but they are all blank.
Screen Shot 2021-06-29 at 12 33 31 PM

What if you have something in each `Story` such as what's coming up, like:
```
<Story name="Standard Text Components">
    <p className="heading--primary" style={{ color: "black" }}>Primary</p>
   <p className="heading--secondary" style={{ color: "black" }}>Secondary</p>
</Story>
```

Might that be helpful to display for the initial viewing of the stories?

src/components/Typography/Typography.stories.mdx
```diff
@@ -0,0 +1,230 @@
...
+
+## Standard Text Components
+
+### Headings
+
+Use the NYPL Design System [Heading](https://nypl.github.io/nypl-design-system/storybook-static/?path=/docs/heading--basic) component to render a standard HTML `<h>` tag. The `level` prop can be used to set the specific `<h>` tag that you require. Each heading level has a default style and size, but those styles can be overridden using the `displaySize` prop.
+
+The display sizes are separate from semantic elements and may be used to override the default size and styling of the semantic HTML heading elements. For example, the `h1` element has the `Primary` style applied by default, but a new style can be applied through the `displaySize` prop. Any `displaySize` style can be applied to any `Heading` element.
+
+`1rem` in the Design System is equal to 16px.
```
EdwinGuzman: This should be higher up in the document and also an explanation why knowing this is important.

src/components/Typography/Typography.stories.mdx
```diff
@@ -0,0 +1,230 @@
...
+
+> <p className="heading--primary" style={{ color: "black" }}>
+>   Primary
+> </p>
+>
+> - font-size: 2.25rem `--font-size-4`
```
EdwinGuzman: I think tables would be good for these sets of data, such as

CSS rule | value | CSS variable name
---|---|---
font-size | 1.75rem | `--font-size-4`

src/components/Typography/Typography.stories.mdx
bigfishdesign13: Done.

src/components/Typography/Typography.stories.mdx
bigfishdesign13: Good suggestion.  Done.

src/components/Typography/Typography.stories.mdx
```diff
@@ -128,58 +170,75 @@ The following CSS variables are available through the DS.
...
+> When making font selections for project based components outside the DS, sans serif fonts display better on computers and mobile devices. Avoid decorative or overly stylized fonts, as they are often difficult to read even for users without visual impairments or reading Standard Text Default
 
 ## Font Colors
 
 All text within a project, unless specifically noted in mockups, should be black on a white background.
 
-In cases where specific colors must be applied to text elements, the DS provides CSS variables for the full NYPL color palette. Refer to the [Colors](https://nypl.github.io/nypl-design-system/storybook-static/?path=/docs/style-guide-colors--colors-brand) section of the Style Guide for details.
+<Preview>
+  <Story name="Font Colors">
+    <>
```
EdwinGuzman, line 189: I see what you meant about the loose text not working inside a `Preview`. I was going to suggest this but glad you found it.

src/components/Typography/Typography.stories.mdx
```diff
@@ -0,0 +1,400 @@
...
+## Font Weight
+
+The DS uses four font weights to render text. The default font-weight is 300.
+
+<Preview>
+  <Story name="Font Weight">
+    <>
+      <p
+        style={{
+          fontSize: "1rem",
```
EdwinGuzman, line 213: Are these combinations already set in scss class names? Just curious because I don't know (for font weight and font size) what the appropriate way to change it in a consuming app would be.

EdwinGuzman (approved):
The update looks great. I'm still slightly unclear how I would go about changing, let's say a `<p>` element's font weight or size in a consuming app. This goes along with using and not using CSS variables. I'm going to approve because I think we said we'd have a "how to use CSS variables" section elsewhere.

src/components/Typography/Typography.stories.mdx
bigfishdesign13: I rewrote the code for the examples to utilize the actual CSS vars.  Not sure why I didn't do that in the first place.  This will not give one example for how to use the CSS vars to override the default styles.

PR #584, opened 2021-06-28
DSD-327: breakpoints style guide

bigfishdesign13:
Fixes [DSD-327](https://jira.nypl.org/browse/DSD-327)

- Updates content in Style Guide `Breakpoints`

### Front End Review:


src/components/Breakpoints/Breakpoints.stories.mdx
```diff
@@ -0,0 +1,48 @@
...
+The following CSS variables are available through the DS.
+
+```
+--breakpoint-xl: 1280px;
+--breakpoint-large: 960px;
+--breakpoint-medium: 600px;
+--breakpoint-small: 320px;
+```
+
+<!-- <Story story={stories.breakpoints} /> -->
```
EdwinGuzman: Are you still trying to do this? If not, this page can just be a documentation page and not an interactive page (which is fine). Then you can remove the other `Breakpoints.stories.tsx` file. This does mean that for new breakpoints this would have to be updated manually but I don't think that's a big deal.

src/components/Breakpoints/Breakpoints.stories.mdx
```diff
@@ -0,0 +1,48 @@
...
+import * as stories from "./Breakpoints.stories.tsx";
+
+<Meta
+  title={getCategory("Breakpoints")}
+  decorators={[withDesign]}
+  parameters={{
+    design: {
+      type: "figma",
+      url:
+        "https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Master?node-id=10975%3A16",
```
EdwinGuzman: With this type of documentation page, the figma addon doesn't show up.

EdwinGuzman (commented):
This looks good. Maybe it'll be good to also have a "how to use CSS variables" section on this breakpoints page. It should be pointed out, however, that the DS classes and components adhere to the breakpoints so developers shouldn't have to mess with them but they should be aware of them.

This is probably not the best time for this but I found two more breakpoints specifically for components:

- `_Breadcrumbs.scss` has `$break-breadcrumbs-mobile: max-width 599px;`
- `_Input.scss` has `$break-input-mobile: 400px;`

I'm not sure if we want to include these in the document and say that these are one-offs that will be normalized later. Or maybe we should make a separate ticket to revisit the layout for those components.

src/components/Breakpoints/Breakpoints.stories.mdx
bigfishdesign13: @EdwinGuzman I was able to get the dynamic inclusion of the classes in the `.mdx` story to work and more or less formatted how I wanted, but the look and feel was different enough that I think I am going to give up on it.  I was intially concerned that losing this functionality would be a problem, but in the end, like you, I feel it is not a big deal to update manually.  I will rmeove this and the `.tsx` file.

src/components/Breakpoints/Breakpoints.stories.mdx
bigfishdesign13: OK.  Gone.

bigfishdesign13:
@EdwinGuzman Please see below...

> This looks good. Maybe it'll be good to also have a "how to use CSS variables" section on this breakpoints page. It should be pointed out, however, that the DS classes and components adhere to the breakpoints so developers shouldn't have to mess with them but they should be aware of them.

I like the idea of including info about using the CSS vars within a consuming app.  However, I will use this same type of doc format on other Style Guide pages, so the `how to` should be a separate page altogether.  I will create a ticket.

> This is probably not the best time for this but I found two more breakpoints specifically for components:
> 
> * `_Breadcrumbs.scss` has `$break-breadcrumbs-mobile: max-width 599px;`
> * `_Input.scss` has `$break-input-mobile: 400px;`
>
> I'm not sure if we want to include these in the document and say that these are one-offs that will be normalized later. Or maybe we should make a separate ticket to revisit the layout for those components.

When I was preparing to update the style guide pages, I found existing DS documentation that said the DS breakpoints should be used for general layout purposes, but component-specific styles would be prefered.  And of cousre, I can't find the page where I read that.  However, in light of that sentiment and for simplicity, I don't think we need to mention these otrher breakpoints in the style guide.  And I do agree that we should revisit the CSS/layout for these two components.  I will create tickets.

EdwinGuzman (approved):
Looks good. I know you removed the figma file link because it wouldn't show up, but maybe it'll be good to have somewhere in the mdx documentation.

PR #586, opened 2021-07-02
DSD-348: Radio helper text

bigfishdesign13:
Fixes [DSD-348](https://jira.nypl.org/browse/DSD-348)

- Adds `helperText` and `errorText` props to `Radio`

### Front End Review:


src/components/Radio/Radio.tsx
```diff
@@ -79,6 +94,11 @@ const Radio = React.forwardRef<HTMLInputElement, RadioProps>((props, ref?) => {
         {...attributes}
       />
       {labelText && showLabel && <Label htmlFor={radioID}>{labelText}</Label>}
+      {(helperText || (errored && errorText)) && showLabel && (
```
EdwinGuzman: I think you can replace `(helperText || (errored && errorText))` with `footnote` since `footnote`'s value comes from either `helperText` or `errorText`.

src/components/Radio/_Radio.scss
```diff
@@ -64,6 +74,17 @@
...
+
+      + .helper-text {
+        color: $gray-dark;
+        font-style: italic;
+      }
     }
   }
 }
+
+// .radio-helper-text {
```
EdwinGuzman: You can delete this.

EdwinGuzman (commented):
This looks good but can you add two stories, one with the helper text and one with the error text? I don't think it's strictly necessary to include but it would be helpful to have static examples (similar to the checked, error, and disabled examples) of how they appear in case updating the initial story through the Controls is missed.

src/components/Radio/Radio.tsx
bigfishdesign13: Yes, that change worked well.

src/components/Radio/_Radio.scss
bigfishdesign13: Yes!  Done.

bigfishdesign13:
@EdwinGuzman Changes made and ready for another review.

EdwinGuzman (approved):

PR #588, opened 2021-07-06
DSD-350: Select refactor

bigfishdesign13:
Fixes [DSD-350](https://jira.nypl.org/browse/DSD-350)

- Refactors `Select` to match `TextInput` formatting
- Adds `errorText`, `helperText`, `labelText`, `required` and `showLabel` props to `Select`
- Deprecates `ariaLabel`, `helperTextId`, `isRequired` and `labelId` props in `Select`

### Front End Review:


src/components/Select/Select.test.tsx
```diff
@@ -51,42 +50,152 @@ describe("Select", () => {
...
+        name="test3"
+      >
+        <option aria-selected={true}>test1</option>
+        <option aria-selected={false}>test2</option>
+      </Select>
+    );
+    expect(wrapper.find("select").prop("aria-label")).to.equal("Aria Label");
+  });
+
+  it("Renders aria-describedby", () => {
```
EdwinGuzman: What makes it render an `aria-describedby`? Is it the `helperText` prop? Add that to this test `it` description, 
"Renders aria-describedby because a ____ prop was passed".

src/components/Select/Select.tsx
```diff
@@ -13,14 +16,22 @@ export interface SelectProps {
...
   /** ID that other components can cross reference for accessibility purposes */
   id?: string;
-  /** Attribute indicating that an option with a non-empty string value must be selected */
-  isRequired: boolean;
-  /** ID of associated label */
+  /** DEPRECATED - Attribute indicating that an option with a non-empty string value must be selected */
+  isRequired?: boolean;
+  /** DEPRECATED - ID of associated label */
   labelId?: string;
+  /** Provides text for a `Label` component if `showLabel` is set to true; populates a `aria-label` sttribute if `showLabel` is set to false. */
```
EdwinGuzman: Typo in `sttribute`.

src/components/Select/Select.tsx
```diff
@@ -29,36 +40,61 @@ export interface SelectProps {
   onBlur?: (event: React.FormEvent) => void;
   /** Passes selects' current value to the React state handler */
   onChange?: (event: React.FormEvent) => void;
+  /** Will add 'aria-required: true' to input */
```
EdwinGuzman: You should add that it also adds the native `required` attribute to the input (since browsers handle that differently).

src/components/Select/Select.tsx
```diff
@@ -13,14 +16,22 @@ export interface SelectProps {
...
+  /** DEPRECATED - ID of associated HelperText */
   helperTextId?: string;
+  /** Populates the HelperErrorText for standard state */
+  helperText?: string;
   /** ID that other components can cross reference for accessibility purposes */
   id?: string;
-  /** Attribute indicating that an option with a non-empty string value must be selected */
-  isRequired: boolean;
-  /** ID of associated label */
+  /** DEPRECATED - Attribute indicating that an option with a non-empty string value must be selected */
```
EdwinGuzman, line 29: I think these labels are great to have, but maybe we can break the alphabetical list convention and put all the deprecated props at the bottom of the list and have one comment such as, "DEPRECATED - the following props are still available to use but are deprecated and will be removed in a future release", or similar.

src/components/Select/Select.tsx
```diff
@@ -80,28 +116,41 @@ const Select = React.forwardRef<HTMLSelectElement, SelectProps>(
...
+          <Label htmlFor={id} optReqFlag={optReqFlag} id={id + `-label`}>
+            {labelText}
+          </Label>
+        )}
+        <select
+          name={name}
+          id={id}
+          className={bem("selectfield", modifiers, blockName, [className])}
+          aria-required={isRequired || required}
+          required={isRequired || required}
```
EdwinGuzman: Nitpick: `isRequired || required` appears three times. Consider making a variable for this.

src/components/Select/_Select.scss
```diff
@@ -2,47 +2,62 @@
...
-  padding: var(--space-xs) var(--space-xl) var(--space-xs) var(--space-s);
-  width: 100%;
+  .selectfield {
+    -moz-appearance: none;
+    -webkit-appearance: none;
+    appearance: none;
+    background-color: var(--ui-white);
+    /* Arrow svg */
+    /* More info here: https://css-tricks.com/probably-dont-base64-svg/ */
+    background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' fill='currentColor' viewBox='0 0 24 24'%3E%3Cpath d='M10.526 12.871L.263 1.676 1.737.324 12 11.52 22.263.324l1.474 1.352L13.474 12.87a2 2 0 01-2.948 0z'/%3E%3C/svg%3E");
```
EdwinGuzman, line 12: This is neat but mentions issues with some browsers that may no longer be a problem. Is there a benefit from doing this rather than rendering an `Icon` component?

src/components/Select/Select.stories.mdx
```diff
@@ -0,0 +1,214 @@
...
+    args={{
+      id: "mySelect",
+      labelText: "What is your favorite color?",
+      showLabel: true,
+      helperText: "Choose wisely.",
+      errorText: "Waaaaaaaa!",
+      errored: false,
+      required: true,
+      disabled: false,
+      selectedOption: "Option Two",
```
EdwinGuzman: This isn't an option and leaving this here doesn't render an option that is selected, so it's better to remove this prop. This still shows up in the `Controls`.

src/components/Select/Select.stories.mdx
```diff
@@ -0,0 +1,214 @@
...
+      </Select>
+    )}
+  </Story>
+</Preview>
+
+<ArgsTable story="Select" />
+
+## Labelling Variations
+
+A Select can be rendered with or without visible labels.
```
EdwinGuzman: Can you add that we do the job of adding an `aria-label` attribute when the label isn't visible so the Select component is still accessible?

src/components/Select/Select.stories.mdx
```diff
@@ -0,0 +1,214 @@
...
+  <Select labelText="What is your favorite color?" showLabel={true}>
+    <option aria-selected={true}></option>
+    <option aria-selected={false}>Red</option>
+    <option aria-selected={false}>Green</option>
+    <option aria-selected={false}>Blue</option>
+    <option aria-selected={false}>Black</option>
+    <option aria-selected={false}>White</option>
+  </Select>
+</Canvas>
+
```
EdwinGuzman, line 124: Can you add a preface here saying that this is how the `helperText` prop gets rendered?

EdwinGuzman: Sorry, I meant this for the next example on line 127 when `showLabel` is set to true.

src/components/Select/Select.stories.mdx
```diff
@@ -0,0 +1,214 @@
...
+  >
+    <option aria-selected={true}></option>
+    <option aria-selected={false}>Red</option>
+    <option aria-selected={false}>Green</option>
+    <option aria-selected={false}>Blue</option>
+    <option aria-selected={false}>Black</option>
+    <option aria-selected={false}>White</option>
+  </Select>
+</Canvas>
+
```
EdwinGuzman, line 154: Similar idea here, it'd be good to say that `errorText` and `errored` were passed for the next examples.

src/components/Select/Select.stories.mdx
```diff
@@ -0,0 +1,214 @@
...
+## Deprecated Implementation
+
```
EdwinGuzman, line 190: Why is this a deprecated implementation? A short explanation would be good to have here.

EdwinGuzman (changes requested):
Good so far but with some comments about adding more details. Also one question about rendering the arrow svg through css rather than through the `Icon` component.

src/components/Select/Select.tsx
EdwinGuzman: This reminds me: maybe we can have a `console.warn` message if any of the deprecated props are used that says something like "These props used in the Select component will be deprecated in a future release: ...... Check the documentation for updated props to use instead."

src/components/Select/Select.test.tsx
bigfishdesign13: Good suggestion.  Done.

src/components/Select/Select.tsx
bigfishdesign13: Done.

src/components/Select/Select.tsx
bigfishdesign13: Done.

src/components/Select/Select.tsx
bigfishdesign13: @EdwinGuzman I will add a `console.warn` message if/when a deprecated prop is used.  I like that a lot.  But for Storybook, what about hiding the props in the `ArgsTable` and including a note about the deprecated props below the ArgsTable?

src/components/Select/Select.tsx
bigfishdesign13: Done.

src/components/Select/_Select.scss
bigfishdesign13: @EdwinGuzman This was how the component was initially put together.  Let me see what I can do.

bigfishdesign13: @EdwinGuzman Replacing the arrow on a `select` input is traditionally done with CSS.  I think refactoring the code to use the Icon component would be overkill.  That being said, having to copy and paste the `svg` text right into the CSS is burdensome.  In a webpack/react setup, how would I directly reference our arrow `svg`?

src/components/Select/Select.stories.mdx
bigfishdesign13: An oversight.  I changed the option values last minute and forgot to also adjust this prop.  Fixed now.

src/components/Select/Select.stories.mdx
bigfishdesign13: Done.

src/components/Select/Select.tsx
EdwinGuzman: Yes, hiding the props in Storybook and adding a note are good ideas.

src/components/Select/_Select.scss
EdwinGuzman: Ah gotcha. Let's leave this like this. It may be more complicated to get webpack involved than doing what's already here.

src/components/Select/Select.stories.mdx
bigfishdesign13: I am not sure we need to explain the details behind the variations.  Adjusting the props in the ArgsTable should suffice.

src/components/Select/Select.tsx
bigfishdesign13: Done.

src/components/Select/Select.stories.mdx
bigfishdesign13: Done.

bigfishdesign13:
@EdwinGuzman I addressed all the comments/suggestions and this PR is ready for a second review, but github still has the PR listed as `Changes Requested`.

EdwinGuzman (approved):
Yeah, the "changes requested" goes away after approving.

PR #593, opened 2021-07-09
DSD-352: Style Guide Buttons

bigfishdesign13:
Fixes [DSD-352](https://jira.nypl.org/browse/DSD-352)

- Adds Style Guide `Buttons`

### Front End Review:


EdwinGuzman (commented):
I think overall this is fine but I'm curious, this seems like it can also go into the regular Button story documentation. Should we have one file for the style guide for some components and one file for the code documentation? Is this the idea for this PR?

bigfishdesign13:
> I think overall this is fine but I'm curious, this seems like it can also go into the regular Button story documentation. Should we have one file for the style guide for some components and one file for the code documentation? Is this the idea for this PR?

@EdwinGuzman I think the regular Button story actually has too much information about the button styles.  For example, the Patterns section doesn't need to be there.  That said, I am trying to find the right balance between style guide pages and the regular story pages and the issue you noted is something I have been thinking about.  How much detail is too much for the Style Guide pages?  For `Typography` and `Iconography`, it was a bit easier as those two do not have explicitly corresponding components.

EdwinGuzman:
Alright, that makes sense. How about removing the `<Story ...>` components so that this guideline shows up as just one single page on the left sidebar and appears as more of a guideline than a list of Storybook stories. You're right about juggling how much data should go into the regular Button (or general component) story compared to a guideline on using them. But, for example, the Button component story has visual guidelines on using `Icon`s but this guideline only has written text and not a visual component.

Also, there's no guideline on using the `disabled` prop here. Is this something that's good enough in the Button component story rather than this guideline?

bigfishdesign13:
@EdwinGuzman I am game for removing the `Story` elements of the page, but I still would like users to know what is covered on the page.  To address that concern, I added a TOC at the top of the page.  Take a look and let me know what you think.

Regarding the button `disabled` state, I don't think that is necessary here.  That is coding/backend, so I think that makes more sense on the `Button` component page.

EdwinGuzman (approved):
Yes, all of that makes sense. One minor comment is that the TOC "Guidelines" should be "General Guidelines" to reflect the non-TOC heading or the heading should be changed from "General Guidelines" to "Guidelines". Either one is fine but they should be the same. I'm approving because this change is minor.

PR #594, opened 2021-07-09
Webpack/React externals dependency

EdwinGuzman:
Fixes [DSD-360](https://jira.nypl.org/browse/DSD-360)

- Allows React hooks to work by installing one version of react and react-dom in the consuming application.
- Add a combination to `webpack.config.js` and `package.json`  that allows apps to import components that use React hooks. @bigfishdesign13 can you try this in a consuming application before it's merged? So far it works but it _might_ mean that react apps need to use the same react version which is `16.14.0`. I'm trying to see if there's a way to make `16.13.0` a minimum and resolve to `16.14.0` for the app.

- Tested with `"@nypl/design-system-react-components": "git+https://github.com/NYPL/nypl-design-system#DSD-152-externals",` in package.json.

### Front End Review:


EdwinGuzman:
Okay, the last commit works! I tried it in both the LCA and DRB:
Screen Shot 2021-07-09 at 5 51 15 PM
Screen Shot 2021-07-09 at 5 52 00 PM

Some updates and why the latest one works:
1. We have to not bundle react through webpack so we have to tell it not to -- that's in the `webpack.config.js` file.
2. We have to set up `peerDependencies` and get a minimum. I choose to go with `>=16.13.0` since most apps should use at least that React version (but it is kind of arbitrary).
3. Before, `react` was a dependency but it should be a devDependency.

I a previous commit, the package worked in DRB since it uses react 16.14.0 (and DS uses 16.14.0) but LCA broke because it is using 16.13.1. So I configured it so it gets the minimum working version to use. If you run `npm ls react` or `npm ls react-dom`, you'll see the hierarchy tree of dependencies and the exact version it installs. If it installs more than one version, that's where the problem happens. With the last commit, running the command `npm ls react` in LCA gets me version 16.13.1 and in DRB it is version 16.14.0.

So it is working as expected -- it is taking the minimum version from the application and from the DS to use so there is no confusion.

This leads to: we haven't set a "this is the minimum working version of react your app needs to use DS". I think now is an appropriate time to say that we want to go with at least a version of 16.13.0. This can be updated since the minimum to use react hooks is version 16.8.0... but at this time point I think apps should use a higher version.

@wluisi What version is the Location finder using and would this approach work?

nypl-wluisi:
@EdwinGuzman I think it should be fine. Currently Location Finder is on React 16.13.1.

PR #596, opened 2021-07-12
DSD-303: Minor Story documentation update for the List component

EdwinGuzman:
Fixes [DSD-303](https://jira.nypl.org/browse/DSD-303)

- This removes the `itemGroups` and `definitions` props from the Controls section in the `List` component.
- This adds an `.mdx` file for the `List` component documentation and updates references.
- There's another round of updates for this component but that will be covered in a different PR for [DSD-361](https://jira.nypl.org/browse/DSD-361).

### Front End Review:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 ## Prerelease
 
+### Changes
+
+- Updates stories for `List` component to use MDX format.
+
 ## 0.23.1
 
 ### Adds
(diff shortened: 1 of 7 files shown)


bigfishdesign13 (changes requested):
I am seeing an issue that appears to be related to how `Icon` is being rendered.  It looks like the `iconRotation` attribute inside of `Icon` is continuously being pushed into the `attributes` array and Storybook gets hung up.  Removing the `iconRotation` prop on line 97 in `List.stories.tsx` resolves the problem in Storybook.  The issue only happens under the `Docs` tab.  The `Canvas` tab seems fine.

EdwinGuzman:
Thanks, updated. I wanted to figure out why the error was occurring but couldn't find anything. The most I could see (paraphrased) is that`useState` was being called in `useEffect` and there were rendering depth issues. It doesn't readily make sense because we're just adding a prop and this doesn't happen in other stories where `Icon` and `IconRotationType` are being used. But something to look into if it happens again.

bigfishdesign13 (approved):
Looks good.  Ship it!

PR #598, opened 2021-07-14
DSD-368: button padding

bigfishdesign13:
Fixes [DSD-368](https://jira.nypl.org/browse/DSD-368)

- changes `Button` to use `16px` for left/right padding for buttons with text
- buttons that only display an icon still use `8px` left/right padding

### Front End Review:


EdwinGuzman (approved):
Look good. I noticed that the icon isn't horizontally aligned, but that's also a visual discrepancy in production.
Screen Shot 2021-07-15 at 10 16 55 AM

PR #599, opened 2021-07-15
DSD-366: dynamic Skeleton Loader

bigfishdesign13:
Fixes JIRA ticket [DSD-366](https://jira.nypl.org/browse/DSD-366)

- Adds new props to `Skeleton Loader` to enable configurable layout

## How has this been tested?

- local build of DS

## Checklist:

## Front End Review:

- [x] View [the example in Storybook](https://deploy-preview-599--stoic-murdock-c7f044.netlify.app/)


src/components/SkeletonLoader/SkeletonLoader.tsx
```diff
@@ -1,7 +1,119 @@
...
+    modifiers = [],
+  } = props;
+
+  const imageModifiers = [];
+
+  attributes["style"] = {
+    width: width,
+  };
+
+  imageModifiers.push(imageAspectRatio);
```
EdwinGuzman: No other value gets added to the `imageModifiers` array. You can add the value in the declaration:
`const imageModifiers = [imageAspectRatio];`

src/components/SkeletonLoader/SkeletonLoader.tsx
```diff
@@ -1,7 +1,119 @@
...
+import bem from "../../utils/bem";
+import {
+  SkeletonLoaderImageRatios,
+  SkeletonLoaderLayouts,
+} from "./SkeletonLoaderTypes";
 
-function SkeletonLoader() {
-  return <div className="skeleton-loader"></div>;
+export interface SkeletonLoaderProps {
+  /** Additional attributes passed to the horizontal rule */
```
EdwinGuzman: Copy/paste issue? When would this prop ever be extended and for what values? It'll be good to have an example of this. Otherwise, I think you can remove `attributes` as a prop and add the `width` value directly to the parent `div`.

src/components/SkeletonLoader/_SkeletonLoader.scss
```diff
@@ -1,26 +1,140 @@
...
-    0px 121px, 0px 148px, 0px 175px;
-  background-repeat: repeat-y;
-  background-size: 50px 230px, /* highlight */ 300px 230px, 250px 230px,
-    250px 230px, 250px 230px, 200px 230px, 200px 230px, 200px 230px;
-  height: 485px; /* change height to see repeat-y behavior */
+@mixin skeleton-highlight {
+  // animation: shine 1s infinite;
+  animation: Pulsate 1000ms cubic-bezier(0.25, -0.5, 1, 0) infinite;
+  background: var(--ui-gray-warm-light);
+  // background-image: linear-gradient(
```
EdwinGuzman: If you don't need these, just remove.

EdwinGuzman:
This looks good but I have some comments. Also, when you first land on the story, because it is initially vertical and shows the image block and it takes up the whole width, it appears like nothing is showing up. Unless if you know to scroll or decide to scroll, it looks off.

Screen Shot 2021-07-15 at 3 13 08 PM

src/components/SkeletonLoader/SkeletonLoader.tsx
bigfishdesign13: Thanks.  Done.

src/components/SkeletonLoader/SkeletonLoader.tsx
bigfishdesign13: Ha.  Yes, that was pasted in and not changed over.  Oops.  Regardless, thanks for the suggestion.  I removed `attributes` altogether.

src/components/SkeletonLoader/_SkeletonLoader.scss
bigfishdesign13: I am notorious for leaving errant commented code blocks in files.  This has now been cleaned up.  Thanks for keeping me honest.

EdwinGuzman:
The update looks good but pinging @wluisi and @hokei for any feedback.

nypl-wluisi (commented):
@bigfishdesign13 This looks great to me. One general question. Would it make sense to have a few more stories to show how you'd pair this with specific components in the DS?

bigfishdesign13:
> @bigfishdesign13 This looks great to me. One general question. Would it make sense to have a few more stories to show how you'd pair this with specific components in the DS?

@EdwinGuzman What did you have in mind?  `SkeletonLoader` mirrors the `Card` components shown in the Figma Main file, but the actual `Card` component here does not.  Not yet, at least.  But yes, once `Card` is updated, I think it would make sense to show some pairing with `Card` and `SkeletonLoader`.

hokei:
Code wise this looks great!  

I think it'd be useful to me to have documentation or advice when to use which loaders.  Should we try to make this loader mimic the general look and shape of the data as it's loaded in (eg: if it's got a header, if it's got a button) 

If we are loading in a list of, say, cards, is it one loader per card, or is it one loader for the whole page? 

If it's one per card, what do we do if we don't know exactly how many cards we are expecting?  

What if our buttons are in a different location than the loader's?  

Which loader should we use if we're loading in, say, a page of text?  

Or a page of text with some images scattered about? 
 
The preciseness of this, especially `headingSize` makes me think that we need to implement this to be as close to the underlying content as possible, but what if we have configurations that aren't supported?  Should we be making PRs here or is it "a do your best" kind of thing?  Or should the mocks be giving us an idea of what the skeleton loader looks like?  

I'm not sure if all of these things can or should be answered in storybook - if there's a place that details best practices for these I'd be super interested to read it!

EdwinGuzman:
@bigfishdesign13 I think your last comment was actually for @wluisi since he asked that.
@hokei good point. I typically think of loaders now as what FB does. They have one generic skeleton loader for all the "card" designs that eventually get loaded.

bigfishdesign13:
> @bigfishdesign13 This looks great to me. One general question. Would it make sense to have a few more stories to show how you'd pair this with specific components in the DS?

@nypl-wluisi What did you have in mind? `SkeletonLoader` mirrors the `Card` components shown in the Figma Main file, but the actual `Card` component here does not. Not yet, at least. But yes, once `Card` is updated, I think it would make sense to show some pairing with `Card` and `SkeletonLoader`.

bigfishdesign13:
@hokei I imagined the loader as a "do your best" kind of thing, which falls in line with what @EdwinGuzman noted in his comment.  That said, recent template mockups do show what the placeholder should look like and how it should be implemented.  That is part of what was driving how this component was constructed.  But all of your questions make sense and show that guidelines for placeholder usage is necessary.

EdwinGuzman:
What are the next steps for this?
Should there be examples of the skeleton loader "as component X before it renders"?

hokei:
> Should there be examples of the skeleton loader "as component X before it renders"?

I think this would be super helpful!

bigfishdesign13:
@hokei and @EdwinGuzman
I added some examples of usage.  I did not specify exactly which components were being replicated, but you get the idea.

EdwinGuzman (approved):
I think the new examples are useful. We can add more later. Approving on my end but wait for others' feedback.

PR #604, opened 2021-07-19
DSD-363: video player

bigfishdesign13:
Fixes JIRA ticket [DSD-363](https://jira.nypl.org/browse/DSD-363)

- Add `VideoPlayer` component

## How has this been tested?

- local build of DS

## Checklist:

## Front End Review:

- [x] View [the example in Storybook](https://deploy-preview-604--stoic-murdock-c7f044.netlify.app/)


src/components/VideoPlayer/VideoPlayer.tsx
```diff
@@ -0,0 +1,81 @@
...
+    attributes = {},
+    className,
+    iframeTitle,
+    modifiers = [],
+    videoId,
+    videoType,
+  } = props;
+
+  let embedCode;
+  let iframeTitleFinal;
```
hokei: Unsolicited Typescript advice - it's not a mistake, per se, but when possible, it's good practice to structure code to allow yourself to use `const` instead of `let`.

For example,  here you could avoid assigning `embedCode` here and just go `const embedCode = ` on L54, and change L29 to `iframeTitle = videoType === VideoPlayerTypes.Vimeo ? "Vimeo video player"  : "YouTube video player"` (and do a similar thing to videoSrc).

src/components/VideoPlayer/VideoPlayer.tsx
```diff
@@ -0,0 +1,81 @@
...
+  let videoSrc;
+
+  {
+    aspectRatio && modifiers.push(aspectRatio);
+  }
+
+  switch (videoType) {
```
nypl-wluisi: @bigfishdesign13 Wouldn't it be better to just have the consuming app pass the video src url, and then we wouldn't need to hardcode the links to vimeo and youtube?

bigfishdesign13: @nypl-wluisi If the direct URL for the video on YouTube or Vimeo was passed, how would you handle the embed code?  The structure of the video viewing page and the video embed page are different.

View: `https://www.youtube.com/watch?v=0fpNksg7-2I`
Embed: `https://www.youtube.com/embed/0fpNksg7-2I`

src/components/VideoPlayer/VideoPlayer.tsx
bigfishdesign13: @hokei Thanks for the suggestions. I made some changes.

src/components/VideoPlayer/VideoPlayer.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+    });
+
+    it("Uses iframeTitle prop to add custom title attribute to iframe", () => {
+      expect(container.find("iframe").prop("title")).to.equal(
+        "My custom iframe title."
+      );
+    });
+  });
+
+  describe("aspect ratio", () => {
```
EdwinGuzman, line 75: You're missing the 16:9 aspect ratio.

src/components/VideoPlayer/VideoPlayer.tsx
```diff
@@ -0,0 +1,82 @@
...
+      allow="accelerometer; autoplay; clipboard-write; encrypted-media; fullscreen; gyroscope; picture-in-picture"
+      allowFullScreen
+      {...attributes}
+    ></iframe>
+  );
+
+  let errorMessage;
+  if (!videoType && !videoId) {
+    console.warn("VideoPlayer requires the `videoType` and `videoId` props");
+    errorMessage =
```
EdwinGuzman: I think these messages should be more user-friendly. These are great for devs to debug what is wrong, but a user can't make much sense out of something like 
Screen Shot 2021-07-21 at 2 53 10 PM

src/components/VideoPlayer/_VideoPlayer.scss
```diff
@@ -0,0 +1,31 @@
...
+    height: auto;
+    padding: var(--space-s);
+  }
+
+  iframe,
+  object,
+  embed {
+    height: 100%;
+    left: 0;
+    margin: atuo;
```
EdwinGuzman: Typo. Hmmm I'm surprised `stylelint` didn't pick this up. Maybe that workflow is broken.

src/components/VideoPlayer/VideoPlayer.tsx
EdwinGuzman: You could try to parse out the video id from the specific video type URL. But, I think it makes sense to pass in just the ID because there are specific parameters that will make video embed renderings consistent throughout different applications.

EdwinGuzman (commented):
I noticed an obvious visual change when changing the aspect ratio in Storybook for the Youtube video but not for the Vimeo video. Is this expected? It appears as if the aspect ratio for the Vimeo type is not working, although the padding around the video itself is changing.

I'm not sure how easy it would be to do, but it's be nice to have the `videoId` prop in Storybook update as the `videoType` prop is updated in the Controls. Right now, if you change the type from Vimeo to Youtube, it'll render an errored Youtube video because the ID isn't right. Or perhaps have a list of "Try out these ids for the Vimeo player and these ids for the Youtube player."

src/components/VideoPlayer/VideoPlayer.tsx
bigfishdesign13: I will make message in the browser more friendly, but I will leave the `console.warn` as is.

src/components/VideoPlayer/VideoPlayer.test.tsx
bigfishdesign13: Fixed.

src/components/VideoPlayer/VideoPlayer.tsx
bigfishdesign13: Fixed.

src/components/VideoPlayer/_VideoPlayer.scss
bigfishdesign13: Fixed.

bigfishdesign13:
> I noticed an obvious visual change when changing the aspect ratio in Storybook for the Youtube video but not for the Vimeo video. Is this expected? It appears as if the aspect ratio for the Vimeo type is not working, although the padding around the video itself is changing.

@EdwinGuzman The way that the videos are rendered when the aspect ratio of hte player wrapper is changed is controlled by YouTune and Vimeo.  YouTube shows letterboxing and Vimeo does not.  And without the letterboxing, it does look a bit odd.  It's up to the consuming app to set the correct aspect ratio for the video that is being added.

EdwinGuzman:
Alright, that makes sense. I think that should be included in the Docs section so that devs explicitly know that they need to use the appropriate aspect ratio for the image they want to render. I think it'll be good to also include the answer to @wluisi 's question in the documentation, e.g. that devs "should use the ID and video type props instead of the full embed URL because ..."

src/components/VideoPlayer/VideoPlayer.stories.mdx
```diff
@@ -0,0 +1,79 @@
...
+        {...args}
+        videoId={args.videoType === "vimeo" ? "474719268" : "nm-dD2tx6bk"}
+      />
+    )}
+  </Story>
+</Preview>
+
+<ArgsTable story="VideoPlayer" />
+
+**Example Vimeo IDs:** `474719268`, `491405018`, `493795778` (square)<br />
```
EdwinGuzman, line 69: What does the "(square)" here mean? To use with the square aspect ratio? I don't see a difference for this video in this aspect ratio.

bigfishdesign13: Yes, that video is square.  And I just realized that the videoId input in the argsTable does not override the videoId that is being set in the code.  Let me see what I can do about that.

bigfishdesign13: There was a bug.  I fixed it.

bigfishdesign13:
@EdwinGuzman I added some additional language to the Docs verbiage.

src/components/VideoPlayer/VideoPlayer.stories.mdx
```diff
@@ -0,0 +1,86 @@
...
+
+# Video Player
+
+| Component Version | DS Version |
+| ----------------- | ---------- |
+| Added             | `0.23.2`   |
+
+<Description of={VideoPlayer} />
+
+The `VideoPlayer` component is used to embed a `Vimeo` or `YouTube` video player on a page. The component requires both the `videoType` and `videoId` props. You can extract the videoId from the video's URL. **IMPORTANT:** Do not try to enter the full video URL as the `videoId` value.
```
EdwinGuzman, line 41: With this lastest commit and this update about not entering a full video URL, I think it'll be good to check the `videoId` prop to verify it's not a full URL (possible ideas: either check that it's a certain pattern for vimeo and youtube, or check that "vimeo.com" and "youtube.com" are not in the string). If a full URL is passed instead of the video id, then render the error message that says the player has not been configured correctly.

bigfishdesign13: I added validation for videoId and also added a new check in the tests.

EdwinGuzman (approved):

PR #616, opened 2021-07-23
DSD-345: enum validation props documentation

EdwinGuzman:
Fixes JIRA ticket [DSD-345](https://jira.nypl.org/browse/DSD-345)

- The original task of the ticket was to add prop validation to the `displaySize` prop in the `Heading` component. It turns out there is type validation for that prop when writing tests and in consuming applications that use Typescript. Non-typescript apps don't get the type validation benefit but it's hard to say that it is the full responsibility of the NYPL DS to add prop type validation in that scenario. This is not to say we don't support non-Typescript apps, we do, but we can only do so much.
- Currently, prop type validation is not supported in MDX. This is the same experience that a non-typescript consuming app has. It _appears_ as if MDX prop type validation is not working but it's just that typescript is not supported in MDX.
- `type` variables are preferred over `enum` variables and eventually, we can make that move. But for now, I outlined why we should stick to string `enums`.

## How has this been tested?

Prop validation was checked in the Library Card App (typescript), Locations app (non-typescript), and Storybook files (MDX).

### Checklist:

### Front End Review:


bigfishdesign13 (approved):
Looks good.

PR #619, opened 2021-07-27
DSD-384, DSD-385: Jest/React Testing Library & Github Actions CI

EdwinGuzman:
Fixes JIRA ticket [DSD-384](https://jira.nypl.org/browse/DSD-384) and also [DSD-385](https://jira.nypl.org/browse/DSD-385). This also partially fixes [DSD-386](https://jira.nypl.org/browse/DSD-386).

## This PR does the following:
- This PR removes Mocha, Chai, Sinon, and Enzyme in favor of Jest and React Testing Library.
- I thought I'd have to update the Github Actions CI but I didn't and it currently passes -- this is ticket DSD-385.
- I did include `jest-axe` for all components but I forgot it was a separate JIRA ticket so now it's bundled with this large PR (sorry). I won't be closing DSD-386 until I get snapshot tests in. That will be a separate PR.

Some notes:
- there are a couple of `TODOS:` because I couldn't easily find a way to convert some tests. I'll take care of that in a separate PR.
- there are also tests that didn't seem right or didn't test the right thing. I removed some of those. I noticed some components need better tests but that's a separate conversation and ticket.
- some tests aren't in the RTL testing methodology but they can be updated later. I wanted to do a 1-to-1 conversion first before writing better tests.
- This is a big PR so we can go through the files together or have a meeting about this (similar to the MDX presentation/walkthrough a while back).

## How has this been tested?

Ran the tests locally and Github Actions CI test script runs and passes.

### Checklist:

- [ ] All new and existing tests passed.

### Front End Review:

Nothing to review on Storybook.

- [ ] View [the example in Storybook]()

PR #620, opened 2021-07-27
DSD-386: Three examples of snapshot testing

EdwinGuzman:
Fixes JIRA ticket [DSD-386](https://jira.nypl.org/browse/DSD-386)

- This adds three examples of snapshot testing through jest.
- To see an error, the easiest way is to update `Notification.tsx` and update the `<aside>` to `<div>`s and then run `npm test`. You'll see the failure and error message.
- These are just examples of how we can add more tests to make sure that the DOM is what we expect and that changes are intended.
- Snapshot files should be committed so don't forget to add them. They automatically get generated.
- Each snapshot test includes further instructions at the top of the file (including how to update tests).
- Part of the ticket was to add `jest-axe` but I prematurely added it to PR #619 .

## How has this been tested?

Locally.

### Checklist:

### Front End Review:


EdwinGuzman:
@bigfishdesign13 added a section to the README on snapshot testing.

bigfishdesign13 (approved):
This looks good to me.

hokei:
This is really great!  My outstanding question is around when someone would use a snapshot test -- because this is a component library and the visuals matter here, should all components have a snapshot test so we're alerted when something changes?

EdwinGuzman:
We wouldn't have a snapshot test for components are that more like guidelines, such as the `Autosuggest`, but yes, if we can get snapshot tests on every component where we expect a specific UI (including with different props), that'd be great.

hokei (approved):

PR #621, opened 2021-07-27
DSD-380/DSD-381: Implementing TSDX and Jest tests

EdwinGuzman:
Fixes JIRA ticket [DSD-380](https://jira.nypl.org/browse/DSD-380) and [DSD-381](https://jira.nypl.org/browse/DSD-381).

- Implements TSDX.
- TSDX is used to set up configurations for a new project, but since we already have configurations, this is used to move over to their more opinionated configurations -- many of which we already use.
- This points to branch PR #620 as an add-on to the jest tests. This allows ticket DSD-381 to be closed. Nothing really had to be done to get it to work since, in the existing PR #619 Jest is working.
- A lot of files were updated because they were linted. Either the linting is slightly different or those files were not linted before.
- The `example/` directory is from TSDX and I haven't really updated it or shown how to use it. I think this can be a separate ticket and PR.
- I updated the README to include documentation on running tests and passing a file path to test a specific file.
- Removed `lodash`. Right now, including it and importing it to use in a component doesn't seem to work. It doesn't load correctly in the es module compiled file. Instead of `lodash`, TSDX does an optimization to import `lodash-es` but this doesn't work for me in a consuming application. To fix this, I added a separate `range` function in the repo as a utility function. This also reduces the repo size since we don't explicitly need `lodash`.
- As a reminder, this uses rollup instead of webpack.
- For more information on all the updates, check the TAD.

## How has this been tested?

- Tested all `npm` scripts locally.
- Tested using `npm link` in a consuming application (in my case, the Library Card App).
- Tested this branch through `package.json` reference: `"@nypl/design-system-react-components": "git+https://git@github.com/NYPL/nypl-design-system.git#DSD-380/tdsx",`
- Tested the Storybook instances to make sure it's the same as before.

### Checklist:

### Front End Review:


bigfishdesign13 (approved):
Looks good.

EdwinGuzman:
@hokei @wluisi this is a big update so I'd like either of you to also review before this gets merged.

.github/workflows/ci.yml
```diff
@@ -15,12 +15,14 @@ jobs:
       CI: true
     steps:
       - uses: actions/checkout@v2
-      - name: Use Node.js 13.x
+      - name: Use Node.js 12.x
```
hokei, line 18: why the downgrade?

.storybook/main.js
```diff
@@ -29,24 +27,21 @@ module.exports = {
...
-      module: {
-        ...config.module,
-        rules: customRules,
-      },
-    };
+  typescript: {
+    // Type-check stories during Storybook build.
+    check: true,
+  },
+  webpackFinal: async (config, { configType }) => {
```
hokei, line 34: So because Storybook is the way it is, it's still being built with webpack while the app is being built with rollup, is that correct?  Is there a danger of running into a problem where stuff is going to work in storybook but not in a consuming app?

example/index.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
```
hokei, line 1: What is this for?  How should someone use the example?

package.json
```diff
@@ -6,36 +6,41 @@
...
-    "prepare": "npm run dist",
+    "start": "tsdx watch",
+    "build": "tsdx build && npm run generate-sass-resources",
+    "test": "tsdx test --passWithNoTests",
+    "test:watch": "tsdx test - --watch",
+    "lint": "tsdx lint",
+    "prepare": "npm run build",
+    "size": "size-limit",
+    "analyze": "size-limit --why",
     "generate-sass-resources": "gulp",
```
hokei, line 33: Was there also a desire to remove gulp?  Is that happening in a different chunk of work?

.github/workflows/ci.yml
EdwinGuzman: This was from a separate PR that got merged into this one. There was a linting issue with the 13.x version but it didn't always happen. I'll eventually bump it up since I haven't seen that issue again.

.storybook/main.js
EdwinGuzman: Yeah, that's the case. I don't think so -- both webpack and rollup transpile/compile the code and minify it. You can still choose how to compile/minify and how to export it. The final output of both, however, is basically the same. I haven't seen any differences between what webpack and rollup produce but that's a good thing to keep an eye on.

example/index.html
EdwinGuzman: This directory got added by TSDX and was discussed how it would be a small playground app where one can test out their component(s).  This is its own conversation on how to best use it so I left it alone but will discuss and add a "how-to" for this in a separate ticket.

package.json
EdwinGuzman: Yes, I haven't found a way to do what gulp does in webpack or rollup that does _two_ processes for the scss files. What I've done so far is try a roll plugin that runs in the same pipeline as the `scss` plugin. It does correctly read an input file and combines all the relevant code into a final `.scss` file, but it can't do that _and_ create the minified `dist/styles.css` file. There is also a separate ticket for this (existing before this conversion) that will be relevant to this.

hokei: Yeah - also I wonder if this is going to change after the chakra migration, it's not clear how much of our custom SCSS we're going to keep.

hokei (approved):

hokei:
This looks good! In trying to build out my mental model of the app:  

Before, we were writing the code in typescript, which would then get built into javascript and then we ran the tests on the js code.  This was ... kind of dumb.  That seems to have been fixed here, we no longer need to run `tsc` before running tests, is that right?  

So we're doing separate build paths for storybook and npm publish, and the storybook is still reading the scss files directly while a consuming app is reading the compiled CSS.  

Still better than what we had before!

EdwinGuzman:
Yeah, and I noticed recently that quite a few of our apps with webpack and mocha follow the similar pattern of compiling with `tsc` into a /lib folder and then running mocha on test files in _that_ directory.

PR #622, opened 2021-07-28
DSD-388: Hero Campaign variant

bigfishdesign13:
Fixes JIRA ticket [DSD-388](https://jira.nypl.org/browse/DSD-388)

- Updates `Hero` to include `Campaign` variant
- Updates `Hero` stories to use `.mdx` format

**Note**
Having five differing variants within a component like this is a lot and moving the stories from `.tsx` to `.mdx` proved to be a big deal due to the validation built into the component. Because of that, I loosened up the restrictions and moved to console warnings rather than throwing errors.  The component is now more composable, but the changes are backward compatible.

## How has this been tested?

- DS Storybook local build
- DS test app

### Checklist:

### Front End Review:

- [x] View [the example in Storybook](https://deploy-preview-622--nypl-design-system-dev.netlify.app/)


src/components/Hero/_Hero.scss
```diff
@@ -189,3 +189,70 @@
...
+  display: flex;
+  justify-content: center;
+  margin-bottom: var(--space-xxl);
+  min-height: 300px;
+  overflow: visible;
+  position: relative;
+
+  @include breakpoint($breakpoint-medium) {
+    // flex-flow: row nowrap;
+  }
```
EdwinGuzman: You can remove this whole block.

src/components/Hero/_Hero.scss
```diff
@@ -189,3 +189,70 @@
...
+      top: var(--space-xxl);
+    }
+
+    + .placeholder {
+      flex: 0 1 200px;
+      margin-left: var(--space-xl);
+    }
+  }
+
+  /* stylelint-disable */
```
EdwinGuzman, line 244: Is this stylelint related to the comment below?

src/components/Hero/Hero.tsx
```diff
@@ -48,84 +40,137 @@ export default function Hero(props: React.PropsWithChildren<HeroProps>) {
...
+    modifiers.push("warning");
   }
 
-  const backgroundImageStyle = backgroundImageSrc
-    ? { backgroundImage: "url(" + backgroundImageSrc + ")" }
-    : { backgroundColor: backgroundColor };
+  let backgroundImageStyle = {};
+  if (heroType === HeroTypes.Primary) {
+    backgroundImageStyle = backgroundImageSrc
+      ? { backgroundImage: "url(" + backgroundImageSrc + ")" }
```
EdwinGuzman: Use string interpolation for this and on line 117.

src/components/Hero/Hero.tsx
```diff
@@ -48,84 +40,137 @@ export default function Hero(props: React.PropsWithChildren<HeroProps>) {
...
-        ? "foregroundColor"
-        : "backgroundColor";
-      throw new Error(
-        `Please provide both foregroundColor and backgroundColor to Hero, only got ` +
-          receivedColor
-      );
-    }
+    contentBoxStyling = {
+      color: foregroundColor,
+      backgroundColor: backgroundColor,
```
EdwinGuzman: Not necessary but you can do the javascript shorthand for objects where the key and value names are the same. So you can do
```
contentBoxStyling = {
  color = foregroundColor,
  backgroundColor,
}
```

src/components/Hero/_Hero.scss
bigfishdesign13: No.  That comment was already in the code when I started working on this.  The stylelink error that is being bypassed was related to the ordering of the classes in the file.

src/components/Hero/Hero.tsx
bigfishdesign13: Cool.  Did not know that.

src/components/Hero/_Hero.scss
bigfishdesign13: Done.

src/components/Hero/Hero.tsx
bigfishdesign13: That's the original code, but it's now cleaned up.

src/components/Hero/Hero.tsx
bigfishdesign13: Done.

src/components/Hero/Hero.tsx
```diff
@@ -48,84 +40,137 @@ export default function Hero(props: React.PropsWithChildren<HeroProps>) {
...
-
-        {image}
-
-        {subHeaderText}
-      </div>
+      {heroType === HeroTypes.Campaign ? (
+        <div
+          className={bem("content", [], heroBaseClass)}
+          style={contentBoxStyling}
+        >
```
EdwinGuzman: This whole block is good but you're repeating the `<div clasName={bem.....` section twice. What you can do is keep that as is and only update the internal DOM by either 1) moving the logic check on line 147 down or 2) declare that logic check above and assign the child blocks to a variable. So something like
```
const childrenToRender = heroType === HeroTypes.Campaign ? <>{image}.....</> : <>{heroType.....}</>;
```

src/components/Hero/Hero.test.tsx
```diff
@@ -2,14 +2,14 @@ import { expect } from "chai";
 import * as Enzyme from "enzyme";
 import * as React from "react";
 
-import Placeholder from "../Placeholder/Placeholder";
+// import Placeholder from "../Placeholder/Placeholder";
```
EdwinGuzman: You can remove this.

src/components/Hero/Hero.stories.mdx
```diff
@@ -0,0 +1,274 @@
...
+    }}
+  >
+    {args =>
+      (args.heroType === HeroTypes.Primary && (
+        <Hero
+          {...args}
+          heading={
+            <Heading
+              level={1}
+              id={"1"}
```
EdwinGuzman: Nits: You can remove the curly braces when the prop is a string literal.

src/components/Hero/Hero.stories.mdx
```diff
@@ -0,0 +1,274 @@
...
+            blockName={"hero"}
+          />
+        }
+        subHeaderText={
+          <div className="hero__body-text">
+            Explore our collection of hundreds of online resources and
+            databases. Use our free online content to help with your research,
+            whether it's finding a single article, tracing a family tree,
+            learning a new language, or anything in between.
+          </div>
```
EdwinGuzman: Since this gets repeated a few times (along with other props), make it a variable you can assign to all the different instances.

src/components/Hero/Hero.tsx
bigfishdesign13: That's actually how I had it initially.  That formatting need some type of wrapper around the children (`<div>` or `<>`), so I wrote it that way that I did to remove unneeded tags.  However, now I see that using `<>` does not insert extra tags.  That being the case, I went that route and dropped all the logic into a variable assignment.

src/components/Hero/Hero.test.tsx
bigfishdesign13: Done.

src/components/Hero/Hero.stories.mdx
bigfishdesign13: Got it.  Thanks for the suggestion.  I cleaned up what I could.

src/components/Hero/Hero.stories.mdx
bigfishdesign13: Done.

EdwinGuzman (approved):
Update looks good.

bigfishdesign13:
@EdwinGuzman 
I merged with `development` and updated the testing.  Please take one more look.  Thanks.

EdwinGuzman (approved):
The update looks good. As we progress further into react testing library, we eventually want to not rely as much on `utils.container.querySelector` and instead test the text/visuals that get rendered.

PR #624, opened 2021-07-28
DSD-346: Breadcrumbs - removing a custom SCSS breakpoint

EdwinGuzman:
Fixes JIRA ticket [DSD-346](https://jira.nypl.org/browse/DSD-346)

- Removes the `$break-breadcrumbs-mobile` custom breakpoint variable. Instead of using a `max-width 599px`, it uses the already configured `$breakpoint-medium` variable and updates the styling rules so it's more mobile-first.

## How has this been tested?

Through storybook with different responsive layouts.

### Checklist:

### Front End Review:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 ### Changes
 
 - Updates the README to include information on the production, development, and "preview" Storybook documentation instances.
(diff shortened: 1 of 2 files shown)


bigfishdesign13 (approved):
This works well.  Curious why this was implemented this way in the first place.

EdwinGuzman:
Not sure... I'm not sure if everything is strictly "mobile-first" but that's an large audit we can do later on.

PR #627, opened 2021-07-29
DSD-122: Fixing bem function bug that crashes Storybook

EdwinGuzman:
Fixes JIRA ticket [DSD-122](https://jira.nypl.org/browse/DSD-122)

- Adds logic check in the `bem` function so that empty arguments are not a problem. This doesn't occur when developing components because Typescript catches this issue and warns you that the argument is wrong or empty. _But_, in the Storybook Controls panel, it doesn't rely on Typescript so an empty `modifiers` argument crashes the function and Storybook.

## How has this been tested?

In the Accordion component, in the Controls panel click on the "Set Object" button for the "modifiers" prop. Since it sets it to an empty object, it throws an error and breaks. Tested this by making sure this is no longer the case.

Screen Shot 2021-07-29 at 3 36 11 PM

### Checklist:

### Front End Review:

- [x] View [the example in Storybook](https://deploy-preview-627--nypl-design-system-dev.netlify.app/)


bigfishdesign13 (approved):
This looks good.  However, I've seen this same type of bug for other props throughout the DS.  I will start to make a list to tyrack where I am seeing the issues.

PR #630, opened 2021-08-02
DSD-400: Jest storybook integration

EdwinGuzman:
Fixes JIRA ticket [DSD-400](https://jira.nypl.org/browse/DSD-400)

- Adds a plugin to Storybook to integrate it with our current Jest unit tests.
- This doesn't add any new tests but only combines a Storybook component with its unit tests. They can be found in the addons panel under the "Test" tab:

Screen Shot 2021-08-02 at 4 06 04 PM

- While you can't see the test code itself, you can see the description of a test. Altogether, you can see how a component is being tested and it will help us figure out if we're missing any tests for a component.

## How has this been tested?

Locally through Storybook.

### Checklist:

### Front End Review:


bigfishdesign13 (commented):
All in all, this works well, btu I am curious about the formatting of the test messages that are generated.  Do we have any countrol over that formatting?  Right now, the text from the `describe` and `it` statements are strung together without separators.  If possible, it would be nice to insert some type of separators.

image

EdwinGuzman:
As far as I could tell, no, there's no way to format it unless if we update a test file's `describe` and `it` clauses. The way it auto-format it could also let us know if any tests don't make sense or if we should update the language so it's consistent.

The last commit makes the tests show up in the preview build. The development workflow automatically picks up the test name but for production, it needs to be explicit. In the last commit, you can see the integration for every test file. It's a bit annoying but after this first step, it should be smoother for newer components.

bigfishdesign13 (approved):
Thanks for the explanation.  Off we go.

PR #631, opened 2021-08-02
DSD-398: video player text elements

bigfishdesign13:
Fixes JIRA ticket [DSD-398](https://jira.nypl.org/browse/DSD-398)

- adds text elements to `VideoPlayer` component.

## How has this been tested?

- DS local build

### Checklist:

### Front End Review:

- [x] View [the example in Storybook](https://deploy-preview-631--nypl-design-system-dev.netlify.app/)


src/components/VideoPlayer/VideoPlayer.test.tsx
```diff
@@ -71,6 +71,34 @@ describe("VideoPlayer", () => {
     });
   });
 
+  describe("text elements", () => {
```
EdwinGuzman, line 74: 👍

src/components/VideoPlayer/VideoPlayer.tsx
```diff
@@ -92,7 +104,16 @@ export default function VideoPlayer(
       {errored ? (
         <span dangerouslySetInnerHTML={{ __html: errorMessage }} />
       ) : (
-        embedCode
+        <>
+          {headingText && <Heading level={2} text={headingText} />}
+          <p>{descriptionText}</p>
```
EdwinGuzman: This should also be a conditional since `descriptionText` is optional. Otherwise, a `<p>` tag will always be rendered even if no `descriptionText` was passed.

bigfishdesign13: Good catch.  A foolish oversight on my part.  Conditional logic has been added.

EdwinGuzman (approved):
Not foolish, it happens!

PR #634, opened 2021-08-05
DSD-396: DatePicker component

EdwinGuzman:
Fixes JIRA ticket [DSD-396](https://jira.nypl.org/browse/DSD-396)

- This ticket _only_ implements the `react-datepicker` plugin as part of the DS `DatePicker` component.
  - This specific ticket/PR doesn't include styling updates.
  - There's a separate ticket to test its functionality/data flow in a consuming application.
- This is based on the component spec document found in the ticket.

## Outstanding questions

These came up while developing which the spec or TAD didn't cover:
- The Figma document says that the date range display is the default but the spec says that the `dateRange` prop is `false` by default. There's a mismatch there.
- Right now, the current date is used as the initial display/value date. Should there be a `value` prop that allows developers to set an initial date? I imagine there's a use case for this somewhere.
- For a date range display, it wasn't clear if the `showLabel` prop would remove the required label text _and_ the "From" and "To" label text. I made it so `showLabel` does remove all three text values from the rendered output. Perhaps we always want to keep the "From" and "To" label in the display?

## How has this been tested?

Locally through Storybook.

### Checklist:

### Front End Review:

- [ ] View [the example in Storybook](https://deploy-preview-634--nypl-design-system-dev.netlify.app/?path=/story/form-elements-datepicker--date-picker)


hokei:
is it possible that people will want to pass different text instead of "From" and "To"?

hokei:
In the "date range" option, on Chrome, when I focus into "From", the "to" floats underneath.  no worries if this is related to stying, want to make sure there's not a functionality bug here

src/components/DatePicker/DatePicker.stories.mdx
```diff
@@ -0,0 +1,150 @@
...
+<Canvas withToolbar>
+  <Story
+    name="DatePicker"
+    args={{
+      dateType: DatePickerTypes.Full,
+      dateFormat: "yyyy-MM-dd",
+      dateRange: false,
+      name: "visit-dates",
+      labelText: "Select the date you want to visit NYPL",
+      helperText: "Note that the Library may be closed on Sundays.",
```
hokei, line 50: does it make sense to merge generic `HelperText` and `HelperTextFrom`?  

(Eg: I imagine `HelperTextFrom` would have a smaller character limit than a generic `HelperText`, no?)

EdwinGuzman:
> is it possible that people will want to pass different text instead of "From" and "To"?

It's possible but for now I'm following the design spec.

> In the "date range" option, on Chrome, when I focus into "From", the "to" floats underneath. no worries if this is related to stying, want to make sure there's not a functionality bug here

Yeah, I get that as well. It is a styling issue.

EdwinGuzman:
Ignore Tugboat failed check for this PR as it's unrelated.

src/components/DatePicker/DatePicker.stories.mdx
EdwinGuzman: Do you mean `HelperTextTo` instead of `HelperTextFrom`? It's possible but having two separate fields will be good for clearer instructions for what dates are expected, such as the figma file where it shows an early "From" date example and a later "To" date example:

Screen Shot 2021-08-06 at 2 28 15 PM

bigfishdesign13: I had a discussion with UX about using just one helper text value, but an explicit decision was made to use two separate helper text values.  Edwin nailed the reasoning in his explanation.

src/components/DatePicker/DatePicker.stories.mdx
```diff
@@ -0,0 +1,150 @@
...
+      helperText="From this date."
+      helperTextTo="To this date."
+      errorText="Please select a valid date range."
+    />
+  </Story>
+</Canvas>
+
+## Month Date Input
+
+It's possible to select dates by the skipping the day value and only selecting
```
bigfishdesign13: Typo: 
...dates by `the` skipping the day...

Remove first `the`.

src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,259 @@
...
+  id,
+  dateRange,
+  showLabel,
+  labelText,
+  className = "",
+  children,
+}) =>
+  dateRange ? (
+    <fieldset id={`dateRange-${id}`} className={className}>
+      {showLabel && <legend>{labelText}</legend>}
```
bigfishdesign13: Similar to `TextInput`, when `showLabel` is set to `false`, appropriate accessibility attributes need to be added.  You might consider dropping `fieldset` altogether and using an altered `div` solutions for both.

**Visible Label**
```html
<div id={`dateRange-${id}`} className={className} role="group" aria-labelledby="the-label">
  <Label id="the-label">The Custom Label</Label>
  ...
```

**Hidden Visible**

```html
<div id={`dateRange-${id}`} className={className} role="group" aria-label="The Custom Label">
```

src/index.ts
```diff
@@ -6,6 +6,7 @@ export { default as Button } from "./components/Button/Button";
 export { ButtonTypes } from "./components/Button/ButtonTypes";
 export { default as Card } from "./components/Card/Card";
 export { default as Checkbox } from "./components/Checkbox/Checkbox";
+export { default as DatePicker } from "./components/DatePicker/DatePicker";
```
bigfishdesign13, line 9: Make sure to also export `DatePickerTypes`.

bigfishdesign13 (commented):
> The Figma document says that the date range display is the default but the spec says that the dateRange prop is false by default. There's a mismatch there.

The Figma file will eventually be changed to match what you are establishing in the React component: single date as default.  The original Figma designs were put together a while ago and need to be updated.

> Right now, the current date is used as the initial display/value date. Should there be a value prop that allows developers to set an initial date? I imagine there's a use case for this somewhere.

Yes, I like this suggestion.  I will update the spec doc.

> For a date range display, it wasn't clear if the showLabel prop would remove the required label text and the "From" and "To" label text. I made it so showLabel does remove all three text values from the rendered output. Perhaps we always want to keep the "From" and "To" label in the display?

The thinking was to hide only the main label and leave the `From` and `To` labels alone, as they really are necessary and it would be difficult for consuming apps to add custom labels for just the date field.  That said, if you think all three should be hidden based on that prop, I am open to ideas.  Let me hear what you are thinking.

src/components/DatePicker/DatePicker.stories.mdx
hokei: Oh I'm wondering if there should be three `helperText` fields like, a `HelperTextFrom` `HelperTextTo` and generic `HelperText` 

Eg: for the example that Edwin posted, perhaps there's a need for a bigger `HelperText` underneath that says "Please note that books in the last ten years have not yet been digitized" or something 


It doesn't make sense to me that you can have a generic `HelperText` and a `HelperTextTo` together, but not `HelperText` and `HelperTextFrom` together.

EdwinGuzman: For the `dateRange` option, the `helperText` prop essentially acts like what you described as the "`helperTextFrom`" prop. It's implied so it can be confusing. For the moment, and based on the component spec, we're only going to use the `helperText` and `helperTextTo` props. I do agree that for some context/use-cases we may want another helper text.

We're continuously going to add addendums to the spec document and TAD document. If we all agree on this update, then we can add more information for this `helperTextFrom` prop update.

This is a new process so I'll set up a meeting to talk about this but to be brief:
- this PR covers the spec/functional requirements
- any updates or concerns should be added in the spec and TAD document in an "Addendum" or "Update" section
- the update would be a new ticket and PR

EdwinGuzman:
> The thinking was to hide only the main label and leave the From and To labels alone, as they really are necessary and it would be difficult for consuming apps to add custom labels for just the date field. That said, if you think all three should be hidden based on that prop, I am open to ideas. Let me hear what you are thinking.

Okay, let's only hide the main label. I'll update this PR.

src/components/DatePicker/DatePicker.tsx
EdwinGuzman: I'll stick to the `fieldset` since that's in the spec for the date range. That sounds good for the accessible label. Will update.

EdwinGuzman:
Updated. For the accessible label in the date range and the `showLabel` prop, I decided to always render the `Legend` element and add a "sr-only" class rule that applies the `screenreader-only` SCSS mixin. This way, the date range's fieldset/legend will always be accessible and when the `showLabel` prop is false, it will only hide it from the screen but not screenreaders.

src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,260 @@
...
+  id,
+  dateRange,
+  showLabel,
+  labelText,
+  className = "",
+  children,
+}) =>
+  dateRange ? (
+    <fieldset id={`dateRange-${id}`} className={className}>
+      <legend className={showLabel ? "sr-only" : ""}>{labelText}</legend>
```
bigfishdesign13: This logic is backwards.  If `showLabel` is true, then the `sr-only` class should not be added.  That class should only be added when `showLabel` is false.

bigfishdesign13 (changes requested):
All in all, this looks good.  From what I can see, just one small logic change is needed.

src/components/DatePicker/DatePicker.tsx
EdwinGuzman: 🤦

bigfishdesign13 (approved):
With that minor logic change, this looks good to me.  I will approve, but feel free to get another review from Crystal or Bill before you merge.

EdwinGuzman:
@hokei @wluisi any other feedback on this before it's merged?

EdwinGuzman:

Link:
- https://pr634-09bxmd3py6koeulla7i9oqglexgw8h89.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/611c332d378c5fa79d8286ae

PR #640, opened 2021-08-12
DSD-414: Chakra UI import

EdwinGuzman:
Fixes JIRA ticket [DSD-414](https://jira.nypl.org/browse/DSD-414)

- Installs Chakra UI and related npm packages (from the [chakra](https://chakra-ui.com/docs/getting-started) installation guide).
- Configures Storybook so that Chakra components get the theme and can render correctly.
- Adds a `theme` directory for all the theme-related configs and also a DS provider.
- This `DSProvider` is based on the `ChakraProvider` and simply exports it with the NYPL-specific theme.
- This NYPL-specific theme still needs to be completed but the starting point is there.
- This adds a "global" and "colors" theme example based on properties and values that already exist in the DS, except now they are a part of the Chakra theme that will be used.
- That's it! No actual components are imported and exported because that's the next step.

## How has this been tested?

I tested this with the NYPL Library Card app by exporting Chakra's `Button` component as `ChakraButton` (since DS already has a `Button` component) in the index file:

```
...
export { default as DSProvider } from "./theme/provider";
export { Button as ChakraButton } from "@chakra-ui/react";
```

_Note that this wasn't committed as it's just an example to show it end-to-end from package to app._
In the LCA, I then rendered the Chakra button (as "Hello Button"):

```
<DSProvider>
  <ChakraButton colorScheme="blue">Hello Button</ChakraButton>
</DSProvider>
```

Screen Shot 2021-08-12 at 12 36 56 PM

### Checklist:

### Front End Review:


README.md
```diff
@@ -142,6 +142,35 @@ Some CSS rules in the Design System, such as the universal focus styling and the
 </div>
 ```
 
+4. Import the `DSProvider` for Chakra-based components
+
+As we start to integrate and compose DS components from Chakra components, we need to get the styles for those components. In order to do so, we need to wrap all the DS components that internally use with Chakra with a simple provider component.
```
hokei: nit: looks like a typo in `wrap all the DS components that internally use with Chakra`

README.md
```diff
@@ -142,6 +142,35 @@ Some CSS rules in the Design System, such as the universal focus styling and the
...
+
+For now, only a few components will need this provider but as we migrate more components to use a similar pattern of styling, the provider will prove to be useful.
+
+You will want to import the `DSProvider` at the root of your application so that:
```
hokei: From a consumer's point of view, it won't be clear what needs `DSProvider` and what doesn't - should we make a breaking change bump here and advise everyone to import it going forward?

EdwinGuzman: We aren't 1.0 yet so almost every version bump is a breaking change (I've noticed a lot already in the LCA since that hasn't been updated in a while).

I do agree it's not clear what components need it and what components don't. Eventually, all components will need it and all apps will have to use the `DSProvider`. We can make that a migration process or just say from this point on that the apps need (when they might actually not). I prefer to make it explicit so app consumers understand why they need the provider component and for what components. This means we'll have to be more explicit but that's okay.

src/theme/foundations/colors.ts
```diff
@@ -0,0 +1,82 @@
+import { Colors } from "@chakra-ui/react";
+
+// Colors taken from "../../styles/01-colors/*.scss"
```
bigfishdesign13: With these values being hardcoded, this adds another layer that we need ot keep an eye on if/when color values are updated.  We should consider using a method to export the SASS vars from the files in `01-colors` into a javascript object.  Additionally, the color SCSS files use a method to convert the SASS vars to the standard CSS vars.  That same functionality should be applied to all of the style files.  These, of course, are tasks for additional tickets.

README.md
bigfishdesign13: I agree.  The statement should apply to all consuming apps, making it easier for devs to digest and then move forward.

EdwinGuzman:

Link:
- https://pr640-1cyzfp5o9mouqwm0dsk0lxabk2hgoxgl.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/611c3252378c5f0acf828697

hokei (approved):
I think we should merge this since we're now working on a lot of things that depend on this and we don't want it to get too out of sync.  It looks good to be moving on with.  

The SCSS colors thing is a thing we need to deal with at some point.  

Screen Shot 2021-08-24 at 2 51 37 PM

I'm seeing this in the tugboat preview - any idea why it's happening? Can fix it here if it's quick, or in a follow up PR.

EdwinGuzman:

Link:
- https://pr640-1cyzfp5o9mouqwm0dsk0lxabk2hgoxgl.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/611c3252378c5f0acf828697

EdwinGuzman:
> The SCSS colors thing is a thing we need to deal with at some point.

Yes, the colors theme has been updating in a later PR.

> I'm seeing this in the tugboat preview - any idea why it's happening? Can fix it here if it's quick, or in a follow up PR.

Yes, I just figured out how to fix this so I'll it soon in _this_ PR.

nypl-wluisi (approved):
This looks good to me!

EdwinGuzman:

Link:
- https://pr640-1cyzfp5o9mouqwm0dsk0lxabk2hgoxgl.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/611c3252378c5f0acf828697

PR #642, opened 2021-08-16
DSD-418: Chakra/DS Button component

EdwinGuzman:
Fixes JIRA ticket [DSD-418](https://jira.nypl.org/browse/DSD-418)

- Uses Chakra's `Button` component to compose the DS `Button` component.
- This is a WIP but setting the PR to view the preview Storybook instance.

### Checklist:

### Front End Review:

- [x] View [the example in Storybook]()


EdwinGuzman:
Update from meeting:
- we'll remove SCSS and move forward with css-in-js
- there's going to be a variant for the red donate button style
- We need to work on the global styles and themes for spacing and other rules ASAP
- I'm going to stick with `buttonType` over `variant` for the prop type (for now)
- for a separate PR we'll document how to write style props and what props are available for DS component and why there's a need to restrict adding style props to Chakra components that are re-exported by DS

I'll continue working on this PR based on the above.

@wluisi @hokei @bigfishdesign13

hokei:
> there's going to be a variant for the red donate button style

Does that mean we're purposely not exposing the other props, or does that mean we're still going to expose them but make it clear in the docs that those other props should be used as a last resort?  @bigfishdesign13 

(PS we don't have to make clear what the chakra props are in the stories, we can just link to Chakra and be like, "these are available but use them as a last resort")

EdwinGuzman:
> Does that mean we're purposely not exposing the other props, or does that mean we're still going to expose them but make it clear in the docs that those other props should be used as a last resort? @bigfishdesign13

For now, none of the style props are actually available for the DS `Button` component since that's our custom component. In this case, we're not exposing the other props but it's not intentional. We just haven't fully decided on whether we should use a `{...rest}` approach for any and all other props or keep using the attribute. Maybe we can hold off on that decision until more components are built and we see, through implementation, what the best approach is.

> (PS we don't have to make clear what the chakra props are in the stories, we can just link to Chakra and be like, "these are available but use them as a last resort")

I'm already doing this (except for the "use them as a last resort" message) in the other PR the re-exports Chakra components directly from the DS - https://pr640-1cyzfp5o9mouqwm0dsk0lxabk2hgoxgl.tugboat.qa/?path=/docs/chakra-layout-box--box
There's a link to the style props documentation in the Story above.

EdwinGuzman:
This is no longer a WIP and is ready for a full review.

EdwinGuzman:

Link:
- https://pr642-0kvfxcldirptzwjy7ykojz9esatodof9.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/611d668c331d463cd063d4de

EdwinGuzman:

Link:
- https://pr642-0kvfxcldirptzwjy7ykojz9esatodof9.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/611d668c331d463cd063d4de

hokei (approved):
This looks good to me!  Most of my questions are around CSS going forward, which seems out of scope for this PR.

src/components/Button/Button.tsx
```diff
@@ -16,84 +22,82 @@ interface ButtonProps {
...
+  } else {
+    variant = getVariant(buttonType);
   }
+
+  styles = useStyleConfig("Button", { variant });
+
+  return (
+    <ChakraButton
+      id={id}
+      className={bem(baseClass, [], "", [className])}
```
nypl-wluisi, line 90: Are we keeping `className` and bem related around temporarily for backwards compatibility reasons?

nypl-wluisi:
This might be pre-existing, but the naming of the props `buttonType` and `type` is a bit confusing. `buttonType` prop controls visual changes for the component itself, where as `type` sets the actual `type` attribute on the button HTML element?

nypl-wluisi:
Search Bar component has some styling issues, related to this button change: https://pr642-0kvfxcldirptzwjy7ykojz9esatodof9.tugboat.qa/?path=/story/components-form-elements-searchbar--basic

search-bar-bug

hokei:
> Search Bar component has some styling issues, related to this button change: 

This is the kind of thing I'd love to figure out if we can catch using automated tests - would a snapshot catch this?

src/components/Button/_Button.scss
```diff
@@ -1,48 +1,12 @@
...
 @mixin button-base {
-  @include space-inset-xs;
-
-  border-radius: 2px;
+  border-radius: 2px !important;
   cursor: pointer;
   display: flex;
   justify-content: center;
-  padding: var(--space-xs) var(--space-s);
+  padding: var(--space-xs) var(--space-s) !important;
```
bigfishdesign13, line 8: What is necessitating the `!important`?

src/theme/components/button.ts
```diff
@@ -0,0 +1,91 @@
+const Button = {
+  // style object for base or default style
+  baseStyle: {
+    borderRadius: "2px",
+    cursor: "pointer",
+    color: "ui.white",
+    justifyContent: "center",
+    padding: "var(--space-xs) var(--space-s)",
```
bigfishdesign13: This is something that is dependent on another PR, but should you replace the `space` CSS vars with the Chakra data objects before merging this?

EdwinGuzman:

Link:
- https://pr642-0kvfxcldirptzwjy7ykojz9esatodof9.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/611d668c331d463cd063d4de

bigfishdesign13 (commented):
I think this looks good as we start shifting away from the SCSS.  I just have some minor questions.

src/components/Button/Button.tsx
EdwinGuzman: Yes, and also because we haven't completely moved away from bem so maybe this should stick around until we no longer need it.

EdwinGuzman:
> This might be pre-existing, but the naming of the props buttonType and type is a bit confusing. buttonType prop controls visual changes for the component itself, where as type sets the actual type attribute on the button HTML element?

Yeah, it's confusing but you're right about how it works.

EdwinGuzman:
> This is the kind of thing I'd love to figure out if we can catch using automated tests - would a snapshot catch this?

Yes, exactly. We just now need to implement more snapshot tests in order to catch these issues.

src/components/Button/_Button.scss
EdwinGuzman: This is only really being used in the `Accordion` component to style the input. Some of Chakra's inline css rules take precedence so it needs the `!important`. But once we update the `Accordion` component, this can be removed.

src/theme/components/button.ts
EdwinGuzman: This is fixed in #648 so I'll leave it there to see the progression of the Chakra integration for this component.

src/components/Button/Button.tsx
nypl-wluisi: Should we consider adding a deprecation flag comment for this and other parts that we're planning on getting rid of? Something like:

```js
// @deprecated since version 2.3.
// Will be deleted in version 3.0.
 ```

Comment (bigfishdesign13) [reply src/components/Button/Button.tsx:90]: @nypl-wluisi Doing this on a code level seems untenable.  If we are going to flag things like this, what about adding a `Deprecated` page to the `Documentation` section in Storybook?

Comment (nypl-wluisi) [reply src/components/Button/Button.tsx:90]: Yeah, that's a good point. We can discuss more, but we need a list somewhere of all props and other things that we plan on no longer using with the switch to Chakra. And then we also need to decide if something like `bem` and `className` are things we are going to move away from, which I think we are?

Comment (EdwinGuzman) [reply src/components/Button/Button.tsx:90]: Yes, I think we should move away from `bem` with this Chakra update but it'll be a slow transition. I do like that the current function with `className={bem(baseClass, [], "", [className])}` (where classname = "newclass") will render `"button newclass". We might have to replace this with 
```
className={`${baseClass} ${className}`}
```

EdwinGuzman:
Snapshot testing catching visual differences:

I reverted the Button component, added a snapshot test for the `SearchBar` component, un-reverted the Button changes, and then ran the tests again and it did fail:

Screen Shot 2021-08-25 at 3 40 27 PM

Ignore the second error with the div/button -- this one fails because the `id` is autogenerated and it will always be different and we need to fix this separately. But you can see that the `button` component has new classes so we can infer that maybe it also changed visually. This is helpful but we need to set time aside first to write snapshot tests in order to catch more of these issues in the future. I need to revert this snapshot test since the `Icon` component is its own issue, but I'll fix the button so it looks right in the `SearchBar` component.

EdwinGuzman:
Updated.

EdwinGuzman:
@wluisi  the latest commit should fix the svg issue in the header nav buttons, but it's global so it should fix it everywhere else.

PR #644, opened 2021-08-17
DSD-424: DatePicker component additional props

EdwinGuzman:
Fixes JIRA ticket [DSD-424](https://jira.nypl.org/browse/DSD-424)

- Adds five additional props to the `DatePicker` component based on its spec document and TAD addendum sections.

## How has this been tested?

With tests and through Storybook.

### Checklist:

### Front End Review:


EdwinGuzman:
Ignore the tugboat failed check.

src/components/DatePicker/_DatePicker.scss
```diff
@@ -5,14 +5,22 @@ fieldset {
 
   &.datePicker {
     @include fieldset-reset;
+    margin-bottom: var(--space-m);
```
bigfishdesign13: This margin is not necessary, as we are expecting consuming apps to use `Form` to layout form fields.

src/components/DatePicker/_DatePicker.scss
```diff
@@ -5,14 +5,22 @@ fieldset {
...
   }
 }
 
 .datePicker {
+  margin-bottom: var(--space-m);
```
bigfishdesign13: Like above, this margin is not necessary, as we are expecting consuming apps to use `Form` to layout form fields.

src/components/DatePicker/_DatePicker.scss
```diff
@@ -5,14 +5,22 @@ fieldset {
 
   &.datePicker {
     @include fieldset-reset;
+    margin-bottom: var(--space-m);
   }
 }
 
 .datePicker {
```
bigfishdesign13: Why `.datePicker` and not `.date-picker` to follow the formatting of other class names?

src/components/DatePicker/_DatePicker.scss
```diff
@@ -5,14 +5,22 @@ fieldset {
...
+  margin-bottom: var(--space-m);
+
   .sr-only {
     @include screenreader-only();
   }
 
+  .form-row,
+  .form-field {
+    margin-bottom: var(--space-xxs);
```
bigfishdesign13: You can set this to `margin-bottom: 0`.

src/components/DatePicker/DatePicker.tsx
```diff
@@ -61,10 +62,26 @@ export interface DatePickerProps extends DatePickerWrapperProps {
...
   /** Populates the `HelperErrorText` component in the "To" `TextInput` component. */
   helperTextTo?: string;
+  /** The initial date value. This must be in the mm/dd/yyyy format. */
+  initialDate?: string;
+  /** The initialTo date value used for date ranges.
+   * This must be in the mm/dd/yyyy format.
+   */
+  initialDateTo?: string;
+  /** The minimum date value that applies to both input fields.
+   * This must be in the mm/dd/yyyy format.
```
bigfishdesign13, line 78: @EdwinGuzman Do you have any concerns that there could be confusion around formatting this string and using the `dateFormat` prop?

src/components/DatePicker/DatePicker.stories.mdx
```diff
@@ -47,9 +47,13 @@ import { DatePickerTypes } from "./DatePickerTypes";
       dateType: DatePickerTypes.Full,
       dateFormat: "yyyy-MM-dd",
       dateRange: false,
+      initialDate: "1/1/2021",
```
bigfishdesign13, line 50: It would be nice to have `initialDate` and `initialDateTo` listed consecutively in the story ArgsTable.

bigfishdesign13 (changes requested):
The additions and functionality look good.  Just a few comments about the CSS.

src/components/DatePicker/DatePicker.tsx
EdwinGuzman: Sort of. The visual representation is the `dateFormat` which is displayed in the input field, but any js-related date information needs to be in the specific format so that the js date parser can work properly. It should still be possible to say "I only want the year to display" but have to pass in an initial date of "1/1/2021". It is a bit annoying but the workaround is more annoying to build in.

src/components/DatePicker/_DatePicker.scss
EdwinGuzman: That's true but I noticed that when I added the `HelperErrorText` component to the whole `DatePicker` component, that it removed the margin-bottom rules. That's why I added it here but I can remove.

src/components/DatePicker/_DatePicker.scss
EdwinGuzman: Yeah, I'll update.

src/components/DatePicker/_DatePicker.scss
EdwinGuzman: Shouldn't we use css variables?

src/components/DatePicker/_DatePicker.scss
EdwinGuzman: Note that some rules like `.datepicker-calendar` come from the React DatePicker package.

src/components/DatePicker/_DatePicker.scss
bigfishdesign13: In this case, we want to kill the margin that is being added by `.form-row`/`.form-field`.  If we had `--space-none`, we could use that, but I don't think that is necessary.  

However, as I think about his more, the `.form-row`/`.form-field` classes are supposed to remove the bottom margin with they are `last-child`.  I will look at this again and see if the CSS in `Form` needs to be tweaked instead.

bigfishdesign13: Yes, the styles in `Form` need to be updated.  For now, set this to 0.

bigfishdesign13 (approved):
The updates look good.

PR #648, opened 2021-08-19
DSD-417: Extending Chakra's base theme with NYPL's theme.

EdwinGuzman:
Fixes JIRA ticket [DSD-417](https://jira.nypl.org/browse/DSD-417)

- Further extends Chakra's base theme with more styles from NYPL.

Some notes:
- The learning curve may take some time.
- After digging into the documentation and working on this, I really like this approach.
- How to use breakpoints is a larger topic and is not included here but the values themselves are a part of the new custom theme.
- The pattern for using sizing values is different and can be confusing. For colors, you can do something like `bg: ui.black`, but for padding, you can't do `padding: space.2`. Instead, you would do something like `py: 2; px: 4` to target padding top/bottom and padding left/right. The number values map to the object key in `"src/themes/foundations/spacing.ts"`. Here, 2 maps to `var(--space-xs)` and 4 to `var(--space-s)`. Check the updated `Button` theme file for an example. 
- This can help reduce a lot of SCSS and CSS vars if we decide on using Chakra to generate those CSS vars for us.
  - At the moment, Chakra will use everything in the custom theme files to generate CSS variables but it will prepend `--chakra`. This means that there may be some overlap. For example, there are two variables for the color `"#C60917"`:

```
--chakra-colors-brand-primary
--brand-primary
```

### How did Chakra generate `--chakra-colors-brand-primary`?

In `"src/themes/foundations/colors.ts"`, the colors theme object is declared as:

```jsx
const colors = {
  // ...
  ui: {
    black: "#000",
    // ...
  },

  brand: {
    primary: "#C60917",
    secondary: "#760000",
    // ...
  }
};
```

Chakra will drill down the object using every key name to create the CSS variable. So for black, it will generate `--chakra-colors-ui-black`. This applies to all styles such as font size, borders, spacing, etc.

This brings up the question: should we rely on Chakra for CSS variables and update the prefix value or keep the existing CSS variables?
- I say we use Chakra and update the prefix to be `"nypl"` so it can generate something like `--nypl-colors-ui-black` or `--nypl-colors-brand-primary`. We could use the `--nypl` namespace although the current CSS variables don't.
- This would mean updating all the values in the documentation but it'll be clearer where they are coming from.

## How has this been tested?

Locally updating values and styles, both through style props and through the `Button` component style object.

### Checklist:

### Front End Review:


EdwinGuzman:

Link:
- https://pr648-nsat5uwplr8mohhpkdvcckgm95dt4zve.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/611ed2a1378c5fc906850c2a

src/theme/foundations/colors.ts
```diff
@@ -1,82 +1,186 @@
...
+ * All Brand colors can be found in Storybook:
+ *   https://nypl.github.io/nypl-design-system/storybook-static/?path=/story/style-guide-colors--colors-brand
+ */
+
+// Reusable variables:
+const grayLight = "#E0E0E0";
+const grayxxlight = "#FAFAFA";
+const brandPrimary = "#C60917";
+const brandSecondary = "#760000";
+const brandPrimarySecondaryObj = {
```
bigfishdesign13: Why not a simpler name?  `brandObj`?

src/theme/foundations/spacing.ts
```diff
@@ -0,0 +1,79 @@
...
+ *     1        |  --space-xxs
+ *     2        |  --space-xs
+ *     4        |  --space-s
+ *     6        |  --space-m
+ *     8        |  --space-l
+ *     12       |  --space-xl
+ *     16       |  --space-xxl
+ *     24       |  --space-xxl
+ */
+export const spacing = {
```
bigfishdesign13, line 34: For typography, "0" aligns with "1rem," which is our base.  What about simplifying the spacing object similarly and using an array of consecutive numbers to match up with the origina NYPL DS spacing values?  Also, can we use negative numbers for the key names?

bigfishdesign13 (commented):
Wow, a lot to digest here.  But it's all good stuf.  I like the idea of moving in a direction to allow Chakra to generate the CSS vars.  That being said, incorporating Chakra is introducing more CSS variables and methods for creating those variables – variables and methods that we can't eliminate – so we need to take it upon ourselves to simplify the NYPL DS side of things.  We can start relying on Chakra for the CSS vars, but we need to start thinking about how to have fewer points that need to be updated manually.

src/theme/foundations/colors.ts
EdwinGuzman: sure

src/theme/foundations/spacing.ts
EdwinGuzman: This spacing file is not for typography-related style rules. Check out the `typography.ts` file where it does have `"0": "1rem"` and it also does include negative font-size values.

EdwinGuzman:

Link:
- https://pr648-nsat5uwplr8mohhpkdvcckgm95dt4zve.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/611ed2a1378c5fc906850c2a

src/theme/foundations/spacing.ts
bigfishdesign13: Yes, that is exactly wehat I noticed in typography.ts.  For consistency, and in an attempt to align with the current naming convension, I was curious if the spacing object names could be reworked in a similar manner.

EdwinGuzman:

Link:
- https://pr648-nsat5uwplr8mohhpkdvcckgm95dt4zve.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/611ed2a1378c5fc906850c2a

src/theme/foundations/spacing.ts
EdwinGuzman: I see what you're saying. I took those values from the existing `.css` files which do match up nicely with Chakra's default values. I think it is technically possible to update the spacing values to be aligned with the typography values and its convention but it seems like it starts on different values on purpose.

src/styles/base/_typography.scss
```diff
@@ -5,35 +5,7 @@
...
 }
-
-.blockquote {
-  border-left: solid 3px var(--ui-gray-light);
-  font-style: italic;
-  margin-left: 1em;
-  padding-left: 1em;
-}
-
-.hr {
```
hokei, line 21: where are these things being set now?

src/theme/foundations/colors.ts
```diff
@@ -1,82 +1,186 @@
...
-   * where a distinct named ui or brand color doesn't quite work.
+   * The following color scales are useful for color edge cases where a
+   * distinct named ui or brand color doesn't quite work. These are used by
+   * Chakra and have been generated through https://themera.vercel.app/.
+   * While these colors are available, we don't recommend to use them but
+   * rather to stick with the utility colors declared above.
+   * @Note If you do need to use a color variant, please contant the DS team.
    */
-  // based on ui/gray/medium
+
```
hokei, line 111: Are they here because Chakra will sometimes use them if nothing else has been set?  

If so, does it make sense to make these colours something very visible, so that any implementer will know that there's a colour being used that DS hasn't set yet?

src/theme/foundations/colors.ts
```diff
@@ -1,82 +1,186 @@
...
+    400: "#969696",
+    500: "#808080",
+    600: "#666666",
+    700: "#4D4D4D",
+    800: "#333333",
+    900: "#1A1A1A",
   },
-  // based on ui/brand/primary
+  /** based on ui.brand.primary = #C60917 */
   red: {
```
hokei, line 126: trying to understand why these shades are the way they are? My gut would make me think that red.100 is #C60917 but it's not?

hokei:
This looks really good!  I think I have the same question you've brought up  before - right now we're implicitly using a little bit of the Chakra CSS strategy, and also the old NYPL CSS strategy.  We should definitely pick one, and I /think/ we should move towards just using the Chakra themeing strategy, since that's already well documented and a lot of people have thought through it.  How would we move towards getting UX-approved values for all of [these](https://github.com/NYPL/nypl-design-system/pull/648/files#diff-ab700234f903526148292ed1a8e5ca93fb54e7a0874c8713cfec5726895d04cbR121)?  The consolidation of the colors started happening because everyone was using a slightly different shade of NYPL red, etc.  So how do we make sure that we can have 8 red colors, but there's one of them that's the "Brand Primary"?  (Am I even interpreting this in the right way?)

src/styles/base/_typography.scss
EdwinGuzman: These were actually not being used... so I just removed them.

There is a component that renders an `hr` but the class name is actually `.horizontal-rule`.

src/theme/foundations/colors.ts
EdwinGuzman: Yeah, these colors are set by default by Chakra. I couldn't find a way to remove or disable them, but if we did disable them, it would break any of these references that Chakra uses internally (it probably calls `blue.500` somewhere in one of its own components).

I added a dev note above that we recommend _not_ to use these colors but to stick to the utility ones above this section which are also declared in Figma.

src/theme/foundations/colors.ts
EdwinGuzman: Same, it was very confusing at first. The shades are all created from https://themera.vercel.app/ based on the initial color. So `#C60917` returns this palette but doesn't include the color itself. I briefly touched upon that in the description above and since we don't recommend using these colors, then it _might_ not be a problem.

EdwinGuzman:
> The consolidation of the colors started happening because everyone was using a slightly different shade of NYPL red, etc. So how do we make sure that we can have 8 red colors, but there's one of them that's the "Brand Primary"? (Am I even interpreting this in the right way?)

Yes, you are interpreting it correctly. The NYPL red "brand primary" color is being set [here](https://github.com/NYPL/nypl-design-system/pull/648/files#diff-ab700234f903526148292ed1a8e5ca93fb54e7a0874c8713cfec5726895d04cbR14). Then it is being set in the `colors.brand.primary` object property which in turn returns the CSS var `--chakra-colors-brand-primary` or can be used as `color: "brand.primary"` in JS styling objects.

I think what we should do is recommend to use those primary values and not use Chakra's `red.500` (or similar) values -- that we recommend against it because NYPL already has the specific red to use and doesn't use shade variations of a color.

hokei:
> I think what we should do is recommend to use those primary values and not use Chakra's red.500 (or similar) values -- that we recommend against it because NYPL already has the specific red to use and doesn't use shade variations of a color.

OK so a real-world situation - A while back, I got feedback from the A11Y report that [the link color](https://github.com/NYPL/nypl-design-system/pull/648/files#diff-ab700234f903526148292ed1a8e5ca93fb54e7a0874c8713cfec5726895d04cbR36) doesn't have enough contrast against the light grey background that we were using.  Would that be a situation where I should use something like [a darker blue](https://github.com/NYPL/nypl-design-system/pull/648/files#diff-ab700234f903526148292ed1a8e5ca93fb54e7a0874c8713cfec5726895d04cbR182) from Chakra?  I think it'd be super tempting to do something like that.  (The solution we ended up doing, which was to use the link_hover color by default and the research colour on hover, is really unsatisfying to me haha)

hokei (approved):

bigfishdesign13:
> I think what we should do is recommend to use those primary values and not use Chakra's `red.500` (or similar) values -- that we recommend against it because NYPL already has the specific red to use and doesn't use shade variations of a color.

@EdwinGuzman & @hokei 
What you are saying is spot on.  In theory, no one should ever use these generated colors.  NYPL has a specific color palette and everything within the NYPL landscape should adhere to that palette.  They are called `brand` colors for a reason.  So yes, we should make recommendations in the docs – IN ALL CAPS, if need be – and do our best to drive the point home.

bigfishdesign13:
> OK so a real-world situation - A while back, I got feedback from the A11Y report that [the link color](https://github.com/NYPL/nypl-design-system/pull/648/files#diff-ab700234f903526148292ed1a8e5ca93fb54e7a0874c8713cfec5726895d04cbR36) doesn't have enough contrast against the light grey background that we were using. Would that be a situation where I should use something like [a darker blue](https://github.com/NYPL/nypl-design-system/pull/648/files#diff-ab700234f903526148292ed1a8e5ca93fb54e7a0874c8713cfec5726895d04cbR182) from Chakra? I think it'd be super tempting to do something like that. (The solution we ended up doing, which was to use the link_hover color by default and the research colour on hover, is really unsatisfying to me haha)

Yes, very tempting to grab one of those genrated colors.  But this is exactly what we want to avoid.  If a color needs to be changed then UX should be notified and the process should get rolling to alter an existing palette color or to generate a new one.

PR #661, opened 2021-08-30
DSD-199: RadioGroup component - Part 1

EdwinGuzman:
Fixes JIRA ticket [DSD-199](https://jira.nypl.org/browse/DSD-199)

- Adds the `RadioGroup` component composed from Chakra's `RadioGroup` component.
- This is a WIP because it depends on the `Radio` component which should be completed first and should be composed from Chakra's `Radio` component.
  - Doing this will clean up the current `Radio` component and follow better patterns.

Edit: Update, there will be another PR that builds on top of the newly refactored `Radio` component. That will be the second part to this PR.

@bigfishdesign13 I'll add comments in this PR which I'm concerned about.

### Checklist:

### Front End Review:


src/components/RadioGroup/RadioGroup.stories.mdx
```diff
@@ -0,0 +1,66 @@
...
+      errorText: "error!",
+      showLabel: true,
+      errored: false,
+      required: false,
+      disabled: false,
+    }}
+  >
+    {(args) => (
+      <RadioGroup {...args}>
+        <Radio value="2" labelText="Radio 2" showLabel />
```
EdwinGuzman: One reason why I want to work on `Radio` first is because the `showLabel` prop is required but it doesn't make sense to _not_ show it. You won't know the value you are selecting.

EdwinGuzman: Also, I think we should stop using the `labelText` prop and make it an explicit child `<Radio value="3">Radio 3</Radio>`

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,114 @@
...
+import HelperErrorText from "../HelperErrorText/HelperErrorText";
+import generateUUID from "../../helpers/generateUUID";
+import { spacing } from "../../theme/foundations/spacing";
+
+export enum RadioGroupLayouts {
+  Vertical = "vertical",
+  Horizontal = "horizontal",
+}
+
+type DirectionMap = { [key: string]: "column" | "row" };
```
EdwinGuzman: These are the values for Chakra's `direction prop`. I think we should use "column"/"row" instead of "vertical"/"horizontal" to be consistent with Chakra's convention. If we're moving to other prop verb usage such as "isDisabled", "isErrored", then it make sense to update this.

@bigfishdesign13 let me know what you think and if you agree, the spec doc should be updated.

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,114 @@
...
+    errored = false,
+    required,
+    disabled,
+  } = props;
+  const baseName = "RadioGroup";
+  const footnote = errored ? errorText : helperText;
+  const direction = directionMap[layout];
+  const newChildren = [];
+
+  React.Children.map(children, (child: React.ReactElement, i) => {
```
EdwinGuzman: I mentioned in the spec doc that we should enforce consumers to use the DS `Radio` component within `RadioGroup` because we can easily control all the styling. Here, if the `RadioGroup` is errored or disabled, it's easy to pass the same "disabled" and "errored" props to the `Radio` component. If it was a native HTML element, "errored" will not work and we won't get the desired style.

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,114 @@
...
+      }
+      newChildren.push(React.cloneElement(child, { ...attributes }));
+    }
+  });
+
+  return (
+    <fieldset id={`radio-group-${id}`} className={`${baseName} ${className}`}>
+      <legend className={showLabel ? "" : "sr-only"}>
+        <span>{labelText}</span>
+        <span>{required ? "required" : "optional"}</span>
```
EdwinGuzman: Typically, the "Required"/"Optional" is part of the `<label>` element but this component uses a `legend`. So we have to now manually redo some of the work from the `Label` component.

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,114 @@
...
+      </legend>
+      <ChakraRadioGroup
+        name={name}
+        defaultValue={defaultValue}
+        onChange={onChange}
+      >
+        <Stack direction={[direction]} spacing={spacing.s}>
```
EdwinGuzman: The `Stack` component is useful here for the layout and is Chakra's pattern so that's why I'm using it here. Note that this does make it easy to complete the requirement of using 1rem (`spacing.s`) for spacing.

src/styles.scss
```diff
@@ -50,6 +50,12 @@
 @import "./components/TextInput/_TextInput.scss";
 @import "./components/Autosuggest/_Autosuggest.scss";
 
+fieldset {
```
EdwinGuzman, line 53: This is temporary but we need better global styles for these rules.

EdwinGuzman:

Link:
- https://pr661-wigyt8r48uanxriut09u83euyjsnw1i0.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/612d4282baa70d3d329337b8

src/components/RadioGroup/RadioGroup.stories.mdx
bigfishdesign13: We included `showLabel` for both `Radio` and `Checkbox` to allow for scenarios where the inputs were used to select a row of tabular data or an image.  In both cases, you wouldn't need a visible label, but you would want something that a screen readers could see.

Regarding `labelText`, to better align with Chakra, yes, I think it would be OK to have that come in as a child.  However, while we don't currently have any designs that require a `Radio` without a label, I think we should retain `showLabel`.  Unless there is a way to do this in the Chakra component that I am not seeing.

src/components/RadioGroup/RadioGroup.tsx
bigfishdesign13: Yes, again for better alignment with Chakra, let's move to `column` and `row`.  On a related note, both `Card` and `SkeletonLoader` use `horizontal` and `vertical` values for their `layout` props.  We'll need to discuss whether this same verbiage change would be applicable there.

src/components/RadioGroup/RadioGroup.tsx
bigfishdesign13: I do agree with you and would love to state that the DS `Radio` component is required, but I worry about pushback from comsuming app devs.

src/components/RadioGroup/RadioGroup.tsx
bigfishdesign13: Yes, exactly.  I created a ticket to add this functionality to `DatePicker`.  Would it make sense to update `Label` and add a method to render as a `<legend>` element?

src/components/RadioGroup/RadioGroup.tsx
bigfishdesign13: Perfect.  This is exactly what we want to do.

src/styles.scss
bigfishdesign13: Hmm. Let's talk about this.

src/components/RadioGroup/RadioGroup.stories.mdx
EdwinGuzman: Okay, let's keep `showLabel` but it should be set to "true" by default.

src/components/RadioGroup/RadioGroup.tsx
EdwinGuzman: I think it should be applicable everywhere just like we decided on the `isDisabled`, `isErrored`, etc. But that can be a refactoring ticket or whenever other components are updated.

src/components/RadioGroup/RadioGroup.tsx
EdwinGuzman: I don't think it's too much to ask to use a DS `Radio` component. If someone is using Chakra, they would use all relevant components including Chakra's `RadioGroup` and `Radio`. If that's their pattern, I think we can follow it and say consumers _must_ use the DS `Radio` component within a `RadioGroup`.

src/components/RadioGroup/RadioGroup.tsx
EdwinGuzman: I would prefer a `Legend` component instead of `Label` rendering a `<label>` or `<legend>` element.

src/components/RadioGroup/RadioGroup.stories.mdx
bigfishdesign13: Yes, agreed.  I am not sure why `false` was the default.

EdwinGuzman:

Link:
- https://pr661-wigyt8r48uanxriut09u83euyjsnw1i0.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/612d4282baa70d3d329337b8

src/components/RadioGroup/RadioGroup.tsx
bigfishdesign13: OK, I am on board with this.

src/components/RadioGroup/RadioGroup.tsx
bigfishdesign13: I created new tickets to update the `layout` prop in `Card` and `SkeletonLoader`.

src/components/RadioGroup/RadioGroup.tsx
bigfishdesign13: OK, we can make a decision on the best approach when we work on [DSD-449](https://jira.nypl.org/browse/DSD-449).

EdwinGuzman:
Is this okay to merge or is something still outstanding?

bigfishdesign13:
> Is this okay to merge or is something still outstanding?

Yes, I think it is OK to merge this.  You had a question about the a11y and the best way to handle hiding something from the screen, but that should not hold up this PR.

EdwinGuzman:

Link:
- https://pr661-wigyt8r48uanxriut09u83euyjsnw1i0.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/612d4282baa70d3d329337b8

PR #670, opened 2021-09-08
DSD-479: Tabs - style update

EdwinGuzman:
Fixes JIRA ticket [DSD-479](https://jira.nypl.org/browse/DSD-479)

- This updates the style of the Tabs component to match Figma and the UX spec document.
- Note: The mobile previous/next buttons are there but don't functionally work. As discussed, it's possible to reuse a "useCarousel"-like React hook in the `Slideshow` component as well as in this `Tabs` component. Will work with @wluisi once the `Slideshow` component is in progress.

## How has this been tested?

Storybook.

### Checklist:

### Front End Review:

- [x] View [the example in Storybook](https://pr670-gzaem3bwedltl0d6j1ihjm334p11sqwv.tugboat.qa/)


src/components/Tabs/Tabs.tsx
```diff
@@ -1,72 +1,43 @@
...
+  useMultiStyleConfig,
   useStyles,
   useTab,
 } from "@chakra-ui/react";
 
-import bem from "../../utils/bem";
 import generateUUID from "../../helpers/generateUUID";
-
-// TODO: later when styling
-// enum TabsType {
```
EdwinGuzman, line 18: Ended up not using variants which makes this simpler.

src/components/Tabs/Tabs.tsx
```diff
@@ -89,14 +60,21 @@ const CustomAnchorTab = React.forwardRef((props: any, _ref) => {
...
-const getElementsFromData = (data, useHash) => {
+const getElementsFromContentData = (data, useHash) => {
   const tabs = [];
   const panels = [];
 
   if (!data?.length) {
     return {};
   }
 
+  if (data?.length > 6) {
```
EdwinGuzman, line 71: As per the UX spec doc.

src/components/Tabs/Tabs.tsx
```diff
@@ -132,45 +110,63 @@ const getElementsFromData = (data, useHash) => {
...
-    index = 0,
-    align = "start",
+    useHash = false,
   } = props;
-  const baseClass = "tabs";
-  // TODO: later when styling
-  // const variant = getVariant(tabsType);
-  // const styles = useStyleConfig("Tabs", { variant });
-  // line,enclosed, enclosed-colored, soft-rounded, solid-rounded, unstyled
+  const styles = useMultiStyleConfig("Tabs", {});
```
EdwinGuzman, line 119: All the styling is handled through `useMultiStyleConfig`.

src/components/Tabs/Tabs.tsx
```diff
@@ -132,45 +110,63 @@ const getElementsFromData = (data, useHash) => {
...
-      defaultIndex={index}
+      defaultIndex={defaultIndex}
       variant="enclosed"
-      // __css={styles}
     >
-      {data ? (
+      {contentData ? (
         <>
-          <TabList>{tabs}</TabList>
+          <Box __css={styles.tablistWrapper}>
```
EdwinGuzman, line 139: The result:
Screen Shot 2021-09-08 at 2 41 31 PM

src/theme/index.ts
```diff
@@ -36,6 +37,7 @@ const theme = extendTheme({
    */
   components: {
     Button,
+    Tabs,
```
EdwinGuzman, line 40: Naming it the same as Chakra's own `Tabs`, allows the existing theme to be extended rather than overriding them.

src/theme/components/tabs.ts
```diff
@@ -0,0 +1,51 @@
...
+        lg: "100%",
+      },
+    },
+    tab: {
+      color: "black !important",
+      paddingInlineStart: 4,
+      paddingLeft: 4,
+      border: "none",
+      borderColor: "ui.black",
+      paddingRight: {
```
EdwinGuzman, line 16: As per the UX spec: 1em padding left and right on mobile or "base" and then 3em on desktop or "md" (600px).

src/theme/components/tabs.ts
```diff
@@ -0,0 +1,51 @@
+const CustomTabs = {
+  baseStyle: {
+    tablist: {
+      borderColor: "ui.black",
+      maxWidth: {
```
EdwinGuzman, line 5: Note how you can add responsive styles by passing in an object rather than a single value for any specific CSS rule. In this case, "base" corresponds to the native mobile viewport and "lg" corresponds to "960px". Chakra does `min-width` by default for mobile-first responsive design.

It's also possible to pass in something like `maxWidth: ["70%", null, "100%"]` but I prefer to use the responsive object to be clearer about the breakpoints, and it's easier to read.

EdwinGuzman:

Link:
- https://pr670-gzaem3bwedltl0d6j1ihjm334p11sqwv.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/613909d7baa70d81ae9b2a49

src/components/Tabs/Tabs.tsx
bigfishdesign13: Makes sense.  At this point, there are no true variations in the overall `Tabs` UI.

src/components/Tabs/Tabs.tsx
bigfishdesign13: The CSS should be adjusted so the tab labels do not wrap.

src/theme/components/tabs.ts
bigfishdesign13: This is, obviously, very important.  We should talk about this more and make sure all of the DS devs are aware of how to handle responsive styles.

bigfishdesign13 (commented):
This looks really good.  The styles, as you have applied them, sync very nicely with the spec doc and the Figme file.  However, I see that some of the styles will need to be adjusted.  Specifically, the bold text for the hover state will need to go.  There are some other styles that also need to be reworked, but we can address those all at once after they are hashed out on the UX side.

Speaking of `hash`, the tab styles get a bit wonky when `useHash` is true.

PR #685, opened 2021-09-20
DSD-470: Chakra Hero

EdwinGuzman:
Fixes JIRA ticket [DSD-470](https://jira.nypl.org/browse/DSD-470)

- Updates the `Hero` component with Chakra components and NYPL theme.
- Note that `hero__heading` in the `Hero.scss` updated the styles for the `Heading` component used internally in the `Hero`. This is harder to do with the Chakra-based theme js file. Instead, I passed an `sx` prop to the `Heading` component to add additional JS-object styles. This is similar to what we did before with the `attributes` prop except this will specifically be used with Chakra's JS-object to extend styles and pass it to its own `sx` prop in the `Heading` component. Let me know if this makes sense and if the name is fine or if it should be renamed.

## How has this been tested?

Locally and Storybook.

### Checklist:

### Front End Review:


EdwinGuzman:

Link:
- https://pr685-njfiszejbmxbclbv1z1hpydcczde5dmy.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/6148b991d7000a2c7b463162

EdwinGuzman:

Link:
- https://pr685-njfiszejbmxbclbv1z1hpydcczde5dmy.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/6148b991d7000a2c7b463162

EdwinGuzman:

Link:
- https://pr685-njfiszejbmxbclbv1z1hpydcczde5dmy.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/6148b991d7000a2c7b463162

EdwinGuzman:

Link:
- https://pr685-njfiszejbmxbclbv1z1hpydcczde5dmy.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/6148b991d7000a2c7b463162

src/components/Hero/Hero.stories.mdx
```diff
@@ -30,220 +29,293 @@ import { getCategory } from "../../utils/componentCategories";
...
+  <Image src="https://placeimg.com/800/400/animals" alt="Image example" />
 );
 
 # Hero
 
 | Component Version | DS Version |
 | ----------------- | ---------- |
 | Added             | `0.2.0`    |
-| Latest            | `0.23.4`   |
+| Latest            | `0.24.2`   |
```
bigfishdesign13: This should be changed to `0.25.1`.

src/components/Hero/Hero.test.tsx
```diff
@@ -8,202 +8,481 @@ import { HeadingLevels } from "../Heading/HeadingTypes";
...
           />
         }
+        subHeaderText="Example Subtitle"
         backgroundImageSrc="https://placeimg.com/1600/800/arch"
       />
     );
-    expect(utils.container.querySelector(".hero--primary")).toBeInTheDocument();
-    expect(utils.container.querySelector(".hero")).toHaveAttribute(
+
+    expect(screen.getByText("Hero Primary")).toBeInTheDocument();
```
bigfishdesign13, line 128: With this type of test, how do you ensure that the UI is styled properly?  This is only checking to see if string value is present onscreen.

src/components/Hero/Hero.tsx
```diff
@@ -1,114 +1,109 @@
...
+const variantMap = {};
+for (const type in HeroTypes) {
+  variantMap[HeroTypes[type]] = HeroTypes[type];
+}
+/**
+ * Map the HeroTypes to the Hero Chakra theme variant object. If a wrong
+ * value is passed (typically in non-Typescript scenarios), then the default
+ * is the "primary" variant.
+ */
+const getVariant = (type) => variantMap[type] || HeroTypes.Primary;
```
bigfishdesign13, line 46: Can the `variantMap` code above be include within the `getVariant` method and converted into a global function that can be used by any component?

src/components/Hero/HeroTypes.tsx
```diff
@@ -1,7 +1,7 @@
 export enum HeroTypes {
-  Campaign = "CAMPAIGN",
-  FiftyFifty = "FIFTYFIFTY",
-  Primary = "PRIMARY",
-  Secondary = "SECONDARY",
-  Tertiary = "TERTIARY",
+  Primary = "primary",
```
bigfishdesign13, line 2: Thank you for converting these to lowercase values!

src/docs/Intro.stories.mdx
```diff
@@ -15,7 +15,6 @@ import getCSSVariable from "../helpers/getCSSVariable";
   image={
```
bigfishdesign13: The `Hero` on the intro page is broken.
image

src/theme/components/hero.ts
```diff
@@ -0,0 +1,247 @@
...
+  paddingBottom: 0,
+  paddingRight: 0,
+  paddingLeft: 0,
+  width: "100%",
+};
+// TODO: Is this being used?
+const placeholder = {
+  "+ .placeholder": {
+    flex: "0 1 200px",
+    marginLeft: 12, // --space-xl
```
bigfishdesign13: For the spacing, I've been using the values that align with the old CSS vars.  So, in this case, I would use `xl`.  I feel those values are a little more concrete than the numeric values.

src/theme/components/hero.ts
```diff
@@ -0,0 +1,247 @@
...
+    minHeight: "350px",
+    content: {
+      bg: "ui.black",
+      color: "ui.white",
+      flex: {
+        base: "0 0 100%",
+        md: "0 0 60%",
+      },
+      maxWidth: { md: "960px" },
+      paddingTop: 16, // --space-xxl
```
bigfishdesign13: Same thing here with the spacing values: `xxl` and `l`.

src/theme/components/hero.ts
```diff
@@ -0,0 +1,247 @@
...
+      _before: {
+        bg: "ui.black",
+        content: `""`,
+        height: "100%",
+        left: "-2000px",
+        position: "absolute",
+        width: "4000px",
+        zIndex: "-1",
+      },
+      ".nypl--books-and-more &": {
```
bigfishdesign13: What exactly does this do?

bigfishdesign13 (changes requested):
The `Hero` component has a lot built in and I can see that this conversion was complex.  All in all, this looks quite good, but there are some things we need to discuss regarding the spacing values.

src/components/Hero/Hero.test.tsx
EdwinGuzman: Right, this type of test doesn't help and it's not class-based anymore. The snapshot tests don't readily work because of Chakra's random id/classname values. So we'll have to fix that in order to get proper styling testing. That may be fixed in Chakra version 1.6.8 so I'll try that then.

src/components/Hero/Hero.tsx
EdwinGuzman: I wanted to do something similar but not all type names are consistent. I don't think it'll matter but I'll try this in a separate PR.

src/docs/Intro.stories.mdx
EdwinGuzman: Ugh yes, forgot the `DSProvider` component.

src/theme/components/hero.ts
EdwinGuzman: Are you suggesting using `"xl"` instead of `12`? Internally, they're both the same in `spacing.ts`:

```
// --space-xl = 48px
xl: "3rem",
12: "3rem",
```

src/theme/components/hero.ts
EdwinGuzman: This is for the cases where a page has a specific class and the `Hero` needs to reflect that page. So something like:

```jsx
<div className="nypl--books-and-more">
  <Hero {...props} />
</div>
```

This CSS rule will take effect for this pattern. We do, however, need to update this and not follow this moving forward, but maybe it's too quick of a move to do this now? It can part of the migration process. If we do need to make this move now, these styles can be variants rather than relying on the parent's class name.

So the above would be:

```jsx
<div >
  <Hero type={HeroTypes.BooksAndMore} {...props} />
</div>
```

src/theme/components/hero.ts
EdwinGuzman: Hmm I see what you're saying now; it's easier to cognitively think about the name values than the numeric values even though they're the same. Is that right? Yeah, I can move to name values if that's what you mean.

EdwinGuzman:

Link:
- https://pr685-njfiszejbmxbclbv1z1hpydcczde5dmy.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/6148b991d7000a2c7b463162

src/components/Hero/Hero.test.tsx
bigfishdesign13: This is related to another question I had.  When we used SCSS, it was nice to see where the styles were coming from and it helped with debugging and adjusting styles.  With the Chakra CSS bundles, there is no way to tell where certain styles are being inherited from and how the CSS hierarchy is structured.  Are you aware of a config setting that would toggle bundling?  And maybe leave some type of trail to see which data object is producing certain styles?

src/components/Hero/Hero.tsx
bigfishdesign13: Maybe now is the time to normalize the naming conventions within the type files?  You already did that here with the type case.

src/theme/components/hero.ts
bigfishdesign13: Yes, that is exactly right.

src/theme/components/hero.ts
bigfishdesign13: Right.  I think this was a nice idea, but it was never fully fleshed out and implemented throughout DS.  I would like to move away from this and make the color assignment more of a manual props process.

EdwinGuzman:

Link:
- https://pr685-njfiszejbmxbclbv1z1hpydcczde5dmy.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/6148b991d7000a2c7b463162

EdwinGuzman:
Updated. Added four new variants with the `Secondary` prefix and added snapshot tests for each variant.

src/theme/components/hero.ts
EdwinGuzman: Updated this.

src/theme/components/hero.ts
EdwinGuzman: Removed the parent wrapper class names in favor of additional types.

bigfishdesign13 (approved):
Looks good!

PR #695, opened 2021-09-27
DSD-497: Global JS styling objects

EdwinGuzman:
Fixes JIRA ticket [DSD-497](https://jira.nypl.org/browse/DSD-497)

- Cleans up all the css-in-js styling JS objects to use name variables rather than numeric values, such as using `"xs"` instead of `2`.
- Creates a `globalMixins.ts` file for "converting" SCSS mixings to JS objects.
- Creates a `global.ts` object for reusable JS styling objects. This file imports and exports `globalMixins`. This is to make it easier to just use `global.ts` when importing objects instead of having to remember to import either `global` or `globalMixins`. Maybe it makes more sense to just import both directly if they are needed.
- SCSS Mixins should not be deleted since they _might_ still be used in consuming apps through the compiled scss file.

### Checklist:

### Front End Review:

### DIFF
--- a/src/theme/components/accordion.ts
+++ b/src/theme/components/accordion.ts
@@ -12,10 +12,10 @@ const buttonStyles = {
   },
 };
 const panelStyles = {
-  paddingTop: 2, // --space-xs
-  paddingRight: 24, // --space-xxxl
-  paddingLeft: 4, // --space-s
-  paddingBottom: 2, // --space-xs
+  paddingTop: "xs",
+  paddingRight: "xxxl",
+  paddingLeft: "s",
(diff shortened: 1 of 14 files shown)


EdwinGuzman:

Link:
- https://pr695-ihs0l7wt4pwvajlkjwuyptyx6bxgjhhe.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/615221c64807c001e13cc3f5

bigfishdesign13 (approved):
I like what's happening here.  The global styling file makes sense so far, but we should keep an eye on it to make sure it does not balloon out of control.

PR #710, opened 2021-10-01
WIP DSD-507: Using chakra for the SearchBar component

EdwinGuzman:
Fixes JIRA ticket [DSD-507](https://jira.nypl.org/browse/DSD-507)

- This is using the children component pattern and I'm wondering if it's better to use the prop pattern instead.
- There are some things I want to discuss before I proceed.
- Some updates are blocked by other components that should be updated first, such as `Form`.
- Creating a PR to see the Tugboat preview.

### Checklist:

### Front End Review:


EdwinGuzman:

Link:
- https://pr710-2rl71du1ury5hwjjw4i8wivnudpjiyty.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/6157346b4807c036654086fe

EdwinGuzman:

Link:
- https://pr710-2rl71du1ury5hwjjw4i8wivnudpjiyty.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/6157346b4807c036654086fe

EdwinGuzman:
@bigfishdesign13 this is ready for a review now.

EdwinGuzman:

Link:
- https://pr710-2rl71du1ury5hwjjw4i8wivnudpjiyty.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/6157346b4807c036654086fe

src/components/SearchBar/SearchBar.tsx
```diff
@@ -1,69 +1,162 @@
...
+  /** The onClick callback function for the `Button` component. */
+  buttonOnClick?: (event: React.MouseEvent | React.KeyboardEvent) => void;
+  /** A class name for the `form` element. */
   className?: string;
-  /** Adds 'method' property to the form */
-  method?: string;
-  /** Adds 'action' property to the form */
-  action?: string;
+  /** The text to display below the form in a `HelperErrorText` component. */
+  helperErrorText?: string;
```
bigfishdesign13, line 34: The data pattern in all other form components is to have separate `helperText` and `invalidText` (nee `errorText`) props.  Additionally, from a UI perspective, the error text always overwrites the helper text when `isInvalid` is `true`.

src/components/SearchBar/SearchBar.tsx
```diff
@@ -1,69 +1,162 @@
...
+}
 
 export interface SearchBarProps {
-  /** Populates aria-label on the form;
-   * defines a hidden string that labels
-   * the SearchBar
-   */
+  /** Adds 'action' property to the `form` element. */
+  action?: string;
+  /** Populates aria-label on the form and labels the entire SearchBar. */
```
bigfishdesign13: This prop only populates the `aria-label` on the form.  It does not create a visual label.  And as we are thinking about this, it shoud be noted that we have removed the `aria` props from all other form components and we are now relying on the component to properly generate the `aria` attributes based on other props.  Should we do the same here and follow the pattern of using `labelText` and `showLabel` props?

src/components/SearchBar/SearchBar.tsx
```diff
@@ -1,69 +1,162 @@
...
+    <Button
+      id={generateUUID()}
+      buttonType={ButtonTypes.Primary}
+      type="submit"
+      onClick={buttonOnClick}
+      disabled={isDisabled}
+    >
+      <Icon
+        name={IconNames.search}
+        modifiers={["small", "icon-left"]}
```
bigfishdesign13: Based on the recent updates to the `Icon` component, this need to be changed.  But I'm sure you already know this.

src/components/SearchBar/SearchBar.tsx
```diff
@@ -1,69 +1,162 @@
...
+        decorative
+      />
+      Search
+    </Button>
+  );
+  // Render the `HelperErrorText` component.
+  const helperErrorTextElem = helperErrorText && (
```
bigfishdesign13: As I mentioned elsewhere, the UI pattern has been for the invalid text to overwrite the helper text when `isInvalid` is true.  Right now, the default invalid text in the `TextInput` is triggering when it is set as invalid, so we'll need to add a method to text the default error text to not show.  I believe we talked about this the other day.

src/theme/components/select.ts
```diff
@@ -27,11 +27,19 @@ const select = {
...
   parts: ["select", "helper"],
   baseStyle: {
+    marginBottom: "xs",
     helper: {
       ...helperTextMargin,
     },
     select,
   },
+  variants: {
+    searchbar: {
```
bigfishdesign13, line 37: I would also like to address the right side border.  I have a query into Kathy about this, but I would like the right side border on the `Select` to be set to 0 and for the borde radius radius to also be set to 0.  Essentially, I would like it to be removed altogether and for the `Select` to appear attached to the `TextInput`.

bigfishdesign13: Similarly, this same type of customization and treatment should be applied to the `Button` in order to make it appear attached the `TextInput`.

src/theme/components/textInput.ts
```diff
@@ -41,6 +42,11 @@ const TextInput = {
       minHeight: "xxl",
     },
   },
+  variants: {
```
bigfishdesign13, line 45: This gets a little more complicated here.  The border radius should be set to 0 on the edges that touch the other elements.  Which means, when the select field is not used, the border radius should not be changed.

bigfishdesign13 (changes requested):
This works well, but there are some changes that we should talk about.

nypl-wluisi:
@bigfishdesign13 @EdwinGuzman 
Can we make this iteration of the "Search Bar" handle a label ("Enter an address ...") above the input, like this?

search-bar-label-above

I'd like to switch over to using the search bar from DS, but I need this layout on several different page types.

EdwinGuzman:
> Can we make this iteration of the "Search Bar" handle a label ("Enter an address ...") above the input, like this? 

Hmm I think this is a good candidate to say that it should be looked at by the UX team first, Figma should be updated, and then we'll make the update. I'm not opposed to making the change but we should try to follow the process.

src/components/SearchBar/SearchBar.tsx
EdwinGuzman: Yeah, I'll update this once the `Icons` PR is merged.

src/components/SearchBar/SearchBar.tsx
EdwinGuzman: Yes, I'll make this update but since we don't ever show the label, then maybe we can say that even though `labelText` is passed, it won't actually show up. So then we won't need the `showLabel` prop.

src/components/SearchBar/SearchBar.tsx
EdwinGuzman: We did but said it'd be a different ticket/PR for both the `Select` and `TextInput` components.

src/components/SearchBar/SearchBar.tsx
bigfishdesign13: OK.  But looking at Bill's comment, adding heading and description elements seem to make more sense.  Similar to both how `VideoPlayer` is handled and how `Slideshow` will be handled.  So yes, add `labelText` for now and `showLabel` can be considered if/when the UX is updated.

bigfishdesign13:
> Can we make this iteration of the "Search Bar" handle a label ("Enter an address ...") above the input, like this?

@nypl-wluisi That descriptor text could also go in as helper text.  That said, I do think the UI for the SearchBar component could be updated.  I'll get that rolling.

nypl-wluisi:
> > Can we make this iteration of the "Search Bar" handle a label ("Enter an address ...") above the input, like this?
> 
> @nypl-wluisi That descriptor text could also go in as helper text. That said, I do think the UI for the SearchBar component could be updated. I'll get that rolling.

Would it be the `<label>` element of the `<input>` ? We can confirm, but for accessibility, that's how I set it up.

EdwinGuzman:
Yes, it should be the `<label>` element for the `<input>` element. This gets trickier when there is a `<select>` component. Should the label be for the select or the input element? It'd get associated with the `form` element through the aria-label but I'm not sure what makes the most sense when there is a `select` as well, i.e. when you click on the label, where does the focus go to? We should have Zach review and give a recommendation.

EdwinGuzman:
@bigfishdesign13 updated.

EdwinGuzman:

Link:
- https://pr710-2rl71du1ury5hwjjw4i8wivnudpjiyty.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/6157346b4807c036654086fe

src/components/SearchBar/SearchBar.tsx
bigfishdesign13: Right.  Thanks for taking care of that.

bigfishdesign13 (approved):
The changes look good.  I will be creating new tickets to address the shift related to the `Select` focus state and to deal with the UX/dev to add a visible label.

PR #712, opened 2021-10-04
DSD-504: Chakra icon

EdwinGuzman:
Fixes JIRA ticket [DSD-504](https://jira.nypl.org/browse/DSD-504)

- Updates the `Icon` component to use Chakra's `Icon` component.
- Updates the CSS style object using NYPL's them.
- Updated all the other components, tests, and documentation where `Icon` is imported.
- *Note*: I think some places can be further refactored but it is more of an issue of the consuming component. For example, some updates to the `Link` component should be another ticket/PR that further updates the combination from this PR.
- I verified all the documentation pages where the `Icon` was updated but it'd be good to go through every single `[component].stories.mdx` page to verify (compared to `development` or `production`) that the changes are internal and the `Icon` displays just as before.

## How has this been tested?

Storybook.

### Checklist:

### Front End Review:


EdwinGuzman:

Link:
- https://pr712-v0jnglyjly0kmwkryv6depxsaogyzjst.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/615b39444807c054f1428060

EdwinGuzman:

Link:
- https://pr712-v0jnglyjly0kmwkryv6depxsaogyzjst.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/615b39444807c054f1428060

src/components/Notification/Notification.tsx
```diff
@@ -94,20 +94,20 @@ export default function Notification(
         return (
           <Icon
             decorative={false}
-            className="notification-icon"
-            name={IconNames.speaker_notes}
-            color={IconColors.section_research_secondary}
-            modifiers={["large"]}
+            className="notification-icon icon"
```
bigfishdesign13: I see that you added the `icon` class here in order to pickup the icon styles written into `_Notification.scss`.  This works, but we also need to include something to cover the custom icons that only show the `chakra-icon` class.  It might be best to remove the `icon` class here and update `_Notification.scss` to target `chakra-icon` instead of `icon`.

src/components/Button/Button.stories.mdx
```diff
@@ -99,16 +104,18 @@ Controls for this Story section to preview a `Button` with the selected icon.
         {args.displayLeft && (
           <Icon
             name={args.iconType}
-            decorative={true}
-            modifiers={["small", "icon-left"]}
+            className="left"
```
bigfishdesign13: If the `align` prop is set, why is the `className` also needed?  That goes for all instances in the story file.

src/components/Icons/Icon.stories.mdx
```diff
@@ -184,59 +192,51 @@ Passing an `iconRotation` prop with a value from `IconRotationTypes` allows
...
 The following icons are for generic purposes. All the available logo names can
 be found in the `IconNames` enum in `src/components/Icons/IconTypes.tsx`.
 
-<Preview withToolbar>
+<Canvas withToolbar>
   <Story name="All Display Icons">{allIconsType(icons)}</Story>
-</Preview>
+</Canvas>
 
 ## All Logo Icons
```
bigfishdesign13, line 241: There are problems with rendering some of the icons in this set.  That is out of scope for this ticket so I will setup another ticket.

src/components/Icons/Icon.tsx
```diff
@@ -1,105 +1,105 @@
...
-  /** Icons designated as decorative will be ignored by screenreaders */
+  /** Overrides default icon color (black). */
+  color?: IconColors;
+  /** Icons designated as decorative will be ignored by screenreaders. True
+   * by default. */
   decorative?: boolean;
-  /** ID that other components can cross reference for accessibility purposes */
-  id?: string;
-  /** This text will be added as a child `<title>` element inside the `<svg>` tag.  It is recommended to do this for increased accessibility. */
-  titleText?: string;
```
bigfishdesign13, line 30: There's no way to apply a text title to the icon?  What happens when an icon is more than just decorative?  How is it presented to a screen reader?

bigfishdesign13 (commented):
Phew!  That touched a lot of pages.  Overall, this looks really good.  I just have some questions sprinkled throughout.

src/components/Notification/Notification.tsx
EdwinGuzman: Good point. This was something I wanted to do when `Notification` is composed using Chakra but I can make this change now.

src/components/Button/Button.stories.mdx
EdwinGuzman: In the `Button` component, it needs more spacing for the left/right margins. This would be good for a Button update but I can make this now as well.

src/components/Icons/Icon.stories.mdx
EdwinGuzman: Yeah, this has always been an issue.

src/components/Icons/Icon.tsx
EdwinGuzman: Forgot to make a note of this:
The `title` attribute in an `svg` element is not accessible. It's useful for other HTML elements but it doesn't work the same for SVGs. The better approach is to add `<title>` HTML elements to the `<svg>` element as its _first_ child. I couldn't find a way to read the SVG component/string and inject a `title` element. I think right now the best approach is to manually update every single SVG element the DS has with a title in their own file. The problem is that in some cases the `titleText` _was_ being passed. Maybe we just make the title static all the time.

To explain the reading issue further, the SVGs we import into `Icon` are all `.svg` files and not React components. This makes it harder to read, decompose, and inject a component or element. I couldn't read the children and put a `<title>` as the first element and push the remaining children after.

We can add a `title` element to each svg file now, but should we agree on a convention? Or for now just put the name of the SVG in the title?

https://css-tricks.com/svg-title-vs-html-title-attribute/

bigfishdesign13: Right, we had determined that `title` was not a proper attribute for `svg` and we were able to insert the proper child element into the `svg`.  It took some time, but it was doable.  Did you do make that work or was it Michael?

Icons can be used for so many diffrerent reasons, so I worry about putting a static value in there.  Let's make sure to have a call to talk about this.

EdwinGuzman: Hmm I don't think this was ever implemented. The previous iteration added `titleText` to the `title` attribute to the `svg` element. But yes, let's talk about this.

bigfishdesign13: Wow, that's very odd.  I do rememebr this working properly.  Then it must have been Michael.  I recall that he ripped things apart and used a new package to handle the `svg`.  The `svg` came in as a React component and we had much more control over it.  Let me see if I can find it in the history.

bigfishdesign13: This was Michael's last commit for this.  There was still an error that was showing in the github tests, but locally it worked fine.

https://github.com/NYPL/nypl-design-system/tree/586ed39c8510f781d65fc7518f0f847c527b12cc/src/components/Icons

EdwinGuzman: I remember that too but I remember that it didn't work. As far as I can tell from that commit, the `iconProps` is setting the `title` attribute to the main `svg` element but no `title` element is being created or inserted.

EdwinGuzman:
@bigfishdesign13 made updates but not for the `title` issue.

src/components/Icons/Icon.tsx
bigfishdesign13: I created a ticket to address this the `title` element.

bigfishdesign13 (approved):
The updates look good.  I created a new ticket to handle the svg `title` element.

PR #703, opened 2021-09-29
DSD-501 & DSD-453: Chakra Image and prop refactor

EdwinGuzman:
Fixes JIRA ticket [DSD-501](https://jira.nypl.org/browse/DSD-501) and [DSD-453](https://jira.nypl.org/browse/DSD-453)

## This PR does the following:
- *This is a work-in-progress to get a sense of it. There are things I want to clean up and add better comments/documentation but I'll do that after getting an initial reaction.*
- Updates the `Image` component to use Chakra.
- Updates/refactors its props.
- Moves styles to its CSS JS object.
- Updates its usage in the `Hero` and `Card` components.

Notes:
- `CardImage` is directly set to `Image`, but `CardImageSizes` and `CardImageRatios` were removed in favor of using `ImageSizes` and `ImageRatios`, respectively.
- There is still SCSS in the `Hero` and `Card` components that update the styling for the `Image` component.
- This was tricky and we probably don't have to do much until we refactor the `Card` component to use Chakra.
- For the `Hero` component, the image overrides already exist in its `hero.ts` CSS-theme JS object. Not much has to be done right now but it could be cleaner. For example, in the `campaign` variant:
```jsx
img: {
  flex: {
    base: "1 1 100%",
    md: "0 0 50%",
  },
  minWidth: "0", // https://github.com/philipwalton/flexbugs/issues/41
  objectFit: "cover",
  width: "100%",
  height: { md: "320px" },
},
```
maybe we can move this to the `image.ts` file as a hero campaign variant. I'm not sure what the best pattern is yet.
- The `Card` component _needs_ the image to be wrapped in divs with specific stylings but the `Hero` component does not need these div wrappers. In fact, it renders pretty badly. By default, the div wrappers are always rendered to make it easier to apply the size and aspect ratio. For the `Hero` component, however, the prop `useImageWrapper` is set to false so that only the `img` element is rendered with no div wrappers. This pattern works nicely but let me know what you think.

## How has this been tested?

Storybook. Check `Hero`, `Card`, and `Image`.

### Checklist:

- [x] I have updated the Storybook documentation accordingly.
- [x] All new and existing tests passed.

### Front End Review:

- [x] View [the example in Storybook](https://pr703-euxm3uflqmmvpnpmpmjb5pn1fy8nndbw.tugboat.qa/)

PR #718, opened 2021-10-07
DSD-505: Chakra Card

EdwinGuzman:
Fixes JIRA ticket [DSD-505](https://jira.nypl.org/browse/DSD-505)

- Updates the `Card` component to use Chakra.
- Note: This is based on the DSD-453 #703 image PR. This is because it's easier to work with and update the `Image` component when it's also using Chakra.
- Instead of updating the `Image` component with a specific `Card` variant type, the `CardImage` extends the `Image` component and adds its own styles on top of the base ones. This is done through the `sx` prop in `Image`.
- The updates in the `Card.tsx` file are really nothing too exciting.
- But take a look at `card.ts` -- this has the most complicated set of styles for the image wrapper and the other elements. This is because there were five modifiers that are now flags to turn on or update certain styles in this file. The `getBodyPaddingStyles` function does a lot to just get a padding value. Hopefully, this can be simplified later on.
- I didn't add anything new and everything should look just as it does on the `development` branch or the QA server. The only difference is that the headers are _now_ the proper styles. This is because the `DSProvider` was missing from certain Storybook blocks but this is fixed now.
- We'll have to chat about the custom image component option. Adding a custom image component strips basically all the styles we want to render. Unfortunately, I don't think we can see this in action until we try it in a real application. I think this can be its own, quite big, ticket (we have to figure out how to add Chakra-based styling to a non-CSS-in-JS image component if that is what is passed).

## How has this been tested?

Storybook

### Checklist:

### Front End Review:


EdwinGuzman:

Link:
- https://pr718-3pixkjnukafuxcezia0fnicbwzme93p6.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/615f6a1ecc9cc4418761216d

src/components/Card/Card.stories.mdx
```diff
@@ -35,12 +35,9 @@ import { getCategory } from "../../utils/componentCategories";
     },
   }}
   argTypes={{
-    blockName: { table: { disable: true } },
     className: { table: { disable: true } },
-    modifiers: { table: { disable: true } },
     id: { table: { disable: true } },
     imageComponent: { table: { disable: true } },
-    padding: { table: { disable: true } },
```
bigfishdesign13, line 43: Even with this removed, the `padding` prop is not showing in the `ArgsTable` in Storybook.

src/components/Card/Card.stories.mdx
```diff
@@ -174,53 +171,55 @@ the default placement and move the image to the last element within a `Card`.
...
-    <Card
-      layout={CardLayouts.Row}
-      center
-      imageSrc="https://placeimg.com/410/210/animals"
-      imageAlt="Alt text"
-      imageAspectRatio={ImageRatios.TwoByOne}
-      imageAtEnd
+<Canvas>
+  <DSProvider>
+    <div
```
bigfishdesign13: If this PR does not make it into the `0.25.1` release, which it looks like it will not, make sure to use `SimpleGrid` here and for other grid layouts on this page.

src/components/Card/Card.tsx
```diff
@@ -1,136 +1,144 @@
...
+    layout,
+  });
+  return (
+    <Image
+      alt={alt}
+      component={component}
+      imageAspectRatio={imageAspectRatio}
+      imageSize={imageSize}
+      src={src}
+      sx={styles}
```
bigfishdesign13: I see you are using the `sx` prop.  What was your reasoning?  If you were running into the same problems that I was, were you able to determine why `__css` was not working and `sx` was needed?

src/components/Image/Image.tsx
```diff
@@ -35,13 +37,14 @@ function ImageWrapper(props: React.PropsWithChildren<ImageWrapperProps>) {
     children,
     imageAspectRatio = ImageRatios.Original,
     imageSize = ImageSizes.Default,
+    sx = {},
```
bigfishdesign13: And here is `sx` again.  I'm curious about this.

src/theme/components/card.ts
```diff
@@ -0,0 +1,167 @@
+const imageSizes = {
+  small: { flex: { md: "0 0 165px" } },
+  medium: { flex: { md: "0 0 225px" } },
+  large: { flex: { md: "0 0 360px" } },
+};
+// This is complicated and can be refactored later...
+const getBodyPaddingStyles = ({ border, hasImage, imageAtEnd, isRow }) => {
```
bigfishdesign13, line 7: You mentioned that you needed to do a lot to handle the padding properly.  Yes, this makes sense.  The original CSS had a lot going on, with some classes overriding other classes simply based on the order within the stylesheet, and I knew this conversion was going to be complex.

src/components/Card/Card.tsx
```diff
@@ -139,27 +147,42 @@ export default function Card(props: React.PropsWithChildren<CardProps>) {
...
       child.props.mdxType === "CardActions"
     ) {
-      cardContents.push(child);
+      const elem = React.cloneElement(child, { key, center, layout });
+      cardContents.push(elem);
     }
   });
 
+  const styles = useMultiStyleConfig("Card", {
+    border,
```
bigfishdesign13, line 156: I _love_ that we can do this.  Logical styles helps SO much!

bigfishdesign13 (commented):
Big lift here, but, overall, this look good.  I have a number of questions and comments sprinkled throughout.  As you noted in your PR comments, let's make sure to talk about the custom image.

src/components/Card/Card.stories.mdx
EdwinGuzman: Yeah... that's because I completely removed `padding` as a prop from `Card`. I didn't see it being used anywhere and there wasn't an example of how to use it, so I didn't think it was prop that we still needed. Do we still need it? I can add it back and add examples.

src/components/Card/Card.tsx
EdwinGuzman: This is slightly different. This is the prop that is coming from `Card` into `Image` and inside the `Image` component, the `sx` prop is used to combine any _extra_ CSS styles with styles coming from `useMultiStyleConfig`. 
https://github.com/NYPL/nypl-design-system/pull/718/files#diff-da73b37ebbe2383290a1d6caf9037c4598e9d500d0632596c8dc877a191ad10fR47

src/components/Image/Image.tsx
EdwinGuzman: In this case, this is the prop name used for any _extra_ CSS styles. The `__css` still works in the Chakra specific component but the styles are combined for the `__css` prop.

src/components/Card/Card.tsx
EdwinGuzman: Yeah, once I figured this out, it made configurations easier to think about and implement.

EdwinGuzman:
I think maybe I should rename the `sx` prop to something else. It is used in Chakra for their own components similar to the `__css` prop for overriding/adding more styles. I wanted to use the same name so it could mean that these are additional styles.

src/components/Card/Card.stories.mdx
EdwinGuzman: Updated this.

src/components/Card/Card.stories.mdx
bigfishdesign13: I reviewed the `Card` component in the Figma Main file and extra padding does not come into play for any of revamped variants.  So no, there is no need to reinstate the `padding` prop.  All good here.

EdwinGuzman:

Link:
- https://pr718-3pixkjnukafuxcezia0fnicbwzme93p6.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/615f6a1ecc9cc4418761216d

bigfishdesign13:
> I think maybe I should rename the `sx` prop to something else. It is used in Chakra for their own components similar to the `__css` prop for overriding/adding more styles. I wanted to use the same name so it could mean that these are additional styles.

Yes.  If that is not actually the Chakra prop, then let's use a unique name.

bigfishdesign13 (approved):
Looks good.

PR #719, opened 2021-10-13
DSD-521: optionally display the helper/invalid text

EdwinGuzman:
Fixes JIRA ticket [DSD-521](https://jira.nypl.org/browse/DSD-521)

- Adds a `showHelperInvalidText` flag to form-related component to optionally display the helper/invalid text in the `HelperErrorText` component.

### Checklist:

### Front End Review:


src/components/SearchBar/SearchBar.tsx
```diff
@@ -83,7 +83,14 @@ export default function SearchBar(props: SearchBarProps) {
     textInputProps,
   } = props;
   const styles = useMultiStyleConfig("SearchBar", { hasSelect: selectProps });
-  const stateProps = { isDisabled, isInvalid, isRequired };
+  const stateProps = {
+    helperText: "",
+    isDisabled,
+    isInvalid,
+    isRequired,
+    showHelperInvalidText: false,
```
bigfishdesign13, line 91: I don't see this prop in the `SearchBarProps` interface.  How is this being used?

bigfishdesign13 (commented):
I only have a question about what is happening in the files for `SearchBar`.  Everything else looks good.

src/components/SearchBar/SearchBar.tsx
EdwinGuzman: `SearchBar` does not accept or use `showHelperInvalidText` as a prop. This is set to `false` in the `stateProps` object that is passed to both the `Select` and `TextInput` components. Those two components _do_ accept this prop. We just never want to show the helper/error text for `Select` or `TextInput` so they are always set to false.

bigfishdesign13: Got it.  Thanks.

bigfishdesign13 (approved):
Question answered.  Everything looks good.

PR #724, opened 2021-10-14
DSD-524: Chakra Notification

EdwinGuzman:
Fixes JIRA ticket [DSD-524](https://jira.nypl.org/browse/DSD-524)

- Updates the `Notification` component to use Chakra.
- Note: I removed the export for the `NotificationContent` and `NotificationHeading` components and they are now props instead.

## How has this been tested?

Storybook.

### Checklist:

### Front End Review:


src/theme/components/notification.ts
```diff
@@ -0,0 +1,95 @@
...
+  baseStyle: ({ centered, notificationType }) => {
+    let color = "ui.black";
+    if (notificationType === NotificationTypes.Announcement) {
+      color = "section.research.secondary";
+    } else if (notificationType === NotificationTypes.Warning) {
+      color = "brand.primary";
+    }
+    return {
+      display: "flex",
+      marginBottom: "s",
```
bigfishdesign13: Change this to `xxs`.  As I look at this in the browser, I am not sure why this margin was originally set as `s`.  That's too large a space.

src/components/Button/Button.tsx
```diff
@@ -26,6 +26,8 @@ interface ButtonProps {
   mouseDown?: boolean;
   /** The action to perform on the `<button>`'s onClick function */
   onClick?: (event: React.MouseEvent | React.KeyboardEvent) => void;
+  /** Optionally pass in additional Chakra-based styles. */
+  sx?: { [key: string]: any };
```
bigfishdesign13: Why were these changes included in this PR?

src/components/Notification/Notification.tsx
```diff
@@ -1,201 +1,183 @@
...
   id?: string;
-  /** Optional prop to control the margin around the `Notification` component */
+  /** Optional prop to control the margin around the `Notification` component. */
   noMargin?: boolean;
-  /** Optional prop to control the coloring of notification text and the visibility of an applicable icon */
-  notificationType?: NotificationTypes;
-  /** Modifiers array for use with BEM. See how to work with modifiers and BEM here: http://getbem.com/introduction/ */
-  modifiers?: string[];
+  /** Content to be rendered in a `NotificationContent` component. */
+  notificationContent: JSX.Element;
```
bigfishdesign13: Why `JSX.Element` for this and `notificationHeading`?  Why not define them as `string`?

bigfishdesign13 (changes requested):
Another complicated conversion.  Overall, it works well.  I have two questions and one requested change.

src/components/Notification/Notification.tsx
EdwinGuzman: Right, I wasn't sure what the right type would be and maybe it can be `string | JSX.Element`.  All the previous examples were plain strings but what if more complicated (such as <b>, <i>, or links) elements should be rendered?

src/components/Button/Button.tsx
EdwinGuzman: Because the `Button` component inside the `Notification` component has updated styles and I'm passing those through this prop - https://github.com/NYPL/nypl-design-system/pull/724/files#diff-1019e818df12ccb2db0eafad4b181840c4eeecd95b057fe177466b21a68d7b54R129

But, as mentioned in an earlier PR, the "sx" name is not the best so I'll update this.

bigfishdesign13: OK, got it.

src/components/Notification/Notification.tsx
bigfishdesign13: Right, `notificationContent` might have more sophisiticated content, so allowing for `JSX.Element` does make sense.  But I think `notificationHeading` can just be `string`.

EdwinGuzman:
Updated.

EdwinGuzman:

Link:
- https://pr724-u1ernhaulbukwkf8kdly8hhkuxytjbpd.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/616891f11ff56a6b8be90167

src/components/Notification/Notification.tsx
bigfishdesign13: I see that `Hero` uses the `string | JSX.Element` pattern, so let's do the same here.  More often than not, the actual content for this section is going to be a plain string.

bigfishdesign13 (changes requested):
Let's go ahead and change the assignment of `notificationContent` to `string | JSX.Element` and bring it into alignment with what had been done in `Hero`.

EdwinGuzman:
Updated and also updated the `sx` prop name to `additionalStyles`. I think that makes it clearer that it's a prop for the DS components whereas the Chakra components use `__css` or `sx`.

EdwinGuzman:

Link:
- https://pr724-u1ernhaulbukwkf8kdly8hhkuxytjbpd.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/616891f11ff56a6b8be90167

bigfishdesign13 (approved):
Very nice.

PR #721, opened 2021-10-13
DSD-523: Updating margins in the TextInput component

EdwinGuzman:
Fixes JIRA ticket [DSD-523](https://jira.nypl.org/browse/DSD-523)

## This PR does the following:
- Removes the bottom margins from the `TextInput` and its `HelperErrorText` component.

### Checklist:

- [ ] I have updated the Storybook documentation accordingly.
- [ ] All new and existing tests passed.

### Front End Review:

- [ ] View [the example in Storybook]()

PR #725, opened 2021-10-14
DSD-509: Chakra HelperErrorText

EdwinGuzman:
Fixes JIRA ticket [DSD-509](https://jira.nypl.org/browse/DSD-509)

- Updates the `HelperErrorText` component with Chakra.

## How has this been tested?

Storybook.

### Checklist:

### Front End Review:


src/components/HelperErrorText/HelperErrorText.tsx
```diff
@@ -28,36 +31,27 @@ export default function HelperErrorText(
...
-    <div
+    <Box
       id={id}
-      className={bem(baseClass, modifiers, blockName)}
-      aria-live={announceAriaLive ? ariaLive : "off"}
+      className={className}
       aria-atomic={ariaAtomic}
-      {...attributes}
+      data-isInvalid={isInvalid}
+      aria-live={announceAriaLive ? ariaLive : "off"}
```
bigfishdesign13, line 51: We don't use this in many of the other components.  Why is that?

src/theme/components/helperErrorText.ts
```diff
@@ -0,0 +1,9 @@
+const helperErrorText = {
+  baseStyle: ({ isInvalid }) => ({
+    marginBottom: "xs",
```
bigfishdesign13: The original component did have `bottom-margin` set to `xs`, but I think this should be removed and the bottom-margin should be 0.  The `Form` component will **now** handle the spacing between form elements.  Once this is set to 0, we will need to review the other components visually to make sure this margin value was not be relied upon.  Off the top of my head, I think `DatePicker` may be the only one that really needs it.

bigfishdesign13 (changes requested):
The requested changes tie into what I was saying in the PR for DSD-523.

src/components/HelperErrorText/HelperErrorText.tsx
EdwinGuzman: Do you mean the `aria-live` attribute itself? I think it fits best for this type of component where the description test might change (often) and is useful to the user and isn't already being presented (such as through a label or link).

EdwinGuzman:
I added the changes from DSD-523 PR #721  in the last commit for this PR. I also updated the `Form` stories so it displays the `DatePicker` component as an example.

src/components/HelperErrorText/HelperErrorText.tsx
bigfishdesign13: Got it.  Makes sense.  Thanks.

src/theme/components/helperErrorText.ts
bigfishdesign13: I had a typo in this comment.  It should have been...
```The Form component will NOW handle the spacing...```

bigfishdesign13 (approved):
This looks good.  We just need to make sure to fix the spacing issues now present in `DatePicker`.

PR #728, opened 2021-10-15
DSD-538: prefix for CSS vars

bigfishdesign13:
Fixes JIRA ticket [DSD-538](https://jira.nypl.org/browse/DSD-538)

- update the prefix for names of CSS vars from `chakra` to `nypl`

## How has this been tested?

- Storybook

### Checklist:

### Front End Review:

- [x] View [the example in Storybook](https://pr728-cw4s1xqg2ojwatihidueawvoqq5wkzy8.tugboat.qa/)


EdwinGuzman:

Link:
- https://pr728-cw4s1xqg2ojwatihidueawvoqq5wkzy8.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/6169f5061ff56a68d9e9a827

EdwinGuzman (commented):
This is looking fine so far but I think all CSS var references should be updated in the "Documentation" section such as in the breakpoints, colors, spacing, etc. to have the new `--nypl` prefix

bigfishdesign13:
@EdwinGuzman As I mentioned in Slack, the `getCSSVariable` method is having trouble with the CSS vars generated by Chakra.  Locally, the method returns a value in some situations and returns nothing in other situations.  But on the Tugboat preview build, it is not working in any situations.

EdwinGuzman:

Link:
- https://pr728-cw4s1xqg2ojwatihidueawvoqq5wkzy8.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/6169f5061ff56a68d9e9a827

src/components/StyleGuide/Buttons.stories.mdx
```diff
@@ -42,17 +53,7 @@ import { Button, ButtonTypes } from "@nypl/design-system-react-components";
...
-For information about using the `Button` component, please see below.
-
-- [General Guidelines](#general-guidelines)
-- [Spacing](#spacing)
-- [Button Types](#button-types)
-- [Patterns](#patterns)
-- [Figma Reference](#figma-reference)
-
-## General Guidelines
+## Guidelines
```
EdwinGuzman, line 56: Can you check the "Spacing" section? Some vars are not updated to use the `--nypl` prefix. Also, in this section, it's passing styles through the `attributes` prop but we want to move away from that. The `ButtonGroup` itself should handle the spacing between buttons.

EdwinGuzman: And the red button with "Button Text" can be a "callout" type instead of updating the background color directly. It's just outdated pattern.

EdwinGuzman:
The `intro.stories.mdx` page has two CSS var colors that should use the `--nypl` prefix for the `Hero` component.

src/components/StyleGuide/Colors.stories.mdx
```diff
@@ -8,329 +8,260 @@ import {
...
+export const uiVarDocs = [];
+export const grayscaleVarDocs = [];
+export const makeColorCard = function (key, value, docArray) {
   docArray.push(
-    <li style={{ textAlign: "center" }}>
-      <UIDocCard backgroundColor={value}>
-        {key}: {value}
-      </UIDocCard>
-    </li>
+    <ColorCard
```
EdwinGuzman: This also need `key={key}`.

EdwinGuzman:
I found out that:

- Chakra computes the custom CSS vars and this takes some time for Chakra to add it to the root of the document.
- This means that `getComputedStyle(document.documentElement).getPropertyValue(...)` does not pick it up on its initial load. I gave it a timeout and it was able to get the `--nypl` _and_ `--chakra` values appropriately. So this is not an issue with what we are doing but how Chakra applies the CSS vars.
- `getCSSVariable` worked before with our previous CSS variables (such as `--ui-black`) because we add CSS variables through SCSS which (I think) is compiled and added to Storybook earlier than Typescript is compiled for the Chakra CSS variables.
- I noticed that `CSSVariablesHelper.tsx` is not being used so let's remove that.
- We only use `getCSSVariable` in stories and is a helper function. Let's remove this function and rely on the native `var` function to get all the CSS variables values. Doing this works on initial load and this is what other devs would use in their apps.

In `Colors.stories.mdx`:
```
const varValue = getCSSVariable(varName);
makeColorCard(varName, varValue, uiVarDocs);

// convert it to:

const varValue = `var(${varName})`;
makeColorCard(varName, varValue, uiVarDocs);
```

EdwinGuzman (approved):

PR #730, opened 2021-10-18
DSD-540: useNYPLTheme hook for theme object

EdwinGuzman:
Fixes JIRA ticket [DSD-540](https://jira.nypl.org/browse/DSD-540)

- Exposes `useNYPLTheme` custom hook that returns a theme object with NYPL-specific styles.
- Internally, it's used from Chakra's `useTheme` hook but the exported theme object restricts the values so only the NYPL values are available.

## How has this been tested?

Through `npm link` and the Library Card Application.
Example of theme object being used in the Library Card app (also written in the Chakra documentation):

Screen Shot 2021-10-18 at 11 09 30 AM

### Checklist:

### Front End Review:


bigfishdesign13 (approved):
This is exactly what I envisioned.  Thank you for putting this together.  I think you included all of the relevant theme objects in the hook.  But wow, I didn't realize is was such a manual process to do so.

PR #735, opened 2021-10-25
DSD-526: Chakra VideoPlayer

bigfishdesign13:
Fixes JIRA ticket [DSD-526](https://jira.nypl.org/browse/DSD-526)

- composes the DS VideoPlayer component using Chakra
- adds new `ComponentWrapper` component to implement common layout patterns

## How has this been tested?

- Storybook in local build of DS

### Checklist:

### Front End Review:

- [x] View [the example in Storybook](https://pr735-3a8fhvpzfkwcv2hi3dmodjzxgwutmvth.tugboat.qa/)


src/components/ComponentWrapper/ComponentWrapper.tsx
```diff
@@ -0,0 +1,44 @@
+import * as React from "react";
+import { Box, useMultiStyleConfig } from "@chakra-ui/react";
+import Heading from "../Heading/Heading";
+import { HeadingLevels } from "../Heading/HeadingTypes";
+import HelperErrorText from "../HelperErrorText/HelperErrorText";
+import Text from "../Text/Text";
+
+export interface StatusBadgeProps {
```
EdwinGuzman: This should be `ComponentWrapperProps`.

src/components/ComponentWrapper/ComponentWrapper.tsx
```diff
@@ -0,0 +1,44 @@
...
+  descriptionText?: string;
+  /** Optional string to set the text for a `Heading` component */
+  headingText?: string;
+  /** Optional string to set the text for a `HelperErrorText` component */
+  helperText?: string;
+}
+
+function ComponentWrapper(props: React.PropsWithChildren<StatusBadgeProps>) {
+  const { children, descriptionText, headingText, helperText } = props;
+  const hasChildren = children ? true : false;
```
EdwinGuzman: Nit: not necessary but you can do `!!children` to get a boolean value from it.

src/components/ComponentWrapper/ComponentWrapper.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const styles = useMultiStyleConfig("ComponentWrapper", { hasChildren });
+
+  if (!children) {
+    console.warn("Component Wrapper has no children.");
+  }
```
EdwinGuzman: You can use the `hasChildren` variable here to reuse it.
Also, it looks like, based on the `hasChildren` flag in the chakra styles, that the margin changes if there are no children. So that makes it seem like it's okay to not pass children because the UI will change accordingly. But here, a warning is logged. I'm wondering if it makes sense to log this if it's acceptable to not pass children. Otherwise, should it be required?

src/components/VideoPlayer/VideoPlayer.test.tsx
```diff
@@ -1,6 +1,7 @@
 import * as React from "react";
 import { render, screen } from "@testing-library/react";
 import { axe } from "jest-axe";
+// import renderer from "react-test-renderer";
```
EdwinGuzman: If you want snapshot tests, you should somehow pass an `id` to the `HelperErrorText` since that gets generated on every new render/test run.

src/components/VideoPlayer/VideoPlayer.test.tsx
```diff
@@ -21,7 +22,11 @@ describe("VideoPlayer", () => {
   describe("YouTube player", () => {
     beforeEach(() => {
       utils = render(
-        <VideoPlayer videoType={VideoPlayerTypes.YouTube} videoId={videoId} />
+        <VideoPlayer
+          videoType={VideoPlayerTypes.YouTube}
+          videoId={videoId}
+          className="video-player"
```
EdwinGuzman: I would prefer if you use the `data-testid` attribute in the component. This would make it clear that `data-testid` is only used for tests, whereas the `className` attribute implies that there are CSS styles for this component.

src/components/VideoPlayer/VideoPlayer.tsx
```diff
@@ -95,31 +73,41 @@ export default function VideoPlayer(
...
 
-  {
-    errored && modifiers.push("errored");
-  }
+  const variant = isInvalid
+    ? "invalid"
+    : getVariant(aspectRatio, VideoPlayerAspectRatios);
+  const styles = useMultiStyleConfig("VideoPlayer", { variant });
+
+  console.log(`variant == ${variant}`);
```
EdwinGuzman: This can be removed.

src/components/ComponentWrapper/ComponentWrapper.tsx
bigfishdesign13: Ugh.  A dead giveaway that I merely made a copy of the `StatusBadge` files.

src/components/ComponentWrapper/ComponentWrapper.tsx
bigfishdesign13: This was something that I thought about as I was putting this together.  You are correct that the UI will adjust accordingly if/when no children are passed.  And because of that, I was wondering if the component should have a different name.  Something like `DefaultBlock`, which would indicate that the component could be used as a standalone.  Of course, with that, I am making assumptions that it might be used standalone.  Thus far, there is no need for anything like that.

After typing this all out, I don't like that idea and I would prefer to make `children` required.

src/components/VideoPlayer/VideoPlayer.test.tsx
bigfishdesign13: OK, another giveaway.  I did have snapshot tests at one point during development, but I realized that some of the child elements within `ComponentWrapper` generated IDs and that the IDs would change every time a snapshot was generated and the tests would subsequently fail.  But I had time this evening, so the IDs are being set and the tests are passing.  Snapshots added.

src/components/VideoPlayer/VideoPlayer.tsx
bigfishdesign13: :)

src/components/VideoPlayer/VideoPlayer.test.tsx
bigfishdesign13: Yes, I was cheating.  Thank you for the suggestion of `data-testid`.  I was not aware of its usage.

src/components/ComponentWrapper/ComponentWrapper.tsx
bigfishdesign13: That's awesome.

EdwinGuzman (approved):

PR #736, opened 2021-10-25
DSD-500: Tabs QA update

EdwinGuzman:
Fixes JIRA ticket [DSD-500](https://jira.nypl.org/browse/DSD-500)

- Updates the `Tabs` component based on feedback from QA.
- This adds `useWindowSize` React hook to listen to the width of the device screen so it can accordingly update the with of the "slide" or tab for the mobile view. This allows the carousel to scroll to the next/previous slide with the right amount.

## How has this been tested?

Storybook.

### Checklist:

### Front End Review:


src/components/Tabs/Tabs.tsx
```diff
@@ -201,10 +215,10 @@ function Tabs(props: React.PropsWithChildren<TabsProps>) {
     <ChakraTabs
       id={id}
       // The following lazy loads each panel whenever it is needed.
```
bigfishdesign13: Why did yoiu move the `isLazy` prop but not the comment that describes the prop?

src/components/Tabs/Tabs.tsx
```diff
@@ -143,19 +143,33 @@ function Tabs(props: React.PropsWithChildren<TabsProps>) {
...
-  // Just an estimate of the tab width for the mobile carousel.
-  const tabWidth = 60;
-  const { prevSlide, nextSlide, carouselStyle } = useCarouselStyles(
+  const { prevSlide, nextSlide, carouselStyle, goToStart } = useCarouselStyles(
     tabProps?.children?.length,
     tabWidth
   );
+  React.useEffect(() => {
+    if (windowDimensions.width > 320) {
+      setTabWidth(mobileTabWidth);
```
bigfishdesign13: If the value of `windowDimensions` is greater than the width of a mobile device, why does this call to `setTabWidth` pass `mobileTabWidth` as the arg?

src/hooks/tests/useCarouselStyles.test.ts
```diff
@@ -90,6 +90,34 @@ describe("useCarouselStyles hook", () => {
...
 
+  it("resets the slide back to the first slide", () => {
+    // Mocking we have an array of five slides.
+    const slides = [1, 2, 3, 4, 5];
+    const { result } = renderHook(() => useCarouselStyles(slides.length));
+    const { nextSlide, goToStart } = result.current;
+
+    // Start at the first slide.
+    expect(result.current.carouselStyle).toEqual({
+      marginLeft: "-0%",
```
bigfishdesign13, line 101: Is the negative sign necessary for this value?

src/hooks/tests/useCarouselStyles.test.ts
```diff
@@ -90,6 +90,34 @@ describe("useCarouselStyles hook", () => {
...
+    act(() => nextSlide());
+    act(() => nextSlide());
+    expect(result.current.carouselStyle).toEqual({
+      marginLeft: "-300%",
+      transition: "all .5s",
+    });
+
+    act(() => goToStart());
+    expect(result.current.carouselStyle).toEqual({
+      marginLeft: "-0%",
```
bigfishdesign13, line 116: Same thing here.  Is the negative sign necessary for this value?

bigfishdesign13 (commented):
Ultimately, this looks good and appears to function well.  I just have a few questions.

src/components/Tabs/Tabs.tsx
EdwinGuzman: Good point, I'll update it.

src/components/Tabs/Tabs.tsx
EdwinGuzman: I'll update, bad variable name.

src/hooks/tests/useCarouselStyles.test.ts
EdwinGuzman: Yes, it's set as ```marginLeft: `-${currentSlide * slideWidth}%`,``` in the style (`useCarouselStyles`) so it can move correctly. But the css value of -0 and 0 are the same so I don't think in that sense it's a big deal.

bigfishdesign13 (approved):
Looks good.

PR #743, opened 2021-10-28
DSD-519: Changelog reminder update

EdwinGuzman:
Fixes JIRA ticket [DSD-519](https://jira.nypl.org/browse/DSD-519)

- Updates the regex for the changelog reminder Github action. _I think_ this will fix it. Checking the code inside the Github action, it creates a `new Regex(...)` with the value we passed. It's missing a forward slash at the end for the regex to work. We'll just have to text it in other PRs to make sure it works.

## How has this been tested?

With new PRs for the Github Action to trigger (or not in this case).

### Checklist:

### Front End Review:


bigfishdesign13 (approved):
Sounds good.

PR #758, opened 2021-11-08
DSD-292 Toggle Component

hokei:
Fixes JIRA ticket DSD-292

- Adds the `Toggle` component to the Design System

### Checklist:

- [x] The focus state doesn't work - any help on that would be very appreciated! 

### Front End Review:


EdwinGuzman:

Link:
- https://pr758-re1qhlzq4xfy2skppvyatiejkowhdbfk.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/618977d4908ae1204a549bae

EdwinGuzman:

Link:
- https://pr758-re1qhlzq4xfy2skppvyatiejkowhdbfk.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/618977d4908ae1204a549bae

hokei:
The datepicker snapshot seems to be failing because it defaults to the current date - which, of course, changes every day.  any ideas on how to fix this?

EdwinGuzman:

Link:
- https://pr758-re1qhlzq4xfy2skppvyatiejkowhdbfk.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/618977d4908ae1204a549bae

EdwinGuzman:
> The datepicker snapshot seems to be failing because it defaults to the current date - which, of course, changes every day. any ideas on how to fix this?

Can you `.skip` it for now. It does change every day but it was written in a way that _would_ work with daily updates. Except that today is also the `15` which is a value being used in the test, but that's my guess. These tests are always difficult but I'll get to it.

src/components/Button/Button.tsx
```diff
@@ -80,7 +80,6 @@ function Button(props: React.PropsWithChildren<ButtonProps>) {
   }
 
   styles = useStyleConfig("Button", { variant });
-
```
EdwinGuzman, line 74: Curious, was this your editor?

src/components/Toggle/Toggle.stories.mdx
```diff
@@ -0,0 +1,180 @@
...
+/>
+
+# Toggle
+
+| Component Version | DS Version |
+| ----------------- | ---------- |
+| Added             | `0.25.3`   |
+| Latest            | `0.25.3`   |
+
+<Description of={Toggle} />
```
EdwinGuzman, line 41: This doesn't do anything. You need to add esdoc style comment above the `Toggle` component function declaration for it to render here.

src/components/Toggle/Toggle.stories.mdx
```diff
@@ -0,0 +1,180 @@
...
+      </VStack>
+      <VStack>
+        <Toggle size={ToggleSizes.Large} defaultChecked={true} labelText="On" />
+        <Toggle size={ToggleSizes.Large} labelText="Off" />
+      </VStack>
+    </HStack>
+  </DSProvider>
+</Canvas>
+
+## Controlled Toggle
```
EdwinGuzman, line 90: This is great but can you add more details and maybe also the "uncontrolled" version? Similar to what's in the `Select` component https://development-6y1v6na1ceqnmojvc5akrm5ewys3cshs.tugboat.qa/?path=/docs/components-form-elements-select--select#getting-select-data-value or `CheckboxGroup` https://development-6y1v6na1ceqnmojvc5akrm5ewys3cshs.tugboat.qa/?path=/docs/components-form-elements-checkboxgroup--checkbox-group#getting-checkbox-data-values

src/components/Toggle/Toggle.stories.mdx
```diff
@@ -0,0 +1,180 @@
...
+      <Toggle isChecked={true} labelText="Controlled Toggle" />
+    </HStack>
+  </DSProvider>
+</Canvas>
+
+## Browser States
+
+<Canvas>
+  <DSProvider>
+    <Grid templateColumns="repeat(4, 1fr)">
```
EdwinGuzman: This is fine but maybe `SimpleGrid` might be easier since you can pass `columns={4}`.

src/components/Toggle/Toggle.stories.mdx
```diff
@@ -0,0 +1,180 @@
...
+      <Box>
```
EdwinGuzman: Why do you need the `Box` here?

src/components/Toggle/Toggle.stories.mdx
```diff
@@ -0,0 +1,180 @@
...
+  </DSProvider>
+</Canvas>
+
+## Layouts
+
+<Canvas>
+  <DSProvider>
+    <Box>
+      <Heading level={3}>Grouped</Heading>
+      <VStack>
```
EdwinGuzman: Pass in `spacing="s"` so it aligns with the Figma design.

src/components/Toggle/Toggle.stories.mdx
```diff
@@ -0,0 +1,180 @@
+import {
+  Meta,
+  Story,
+  Canvas,
+  ArgsTable,
+  Description,
+} from "@storybook/addon-docs/blocks";
+import { withDesign } from "storybook-addon-designs";
+import { HStack, VStack, Grid } from "@chakra-ui/react";
```
EdwinGuzman: Add `Box` here.

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,123 @@
...
+      isDisabled = false,
+      isInvalid = false,
+      isRequired = false,
+      labelText,
+      name,
+      size = ToggleSizes.Large,
+    } = props;
+    const footnote = isInvalid ? invalidText : helperText;
+    const ariaAttributes = {};
+    const onChange = props.onChange || onChangeDefault;
```
EdwinGuzman: Nit: get `onChange` from the destructure above.

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,123 @@
...
+                  defaultChecked,
+                })}
+            {...ariaAttributes}
+            __css={switchStyles}
+          />
+          <Box __css={styles.label}>
+            <FormLabel htmlFor={id}>{labelText}</FormLabel>
+          </Box>
+        </Box>
+        {(invalidText || helperText) && (
```
EdwinGuzman: You can use `footnote` here instead of `(invalidText || helperText)`.

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,123 @@
...
+          <Box __css={styles.helper}>
+            <HelperErrorText isInvalid={isInvalid} id={`${id}-helperText`}>
+              {isInvalid ? invalidText : helperText}
```
EdwinGuzman: You can use `footnote` here instead of `{isInvalid ? invalidText : helperText}`.

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,123 @@
...
+                  onChange,
+                }
+              : {
+                  defaultChecked,
+                })}
+            {...ariaAttributes}
+            __css={switchStyles}
+          />
+          <Box __css={styles.label}>
+            <FormLabel htmlFor={id}>{labelText}</FormLabel>
```
EdwinGuzman: Use the `DS`' `Label` rather than `FormLabel`.

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,123 @@
...
+    const switchStyles = useStyleConfig("Switch");
+    ariaAttributes["aria-label"] =
+      labelText && footnote ? `${labelText} - ${footnote}` : labelText;
+    const variant = isDisabled ? "disabled" : isInvalid ? "error" : "standard";
+
+    return (
+      <>
+        <Box
+          display="flex"
+          alignItems="center"
```
EdwinGuzman: Can you move `alignItems="center"` to the `styles` theme object?

src/theme/components/toggle.ts
```diff
@@ -0,0 +1,66 @@
+import { helperTextMargin } from "./global";
+
+// Style object for the Checkbox's label
```
EdwinGuzman: Should be `Toggle's`, right?

src/theme/components/toggle.ts
```diff
@@ -0,0 +1,66 @@
...
+    disabled: {
+      track: {
+        bg: "ui.gray.medium",
+      },
+    },
+    error: {
+      track: {
+        bg: "red.200",
+        _checked: {
+          bg: "red.500",
```
EdwinGuzman: Chakra exposes these values and they're available but we don't recommend using them and we don't use them ourselves. Use the `ui.error.primary` or `ui.error.secondary` values instead.

src/theme/components/toggle.ts
```diff
@@ -0,0 +1,66 @@
...
+          opacity: 1,
+        },
+      },
+    },
+    disabled: {
+      track: {
+        bg: "ui.gray.medium",
+      },
+    },
+    error: {
```
EdwinGuzman: This doesn't need to be a variant. When you pass in the `isInvalid` prop to a Chakra component, you can update that state by targeting the `_invalid` key. Above in the `baseStyle`, you can add:
```
baseStyle: {
  opacity: 0.4,
  track: {
    p: "4px",
  },
  _invalid: {
    track: {
      bg: "red.200",
      _checked: {
        bg: "red.500",
      },
    },
  },
},
```

src/theme/components/toggle.ts
```diff
@@ -0,0 +1,66 @@
+import { helperTextMargin } from "./global";
+
+// Style object for the Checkbox's label
+const baseStyleLabel = {
+  fontSize: "sm",
```
EdwinGuzman: This works but should be `"-1"` (I think). The values to use for font sizes are in the `typography.ts` file.

src/theme/components/toggle.ts
EdwinGuzman: You might be able to write everything without needing to have variants involved.

src/components/Toggle/Toggle.stories.mdx
```diff
@@ -0,0 +1,180 @@
...
+
+<Canvas>
+  <Story
+    name="Toggle"
+    args={{
+      labelText: "Test Label",
+      size: "lg",
+      defaultChecked: false,
+      helperText: "This is the helper text!",
+      id: "",
```
EdwinGuzman: For the purposes of Storybook, disable this `id` prop by adding it to the `argsType` prop on line 28. I was trying to toggle by clicking on the label and it wasn't working. It failed because it's set to `""` by default and so the label/input linking doesn't work.

EdwinGuzman (changes requested):
Needs some updates but on the right track.

src/components/Button/Button.tsx
hokei: I probably put a console.log here to see what was going on and then took it out and it registered as changed.

src/components/Toggle/Toggle.stories.mdx
hokei: somehow without `Box` it doesn't show up right?  Is `VStack` considered a block element?

src/components/Toggle/Toggle.stories.mdx
hokei: For some reason auto-import doesn't work on mdx files for me - have you found a VSCode extension that will help with this?

src/theme/components/toggle.ts
```diff
@@ -0,0 +1,63 @@
...
+      _disabled: {
+        bg: "ui.gray.medium",
+        _checked: {
+          opacity: 0.4,
+        },
+      },
+    },
+    thumb: {
+      _disabled: {
+        bg: "ui.error.primary",
```
hokei, line 47: Not sure why this isn't working?  Putting a color under `thumb` works

hokei:
For some reason I'm still not able to focus on the toggle via tabbing and keyboard - anyone else getting that?

EdwinGuzman:

Link:
- https://pr758-re1qhlzq4xfy2skppvyatiejkowhdbfk.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/618977d4908ae1204a549bae

hokei:
Screen Shot 2021-12-02 at 9 21 20 AM

How do you get snapshots not to fail on auto-generated classnames?

src/components/Toggle/Toggle.test.tsx
```diff
@@ -0,0 +1,140 @@
...
+  it("Sets the Toggle's ID", () => {
+    render(<Toggle id="inputID" labelText="Test Label" />);
+    expect(screen.getByRole("checkbox")).toHaveAttribute("id", "inputID");
+  });
+
+  it("Calls the UUID generation function if no id prop value is passed", () => {
+    generateUUIDSpy = jest.spyOn(generateUUID, "default");
+    expect(generateUUIDSpy).toHaveBeenCalledTimes(0);
+    render(<Toggle labelText="Test Label" />);
+    expect(generateUUIDSpy).toHaveBeenCalledTimes(2);
```
hokei: I can't figure out why this was called twice - it makes me nervous that I made a mistake and it's rendering twice.  Thoughts?

EdwinGuzman:

Link:
- https://pr758-re1qhlzq4xfy2skppvyatiejkowhdbfk.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/618977d4908ae1204a549bae

src/components/Toggle/Toggle.stories.mdx
EdwinGuzman: Yeah, `VStack` renders a "div" which is what `Box` renders. `VStack` just has extra props added to it.

src/components/Toggle/Toggle.stories.mdx
EdwinGuzman: I haven't. My vscode also doesn't tell me if there are syntax or variable issues in the file so it's a bit rough at times.

EdwinGuzman:
BTW, don't worry about the "CI / Lint" Github Action. It currently runs the "CI / Test" action but will be fixed later.

src/theme/components/toggle.ts
EdwinGuzman: Yeah, that's annoying. Maybe it doesn't pick this up directly.

EdwinGuzman:
> How do you get snapshots not to fail on auto-generated classnames?

Run `npm test -- --updateSnapshot`.

EdwinGuzman:
It doesn't look like the `invalid` state is right.

> For some reason I'm still not able to focus on the toggle via tabbing and keyboard - anyone else getting that?

It works! The outline is just not visible. You can tab into it and toggle by pressing the "space" button, it's just not obvious because of the missing outline. Not sure how it disappeared though.

src/theme/components/toggle.ts
EdwinGuzman: Oh wait, I think this should be `_invalid`. This is related to my other comment about the invalid state not displaying correctly.

hokei:
> Run npm test -- --updateSnapshot 

Does this need to be run constantly, or only just there are changes to the component?  What triggers the classnames to regenerate?

EdwinGuzman:
> Does this need to be run constantly, or only just there are changes to the component? What triggers the classnames to regenerate?

Only when there are stylistic changes to a component you are intentionally doing. If a test fails because you accidentally or intentionally updated the background color (for example), then that's great that it pointed it out. But if you purposely made the change and forgot to update the test, it is a bit annoying because it's a false alarm. Only then would you want to update the snapshot tests.

hokei:
Ah ok!  I now understand why my snapshots were failing - somehow I thought they were regenerating on a rebuild or something and kind of panicked

src/components/DatePicker/DatePicker.test.tsx
```diff
@@ -543,7 +543,7 @@ describe("DatePicker", () => {
     // Note: Have to add initial dates so that the snapshot tests always
     // pass. Otherwise, it'll use the _current_ date which changes
     // based on the day it is tested and is not what we want.
-    it("renders the UI snapshot correctly", () => {
+    it.skip("renders the UI snapshot correctly", () => {
```
EdwinGuzman: When you merge from `development`, this can be removed.

EdwinGuzman: Including the update below in this file.

src/components/Toggle/Toggle.stories.mdx
```diff
@@ -0,0 +1,191 @@
...
+  parameters={{
+    design: {
+      type: "figma",
+      url:
+        "https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Main?node-id=31581%3A302",
+    },
+    jest: ["Toggle.test.tsx"],
+  }}
+  argTypes={{
+    id: { table: { disable: true } },
```
EdwinGuzman: Can you do `control: false` instead of `table: { disable: true }`? It's explained more in #788 , but I think it's better to display all props and to disable the control rather than hide it completely. If props are hidden, devs won't know what's available unless they look at the source.

src/components/Toggle/Toggle.test.tsx
EdwinGuzman: Every DS component calls the `generateUUID` function unless an `id` is passed to it. In this case, it's the `Label` component that needs an explicit `id` or else it'll call this function. You _can_ make it called twice and add a comment that the second call is from the `Label` or you can pass an `id` value similar to the value for the `HelperErrorText` component.

src/theme/components/toggle.ts
EdwinGuzman: The `track` and `label` in Chakra's `Switch` have "invalid" style objects but the `thumb` does not... so what you can do is add a specific rule to the `track` object:

```
track: {
  ...
  _invalid: {
    border: "1px solid",
    borderColor: "ui.error.primary",
    bg: "inherit",
    "> span": {
      bg: "ui.error.primary",
    },
  },
}
```

EdwinGuzman: Screen Shot 2021-12-06 at 5 57 54 PM

src/theme/components/toggle.ts
```diff
@@ -0,0 +1,63 @@
...
+      },
+    },
+  },
+  defaultProps: {
+    colorScheme: "white",
+  },
+};
+
+const Toggle = {
+  parts: ["helper", "switch"],
```
EdwinGuzman: This can just have `"helper"`.

EdwinGuzman:
> For some reason I'm still not able to focus on the toggle via tabbing and keyboard - anyone else getting that?

This works functionally... but I'm not sure why the style doesn't show up. It seems to be, from Chakra's site and inspecting the element, that it's the `track` that gets the outline or box-shadow when the component is focused. I can't seem to add any new styles to it but I'll try again soon. @bigfishdesign13 , it should be `box-shadow` for the focus outline, right?

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,120 @@
...
+                }
+              : {
+                  defaultChecked,
+                })}
+            {...ariaAttributes}
+            __css={switchStyles}
+          />
+          <Box __css={styles.label}>
+            <Label htmlFor={id}>{labelText}</Label>
+          </Box>
```
EdwinGuzman: You don't need this. Chakra's `Switch` component is a label wrapping an input. So you can do:
```
...
  __css={switchStyles}
>
  {labelText}
</Switch>
```

Right now, there are two labels rendering so we should make use of the existing one coming from Chakra.

EdwinGuzman:
For the latest round of updates:
- the extra label should be removed,
- targeting the `thumb` key doesn't seem to work in the theme object so you can add a direct "> span" CSS rule,
- tabbing and using the keyboard to toggle the input works, but the focus/box-shadow is gone,
- the test with the `generateUUIDSpy` being called twice can be fixed if the label is removed or if it is passed an `id`.

src/theme/components/toggle.ts
```diff
@@ -0,0 +1,63 @@
...
+const baseStyle = {
+  alignItems: "center",
+  label: baseStyleLabel,
+  helper: baseStyleHelper,
+};
+
+const Switch = {
+  baseStyle: {
+    opacity: 0.4,
+    track: {
```
bigfishdesign13, line 14: To address the focus problem, add the following to `track`:
```
_focus: {
  outline: "2px solid",
  outlineColor: "ui.focus",
  outlineOffset: "2px",
  zIndex: "9999",
},
```

Similar to `activeFocus` in global.ts, it would make sense to also add this style object for reuse.  Maybe call it `activeFocusNoBorder`.

Cc @EdwinGuzman

src/theme/components/toggle.ts
hokei: how did you figure out that the "invalid" object was missing in "thumb" ?

src/components/Toggle/Toggle.tsx
hokei: So looking there's something strange going on here.  When I just do the suggestion above, the text appears outside the `label` element. 
Screen Shot 2021-12-15 at 3 16 24 PM

Looking at the `switch` documentation, it looks like they have two `label` elements - https://chakra-ui.com/docs/form/switch#usage

Does this look like a bug in Chakra?  I'm happy to send them a PR if that's the case
Screen Shot 2021-12-15 at 3 21 54 PM

EdwinGuzman:

Link:
- https://pr758-re1qhlzq4xfy2skppvyatiejkowhdbfk.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/618977d4908ae1204a549bae

EdwinGuzman:

Link:
- https://pr758-re1qhlzq4xfy2skppvyatiejkowhdbfk.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/618977d4908ae1204a549bae

EdwinGuzman:

Link:
- https://pr758-re1qhlzq4xfy2skppvyatiejkowhdbfk.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/618977d4908ae1204a549bae

src/components/Toggle/ToggleSizes.tsx
```diff
@@ -0,0 +1,4 @@
+export enum ToggleSizes {
+  Large = "lg",
```
bigfishdesign13: To follow the same patterns as other components, the string values should not be abbreviated.  `Icon` and `Image` are two components that use the patterns that I am recommending.
```
export enum ToggleSizes {
  Large = "large",
  Small = "small",
}
```

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,118 @@
...
+};
+
+/**
+ * Component that renders Chakra's `Switch` component along with NYPL defaults.
+ */
+const Toggle = React.forwardRef<HTMLInputElement, ToggleProps>(
+  (props, ref?) => {
+    const {
+      additionalStyles = {},
+      invalidText,
```
bigfishdesign13: Move below `id` in list of props.

src/index.ts
```diff
@@ -113,6 +113,7 @@ export {
   TextInputRefType,
 } from "./components/TextInput/TextInput";
 export { TextInputTypes } from "./components/TextInput/TextInputTypes";
+export { default as Toggle } from "./components/Toggle/Toggle";
```
bigfishdesign13, line 120: Make sure to also export the `ToggleSizes` enum.

src/components/Toggle/Toggle.stories.mdx
```diff
@@ -0,0 +1,174 @@
...
+| Component Version | DS Version |
+| ----------------- | ---------- |
+| Added             | `0.25.3`   |
+| Latest            | `0.25.3`   |
+
+<Description of={Toggle} />
+
+The Toggle component is used as an alternative for the Checkbox component and returns a boolean response. Like the Checkbox component, the Toggle component can display a label to the right of the toggle and helper/error text below the toggle.
+
+## Best Practices
```
bigfishdesign13: This can be demoted to `### Best Practices` or even just to **bold** text.  I feel bold might be the best option, as this is just featured info within the description/introduction section.

src/components/Toggle/Toggle.stories.mdx
```diff
@@ -0,0 +1,174 @@
...
+    {(args) => <Toggle {...args} />}
+  </Story>
+</Canvas>
+
+<ArgsTable story="Toggle" />
+
+## Sizes
+
+<Canvas>
+  <DSProvider>
```
bigfishdesign13, line 74: Flips these so the Large set is first and use `SimpleGrid` to spread out the groups.  Also, similar to the `Browser States` section, add a `Heading` for each set (Large, Small).

bigfishdesign13 (changes requested):
Looking good.  I have some requests focused on the Story presentation.

src/theme/components/toggle.ts
EdwinGuzman: On the [Switch page](https://chakra-ui.com/docs/form/switch ), there is a link to the style source: https://github.com/chakra-ui/chakra-ui/blob/main/packages/theme/src/components/switch.ts

Typically, Chakra components should have this `_invalid` state available to them so I _think_ I just added it and it didn't work. If not that... then I probably saw that in the source code, the "data-invalid" attribute gets added to Chakra divs. That's a good indication of whether the underscore helper keys are available or not. If you try passing in `isRequired`, you should see `data-required`. It seems that's how Chakra keeps track of this. I could also be wrong and thinking of something else but I can't say I remember exactly how I came to the conclusion above.

EdwinGuzman:
The last update looks great! I would say, maybe add a gray border on the default state because when you turn on `isInvalid`, there's a subtle change in the layout. The text shifts just a little bit.

This is not a PR problem but a design problem: @bigfishdesign13 I noticed that the `isRequired` prop doesn't visually change the component and the Figma designs don't have "Required" text in any example. Functionally, this works fine so this can be a ticket for a future iteration.

src/components/Toggle/ToggleSizes.tsx
hokei: Because "lg" and "sm" are specifically the values expected by Chakra, I added this: https://github.com/NYPL/nypl-design-system/pull/758/files#diff-7ae8cfe58c40cccbd1d760ad907b63d018daeebebb15d0c63a33a8d15c38cfbcR91 which I'm pointing out in case there's a specific pattern we want to follow

src/components/Notification/Notification.stories.mdx
```diff
@@ -27,8 +27,7 @@ import DSProvider from "../../theme/provider";
   parameters={{
     design: {
       type: "figma",
-      url:
-        "https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/?node-id=10734%3A812",
+      url: "https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/?node-id=10734%3A812",
```
hokei, line 30: Are these figma changes auto-generated?  Wondering if it's a bad merge somewhere - they look right but should I revert?

src/components/Toggle/ToggleSizes.tsx
EdwinGuzman: I'm okay with this. I think this may be the first time we use Chakra's own `size` prop to set the specific size we want.

src/components/Notification/Notification.stories.mdx
EdwinGuzman: It might be prettier making this change.

EdwinGuzman:

Link:
- https://pr758-re1qhlzq4xfy2skppvyatiejkowhdbfk.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/618977d4908ae1204a549bae

EdwinGuzman:
One minor thing before I approve: the label extends the full width of the container so you can click on the white space and it will trigger the input. You can add `width: "fit-content",` to the label (I think) or the component itself so that doesn't happen.

src/components/Toggle/ToggleSizes.tsx
bigfishdesign13: If we can rely on the typescript enums, then yes, this is fine.  But what about scenarios where the enums are not being used?  Then our pattern changes.

bigfishdesign13:
> This is not a PR problem but a design problem: @bigfishdesign13 I noticed that the `isRequired` prop doesn't visually change the component and the Figma designs don't have "Required" text in any example. Functionally, this works fine so this can be a ticket for a future iteration.

@EdwinGuzman The `Toggle` component is treated the same as `Checkbox` and `Radio`, where nothing is indicated visually when the individual inputs are set as required.  Do we need a `ToggleGroup` component?

EdwinGuzman:
> The Toggle component is treated the same as Checkbox and Radio, where nothing is indicated visually when the individual inputs are set as required. Do we need a ToggleGroup component?

Yeah, that seems like the pattern we are taking so maybe we should make a separate ticket for that grouping.

Edit: Actually, I think in most use cases, a page or a form would only need one `Toggle`. If this is the case and it should be required, there's no visual indication that it is required. But you're right that the `Checkbox` can be used individually and that has this same concern (not the `Radio` since having a single one doesn't make sense).

src/components/Notification/Notification.stories.mdx
hokei: The only difference was the line break - my vscode mdx lint settings are off and I'm not quite sure how to fix it.  If it isn't a huge deal I'd rather not deal with it right now either

EdwinGuzman:

Link:
- https://pr758-re1qhlzq4xfy2skppvyatiejkowhdbfk.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/618977d4908ae1204a549bae

EdwinGuzman (approved):
Great! Just remember to update from `development` before merging.

src/components/Toggle/Toggle.stories.mdx
```diff
@@ -0,0 +1,176 @@
...
+    className: { control: false },
+  }}
+/>
+
+# Toggle
+
+| Component Version | DS Version |
+| ----------------- | ---------- |
+| Added             | `0.25.3`   |
+| Latest            | `0.25.3`   |
```
bigfishdesign13: These both need to be changed to `0.25.8`.

bigfishdesign13 (approved):
Looks good, but make sure to update the version numbering on the story page before you merge.

EdwinGuzman:

Link:
- https://pr758-re1qhlzq4xfy2skppvyatiejkowhdbfk.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/618977d4908ae1204a549bae

EdwinGuzman:

Link:
- https://pr758-re1qhlzq4xfy2skppvyatiejkowhdbfk.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/618977d4908ae1204a549bae

PR #857, opened 2022-02-07
DSD-759: Fixing Breadcrumbs issues

EdwinGuzman:
Fixes JIRA ticket [DSD-759](https://jira.nypl.org/browse/DSD-759)

- Updates the `Breadcrumbs` aria-label to "Breadcrumbs". This should not be modified and having more than one `<nav>` element with the same aria-label is not correct. We can't prevent this from the DS but can only give the pattern for using it. https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html
- Updated a CSS pseudo-class that fails because of the `Emotion` package and how styled components are rendered on the server and client side.
- Note, there server/client id not matching issue is not related to the component but just the app passing in ids to the correct components.

## How has this been tested?

Locally and Storybook.

- There should only be one `<nav>` HTML element with aria-label set to "Breadcrumbs". Since the DS `Breadcrumbs` component renders this landmark, only one should be rendered on the page.

### Checklist:

### Front End Review:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 ## Prerelease
 
+### Fixes
+
+- Fixing `Breadcrumbs` related logging issue with a CSS pseudo-selector and setting the `aria-label` to "Breadcrumbs". Adding an accessibility test that should fail when more than one `Breadcrumbs` component is rendered on a page since that landmark should only be rendered once on a web page.
+
 ## 0.25.9 (February 3, 2022)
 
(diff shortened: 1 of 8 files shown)


EdwinGuzman:

Link:
- https://pr857-mpwklchmwpfjjm9xlykvvceprwn6tvb0.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/62017c92046a4b8d29aa95dd

bigfishdesign13 (approved):
Make sure to bump up the `Latest Version` number on the story page, but otherwise it looks good.

PR #863, opened 2022-02-10
DSD-763: Notification a11y fixes

EdwinGuzman:
Fixes JIRA ticket [DSD-763](https://jira.nypl.org/browse/DSD-763)

- Adds an `aria-label` to the `Notification` component.
- Adds accessibility documentation for the `Notification` component.
- Note: the `aside` and `role="complementary"` are both the same so using both is redundant and why `role="complementary"` is left out.

## How has this been tested?

Locally and in the  DS test app.

- This component itself is not an accessibility concern but how it is used _is_. We have to make it clear that it should not be used within `header` or `footer` HTML landmark elements. _That_ is the actual accessibility issue of rendering a landmark inside another landmark. Even though this is the rule, rendering a `Notification` component inside a `main` landmark element is acceptable.

### Checklist:

### Front End Review:


EdwinGuzman:

Link:
- https://pr863-kcjqgayt2dq7bdtgr972dgykmyjmzlcd.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/62056b30f8ba3c5cfaa6e1e6

src/components/Notification/Notification.tsx
```diff
@@ -30,6 +30,8 @@ type BasePropsWithoutAlignText = Omit<BaseProps, "alignText">;
 type BasePropsWithoutCentered = Omit<BaseProps, "centered">;
 
 export interface NotificationProps extends BasePropsWithoutAlignText {
+  /** Label used to describe the `Notification`'s aside HTML element. */
+  ariaLabel?: string;
```
bigfishdesign13, line 34: We've gone out of our way to remove explicit `aria-label` props from other components.  I'm wondering if there is another way we can handle this here.  The pattern we've used elsewhere is to require some type of title text, which would in turn be used to populate `aria-label`, and then control the visibility of the title element with another boolean prop.  I feel if we go that route here, then a better prop name would be needed for `  notificationHeading`.  What do you think?

bigfishdesign13 (commented):
I have some concerns about this update.  Please see my comment.

src/components/Notification/Notification.tsx
EdwinGuzman: Yeah, that's a good point for our DS pattern. I was thinking that it might make sense to use the text from the `notificationHeading` prop but it's not always applicable and the heading is not required. I was trying to follow the description from [DSD-209](https://jira.nypl.org/browse/DSD-209) but also what's in [Metronome](http://themetronome.co/components/notification/). Specifically, the example aria-labels in Metronome have "Library-wide notification" and "Location notification" which are good but perhaps not the best to display in the actual `notificationHeading` prop. 

So I was thinking of the case where it might be something like
```
<Notification
  ariaLabel="Library-wide notification"
  notificationHeading="Snow day closure"
  notificationContent="All NYPL branches will have a delay opening on ...."
/>
```
and the case where there are multiple `Notifications` with the same heading string but different content. The aria-labels for each `aside` _should_ be unique.

We can use the `notificationHeading` prop and add more patterns for users to implement which works for me.

bigfishdesign13: The original docs for `Notification` were very narrow in scope and the a11y examples that you see in Metronome right now are remnants that were never updated.  So yes, the cases that you are thinking about start to fall apart.

Do you want to have a call about how to execute this change?  Or do you have a sense of how you would refactor for this?

EdwinGuzman: I don't. Let's set up a call about it soon.

EdwinGuzman:

Link:
- https://pr863-kcjqgayt2dq7bdtgr972dgykmyjmzlcd.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/62056b30f8ba3c5cfaa6e1e6

EdwinGuzman:

Link:
- https://pr863-kcjqgayt2dq7bdtgr972dgykmyjmzlcd.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/62056b30f8ba3c5cfaa6e1e6

EdwinGuzman:
This has been updated based on the updated AC in https://jira.nypl.org/browse/DSD-763

EdwinGuzman:

Link:
- https://pr863-kcjqgayt2dq7bdtgr972dgykmyjmzlcd.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/62056b30f8ba3c5cfaa6e1e6

src/theme/components/notification.ts
```diff
@@ -2,31 +2,34 @@ import { NotificationTypes } from "../../components/Notification/NotificationTyp
...
+      notificationType === NotificationTypes.Warning
+    ) {
+      bg = "ui.gray.x-light-cool";
     }
     return {
       bg,
       display: "flex",
       fontSize: "-1", // slightly smaller than the default size
-      padding: "var(--nypl-space-s) var(--nypl-space-l)",
+      padding: "var(--nypl-space-s)",
```
bigfishdesign13: This padding attribute as well as the conditional attributes on lines 18 and 19 should be moved into the styles for the `container`.  The content area here need to act the same as other full width component (ex. `Breadcrums` and the content area of `Template`) where there is a `1280px` max-width and a constant `16px` padding on both sides.  I did some testingm locally and it seems to work fine when the attributes are moved in with the other styles for `container`.

src/theme/components/notification.ts
```diff
@@ -62,6 +65,10 @@ const NotificationContent = {
         ? "calc(var(--nypl-space-m) + var(--nypl-space-s))"
         : null,
       width: "100%",
+      // Links should always be black and underlined.
+      a: {
+        color: "ui.black",
```
bigfishdesign13, line 70: I did not state this initially (truthfully, I am just thinking of it now), but the `hover` state also needs to be black.  There should be no color change, ever.  We don’t have `visited` or `active` colors, so it is only `hover` that we need to worry about.

bigfishdesign13 (changes requested):
Most changes are good, but a few of the styles need to be adjusted.

src/theme/components/notification.ts
bigfishdesign13: I was also curious why this component is using CSS vars and not the theme object.

EdwinGuzman: The above makes sense. I think it's how we initially wrote this before using the theme object.

EdwinGuzman:
Updated.

bigfishdesign13 (approved):
Very nice.

PR #898, opened 2022-03-14
Dsd 851/pagination current page update

oliviawongnyc:
Fixes JIRA ticket [DSD-851](https://jira.nypl.org/browse/DSD-851).

- Replaces the `Pagination` component's `initialPage` prop with `currentPage`, a value that updates the selected page programmatically without the user explicitly requesting it.
- Updates tests, documentation, and Storybook to reflect this change.
- Moves logic from Pagination.stories.tsx to utils.ts, for simplifying and streamlining.

## How has this been tested?

- Tested locally by linking the DS and the Test App and checking the Search and Filter page.
- All new and existing unit tests pass.

- No, but it should be noted that this change allows the selected page in the `Pagination` component to be changed without the user specifically asking for the page to be changed – for example, if the user is on page 3, but should be brought back to page 1 of a set of results after a new search.

### Checklist:

### Front End Review:

- [x] View [the example in Storybook]()


EdwinGuzman:

Link:
- https://pr898-o4azsjui7dpl8xhdfs6zybijagsl926p.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/622f91f5bce4cc1c7350c833

src/components/Pagination/Pagination.stories.mdx
```diff
@@ -90,7 +91,7 @@ const getPageHref = (selectedPage: number) => {
 
 It is preferred to use the `Pagination` component with the `getPageHref`
 function passed as a prop and that the URL is updated whenever a page is
-changed. However, the optional `onPageChange` prop is available in case URL
+changed. However,å the optional `onPageChange` prop is available in case URL
```
EdwinGuzman: Typo character after "However"

src/components/Pagination/Pagination.stories.mdx
```diff
@@ -27,7 +29,7 @@ import { getCategory } from "../../utils/componentCategories";
     className: { control: false },
     getPageHref: { control: false },
     id: { control: false },
-    initialPage: { table: { defaultValue: { summary: 1 } } },
+    currentPage: { table: { defaultValue: { summary: 1 } } },
```
EdwinGuzman: Nit: order this alphabetically.

EdwinGuzman (changes requested):
This is very close but the "Pagination with Unchanging URL" doesn't work like it use to: when you click on a number link or "previous"/"next", the `Pagination` component flashes but the page doesn't change. The resulting current page is updated correctly, but I think your `useEffect` where it does `currentPage !== selectedPage` doesn't update correctly. For this example, the `currentPage` is 7 and doesn't change, but the `selectedPage` does. So if I click on "8", the console logs "Current page is 8" but because it's not 7, the internal state sets it back to 7 so the UI does not change.

Does this make sense?

src/components/Pagination/Pagination.stories.mdx
```diff
@@ -73,14 +75,13 @@ const getPageHref = (selectedPage: number) => {
     name="Pagination with URL Updates"
```
EdwinGuzman, line 79: Update line 42 to version `0.25.12`.

EdwinGuzman:
We chatted about this but adding it here so it doesn't get lost in Slack:

- keep the `initialPage` prop and add `currentPage` to force the component to go to a certain page.
- add a section in the storybook doc that goes into the `currentPage` prop: when and how you’ll it, etc
- add an example: you can probably replicate the existing client-side pagination example but render a button that when you click on it, it updates a value that goes into the `currentPage` prop so you can see this forced update in action
- log a warning when `currentPage` is used with the `getPageHref` function

EdwinGuzman:

Link:
- https://pr898-o4azsjui7dpl8xhdfs6zybijagsl926p.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/622f91f5bce4cc1c7350c833

src/components/Pagination/Pagination.stories.mdx
```diff
@@ -110,13 +114,60 @@ const onPageChange = (currentPage: number) => {
...
+        console.log(`Current page: ${selectedPage}`);
+      },
+      pageCount: 100,
+    }}
+  >
+    {(args) => <Pagination {...args} />}
+  </Story>
+</Canvas>
+
+### Forcing a Page Change Programatically
```
EdwinGuzman: "Programmatically" has two 'm's.

src/components/Pagination/Pagination.stories.mdx
```diff
@@ -110,13 +114,60 @@ const onPageChange = (currentPage: number) => {
...
+
+Generally, the selected page will change when a user clicks on a page number or
```
EdwinGuzman: This is great but I think the last sentence should be first to make it clear that this only works for the clientside unchanging URL version of the component. Also, make a note that it can only be used in combination with the `onPageChange` function.

src/components/Pagination/Pagination.stories.mdx
```diff
@@ -110,13 +114,60 @@ const onPageChange = (currentPage: number) => {
...
+when you want to programatically change the page without the user explicitly
+requesting it (for example, if you want a user to be brought back to page 1 after
+entering a new search term). In this case, you can use the `currentPage` prop to
+force this change. Note that the `currentPage` prop only works with the unchanging
+URL `Pagination`.
+
+The following example shows that the user can be brought to page 1 without clicking
+on the `Pagination` component.
+
+<Canvas withToolbar>
```
EdwinGuzman: This is more of a further example and explanation of the previous Story, so I don't think it needs this Story here. We also don't want to have Controls for this since it's not a variant. You can swap the `Story` wrapper for a `DSProvider` wrapper and not have the `args`; it's fine to hardcode the props for this example.

```
<Canvas>
  <DSProvider>
    {() => {
      const [page, ...
      // ...
    }}
  </DSProvider>
</Canvas>

You can also make the internal `{() => {....}}` into a component if you want.
```

src/components/Pagination/Pagination.test.tsx
```diff
@@ -1,4 +1,6 @@
 import * as React from "react";
+// import { render, screen } from "@testing-library/react";
```
EdwinGuzman: Okay to remove this.

src/components/Pagination/Pagination.tsx
```diff
@@ -10,14 +10,19 @@ import generateUUID from "../../helpers/generateUUID";
...
   /** The callback function that takes a page number and returns a string
    * to use for a link's `href` attribute. This is used when the current
    * page should refresh when navigating. */
   getPageHref?: undefined | ((pageNumber: number) => string);
   /** ID that other components can cross reference for accessibility purposes. */
   id?: string;
-  /** The current page selected. */
-  initialPage: number;
+  /** The initially selected page (default value is 1). If using with `getPageHref`,
+   * this value will reset with each new page refresh. */
```
EdwinGuzman: The second sentence depends on a consuming app's implementation, so it's okay to remove or it should be rephrased.

src/components/Pagination/Pagination.tsx
```diff
@@ -31,24 +36,39 @@ export interface PaginationProps {
...
-  const [currentPage, setCurrentPage] = useState<number>(initialPage);
+  const refCurrentPage = useRef(currentPage);
+  const [selectedPage, setSelectedPage] = useState<number>(initialPage);
   const styles = useMultiStyleConfig("Pagination", {});
-  const previousPageNumber = currentPage - 1;
-  const nextPageNumber = currentPage + 1;
+  const previousPageNumber = selectedPage - 1;
+  const nextPageNumber = selectedPage + 1;
+
+  React.useEffect(() => {
```
EdwinGuzman, line 54: Add a short comment about how this works.

src/components/Pagination/Pagination.tsx
```diff
@@ -31,24 +36,39 @@ export interface PaginationProps {
...
     return null;
   }
   if (getPageHref && onPageChange) {
     console.warn(
-      "Props for both `getPageHref` and `onPageChange` are passed. Will default to using `getPageHref`."
+      "NYPL Reservoir Pagination: Props for both `getPageHref` and `onPageChange` are passed. Will default to using `getPageHref`."
+    );
+  }
+
+  if (getPageHref && currentPage) {
```
EdwinGuzman, line 71: 👍  can you add a test for this?

EdwinGuzman (commented):
Great update; it needs some changes in the Storybook file, one new test, and comment updates in the component file. The component itself works well!

src/components/Pagination/Pagination.stories.mdx
oliviawongnyc: At least I was consistent ;)!

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 - Updates the `SearchBar`'s `textInputProps` prop object to not require the `placeholder` property.
 - Updates the `Logo` component to include new variants for `FirstBook` and `Open eBooks`.
 - Updates the border and text styles for the `Select` component.
+- Adds a `currentPage` prop to the `Pagination` component, a value that updates the selected page programmatically without the user explicitly requesting it.
```
EdwinGuzman, line 24: This might be better in the `### Adds` section above.

EdwinGuzman (approved):
👍

PR #909, opened 2022-03-21
DSD-860: Updating the Image's caption text to font size 12px

EdwinGuzman:
Fixes JIRA ticket [DSD-860](https://jira.nypl.org/browse/DSD-860)

- Uses the `text.tag` design token to update the `Image`'s caption and credit.

## How has this been tested?

Storybook and updated snapshot tests.

- N/A

### Checklist:

### Front End Review:


src/theme/components/image.ts
```diff
@@ -85,6 +85,7 @@ const CustomImage = {
     },
     figcaption: {
       fontStyle: "italic",
+      fontSize: "text.tag",
```
oliviawongnyc, line 88: Why are we switching which element the fontSize is added to?

EdwinGuzman: The element where `figcaption` is applied is the parent of the element where `captionWrappers` is applied. So it covers both "captionWrappers".

oliviawongnyc: Ah, okay. Thank you!

oliviawongnyc (approved):

bigfishdesign13 (approved):
Looks good.  Simple.

PR #907, opened 2022-03-21
DSD-847: Storybook Sidebar

EdwinGuzman:
Fixes JIRA ticket [DSD-847](https://jira.nypl.org/browse/DSD-847)

## This PR does the following:

- Updates Storybook's sidebar to:
  - Welcome
  - Chakra UI
  - Style Guide
  - Components
    - Chakra Exports
  - Hooks
  - Development Guide

I wanted to remove the `Autosuggest` story which was the only one under the existing "Development Guide", but it wasn't easy to. We _might_ still need this story for those using an `Autosuggest`-like component in their app, like the Library Card App.

The `useCarouselStyles` page is empty for now. It's just for the category to show up in the sidebar.

## How has this been tested?

Locally.

## Accessibility concerns or updates

- N/A

### Checklist:

- [ ] I have updated the Storybook documentation accordingly.
- [ ] I have added relevant accessibility documentation for this pull request.
- [ ] All new and existing tests passed.

### Front End Review:

- [ ] View [the example in Storybook]()

PR #911, opened 2022-03-21
DSD-863: Adding mainContent id to the TemplateAppContainer and a11y guides

EdwinGuzman:
Fixes JIRA ticket [DSD-863](https://jira.nypl.org/browse/DSD-863)

- Adds a `contentId` prop to the `TemplateAppContainer` component and adds an `id` prop to the `TemplateContent` component. The default value of these `id`s are both set to "mainContent" and it will render as an attribute on the `main` HTML element. This is used as the target for the skip navigation link in consuming applications.
- Adds an `Accessibility Guide` section to Storybook with a "Skip Navigation" page: https://pr911-zocq9v2lgrhu7brmfjuiyis9wjhfcey3.tugboat.qa/?path=/story/accessibility-guide-skip-navigation--page
  - PR #907 is open and updates the sidebar. I think adding an "Accessiblity Guide" might be useful but let me know what you think. It should be more general guides for consuming applications and how it fits with DS components.

**Do note that the NYPL Header may render _any_ skip navigation string.** It is currently set to "mainContent" in the DS Test app but there are some instances where it is "main-content". The [Staff Picks](https://www.nypl.org/books-more/recommendations/staff-picks/adults) site, for example, uses "main-content" whereas the [homepage](https://www.nypl.org/) uses "mainContent".

Once, and if, we build the NYPL Header and perhaps a separate "SkipNavigation" DS component, we will have more control over this. This is why I made the id customizable but set the default to `"mainContent"` in the `TemplateAppContainer` and `TemplateContent` components. For the moment, this is an issue in our own DS Test app and, similarly, each app is responsible to configure this correctly.

## How has this been tested?

Locally and through Storybook.

- Adds information on the skip navigation link and the `main` element's id attribute.

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/Template/Template.stories.mdx
+++ b/src/components/Template/Template.stories.mdx
@@ -60,7 +60,43 @@ import { getCategory } from "../../utils/componentCategories";
 | Component Version | DS Version |
 | ----------------- | ---------- |
 | Added             | `0.3.6`    |
-| Latest            | `0.25.12`  |
+| Latest            | `0.25.13`  |
+
+## Table of Contents
+
+- [Accessibility for TemplateAppContainer](#accessibility-for-templateappcontainer)
+- [TemplateAppContainer Component](#templateappcontainer-component)
+- [TemplateAppContainer Props](#templateappcontainer-props)
+- [Template Children Components](#template-children-components)
+- [Template Children Props](#template-children-props)
+- [Full Example with Template Children Components](#full-example-with-template-children-components)
+
+## Accessibility
+
+**The `TemplateAppContainer` component is the recommended way to render the entire
+application.** Therefore, this accessibility section is specifically for the
+`TemplateAppContainer` but the same rules apply to the individual "Template"
+components.
+
+If you need to render an alert or notification at the top of the page with an
+`aside` HTML element or HTML element with the `role="complementary"` attribute,
+then pass that alert or notification component to the `aboveHeader` prop. These
+elements should _not_ be rendered in the `header` HTML section since that's an
+accessibility violation.
+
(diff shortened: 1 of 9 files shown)


EdwinGuzman:

Link:
- https://pr911-zocq9v2lgrhu7brmfjuiyis9wjhfcey3.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/6238f869bce4cc1c73f02b38

src/components/Template/Template.stories.mdx
```diff
@@ -60,7 +60,43 @@ import { getCategory } from "../../utils/componentCategories";
...
+`TemplateAppContainer` but the same rules apply to the individual "Template"
+components.
+
+If you need to render an alert or notification at the top of the page with an
+`aside` HTML element or HTML element with the `role="complementary"` attribute,
+then pass that alert or notification component to the `aboveHeader` prop. These
+elements should _not_ be rendered in the `header` HTML section since that's an
+accessibility violation.
+
+The `TemplateAppContainer` component renders as a `main` HTML element with a
```
bigfishdesign13: This needs to be reworded.  `TemplateAppContainer` doesn’t itself render as the `main` HTML element.  The `main` element is a child element within the full DOM of the page template.

bigfishdesign13 (approved):
Overall, this looks good and works well.  I just have one comment about polishing some of the docs, but otherwise it is good to go.

I do like the idea of have a general `Accessibility` section in the sidebar, but I am wondering about how to organize those docs.  Do we need separate pages for various `a11y` topics?  Or can we get away with one page that talks about NYPL philosophy and approach to `a11y` and then have sub-sections right on that single page?

PR #999, opened 2022-05-09
DSD-936: Login dropdown

oliviawongnyc:
Fixes JIRA ticket [DSD-936](https://jira.nypl.org/browse/DSD-936).

- Creates a log in dropdown
- Modularizes Header code
- Modularizes header theme files
- Merges in development
- Fixes and adds tests

## This PR does not do the following:

- Creates search dropdown
- Creates dropdowns in the mobile view

## How has this been tested?

- This change has been tested in Storybook
- Existing and new tests pass

- None at the moment.

### Checklist:

### Front End Review:


EdwinGuzman:

Link:
- https://pr999-zq6kg1chg3afea6fikhia47g01bqo42r.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/62797713fcbc3532a76b0ad9

src/theme/components/headerUpperNav.ts
```diff
@@ -0,0 +1,117 @@
...
+    svg: {
+      position: "relative",
+      height: "8px",
+      top: "1px",
+    },
+  },
+  "#donateButton": {
+    alignItems: "center",
+    bg: "#ed1c24",
+    borderRadius: "var(--nypl-space-xxs)",
```
EdwinGuzman: The spacing values aren't meant to be used for the radius. You can create a new value in `/src/theme/foundations/radii.ts`. You can call it "headerDonate" and set it to `"0.25rem"` (which is xxs), just to make it clear it's a header-specific value (for now).

src/components/Header/Header.tsx
```diff
@@ -1,31 +1,33 @@
...
-import Icon from "../Icons/Icon";
 import Link from "../Link/Link";
-import List from "../List/List";
 import Logo from "../Logo/Logo";
+import HorizontalRule from "../HorizontalRule/HorizontalRule";
 import Notification from "../Notification/Notification";
 import useWindowSize from "../../hooks/useWindowSize";
+import HeaderUpperNav from "./components/UpperNav";
+import HeaderLowerNav from "./components/LowerNav";
+import HeaderBreakpointSmall from "./components/BreakpointSmall";
```
EdwinGuzman: You can remove "Header" from the variable name since we know it's only going to be used here and is in the sub-directory. Let's rename "BreakpointSmall" to "Mobile", including the file. "Mobile" is more generic and if the breakpoint changes for whatever reason, we don't want to involve updating the file name.

src/components/Header/Header.tsx
```diff
@@ -44,134 +46,32 @@ export const Header = chakra(() => {
...
-          </VStack>
-        </Flex>
-      ) : (
-        <Flex>
-          <Box>
+        <HStack
+          marginX={windowDimensions.width > breakpointLarge ? "50px" : "0px"}
+          justifyContent="space-between"
+        >
+          <Link href="/" __css={styles.logo}>
```
EdwinGuzman: Add `aria-label="The New York Public Library"` to the `Link` component.

src/components/Header/components/BreakpointSmall.tsx
```diff
@@ -0,0 +1,53 @@
...
+  Spacer,
+  useMultiStyleConfig,
+} from "@chakra-ui/react";
+
+import Link from "../../Link/Link";
+import Icon from "../../Icons/Icon";
+import Logo from "../../Logo/Logo";
+
+const HeaderBreakpointSmall = chakra(() => {
+  const styles = useMultiStyleConfig("Header", {});
```
EdwinGuzman: Create a new theme file for the mobile styles.

src/components/Header/components/UpperNav.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+import Button from "../../Button/Button";
+import Icon from "../../Icons/Icon";
+import Link from "../../Link/Link";
+import List from "../../List/List";
+
+const HeaderUpperNav = chakra(({ logInOpen, setLogInOpen }) => {
+  const styles = useMultiStyleConfig("HeaderUpperNav", {});
+  return (
+    <List
```
EdwinGuzman: Navigations should be wrapped in a `nav` element. When you do so, add `aria-label="Header top links"`. This is the aria label on the current header top nav.

src/components/Header/components/LowerNav.tsx
```diff
@@ -0,0 +1,73 @@
...
+import List from "../../List/List";
+import TextInput from "../../TextInput/TextInput";
+
+const HeaderLowerNav = chakra(() => {
+  const styles = useMultiStyleConfig("HeaderLowerNav", {});
+
+  const [searchOpen] = useState<boolean>(false);
+
+  return (
+    <List
```
EdwinGuzman: Navigations should be wrapped in a `nav` element. When you do so, add `aria-label="Main Navigation"`. This is the aria label on the current header lower nav.

src/theme/components/header.ts
```diff
@@ -104,23 +43,15 @@ const horizontalRule = {
...
-    icons,
-    horizontalRule,
+  parts: ["container", "notification", "logo", "icons", "horizontalRule"],
+  baseStyle: () => {
+    return {
+      container,
+      notification,
+      logo,
+      icons,
+      horizontalRule,
```
EdwinGuzman: Move the horizontal rule down a bit so that it doesn't overlap with the Search button.
Screen Shot 2022-05-10 at 12 06 05 PM

src/components/Header/Header.tsx
```diff
@@ -44,134 +46,32 @@ export const Header = chakra(() => {
...
               id="header-nypl-logo"
-              name="nyplFullBlack"
-              size="xxsmall"
+              name={
+                windowDimensions.width > breakpointLarge
+                  ? "nyplFullBlack"
+                  : "nyplLionBlack"
+              }
+              size={
+                windowDimensions.width > breakpointLarge ? "small" : "xxsmall"
```
EdwinGuzman: The mobile name and size arent' used because the `Logo` in the `HeaderBreakpointSmall` is rendered rather than this one in smaller widths.

src/components/Header/components/UpperNav.tsx
```diff
@@ -0,0 +1,97 @@
...
+          >
+            {logInOpen ? "Close" : "Log In"}
+            <Icon
+              align="right"
+              name={logInOpen ? "close" : "arrow"}
+              size="small"
+            />
+          </Button>
+          <List
+            id={logInOpen ? "logInMenuOpen" : "logInMenuClosed"}
```
EdwinGuzman: `id`s should be static, so give it only one `id` value. Hiding this should be dealt with differently.

src/theme/components/headerLowerNav.ts
```diff
@@ -0,0 +1,76 @@
+const lowerNav = {
+  alignItems: "center",
+  fontSize: "18px",
+  marginLeft: "auto",
+  whiteSpace: "nowrap",
+  li: {
+    letterSpacing: ".02em",
+    marginLeft: "var(--nypl-space-xxs)",
```
EdwinGuzman: This can be just "xxs".

src/theme/components/headerLowerNav.ts
```diff
@@ -0,0 +1,76 @@
+const lowerNav = {
+  alignItems: "center",
+  fontSize: "18px",
```
EdwinGuzman: In `typography.ts`, the "18px" value is the same as "1". In this case, you can do `fontSize: "1"` to stick with the theme tokens.

src/theme/components/headerUpperNav.ts
```diff
@@ -0,0 +1,117 @@
+const upperNav = {
+  alignItems: "baseline",
+  display: "flex",
+  marginTop: "var(--nypl-space-s)",
```
EdwinGuzman: Use "s".

EdwinGuzman (changes requested):
Added feedback for updates.

There are two things we should address before moving forward.
1. The layout for this is:
```mermaid
flowchart LR

A[Wrapper] --> B{Viewport}
B -->|Desktop| C[Logo -> upper/lower navs]
B -->|Mobile| D[Logo -> icon nav]
```
whereas I think it should be:
```mermaid
flowchart LR

A[Wrapper] --> B[Logo]
B --> C{Viewport}
C -->|Desktop| D[upper/lower navs]
C -->|Mobile| E[icon nav]
```

Or,
```
<Box>
    <Notification>
    <HStack>
        <Logo />
        <VStack>
            { "on mobile view" ?
                  "the mobile icons" : 
                  "the two desktop nav lists"}
        </VStack>
    </HStack>
</Box>
```

This way, there's only one `Logo` at any time and only the nav links change.

2. Lines 17-66 in `UpperNav.tsx` should be its own component. Eventually, there can be an "isMobile" boolean prop that renders it with mobile styling so it can be reused in the mobile viewport. For now, make it a separate component.

src/theme/components/header.ts
oliviawongnyc: Hmm, mine doesn't do that...I wonder why?
Screen Shot 2022-05-11 at 2 03 13 PM
t

src/components/Header/Header.tsx
oliviawongnyc: I might be misunderstanding this comment, but here I'm trying to say if the window with is bigger than breakpointLarge, render size small, if window is smaller than breakpointLarge but bigger than breakpointMedium, render size xxsmall, if it's smaller than breakpointMedium, render Mobile.

EdwinGuzman: Yeah, your explanation makes sense but what I was trying to say is that the Logo here wasn't rendered but instead the Logo in the Mobile component was being rendered for small widths. So instead of having to render two different Logo components in two separation locations that depend on the size of the window, it is better to render Logo once and update it's size. Now this is fixed because Logo is being rendered once in `Header` and no longer in `Mobile`.

src/theme/components/header.ts
EdwinGuzman: This is no longer a problem. I think you fixed it while making updates.

src/components/Header/Header.tsx
```diff
@@ -1,31 +1,36 @@
...
-  const styles = useMultiStyleConfig("Header", {});
   const windowDimensions = useWindowSize();
   // These values are based on the breakpoints md and lg
   // from the NYPL theme object.
   const breakpointMedium = 600;
   const breakpointLarge = 960;
 
+  const styles = useMultiStyleConfig("Header", {
+    width: windowDimensions.width,
+    breakpointMedium,
```
EdwinGuzman: Instead of passing these arguments and doing the logic check in the theme file, make a boolean for the logic check (`with < breakpointMedium`) at this top level and pass that boolean down. Same idea for the `windowDimensions.width > breakpointLarge`.

src/theme/components/headerMobile.ts
```diff
@@ -0,0 +1,12 @@
+const HeaderMobile = {
+  baseStyle: {
+    icons: {
+      marginTop: "var(--nypl-space-xs)",
+      svg: {
+        marginLeft: "var(--nypl-space-m)",
```
EdwinGuzman: Make the values `"xs"` and `"m"` respectively for marginTop and marginLeft.

EdwinGuzman (commented):
Some more minor updates and fix the snapshot tests.

EdwinGuzman (approved):

PR #1006, opened 2022-05-16
DSD-940: Header's SitewideAlerts component

EdwinGuzman:
Fixes JIRA ticket [DSD-940](https://jira.nypl.org/browse/DSD-940)

- Adds the `SitewideAlerts` component that fetches and renders NYPL sitewide alerts in the `Header` component.

## How has this been tested?

Locally in Storybook and in tests.

- The `Notification` component needs to be set above the HTML `header` element so there are no accessibility issues with nested landmarks.

### Checklist:

### Front End Review:


src/components/Header/components/SitewideAlerts.test.tsx
```diff
@@ -0,0 +1,139 @@
...
+    {
+      attributes: {
+        "alert-text": {
+          en: {
+            text: '<p>Masks are encouraged at all NYPL locations. <a href="https://nypl.org/reopening" rel="nofollow">Learn more and find information about our hours and services.</a></p>\n',
+            type: "text-single",
+          },
+          type: "text-group",
+        },
+        "closed-reason": {
```
oliviawongnyc, line 23: What is the difference between the closed-reason and the alert-text?

src/components/Header/components/SitewideAlerts.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const fetchErrorMessage =
+    "NYPL Reservoir SitewideAlerts: There was an error fetching NYPL sitewide alerts.";
+  const [alerts, setAlerts] = useState<Alert[]>([]);
+
+  // Make a request to the NYPL API endpoint for sitewide alerts, parse
+  // the data, and set it to the local state.
+  useEffect(() => {
+    fetch(alertsApiUrl)
+      .then((response) => {
+        if (response.status >= 200 && response.status < 300) {
```
oliviawongnyc, line 24: The response code could be something other than 200?

src/components/Header/components/SitewideAlerts.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  // Great, we have NYPL alerts data. Each alert is rendered as a list
+  // item in an unordered list. Note that the alert description may
+  // contain HTML which is rendered as-is.
+  const getAlertsElems = (data: Alert[]) => {
+    return (
+      <List
+        noStyling
+        type="ul"
+        __css={{
```
oliviawongnyc: Is there a reason you're not adding this to the Header theme file, or a separate, SitewideAlerts theme file?

oliviawongnyc (approved):

src/components/Header/components/SitewideAlerts.test.tsx
EdwinGuzman: I don't know. This is a mock response that's coming from the real API. The `Header` uses `alert-text` so it's better to use that for now. Eventually, this API will change but we must wait for that.

src/components/Header/components/SitewideAlerts.tsx
EdwinGuzman: Yes. We have no control of the API so if it is down, we'll get a 5xx error. If we send the wrong URL query params, we could get a 4xx client error. We could also get a successful 2xx status but still not get the correct data we want. This mostly depends on how well the API was built and how it returns the data and its status. To be careful, we can check for the successful status range and otherwise throw an error ourselves. Regardless, instead of crashing for any bad scenario, we catch any of those errors and just log a warning.

Resource here:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status

src/components/Header/components/SitewideAlerts.tsx
EdwinGuzman: Yeah, that's a good idea. I didn't want to create another new theme file for something small.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 - Adds static `Header` component with login dropdown (functionality to be added later).
 - Adds @chakra-ui/focus-lock packages to dependencies and peerDependencies.
 - Implements focus lock in `Login` component in the `Header`'s internal `Login` component.
+- Adds the `SitewideAlerts` component used internally in the `Header component. This component dynamically fetches and renders NYPL sitewide alerts.
```
bigfishdesign13, line 15: Typo: add closing tick mark for `Header`

bigfishdesign13 (approved):
Looks good.  Just make sure to fix the typo before you merge.

PR #1019, opened 2022-05-25
DSD-983: Adding mobile Header nav menu

EdwinGuzman:
Fixes JIRA ticket [DSD-983](https://jira.nypl.org/browse/DSD-983)

- Adds `MobileNav` and `MobileNavButton` to implement the `Header`'s 
mobile nav menu.
- @bigfishdesign13 we need the NYPL logo that is _just_ the text.
Screen Shot 2022-05-25 at 5 05 42 PM
- @bigfishdesign13 I also have some concerns throughout which I'll add comments for.

## How has this been tested?

Locally in Storybook.

- Just minor concern about the logo not being decorative and having the right title.

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/Header/components/Mobile.tsx
+++ b/src/components/Header/components/Mobile.tsx
@@ -2,6 +2,7 @@ import React from "react";
 import { chakra, Flex, HStack, useMultiStyleConfig } from "@chakra-ui/react";
 
 import Icon from "../../Icons/Icon";
+import MobileNavButton from "./MobileNavButton";
 import SearchButton from "./SearchButton";
 
 const Mobile = chakra(() => {
@@ -22,12 +23,7 @@ const Mobile = chakra(() => {
           title="NYPL Locator"
         />
         <SearchButton isMobile />
-        <Icon
-          id="hamburger-icon"
-          name="utilityHamburger"
-          size="medium"
-          title="Menu"
-        />
+        <MobileNavButton />
       </HStack>
     </Flex>
   );
--- a/src/components/Header/components/MobileNav.tsx
+++ b/src/components/Header/components/MobileNav.tsx
@@ -69,17 +69,25 @@ const MobileNav = chakra(() => {
           href="#"
           borderTop="1px solid rgb(54, 54, 54)"
           borderRight="1px solid rgb(54, 54, 54)"
-          gridColumn="2 / 1"
+          gridColumn="1 / span 1"
         >
           Get a Library Card
         </Link>
-        <Link href="#" borderTop="1px solid rgb(54, 54, 54)" gridColumn="2 / 2">
+        <Link
+          href="#"
+          borderTop="1px solid rgb(54, 54, 54)"
+          gridColumn="2 / span 1"
+        >
           Get Email Updates
         </Link>
-        <Link href="#" borderTop="1px solid rgb(54, 54, 54)" gridColumn="1 / 3">
+        <Link
+          href="#"
+          borderTop="1px solid rgb(54, 54, 54)"
+          gridColumn="1 / span 2"
+        >
           Shop NYPL
         </Link>
-        <Link href="#" gridColumn="1 / 3">
+        <Link href="#" gridColumn="1 / span 2">
           Donate
         </Link>
       </SimpleGrid>
--- a/src/components/Header/components/MobileNav.tsx
+++ b/src/components/Header/components/MobileNav.tsx
@@ -1,0 +1,90 @@
+import {
+  Box,
+  chakra,
+  Flex,
+  Spacer,
+  useMultiStyleConfig,
+} from "@chakra-ui/react";
+import React from "react";
+
+import Link from "../../Link/Link";
+import List from "../../List/List";
+import Logo from "../../Logo/Logo";
+import SimpleGrid from "../../Grid/SimpleGrid";
+
+/**
+ * The Header navigation for the mobile view.
+ */
+const MobileNav = chakra(() => {
+  const styles = useMultiStyleConfig("HeaderMobileNav", {});
+
+  return (
+    <Box __css={styles}>
+      <Flex>
+        <Box>
+          <Logo
+            name="nyplLionWhite"
+            size="xxsmall"
+            marginTop="15px"
+            marginLeft="15px"
+            decorative={false}
+            title="NYPL Header Logo"
+          />
+        </Box>
+        <Spacer />
+        <nav aria-label="Header mobile links">
+          <List
+            id="header-mobile-nav"
+            listItems={[
+              <Link href="#" key="booksMusicMoviesLink">
+                Books/Music/Movies
+              </Link>,
+              <Link href="#" key="researchLink">
+                Research
+              </Link>,
+              <Link href="#" key="educationLink">
+                Education
+              </Link>,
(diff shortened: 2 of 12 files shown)


src/components/Header/components/MobileNav.tsx
```diff
@@ -0,0 +1,90 @@
...
+                Get Help
+              </Link>,
+            ]}
+            noStyling
+            type="ul"
+            __css={styles.sideNav}
+          />
+        </nav>
+      </Flex>
+      <SimpleGrid gap="0" data-testid="bottomLinks" __css={styles.bottomLinks}>
```
EdwinGuzman, line 67: By adding `__css={styles.bottomLinks}`, the `SimpleGrid` component somehow lost its `display: "grid"` style. Have you seen this before? I had to manually add that rule in "bottomLinks".

src/components/Header/components/MobileNav.tsx
```diff
@@ -0,0 +1,90 @@
...
+        <Link
+          href="#"
+          borderTop="1px solid rgb(54, 54, 54)"
+          borderRight="1px solid rgb(54, 54, 54)"
+          gridColumn="2 / 1"
```
EdwinGuzman: Adding `gridColumn` got complicated. In the Chakra study group we were using `GridItem`. I'm thinking we can expose and use this component or maybe make our own. What do you think?

I'm pretty sure my `gridColumn` values are wrong but it works so... magic?

src/components/Header/components/MobileNav.tsx
```diff
@@ -0,0 +1,90 @@
...
+        >
+          Get a Library Card
+        </Link>
+        <Link href="#" borderTop="1px solid rgb(54, 54, 54)" gridColumn="2 / 2">
+          Get Email Updates
+        </Link>
+        <Link href="#" borderTop="1px solid rgb(54, 54, 54)" gridColumn="1 / 3">
```
EdwinGuzman: Continuing the comment above, I'm not sure why "1 / 3" works. I was expecting to do "1 / 2" so that this link spans two columns. I think the GridItems "span" prop was easier to use.

src/theme/components/headerMobileNav.ts
```diff
@@ -0,0 +1,59 @@
...
+        color: "ui.white",
+        textDecoration: "none",
+        _hover: {
+          color: "ui.white",
+          textDecoration: "none",
+        },
+      },
+    },
+    bottomLinks: {
+      display: "grid",
```
EdwinGuzman, line 28: This is what I meant regarding that `SimpleGrid` seemed to have lost this rule when adding the `__css` prop.

src/components/Header/components/MobileNav.tsx
bigfishdesign13: No, I have not seen that before.  I am curious what else is in `bottomLinks`.  I will check when I review the files.

src/components/Header/components/Mobile.tsx
```diff
@@ -22,12 +23,7 @@ const Mobile = chakra(() => {
           title="NYPL Locator"
         />
         <SearchButton isMobile />
-        <Icon
-          id="hamburger-icon"
-          name="utilityHamburger"
-          size="medium"
-          title="Menu"
-        />
+        <MobileNavButton />
```
bigfishdesign13, line 26: Have you considered creating a general button component that could then be configured and used for each of the buttons that will be in this nav?  So far, you have `SearchButton` and `MobileNavButton`, but I assume you will have at least two more in the end.

src/components/Header/components/MobileNav.tsx
bigfishdesign13: `GridItem` is used with `Grid` – it's not intended for `SimpleGrid` – so I don't think we would gain anything by exposing it.  At least, not without also exposing `Grid`.

Regarding `gridColumn`, I have not yet mastered that atrtibute, so yes, it's all magic as far as I'm concerned.

src/components/Header/components/MobileNav.tsx
bigfishdesign13: `span` can be used here as well.  That is not specific to the Charkra component.  That is a valid value for the CSS `grid-column` attribute.

https://developer.mozilla.org/en-US/docs/Web/CSS/grid-column

src/theme/components/headerMobileNav.ts
bigfishdesign13: When did you add the `gridTemplateColumns` attribute?

src/components/Header/components/MobileNav.tsx
bigfishdesign13: I was not able to pinpoint why this is happening.  I looked at nesting, props on child elements, etc., but I could not find anything specifically tied to this issue.  Despite not finding anything, I feel it's related to nesting and the mobile sizing.

Let's move forward with what you have done and keep our eyes open for possible causes.

bigfishdesign13 (commented):
I responded to your comments.

src/components/Header/components/MobileNav.tsx
EdwinGuzman: Ha, magic indeed. But if `GridItem` is using the grid from it's parent, then it shouldn't matter who the parent is as long as they implement grid, no? 

Update: This actually does work:
```
<SimpleGrid gap="0" data-testid="bottomLinks" __css={styles.bottomLinks}>
        <GridItem colSpan={1}>
          <Link
            href="#"
            borderTop="1px solid rgb(54, 54, 54)"
            borderRight="1px solid rgb(54, 54, 54)"
          >
            Get a Library Card
          </Link>
        </GridItem>
        <GridItem colSpan={1}>
          <Link href="#" borderTop="1px solid rgb(54, 54, 54)">
            Get Email Updates
          </Link>
        </GridItem>
        <GridItem colSpan={2}>
          <Link href="#" borderTop="1px solid rgb(54, 54, 54)">
            Shop NYPL
          </Link>
        </GridItem>
        <GridItem colSpan={2}>
          <Link href="#">Donate</Link>
        </GridItem>
      </SimpleGrid>
```

I'm _not_ going to use this but something for us to think about. This is much simpler than doing `gridColumn="2 / 1"`.

src/components/Header/components/MobileNav.tsx
EdwinGuzman: This is helpful! thanks.

src/components/Header/components/Mobile.tsx
EdwinGuzman: YEah, but the refactor would be later.

EdwinGuzman:
BTW, I'll update the SVG mobile logo once `development` is merged into the main `Header` branch. So that'll be a while until that change makes it way to this PR branch.

src/components/Header/components/MobileNav.tsx
bigfishdesign13: > But if GridItem is using the grid from it's parent, then it shouldn't matter who the parent is as long as they implement grid, no?

I would need to play with this to know for sure.  I am just curious if there is validation built into the parent element that is used to dictate which components `GridItem` can be used within.

bigfishdesign13 (approved):
Knowing that we will polish this up later, I will approve this PR.

PR #1022, opened 2022-05-27
DSD-980: Visually hidding the Radio's label when showLabel is set to false

EdwinGuzman:
Fixes JIRA ticket [DSD-980](https://jira.nypl.org/browse/DSD-980)

- Visually hides the `Radio`'s label text when `showLabel` is set to false.
- I tried to pass in the `showLabel` boolean to the theme object but Chakra's `Radio` component is difficult to update and override. It also doesn't work like the rest of the Chakra's form-components, such as input, select, and checkbox -- all of which behave in a similar way.
- Adds axe tests to verify that other form components are accessible when `showLabel` is set to false.

## How has this been tested?

Locally in Storybook

- Fixes an accessibility issue where the `aria-label` was not being added to the `input` HTML element even though it was passed as a prop. It's because of how Chakra renders its `Radio` component.

### Checklist:

### Front End Review:


src/components/Radio/Radio.tsx
```diff
@@ -68,7 +69,7 @@ export const Radio = chakra(
       value,
       ...rest
     } = props;
-    const styles = useMultiStyleConfig("Radio", {});
+    const styles = useMultiStyleConfig("Radio", { showLabel });
```
bigfishdesign13, line 72: I didn’t see `showLabel` being used in `radio.ts`.  Does it need to be passed?

bigfishdesign13 (approved):
This looks good and works well.  Approved, but I had a question about the prop you are passing into the theme object.

bigfishdesign13:
The `Latest Version` number should be bumped to `1.0.3`.

PR #1031, opened 2022-06-01
DSD-985: Header login functionality

oliviawongnyc:
Fixes JIRA ticket [DSD-985.](https://jira.nypl.org/browse/DSD-985)

- Adds login functionality, which includes reading cookie values with new package, js-cookie, making an API call for patron details, and showing a different UI to logged in users.

## How has this been tested?

- New and existing tests pass.
- This update has also been tested locally.

- Added a listener that closes the login menu when escape key is pressed.

### Checklist:

### Front End Review:


package.json
```diff
@@ -64,6 +64,7 @@
     "@emotion/styled": ">=11.3.0",
     "framer-motion": "^4.1.17",
     "he": "1.2.0",
+    "js-cookie": "^3.0.1",
```
EdwinGuzman: Remove the caret.

src/theme/components/headerUpperNav.ts
```diff
@@ -70,19 +69,42 @@ const upperNav = {
       display: "inline-block",
       color: "ui.white",
       fontSize: "14px",
-      fontWeight: 500,
+      fontWeight: 400,
```
EdwinGuzman: Use `"regular"` instead of `400`.

src/components/Header/headerUtils.ts
```diff
@@ -48,3 +49,55 @@ export const parseAlertsData = (data: any): Alert[] => {
...
+  return { cookieValue, accessToken };
+};
+
+/**
+ * fetchPatronData uses the patronApiUrl combined with the
+ * access_token from the nyplIdentityPatron cookie to fetch
+ * the patron's information from the server.
+ */
+export const fetchPatronData = (accessToken, cb) => {
+  const patronApiUrl = "https://platform.nypl.org/api/v0.1/auth/patron/tokens/";
```
EdwinGuzman: This probably won't be used outside of this function but make this a global variable similar to `alertsApiUrl`.

src/components/Header/headerUtils.ts
```diff
@@ -48,3 +49,55 @@ export const parseAlertsData = (data: any): Alert[] => {
...
+        throw new Error(fetchErrorMessage);
+      }
+    })
+    .then(cb)
+    .catch(() => {
+      console.warn(fetchErrorMessage);
+    });
+};
+
+export const extractPatronName = (data) => {
```
EdwinGuzman: Add a short comment and give `data` a type, or maybe just leave it as `any` for now.

src/components/Header/Header.tsx
```diff
@@ -35,6 +40,32 @@ export const Header = chakra(() => {
...
 
   const [loginOpen, setLoginOpen] = useState<boolean>(false);
+  const [patronDataReceived, setPatronDataReceived] = useState<boolean>(false);
+  const [patronName, setPatronName] = useState<string>("");
+
+  // Allows user to use esc key to close the login menu.
+  useEffect(() => {
+    const close = (e) => {
+      if (e.keyCode === 27) {
+        setLoginOpen(false);
```
EdwinGuzman, line 51: This is nice. We can reuse this for all the other button/menus.

src/components/Header/Header.test.tsx
```diff
@@ -128,3 +145,124 @@ describe("Header", () => {
...
 });
+
+describe("Patron is logged in", () => {
+  beforeAll(async () => {
+    jest.spyOn(Cookies, "get").mockReturnValue(mockCookieValue);
+
+    // This is assuming that the sitewide alerts fetch call runs
+    // before the patron information call. If the alerts call runs
+    // after the patron information call, the wrong response will be
+    // returned for each and this test will fail.
```
EdwinGuzman, line 156: Can you add a `TODO` after this comment that says something like "fix the indeterministic fetch calls".

src/components/Header/Header.test.tsx
```diff
@@ -1,12 +1,29 @@
...
+      id: "0123456",
+      names: ["PATRON, JANE"],
+    },
+  },
+  count: 1,
+  debugInfo: [],
+  statusCode: 200,
+};
+
+const mockCookieValue = '{"token_type":"Bearer", "access_token":"123abc"}';
```
EdwinGuzman, line 25: Let's also verify that we are calling the correct key in the browser's cookie. I'll describe it below in the `"Patron is logged in"` describe block.

src/components/Header/Header.test.tsx
```diff
@@ -128,3 +145,124 @@ describe("Header", () => {
     expect(header).toMatchSnapshot();
   });
 });
+
+describe("Patron is logged in", () => {
```
EdwinGuzman, line 149: Set a variable to store the `Cookies.get` function before `beforeAll`:
`const realGet = Cookies.get;`

Instead of spying on the "get" function, mock it:
`Cookies.get = jest.fn().mockReturnValue(mockCookieValue);`

Restore the original `Cookies.get` function:
```
afterAll(() => {
    Cookies.get = realGet;
  });
```

and finally test:
`expect(Cookies.get).toHaveBeenCalledWith("nyplIdentityPatron");


I think it's enough if this is done once instead of all three tests but that's up to you.

src/components/Header/components/Login.test.tsx
```diff
@@ -0,0 +1,55 @@
+import * as React from "react";
+import { render, screen } from "@testing-library/react";
+
+import Login from "./Login";
+
+describe("Login", () => {
+  const setLoginOpen = jest.fn();
+  it("renders the logged out UI if no patron name", () => {
```
EdwinGuzman:
```
"... if there is no `patronName` value"
```

src/components/Header/components/Login.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <Login loginOpen={true} patronName="" setLoginOpen={setLoginOpen} />
+    );
+
+    const links = screen.getAllByRole("link");
+
+    expect(links.length).toEqual(2);
+    expect(links[0]).toHaveTextContent(/log into the catalog/i);
+    expect(links[1]).toHaveTextContent(/log into the research catalog/i);
+  });
+  it("renders the logged in UI if patron name", () => {
```
EdwinGuzman:
```
"... if there is a `patronName` value"
```

src/components/Header/components/Login.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+        setLoginOpen={setLoginOpen}
+      />
+    );
+
+    const links = screen.getAllByRole("link");
+
+    expect(links.length).toEqual(3);
+    expect(links[0]).toHaveTextContent(/go to the catalog/i);
+    expect(links[1]).toHaveTextContent(/go to the research catalog/i);
+    expect(links[2]).toHaveTextContent(/log out/i);
```
EdwinGuzman, line 61: This should also test that the message "You are logged in as: Patron, Jane A" is in the UI.

src/components/Header/Header.test.tsx
```diff
@@ -1,12 +1,29 @@
...
 import Header from "./Header";
 import { refineryResponse } from "./components/SitewideAlerts.test";
 
+const patronResponse = {
+  data: {
+    patron: {
+      barCodes: ["01234567890123"],
+      emails: ["test@nypl.org"],
+      id: "0123456",
+      names: ["PATRON, JANE"],
```
EdwinGuzman, line 17: Okay, so it seems there is no issue if there's a middle initial, right?

src/components/Header/components/Login.tsx
```diff
@@ -1,61 +1,97 @@
...
-                id="log-in-catalog-icon"
-                name="utilityAccountFilled"
-                size="medium"
-                title="Log in to your account"
-              />
-              Log Into The Catalog
-            </Link>,
+        <VStack className="loginMenu">
+          {patronName && (
+            <Box id="patronGreeting">
```
EdwinGuzman: In the current header, when this menu is opened, this text is focused to automatically. It should be replicated here. You can add a `ref` (and maybe a tabIndex value?) to this `Box` and through a `useEffect`, focus on this element when `loginOpen` is true.

EdwinGuzman (changes requested):
Some changes are needed but this is great!
Visually, the styling for the logged in message should be slightly tweaked:
Screen Shot 2022-06-03 at 5 09 50 PM

src/components/Header/Header.test.tsx
oliviawongnyc: Thought about it some more and seems like no!

src/components/Header/Header.tsx
oliviawongnyc: Adjusted to make work for more browsers.

src/components/Header/Header.test.tsx
```diff
@@ -128,3 +145,131 @@ describe("Header", () => {
...
+          json: () => Promise.resolve(patronResponse),
+        })
+      );
+
+    await waitFor(() => render(<Header />));
+  });
+  afterAll(() => {
+    Cookies.get = realGet;
+  });
+  it.only("displays logged in view", async () => {
```
EdwinGuzman: Remove `.only`.

src/components/Header/components/Login.tsx
```diff
@@ -1,61 +1,110 @@
...
-              />
-              Log Into The Catalog
-            </Link>,
+        <VStack className="loginMenu">
+          {patronName && (
+            <Box
+              id="patronGreeting"
+              data-testid="patronGreeting"
+              ref={greeting}
+              tabIndex={-1}
```
EdwinGuzman: Set this to `{0}`. This is similar to what's in prod: https://github.com/NYPL/dgx-header-component/blob/master/src/components/MyNypl/MyNypl.jsx#L83

The `-1` is correct and allows us to programmatically focus on the element, but we also want to tab on this element within the focus trap.

EdwinGuzman (approved):
Looks good! Just remove the `.only` in the test and update the tabIndex value to `0`.

PR #1035, opened 2022-06-02
DSD-997: Optimize npm release bundle

EdwinGuzman:
Fixes JIRA ticket [DSD-997](https://jira.nypl.org/browse/DSD-997)

- Slightly decreases the bundle size for the npm package. More could be done?
- The size of the bundled file is actually the same. If we want to optimize that, it'd be code changes.
- What this does is remove the `src` folder from the npm package. This is not needed by consuming applications since they already have the `dist` folder. They also don't need test or storybook files. Doing this reduces the npm package size. Notice how in the screenshots below, the number of files in the npm package went from 389 to 122.

Before:
existing

After removing `"src"` from the `"files"` object in `package.json`:
after-update

It's better and less than 1MB but it's still quite large. Chakra, for example, has 105KB "unpacked" size but _I think_ it's because they export/publish every component on an individual basis. Comparisons are bad, anyway, but we should keep trying to reduce the size of the bundle. Maybe it's time to move away from TSDX and use a faster/better bundler.

## How has this been tested?

By making a test npm release version [`1.0.3-beta`](https://www.npmjs.com/package/@nypl/design-system-react-components/v/1.0.3-beta). I cannot deprecate this version but the `-beta` should be enough to let others know not to use it.

- N/A

### Checklist:

### Front End Review:


bigfishdesign13 (approved):
Nice to see these improvements.  We started using TSDX because of its support for creating packages.  Are there other decent options out there with similar benefits?

EdwinGuzman:
I think there are other options for us to bundle the code. TSDX was a good option for us to get started and get to where we are but now we should figure out how to really optimize the distributed, minified code bundle.

PR #1036, opened 2022-06-02
DSD-957: Multiselect Group Component

isastettler:
Fixes JIRA ticket [DSD-957](https://jira.nypl.org/browse/DSD-957)

- Adds the `MultiSelectGroup` component
- Adds tests for `MultiSelectGroup` component
- Adds `MultiSelectGroup` story to storybook
- Fixes the `MutliSelect` story width prop
- Adds white background color to `MultiSelect` `menuContainer` style

## How has this been tested?

- Unit tests in DS.
- Locally through storybook.

-

### Checklist:

### Front End Review:

- [ ] Review the [Vercel preview deployment](https://nypl-design-system-nm7y6l0n9-nypl.vercel.app/?path=/story/welcome--page).


src/components/MultiSelect/MultiSelect.stories.mdx
```diff
@@ -26,25 +26,35 @@ import * as stories from "./MultiSelect.stories.tsx";
...
+        "The action to perform for a mixed state checkbox (parent checkbox). <br /> `onMixedStateChange?: (event: React.ChangeEvent<HTMLInputElement>) => void;`",
     },
     selectedItems: { control: false },
-    variant: { control: false, defaultValue: { summary: "listbox" } },
+    variant: { control: false },
     width: {
+      description: "Value used to set the width for the MultiSelect component",
+      control: "radio",
+      options: ["default", "fitContent", "full"],
+      defaultValue: "default",
```
EdwinGuzman: The default value is already added in the `table` object on the following line. So this is not needed here.

src/components/MultiSelect/MultiSelectListbox.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { forwardRef } from "react";
-import { chakra } from "@chakra-ui/react";
 import {
   Box,
+  chakra,
```
EdwinGuzman, line 4: You should do this same import update in `MultiSelectDialog.tsx`.

src/components/MultiSelectGroup/MultiSelectGroup.stories.mdx
```diff
@@ -0,0 +1,101 @@
...
+      type: "figma",
+      url: "https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Main?node-id=43593%3A24611",
+    },
+    jest: ["MultiSelectGroup.test.tsx"],
+  }}
+  argTypes={{
+    className: { control: false },
+    id: { control: false },
+    isBlockElement: {
+      defaultValue: undefined,
```
EdwinGuzman: This `defaultValue` outside of the `table` object is deprecated. So remove (line 36, 40, 44) these but keep the `defaultValue: { summary: .... } ` inside of the `table` key.

src/components/MultiSelectGroup/MultiSelectGroup.stories.mdx
```diff
@@ -0,0 +1,101 @@
...
+      control: "radio",
+      options: [true, false, undefined],
+      tabel: { defaultValue: { summary: undefined } },
```
EdwinGuzman: Typo: it should be "table". Fix this on line 41 and 45 as well.

src/components/MultiSelectGroup/MultiSelectGroup.stories.mdx
```diff
@@ -0,0 +1,101 @@
...
+    },
+  }}
+/>
+
+# MultiSelectGroup
+
+| Component Version | DS Version |
+| ----------------- | ---------- |
+| Added             | `1.0.1`    |
+| Latest            | `1.0.1`    |
```
EdwinGuzman: Update these to `1.1.0` for now.

src/components/MultiSelectGroup/MultiSelectGroup.stories.mdx
```diff
@@ -0,0 +1,101 @@
...
+</Canvas>
+
+<ArgsTable story="MultiSelectGroup" />
+
+## Accessibility
+
+The `MultiSelectGroup` renders a group of `MultiSelect` components that are wrapped in
+a `<fieldset>` element. The `<fieldset>` element renders a `<legend>` element that
+can be visually hidden through the `showLabel` prop. If the `<legend>` is hidden the `aria-label`
+attribute will be added to the `MultiSelectGroup` component.
```
EdwinGuzman: > If the `<legend>` is hidden the `aria-label` attribute will be added to the `MultiSelectGroup` component.

This isn't exactly true. The `aria-label` attribute will be added to the `div` inside of the fieldset but not to a separate `MultiSelectGroup` component.

src/components/MultiSelectGroup/MultiSelectGroup.stories.mdx
```diff
@@ -0,0 +1,101 @@
...
+For accessibility information on the `MultiSelect` component, see the [MultiSelect Accessibility](https://nypl-design-system-vvscfj2pn-nypl.vercel.app/?path=/docs/components-form-elements-multiselect--multi-select-list-box) documentation.
+
+## Layout Patterns
+
+The width of the MultiSelect components can be passed through the `multiSelectWidth` prop.
```
EdwinGuzman: "...  passed through the `multiSelectWidth` prop, with values of `"default"`, `"fitContent"`, and `"full"`.

src/components/MultiSelectGroup/MultiSelectGroup.stories.mdx
```diff
@@ -0,0 +1,101 @@
...
+On screen width under 600px width the `layout` is automatically set to "column" and the `multiSelectWidth` to "full".
+
+<Canvas>
+  <DSProvider>
+    <stories.MultiSelectGroupLayoutStory />
```
EdwinGuzman, line 92: Maybe add more examples? Such as where `layout` is "column" and `multiSelectWidth` is "full", or where `layout` is "row" and `multiSelectWidth` is "full". Not all permutations need to be shown but more examples would be helpful.

src/theme/components/multiSelect.ts
```diff
@@ -22,18 +22,14 @@ const multiSelectWidths = {
...
   parts: [
     "actionButtons",
+    "menu",
     "menuButton",
     "menuContainer",
-    "menu",
     "menuChildren",
   ],
-  baseStyle: ({ width = "default", isBlockElement, isOpen }) => ({
+  baseStyle: ({ isBlockElement, isOpen, width }) => ({
```
EdwinGuzman: I think it's okay to have the default for `width` be set to `"default"`. It does propagate all the way down from the `MultiSelect` main component, but there's no harming adding it back in.

src/components/MultiSelectGroup/MultiSelectGroup.tsx
```diff
@@ -0,0 +1,103 @@
...
+          id={`${id}-multiselect-group`}
+          legendText={labelText}
+          isLegendHidden={!showLabel}
+          {...rest}
+        >
+          <Stack
+            id={id}
+            ref={ref}
+            data-testId="multi-select-group"
+            {...(!showLabel && { "aria-label": labelText })}
```
EdwinGuzman: This is easier to read as `aria-label={!showLabel ? labelText : undefined}`

src/components/MultiSelectGroup/MultiSelectGroup.stories.mdx
```diff
@@ -0,0 +1,101 @@
...
+    },
+    jest: ["MultiSelectGroup.test.tsx"],
+  }}
+  argTypes={{
+    className: { control: false },
+    id: { control: false },
+    isBlockElement: {
+      defaultValue: undefined,
+      control: "radio",
+      options: [true, false, undefined],
```
EdwinGuzman: `isBlockElement` is listed as boolean prop but there's still the option to pass `undefined`. This should be removed.

src/components/MultiSelectGroup/MultiSelectGroup.test.tsx
```diff
@@ -0,0 +1,207 @@
...
+        id="MultiSelectGroup"
+        labelText="MultiSelectGroup example"
+        showLabel={false}
+        multiSelectWidth="default"
+      >
+        {multiSelectItems.map((multiSelectItem) => (
+          <MultiSelect
+            key={multiSelectItem.id}
+            id={multiSelectItem.id}
+            variant={"listbox"}
```
EdwinGuzman: Nit: You can remove braces from string values `variant="listbox"`.

src/components/MultiSelectGroup/MultiSelectGroup.test.tsx
```diff
@@ -0,0 +1,207 @@
...
+            label={multiSelectItem.name}
+            items={multiSelectItem.items}
+            selectedItems={{}}
+            onChange={handleChangeMock}
+            onClear={() => "clear"}
+          />
+        ))}
+      </MultiSelectGroup>
+    );
+    expect(screen.getByTestId("multi-select-group")).not.toHaveAttribute(
```
EdwinGuzman, line 105: Can you also add an expectation that the text is visible in the document since the fieldset does render it?

src/components/MultiSelectGroup/MultiSelectGroup.tsx
```diff
@@ -0,0 +1,103 @@
...
+        <Fieldset
+          id={`${id}-multiselect-group`}
+          legendText={labelText}
+          isLegendHidden={!showLabel}
+          {...rest}
+        >
+          <Stack
+            id={id}
+            ref={ref}
+            data-testId="multi-select-group"
```
EdwinGuzman: Typo: this should be `data-testid` with lowercase `"i"`.

src/components/MultiSelectGroup/MultiSelectGroup.tsx
```diff
@@ -0,0 +1,103 @@
...
+import MultiSelect, { MultiSelectWidths } from "../MultiSelect/MultiSelect";
+import useNYPLBreakpoints from "../../hooks/useNYPLBreakpoints";
+
+export interface MultiSelectGroupProps {
+  children: React.ReactNode;
+  /** Additional className to use. */
+  className?: string;
+  id: string;
+  /** Optional prop to overwrite default behavior of the `MultiSelect` component, if nothing gets passed, it is set to `undefined`. */
+  isBlockElement?: boolean;
```
EdwinGuzman: If the comment is true, then this should be of type `boolean | undefined`. How about you set it just as `boolean` and give it a default value? On line 38 `isBlockElement: false` (I'm not sure if this is the correct default value for this component but the single `MultiSelect` component has a default value of false). @bigfishdesign13 thoughts on the default value?

This will mean the code on lines 50-56 will be updated.

src/components/MultiSelectGroup/MultiSelectGroup.stories.tsx
```diff
@@ -0,0 +1,295 @@
...
+export const MultiSelectGroupStory: Story<MultiSelectGroupProps> = (args) => {
+  const [selectedItems, setSelectedItems] = React.useState({});
+
+  function handleChange(itemId: string, groupId: string) {
+    let itemIds;
+    // Check if the id already exists in the state
+    if (
+      selectedItems[groupId] !== undefined &&
+      // @TODO Temporary hack to make availability multiselect use radios.
+      groupId !== "availability"
```
EdwinGuzman: I don't understand this, "availability" is not used as an id value in any of the groups.

src/components/MultiSelectGroup/MultiSelectGroup.stories.tsx
```diff
@@ -0,0 +1,295 @@
...
+export const MultiSelectGroupLayoutStory: Story<MultiSelectGroupProps> = () => {
+  const [selectedItems, setSelectedItems] = React.useState({});
+
+  function handleChange(itemId: string, groupId: string) {
+    let itemIds;
+    // Check if the id already exists in the state
+    if (
+      selectedItems[groupId] !== undefined &&
+      // @TODO Temporary hack to make availability multiselect use radios.
+      groupId !== "availability"
```
EdwinGuzman: I don't get this either.

src/components/MultiSelectGroup/MultiSelectGroup.stories.tsx
```diff
@@ -0,0 +1,295 @@
...
+    let newSelectedItems = {};
+    for (let key of Object.keys(selectedItems)) {
+      if (key !== groupId) {
+        newSelectedItems[key] = selectedItems[key];
+      }
+    }
+    setSelectedItems(newSelectedItems);
+  }
+  return (
+    <>
```
EdwinGuzman: Instead of a fragment and separating with a `<br />`, use Chakra's `VStack`:
```
<VStack gap="grid.m" align="start">
    <MultiSelectGroup ... ></MultiSelectGroup>
    <MultiSelectGroup ... ></MultiSelectGroup>
</VStack>
```

EdwinGuzman (changes requested):
This is nice! Added comments in PR but have additional questions:

1. Can you mix "listbox" and "dialog" variants in a `MultiSelectGroup`? From the examples, it seems like you cannot but it's not explained if that is allowed.
2. There's little guidance on how to use the `MultiSelectGroup` with multiple `MultiSelect`.  The `MultiSelect` storybook doc shows the expected shape of the items and how to keep track of state with the `useMultiSelect` hook. The `MultiSelectGroup` doesn't have the equivalent and I noticed that the event handlers for `onChange` and `onMixedStateChange` in `MultiSelectGroup.stories.tsx` have _slightly_ different implementation than the hook. Should these event function implementations be added to the doc to explain to users how to keep track of state for all `MultiSelect` components in a single `MultiSelectGroup`?
3. Perhaps a `useMultiSelectGroup` hook would be useful to have.

nypl-wluisi:
@isastettler @EdwinGuzman I edited my previous comment recommending `useMultiSelect` hook in the `MultiSelectGroup` stories to replace the internal state management. The problem is that hook can only handle 1 `MultiSelect` not multiple. I have a newer version of that hook I've been working on that could replace it, as it can handle 1 or 5 `MultiSelect` components. I can show Isa the code for that on Friday, because it's needed here.

EdwinGuzman:
Is it "1 or 5" or "1 to 5"? Why not more? If a max of five `MultiSelect` should be used, then  there should be a warning in the component that checks for more than five child `MultiSelect`.

I noticed the same for the `useMultiSelect` hook. At first, I thought it could be used for multiple components, especially how the internal data structure is shaped. But now, in this PR, it doesn't seem to work unless a few things are updated. I'd suggest simplifying `useMultiSelect` for one `MultiSelect` and `useMultiSelectGroup` for multiple ones. This is similar to how Chakra has `useCheckbox` and `useCheckboxGroup`.

- https://v1.chakra-ui.com/docs/styled-system/component-hooks/use-checkbox
- https://v1.chakra-ui.com/docs/styled-system/component-hooks/use-checkbox-group

src/components/MultiSelect/MultiSelect.stories.mdx
isastettler: removed

src/components/MultiSelect/MultiSelectListbox.tsx
isastettler: updated MultiSelectDialog.tsx

src/components/MultiSelectGroup/MultiSelectGroup.stories.mdx
isastettler: removed

src/components/MultiSelectGroup/MultiSelectGroup.stories.mdx
isastettler: fixed

src/components/MultiSelectGroup/MultiSelectGroup.stories.mdx
isastettler: updated

src/components/MultiSelectGroup/MultiSelectGroup.stories.mdx
isastettler: Thank you for clarifying. I changed the text to "…`aria-label` attribute with `label` value will be rendered instead"

src/components/MultiSelectGroup/MultiSelectGroup.stories.mdx
isastettler: updated

src/theme/components/multiSelect.ts
isastettler: updated

src/components/MultiSelectGroup/MultiSelectGroup.tsx
isastettler: updated

src/components/MultiSelectGroup/MultiSelectGroup.test.tsx
isastettler: fixed

src/components/MultiSelectGroup/MultiSelectGroup.test.tsx
isastettler: I added the expect for the label text, however I noticed that the `<legend>` tag and text are still present on page even when hidden so it would pass even if the label is not visual on the page.

src/components/MultiSelectGroup/MultiSelectGroup.tsx
isastettler: fixed

src/components/MultiSelectGroup/MultiSelectGroup.stories.tsx
isastettler: updated

src/components/MultiSelectGroup/MultiSelectGroup.tsx
isastettler: As of now, the isElementBlock behavior is so that when the `layout` prop gets set to `"column"`, the isBlockElement will be set to true by default. Having no default value of the `isBlockElement` prop in `MultiSelectGroup`  allows the user to set `isBLockElement=false` even when `layout="column"`. 
However, if that is not an option that should be available to the user, `isBlockElement` could be set to `false` by default (the same default as on the `MultiSelect` component). Then the logic in MultiSelectGroup could be `const finalIsBlockElement = layout === "column" ? true : isBlockElement;`
@nypl-wluisi @bigfishdesign13, how would you like me to proceed?

src/components/MultiSelectGroup/MultiSelectGroup.stories.mdx
EdwinGuzman: This is good. I think you mean `labelText` value and not just `label`.

src/components/MultiSelectGroup/MultiSelectGroup.test.tsx
EdwinGuzman: That's correct. For the `fieldset` and `legend` structure, the `legend` must always be in the DOM, so what is done is to visually hide the text when `showLabel` is false. So this test is good as is now, thanks.

src/components/MultiSelectGroup/MultiSelectGroup.stories.tsx
isastettler: I removed this

src/components/MultiSelectGroup/MultiSelectGroup.stories.tsx
isastettler: removed

src/components/MultiSelectGroup/MultiSelectGroup.stories.mdx
isastettler: Yes, thank you. I updated this.

src/components/MultiSelectGroup/MultiSelectGroup.stories.mdx
isastettler: Added the two patterns mentioned. I also added a line to point out the default behavior of the `MultiSelect` components, when `layout = "column` is passed: 

> Note: when `layout = "column"` is passed, the MultiSelect components will have `isBlockElement` set to `true` by default.

src/components/MultiSelectGroup/MultiSelectGroup.tsx
```diff
@@ -0,0 +1,103 @@
...
+import MultiSelect, { MultiSelectWidths } from "../MultiSelect/MultiSelect";
+import useNYPLBreakpoints from "../../hooks/useNYPLBreakpoints";
+
+export interface MultiSelectGroupProps {
+  children: React.ReactNode;
+  /** Additional className to use. */
+  className?: string;
+  id: string;
+  /** Optional prop to overwrite default behavior of the `MultiSelect` component, if nothing gets passed, it is set to `undefined`. */
+  isBlockElement?: boolean | undefined;
```
EdwinGuzman: Let's wait to hear what @bigfishdesign13 and @nypl-wluisi think about the `isBlockElement`. Doing this goes against the pattern set in the DS and it feels like a special case that can be updated to not confuse devs.

bigfishdesign13: @EdwinGuzman In `ButtonGroup`, `CheckboxGroup` and `RadioGroup` we include props that mirror some of the props on the child components.  Passing one of those props to the group component sets the value for the corresponding prop on the child components.  I felt that using that same pattern here for `isBlockElement` would make sense.  How do you see this implementation in `MultiSelectGroup` as going against the DS patterns?

EdwinGuzman: That pattern is fine. The issue is with declaring the boolean prop to be of type `boolean | undefined`.

bigfishdesign13: @isastettler @EdwinGuzman 
The logic for `isBlockElement` that was described on Friday is appropriate: the default behavior for column shoudl be `true` and the default behavior for row shoudl be `false`.  That said, I know there were some syntactical issues that required us to use `undefined` for values in some other components, but I am not sure if that is necessary here.

nypl-wluisi: @bigfishdesign13 I think we should get rid of the `isBlockElement` prop on this component and have that setting controlled by the prop value of `layout`. This will solve the problem Edwin is highlighting and also make the component easier to use from a dev perspective.

bigfishdesign13: Yes, I was thinking that as well.  Let’s go ahead and do that.  The `isBlockElement` value for the child `MultiSelect` components can be set based on the value of `layout`.

isastettler: Sounds good. The `isBlockElement` prop is now removed from the `MultiSelectGroup` component. It will be passed on to the `MultiSelect` child components conditionally to the layout.

EdwinGuzman:
The update looks good. The only thing left is the custom code for state storage that should be a hook.

EdwinGuzman (approved):

EdwinGuzman:
Approving. `useMultiSelect` hook work will be done in a subsequent ticket and PR.

nypl-wluisi:
Ticket for new `useMultiSelect` hook is here: https://jira.nypl.org/browse/DSD-1135

PR #1033, opened 2022-06-01
DSD-999: Fixing bugs found after installing v1.0.2 in Turbine

EdwinGuzman:
Fixes JIRA ticket [DSD-999](https://jira.nypl.org/browse/DSD-999)

## This PR does the following:

- Fixes bug found after installing version `1.0.2` in the Turbine test app.
- Fixes the border color for the `Accordion` button.
Screen Shot 2022-06-01 at 4 25 14 PM

- Fixes a false warning message logged from the `Table` component.
- Fixes the background color for the `Toggle` button.
Screen Shot 2022-06-01 at 4 21 29 PM

## How has this been tested?

Locally and in Turbine

## Accessibility concerns or updates

- N/A

### Checklist:

- [ ] I have updated the Storybook documentation accordingly.
- [ ] I have added relevant accessibility documentation for this pull request.
- [ ] All new and existing tests passed.

### Front End Review:

- [ ] Review the Vercel preview deployment once it is ready.

PR #1037, opened 2022-06-02
DSD-934: responsive Table

bigfishdesign13:
Fixes JIRA ticket [DSD-934](https://jira.nypl.org/browse/DSD-934)

- Updates the `Table` component to be responsive for a mobile viewport.

- Nothing new was added to the story, but I am wondering if something should be added to say that the mobile layout is vertical (i.e. no longer horizontal).  I don’t think we say antyhing like that on any of the other stories, but watching all the table cells flip from horizontal to vertical is quite striking.

## How has this been tested?

- All a11y considerations that were built into the desktop layout also apply to the mobile layout, so there are no a11y concerns with this update.

### Checklist:

### Front End Review:


src/theme/components/customTable.ts
```diff
@@ -1,46 +1,131 @@
...
   _last: {
-    paddingRight: showRowDividers ? "m" : "0",
+    paddingRight: showRowDividers ? { base: "0", md: "m" } : null,
+    borderBottom: showRowDividers ? "1px solid" : "none",
+    borderColor: cellBorderColorStyles,
+  },
+  "> span": {
+    flexBasis: "50%",
+    paddingBottom: { base: "s", md: "0" },
+    // paddingLeft: "0",
```
EdwinGuzman: This can be removed.

src/components/Table/Table.tsx
```diff
@@ -110,17 +117,28 @@ export const Table = chakra((props: React.PropsWithChildren<TableProps>) => {
       }
     }
 
+    const cellContent = (key: number, column: string | JSX.Element) => {
+      return windowDimensions.width <= breakpointMedium ? (
+        <>
+          <Box as="span">{columnHeaders[key]}</Box>
```
EdwinGuzman: If nothing else is added to these `Box`es, then I think having direct `<span>`s should be okay.

EdwinGuzman (changes requested):
This looks good and the reusability in the `customTable.ts` is nice. I agree that some documentation would be helpful to let users know what to expect on a mobile view. Something similar is written for the `Tabs` mobile view: https://nypl.github.io/nypl-design-system/reservoir/v1/?path=/docs/components-overlays-switchers-tabs--tabs-with-controls#responsive-mobile-carousel

I encountered this problem before and fixed it in PR #1033 regarding the Table warning message: the `Table`'s `columnHeaders` prop is optional. If this is missing, the mobile view is off:

Screen Shot 2022-06-03 at 3 04 40 PM

Maybe it should be required or recommended to use it.

EdwinGuzman (approved):

PR #1038, opened 2022-06-06
DSD-984: useNYPLBreakpoints hook

EdwinGuzman:
Fixes JIRA ticket [DSD-984](https://jira.nypl.org/browse/DSD-984)

- Creates a `useNYPLBreakpoints` hook to replace manually doing conditional checks with the `useWindowSize` hook and using magic numbers for breakpoint values.
- Now, instead of setting a setting a temporary value to "600", using `useEffect`, and the width from `useWindowSize`, this hook returns all of that combined into a single boolean value.

```
const {
  isLargerThanSmall,
  isLargerThanMedium,
  isLargerThanLarge,
  isLargerThanXLarge
} = useNYPLBreakpoints();
```

And now this can be used such as `{isLargerThanMedium ? <DesktopView /> : <MobileView />}`, for example.

## How has this been tested?

Locally in storybook and unit tests.
Note: a file was brought in from Chakra to help mocking the `MediaMatch` API.

- N/A

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/ButtonGroup/ButtonGroup.tsx
+++ b/src/components/ButtonGroup/ButtonGroup.tsx
@@ -3,7 +3,7 @@ import * as React from "react";
 
 import Button from "../Button/Button";
 import { LayoutTypes } from "../../helpers/types";
-import useWindowSize from "../../hooks/useWindowSize";
+import useNYPLBreakpoints from "../../hooks/useNYPLBreakpoints";
 
 export type ButtonGroupWidths = "default" | "full";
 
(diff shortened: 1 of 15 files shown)


src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -39,25 +39,9 @@ export const ButtonGroup = chakra(
...
-        setFinalButtonWidth("full");
-        setFinalLayout("column");
-      } else {
-        // Otherwise, set the layout and button width to the values
-        // passed in via the `buttonWidth` and `layout` props.
-        setFinalButtonWidth(buttonWidth);
-        setFinalLayout(layout);
-      }
-    }, [buttonWidth, layout, windowDimensions.width]);
+    const { isLargerThanMedium } = useNYPLBreakpoints();
```
bigfishdesign13: `isLargerThanMedium` is used three times right here in this file.  And I am guessing that is predominantly what I am going to see throughout the rest of the code changes, as that is the breakpoint we target for mobile-to-desktop UI format changes.  Would it make sense to also include something like `isLargerThanMobile`?

src/hooks/useNYPLBreakpoints.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * This hook is used to determine if the current screen size is larger than
+ * the specific NYPL breakpoint values. The returned value is an object with
+ * boolean values for each breakpoint.
+ */
+const useNYPLBreakpoints = () => {
+  const [
+    isLargerThanSmall,
+    isLargerThanMedium,
```
bigfishdesign13, line 11: Echoing my comment from `ButtonGroup.tsx`, would it make sense to also include something like isLargerThanMobile in addition to these four base options?

bigfishdesign13 (commented):
I really like this.  I have often wondered what we could add to eliminate the hardcoded breakpoint number values in the code.  

Ultimately, this looks good, but I have a question about a possible addition.

EdwinGuzman:
Updated with a `isLargerThanMobile` boolean value.

bigfishdesign13 (approved):
Awesome.  Ship it!

PR #1041, opened 2022-06-08
DSD-1001: Updated main SCSS file

EdwinGuzman:
Fixes JIRA ticket [DSD-1001](https://jira.nypl.org/browse/DSD-1001)

- The biggest update is that the `styles.scss` file now imports the `react-datepicker`'s uncompiled SCSS so _we_ can compile and minify it. Importing the already compiled CSS seems to fail _only_ in parcel-type apps.
- I want to remove the `breakpoint` scss package dependency which we do not use anymore. It's only there because some scss mixins use it but even those mixins are not used in the app. I did notice that SCOUT imports and declars this `breakpoint` package so it's not used right. We should let apps handle that if they need it.
- `breakpoint` was moved to `resources.scss` which isn't also used much in DRB, SCOUT, MLN, Turbine, or the parcel Library Registry app. I think we can do away with this unless if we think the DS should also be responsible for hosting useful scss mixins.
- The readme was updated with better and more information on how to import the compiled CSS file in TSX or SCSS/CSS.
- Renamed and moved the `Placeholder`'s styles file.

## How has this been tested?

Tested this branch in but the parcel Library Registry app and in the NextJS Turbine app. For both apps, I made sure that importing through the JS file and SCSS/CSS file works as expected.

- N/A

### Checklist:

### Front End Review:


README.md
```diff
@@ -196,22 +155,6 @@ Find more information about the Design System's internal use of Chakra to create
 
 Chakra was integrated into the Design System in version `0.25.0`. For those looking to update to a version greater than or equal `0.25.0`, check out our [Chakra Migration Guide](/CHAKRA_MIGRATION_GUIDE.md).
 
-## CDN
```
EdwinGuzman, line 199: This was here twice, probably due to a merge conflict and it went unnoticed.

src/components/Placeholder/_Placeholder.scss
```diff
@@ -22,12 +19,19 @@
...
-      margin-bottom: var(--nypl-space-xxxl);
-    }
   }
 
   &--short {
     padding: 1rem 1rem;
   }
 }
+
+@media (min-width: $nypl-breakpoint-large) {
```
EdwinGuzman, line 29: Reducing the `breakpoint` package's dependency.

src/styles/base/_02-mixins.scss
```diff
@@ -40,10 +40,6 @@
   width: 100%;
 }
 
-.container {
```
EdwinGuzman, line 43: Not being used.

README.md
```diff
@@ -267,6 +210,101 @@ $ npm run build-storybook:v1
 
 You can then view `/reservoir/v1/index.html` in your browser. _Make sure not to commit this directory_.
 
+## Accessibility
```
EdwinGuzman, line 225: This was moved, not updated.

README.md
```diff
@@ -138,7 +57,47 @@ const ApplicationContainer = (props) => {
 };
 ```
 
-4. Use DS components!
+3. Import the minified styles `/dist/styles.css` file in your app.
```
EdwinGuzman: This is the important and new info to review.

README.md
```diff
@@ -13,108 +13,27 @@ Storybook documentation
...
-| 4.                | [Using Chakra UI Components](#using-chakra-ui-components)                           |
-| 5.                | [Storybook](#storybook)                                                             |
-| 6.                | [Typescript Usage](#typescript-usage)                                               |
-| 7.                | [Unit Testing](#unit-testing)                                                       |
-| 8.                | [Accessibility](#accessibility)                                                     |
+| 1.                | [Using the Design System in Your Product](#using-the-design-system-in-your-product) |
+| 2.                | [Using Chakra UI Components](#using-chakra-ui-components)                           |
+| 3.                | [Storybook](#storybook)                                                             |
+| 4.                | [Accessibility](#accessibility)                                                     |
+| 5.                | [Contributing Quickstart](#contributing-quickstart)                                 |
```
EdwinGuzman, line 20: Moved the contributing info below all the "how to use" sections. Most devs reading the README will be interested in using it first before contributing back.

hokei:
This makes a lot of sense to me tho I'll let someone else approve - in particular, I agree with moving the "breakpoint" dependency - I don't think, Long Term, Reservoir should be responsible for maintaining scss mixins as it doesn't use them directly.

README.md
oliviawongnyc: As someone who hasn't done this a lot before, I might have read this line and tried to add `import "/dist/styles.css"` directly into my app. Maybe add a line, "How to do this will depend on the stack of your application." Then, you can delete that line from below.

README.md
```diff
@@ -138,7 +57,47 @@ const ApplicationContainer = (props) => {
...
+```scss
+// main.scss (for example)
+@import "@nypl/design-system-react-components/dist/styles.css";
+
+body {
+  ....
+}
+```
+
+Note: Using tilde to import scss/css is no longer a best practice for apps using _recent_ versions of webpack or parcel. For apps using parcel, prepend the string import with `npm:` such as :
```
oliviawongnyc: Since we don't see a tilde before this point, this line may be confusing. You could remove it, or if you think it may be applicable to some, then add a line after saying "But, if you are using an older bundler version, import the styles like so: ...."

README.md
```diff
@@ -138,7 +57,47 @@ const ApplicationContainer = (props) => {
...
+
+```scss
+@import "npm:@nypl/design-system-react-components/dist/styles.css";
+```
+
+4. Optionally import the `/dist/resources.scss` file in your main _scss_ file. This file contains useful SCSS mixins that can be used for global or component style rules.
```
oliviawongnyc: Add an example here?

oliviawongnyc (approved):

oliviawongnyc:
Tested this fix in my Parcel application and was able to import styles and render a DatePicker component flawlessly. Thank you!

README.md
bigfishdesign13: Good call.  I think it’s important to have clear `how to use` docs.

README.md
```diff
@@ -267,6 +210,101 @@ $ npm run build-storybook:v1
 
 You can then view `/reservoir/v1/index.html` in your browser. _Make sure not to commit this directory_.
 
+## Accessibility
+
+### Development and Storybook
+
+The Reservoir Design System is built with accessibility in mind. By using Chakra UI as our foundational base, the custom DS components built with Chakra have accessibility concerns already implemented. On top of built-in accessible elements, DS components internally work to link labels with input elements, to add correct `aria-*` attributes, to visually hide text but still associate it with the correct element for titles and descriptions, to auto-generate a random `id` attribute if none was passed, and much more.
```
bigfishdesign13: The note about the randomly generated IDs is no longer true and should be removed.

README.md
```diff
@@ -267,6 +210,101 @@ $ npm run build-storybook:v1
...
+- `@storybook/addon-a11y` for real-time accessibility testing in the browser through Storybook. _Every_ component has a tab that displays violations, passes, and incomplete checks performed by `axe-core`.
+
+If applicable, DS components have section(s) on accessibility in their Storybook documentation. For example, in the `Slider`'s [Storybook file](/src/components/Slider/Slider.stories.mdx), there are two "Accessibility" sections for each of the two `Slider` types, "single" and "range". This gives an explanation on additional changes we made to make the combination of elements in the `Slider` component accessible.
+
+### Product Requirements
+
+The Reservoir Design System provides accessible stories, but real live data can necessitate additional accessibility requirements beyond what we're committed to in our generic, extensible components. To ensure your products' final result is accessible, please adhere to the accessibility requirements put together by NYPL's accessibility coordinator on [Metronome](http://themetronome.co/).
```
bigfishdesign13: Should this statement say `real life data`? Or was it intended to be `real live data`?

src/components/Placeholder/_Placeholder.scss
bigfishdesign13: `Placeholder` should be refactored to use a theme file.  Not here necessarily.  We should just keep that in mind.

src/resources.scss
```diff
@@ -1,3 +1,5 @@
+@import "~breakpoint-sass/stylesheets/breakpoint";
```
bigfishdesign13, line 1: Just for clarification (again): `resources.css` is intended to be used by consuming apps and `styles.css` is for internal use?

bigfishdesign13 (approved):
How much do you know about how the consuming apps are actually using the SCSS files?  Let’s setup a ticket to figure this out and then, yes, let’s move away from these files.  We no longer use SCSS, so I don’t think we should be maintaining a library of mixins for consuming apps.

On a related not, the LibCal app is loading `styles.css`, but I don’t think it really needs to any more.  I dealt with a couple of UI issues last year related to those styles.
https://libcal.nypl.org/appointments/schwarzman

src/resources.scss
EdwinGuzman: No, they're both meant to be used by consuming apps but `resources.scss` is optional. It contains scss mixins that are not necessary for the DS to work.

EdwinGuzman:
> How much do you know about how the consuming apps are actually using the SCSS files? Let’s setup a ticket to figure this out and then, yes, let’s move away from these files. We no longer use SCSS, so I don’t think we should be maintaining a library of mixins for consuming apps.

Some apps are still pulling in the `styles.css` file but maybe it's because they don't know that it's okay to not use it. It does contain `DatePicker` styles so it is needed if they want to use that component.

> On a related not, the LibCal app is loading styles.css, but I don’t think it really needs to any more. I dealt with a couple of UI issues last year related to those styles.
https://libcal.nypl.org/appointments/schwarzman

I don't know how this app works but it looks like it's a script tag. If it's removed, nothing looks bad so yeah, we can make the recommendation to remove this. Looks like they don't use any DS components.

EdwinGuzman:
This was approved but the latest update can be reviewed if need be.

PR #1044, opened 2022-06-10
DSD-966: Mobile Header Login

EdwinGuzman:
Fixes JIRA ticket [DSD-966](https://jira.nypl.org/browse/DSD-966)

- Updates the desktop login menu in the `Header` component to render for the mobile view.
- Refactors a bit of the code.
- There are still a few places that need to be updated but will be a different ticket; didn't want to keep this growing.

## How has this been tested?

Storybook.

- N/A

### Checklist:

### Front End Review:


src/components/Header/components/LoginButton.tsx
```diff
@@ -0,0 +1,69 @@
...
+    });
+
+    useEffect(() => {
+      if (patronName && isLoginOpen) {
+        greetingRef.current.focus();
+      }
+    }, [isLoginOpen, patronName]);
+
+    return (
+      <FocusLock isDisabled={!isLoginOpen}>
```
oliviawongnyc, line 46: FYI, focus trap is slightly different in the current header than we have it. For both desktop and mobile, when the login drop down is opened, the focus jumps to the "Log Into the Catalog" button rather than staying on the open/close button.

src/theme/components/headerLogin.ts
```diff
@@ -0,0 +1,109 @@
...
+      textAlign: "left",
+      textTransform: "none",
+      _focus: {
+        boxShadow: isMobile ? null : "1px 1px 2px 2px #0f465c",
+        outline: isMobile ? "inset" : "none",
+        outlineStyle: isMobile ? "solid" : null,
+        outlineWidth: isMobile ? "0.1875em" : null,
+      },
+      ".greeting": {
+        fontStyle: "italic",
```
oliviawongnyc: Oddly enough, in the current Header the greeting is not italicized in mobile view.
Screen Shot 2022-06-14 at 3 17 32 PM

src/theme/components/headerLogin.ts
```diff
@@ -0,0 +1,109 @@
...
+      bg: isMobile ? "ui.black" : "ui.white",
+      borderRadius: isMobile ? "0" : "28px",
+      color: isMobile ? "ui.white" : "#1B7FA7",
+      fontSize: isMobile ? "18px" : null,
+      padding: isMobile ? "30px" : null,
+      textDecoration: isMobile ? "underline" : null,
+      textTransform: isMobile ? "uppercase" : null,
+      width: isMobile ? "100%" : "140px",
+      _focus: {
+        boxShadow: isMobile ? null : "1px 1px 2px 2px #0f465c",
```
oliviawongnyc: I think there should be a boxShadow for mobile as well. It seems like there is one in the current Header.

src/theme/components/headerLogin.ts
```diff
@@ -0,0 +1,109 @@
...
+        outline: isMobile ? "inset" : "none",
```
oliviawongnyc, line 103: For me, the outline around the log out button on Mobile seems a bit off. The top line is over the red buttons above and the button line is invisible (I think this is where the shadow would come in?). See image.
Screen Shot 2022-06-14 at 3 23 04 PM

src/components/Header/components/LoginButton.test.tsx
```diff
@@ -0,0 +1,201 @@
...
+        <LoginButton
+          isLoginOpen
+          isMobile
+          patronName="PATRON, JANE A"
+          setIsLoginOpen={setIsLoginOpen}
+        />
+      );
+
+      const greetingContainer = screen.getByTestId("patronGreeting");
+      expect(greetingContainer).toHaveFocus();
```
oliviawongnyc, line 225: This is cool – will have to remember!

oliviawongnyc (approved):
Looks good! A couple of comments on style and focus.

oliviawongnyc (changes requested):
Edwin, I expected better! Jk, just a few comments.

EdwinGuzman:
Updated

src/components/Header/Header.tsx
```diff
@@ -39,7 +39,7 @@ export const Header = chakra(() => {
     isWidthLarge,
   });
 
-  const [loginOpen, setLoginOpen] = useState<boolean>(false);
+  const [isLoginOpen, setIsLoginOpen] = useState<boolean>(false);
```
hokei, line 45: Is it a requirement that an open login should persist between desktop and mobile?  The current NYPL header doesn't do this.

hokei: Also because if state doesn't need to persist between the two, then this state can live in `<LoginButton>`

src/components/Header/components/Login.tsx
```diff
@@ -1,113 +1,92 @@
...
-      </Button>
-      {loginOpen && (
-        <VStack className="loginMenu">
-          {patronName && (
-            <Box
-              id="patronGreeting"
-              data-testid="patronGreeting"
-              ref={greeting}
-              tabIndex={0}
+const Login = chakra(
```
hokei, line 10: I wonder if we can namespace the components under `Header/components`.  While they aren't going to be exported, `Login` and `Mobile`, in particular, can be pretty confusing, and I think `Header.Login` or `NYPLHeader.Login`, etc, would make it clear that this shouldn't be used in trying to make a create a generic login component.

src/components/Header/components/LoginButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+        }
+      }
+    }, [isLoginOpen, patronName]);
+
+    return (
+      <FocusLock isDisabled={!isLoginOpen}>
+        <Button
+          aria-label={desktopButtonLabel}
+          buttonType="link"
+          id="loginButton"
```
hokei, line 50: Are the ids vestiges of an earlier time?  Can we remove them?

src/theme/components/headerLogin.ts
bigfishdesign13: The focus outline for Reservoir components is setup with a larger gap between the outline and the element, so that is what you are seeing.  And that if fine.  However, the outline on the first button is not layering properly when it has focus.  The z-index value that is part of the default focus styles is not doing its thing because the `a` element is inside the `li` element.  Of the top of my head, I am not sure how to fix this.

image

bigfishdesign13 (approved):
Ultimately, this works and does its thing.  There a number of styling issues that will need to be dealt with in the end, but it is functional and will work well for the POC.

src/components/Header/Header.tsx
EdwinGuzman: Right, it's not. Some of this has been updated in the latest PR so I'll keep iterating on a later PR.

src/components/Header/components/Login.tsx
EdwinGuzman: YEah, that's a good idea.

src/components/Header/components/LoginButton.tsx
EdwinGuzman: Will remove.

oliviawongnyc (approved):

bigfishdesign13:
@EdwinGuzman Are there additional changes you would like to make here?

EdwinGuzman:
No, the rest of updates are going in a new PR.

PR #1063, opened 2022-06-24
DSD-665: updated doc for Colors Style Guide

bigfishdesign13:
Fixes JIRA ticket [DSD-665](https://jira.nypl.org/browse/DSD-665)

- Updates the docs for the `Colors` Style Guide to include additional data related to contrast and WCAG compliance.
- Exports `ColorCard`.
- The `ColorCard` component is not yet responsive.  That will need to be added in a future update. 
- Use the `UI Fills` and `Brand Fills` screens in the Figma Main file as reference: https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/?node-id=10733%3A56919

## How has this been tested?

- Added a11y details related to contrast ratio and WCAG compliance to the `Colors` Style Guide docs.

### Checklist:

### Front End Review:


src/components/StyleGuide/Colors.stories.mdx
```diff
@@ -6,82 +6,377 @@ import SimpleGrid from "../Grid/SimpleGrid";
 import { getCategory } from "../../utils/componentCategories";
 import DSProvider from "../../theme/provider";
 
+export const defaultData = {
```
EdwinGuzman: A better name for this would be "colorContrastData".

src/components/StyleGuide/Colors.stories.mdx
```diff
@@ -6,82 +6,377 @@ import SimpleGrid from "../Grid/SimpleGrid";
...
   ],
 };
 
-export const makeColorCard = function (key) {
-  const cssValue = `var(${key})`;
+export const makeColorCard = function (data) {
+  const name = data.name;
+  const dataBlack = data.dataBlack;
+  const dataWhite = data.dataWhite;
+  const textColor = data.textColor;
```
EdwinGuzman: Use variable destructing here:
```
const { dataBlack, dataWhite, name, textColor } = data;
```

src/components/StyleGuide/Colors.stories.mdx
```diff
@@ -6,82 +6,377 @@ import SimpleGrid from "../Grid/SimpleGrid";
...
+      dataBlack={dataBlack}
+      dataWhite={dataWhite}
+      textColor={textColor}
+      key={name}
+    />
   );
   return card;
 };
-
 export const getColorCards = (cat) => {
```
EdwinGuzman: Can this be updated to "category" rather than "cat"?

src/components/StyleGuide/Colors.stories.mdx
```diff
@@ -6,82 +6,377 @@ import SimpleGrid from "../Grid/SimpleGrid";
...
+    },
   ],
   education: [
-    "--nypl-colors-section-education-primary",
-    "--nypl-colors-section-education-secondary",
+    {
+      name: "section.education.primary",
+      dataBlack: ["3.94", "Fail", "AA"],
+      dataWhite: ["5.33", "AA", "AAA"],
+      textColor: "ui.white",
```
EdwinGuzman, line 134: Why is some information here and not in the `defaultData` array?

EdwinGuzman (changes requested):
This looks great and I really like this but some concerns:
- I get in general what the table means but someone looking at it won't know what "6.08:1" means. There's no context for the color contrast ratio value and that should be a heading in the table.
- The first value for "ui.white" and "ui.black" should also have a heading for "font color" or something.
- In general, there should be more information before the "Brand Colors" section that explains what the table is and how to read it. "AA" and "AAA" doesn't mean much without context or links to resources for further information. Also something like, "these are the combinations that pass and don't pass accessibility tests", "we don't recommend to use those combinations because they are hard to read", "if you use it, make sure the font size is appropriate", etc.

Questions:
- Why do you need to export `ColorCard`?
- The data being shown is tabular data. Why not use the `Table` component instead of making a custom layout? With the `Table` component, the "row headers" can be used since that is also how the table should be read.

src/components/StyleGuide/Colors.stories.mdx
bigfishdesign13: I created `defaultData` to house objects that would be used more than once.  Right now, those that are not included in `defaultData` only appear that one time.  This follows the pattern that we use to handle typography styles.

bigfishdesign13:
@EdwinGuzman See below…

> This looks great and I really like this but some concerns:
> 
> * I get in general what the table means but someone looking at it won't know what "6.08:1" means. There's no context for the color contrast ratio value and that should be a heading in the table.
> * The first value for "ui.white" and "ui.black" should also have a heading for "font color" or something.
> * In general, there should be more information before the "Brand Colors" section that explains what the table is and how to read it. "AA" and "AAA" doesn't mean much without context or links to resources for further information. Also something like, "these are the combinations that pass and don't pass accessibility tests", "we don't recommend to use those combinations because they are hard to read", "if you use it, make sure the font size is appropriate", etc.

I just did a lot of work on these same color cards in the Figma file and the questions you raise here also came up in my mind as I was working in Figma.  I started to address those concerns there and I will mirror those same changes here.

> 
> Questions:
> 
> * Why do you need to export `ColorCard`?

If we are going to include something similar in Turbine, we should just reuse what we already created here.

> * The data being shown is tabular data. Why not use the `Table` component instead of making a custom layout? With the `Table` component, the "row headers" can be used since that is also how the table should be read.

The spacing in the `Table` component didn’t match the smaller footprint I desired, but let me see if I can tweak the styles to get it to do what I need.

src/components/StyleGuide/ColorCard.tsx
```diff
@@ -1,43 +1,174 @@
...
-/**
- * ColorCard
- * Component only used for Storybook.
- */
-export default function ColorCard(
-  props: React.PropsWithChildren<ColorCardProps>
-) {
-  const { backgroundColor, colorName = "" } = props;
+export interface DataTableProps {
+  /** data related to black against current color */
```
EdwinGuzman: Nit: should be sentences with capitalization and period.

src/components/StyleGuide/ColorCard.tsx
```diff
@@ -1,43 +1,174 @@
...
+    dataWhite[2] === "AAA") && <SuccessIcon />;
+  const blackSmallTextSuccess = (dataBlack[1] === "AA" ||
+    dataBlack[1] === "AAA") && <SuccessIcon />;
+  const blackLargeTextSuccess = (dataBlack[2] === "AA" ||
+    dataBlack[2] === "AAA") && <SuccessIcon />;
+  const tableData = [
+    [
+      <>
+        <span style={{ color: "white", padding: 0 }}>ui.white</span>
+      </>,
```
EdwinGuzman: The fragment is not needed since the span is a single element.

src/components/StyleGuide/ColorCard.tsx
```diff
@@ -1,43 +1,174 @@
...
+      <>
+        {dataWhite[2]}
+        {whiteLargeTextSuccess}
+      </>,
+    ],
+    [
+      <>
+        <span style={{ color: "black", padding: 0 }}>ui.black</span>
+      </>,
```
EdwinGuzman: The fragment is not needed since the span is a single element.

src/components/StyleGuide/Colors.stories.mdx
```diff
@@ -91,33 +394,59 @@ export const getColorCards = (cat) => {
...
-| ---------------------------------------- |
-| 1. [Brand Colors](#brand-colors)         |
-| 2. [Section Colors](#section-colors)     |
-| 3. [Utility Colors](#utility-colors)     |
-| 3. [Grayscale Colors](#grayscale-colors) |
-| 4. [Figma Reference](#figma-reference)   |
+| Table of Contents                              |
+| ---------------------------------------------- |
+| 1. [Overview](#overview)                       |
+| 2. [Brand Fills](brand-fills)                  |
```
EdwinGuzman: This and "UI Fills" are missing the pound `#` sign for the heading link.

src/components/StyleGuide/Colors.stories.mdx
```diff
@@ -91,33 +394,59 @@ export const getColorCards = (cat) => {
...
+must be at least 4.5:1. To meet the `WCAG AAA` standard, the contrast ratio must
+be at least 7:1.
 
-## Brand Colors
+## Brand Fills
 
-Brand fills are the NYPL primary and secondary colors.
+These colors are connected to the NYPL brand and are not intended to be generic.
+
+### NYPL Brand Colors
```
EdwinGuzman, line 432: I find it a bit odd that the TOC includes h2s and h3s while it appears as if they are all at the same level. Maybe the TOC for these h3s should skip the number prefix?

src/components/StyleGuide/Colors.stories.mdx
```diff
@@ -91,33 +394,59 @@ export const getColorCards = (cat) => {
...
+| 3. [NYPL Brand Colors](#nypl-brand-colors)     |
+| 4. [NYPL Section Colors](#nypl-section-colors) |
+| 5. [UI Fills](ui-fills)                        |
+| 6. [Utility Colors](#utility-colors)           |
+| 7. [Grayscale Colors](#grayscale-colors)       |
+| 8. [Figma Reference](#figma-reference)         |
+
+## Overview
+
+The Reservoir Design System (DS) the provides the NYPL branded and approved
```
EdwinGuzman: Typo, extra "the" in `"(DS) the provides"`.

src/components/StyleGuide/Colors.stories.mdx
```diff
@@ -91,33 +394,59 @@ export const getColorCards = (cat) => {
...
+color palette. The NYPL color palette is used to style the DS components and
+is also available for consuming applications in the form of javascript and
+CSS design tokens. Please refer to the [Design Tokens](http://localhost:6006/?path=/story/style-guide-design-tokens--page)
```
EdwinGuzman: I'm surprised the link works on Vercel. You can update it to `[Design Tokens](/?path=/story/style-guide-design-tokens--page)`

src/components/StyleGuide/Colors.stories.mdx
```diff
@@ -91,33 +394,59 @@ export const getColorCards = (cat) => {
...
+page for information about consuming and using all style tokens.
+
+Whenever using these colors, designer must ensure that the foreground and
+background colors pass, at a minimum, the `WCAG AA` accessibility guidelines for
+color contrast. For reference, to meet the `WCAG AA` standard, the contrast ratio
+must be at least 4.5:1. To meet the `WCAG AAA` standard, the contrast ratio must
```
EdwinGuzman: Can you say `WCAG 2.1 AA` and `WCAG 2.1 AAA` and link to https://www.w3.org/TR/WCAG21/#contrast-minimum?

EdwinGuzman (changes requested):
Just very minor changes.

src/components/StyleGuide/Colors.stories.mdx
bigfishdesign13: I was also a little worried about that.  I converted the TOC format to now match the format on the story pages, which allows for better nesting and visual clarity.  I will add new ticket to do the same for the other `Style Guide` pages.

src/components/StyleGuide/Colors.stories.mdx
bigfishdesign13: Oops. :(

EdwinGuzman (approved):

PR #1066, opened 2022-06-29
DSD-1045: Header component naming

EdwinGuzman:
Fixes JIRA ticket [DSD-1045](https://jira.nypl.org/browse/DSD-1045)

- ~While setting up full pages to make old vs new `Header` component comparisons in [Turbine](https://nypl-ds-test-1ef6b2009-nypl.vercel.app/fullPages/header-mobile), I noticed that the desktop DS `Header` has rendering issues caused by different initial values from the `useNYPLBreakpoints` hook. Now, that hook stores its values internally through state and updates it on a useEffect so that they always start out as false and then update based on the browser's viewport width. Before, it would _immeditately_ tell you if something was "larger than mobile" (for example). This is fine on the client but the server doesn't know about media queries or viewport widths so it would be false.~
  - The above was fixed in a different PR. 
- The specific error can be seen on this page: https://nypl-ds-test-1ef6b2009-nypl.vercel.app/fullPages/header
  - The initial view does not render properly but if the browser is resized, it will fix itself. 
  - The above was fixed in a different PR.
- Removed the `HeaderComponent.` prefixes because it caused circular dependencies. They were only seen and logged to the console when using `npm link`. The main `Header` component still uses this but individual header-only components don't use the prefix and import to avoid the circular dependency.

## How has this been tested?

Locally and in Turbine.

- N/A

### Checklist:

### Front End Review:


src/components/Header/components/LoginButton.tsx
```diff
@@ -64,7 +64,7 @@ const LoginButton = chakra(({ isMobile = false }: LoginButtonProps) => {
           />
         </Button>
         {isOpen && (
-          <HeaderComponents.Login
+          <Login
```
bigfishdesign13: I understand that you need to remove `HeaderComponents.`, but I now worry that the component names are too generic.  How would you feel about changing the component names to include `Header` as part of the full component name?  `HeaderLoginButton`, `HeaderUpperNav`, etc.

I’m thinking about this because we’re starting to work non-atomic level components and we could potentially see other components with similar names.

bigfishdesign13 (commented):
I have a question about naming conventions.

src/components/Header/components/LoginButton.tsx
EdwinGuzman: Yeah, I'll update this PR with that change. The namespacing, unfortunately, didn't work as well because of the circular dependencies.

oliviawongnyc (approved):
I don't know what was going on with the circular dependencies, but this looks good to me!

EdwinGuzman:
I originally had an `index.ts` file to namespace all the components. This was helpful when importing it in the `Header.tsx` component and doing something like 
```
<HeaderComponents.LoginButton ... />
<HeaderComponents.SearchButton ... />
```

The problem is that I was doing the same in components within this file. So the `index.ts` file contained both `LoginButton` and `Login`, for example. `LoginButton` imports `Login` but I think I had something in `Header` that `Login` used. So it was `Header` -> `LoginButton` -> `Login` -> `Header`. I don't remember the _exact_ components where this circular dependency was an issue. In any case, the prefix will make these component names clearer for their intent.

PR #1074, opened 2022-07-06
DSD-1050: Header timeout functions

EdwinGuzman:
Fixes JIRA ticket [DSD-1050](https://jira.nypl.org/browse/DSD-1050)

- Reimplements the code to completely log out a patron from Encore/Catalog in the Header.
- The code is based on https://github.com/NYPL/dgx-header-component/blob/master/src/utils/encoreCatalogLogOutTimer.js

The logic flow is as follows:
- The `Header` component mounts
- The patron is not logged in:
  - Cool, delete the `nyplIdentityPatron` and `VALID_DOMAIN_LAST_VISITED` cookies if they exist, just to make sure we log the patron out completely.
- The patron is logged in:
  - The patron is on a "valid" domain or the `VALID_DOMAIN_LAST_VISITED` is not available
    - Create the `VALID_DOMAIN_LAST_VISITED` with the current time and log the patron out in 30 minutes.
  -  The patron is not on a "valid" domain
      - Get the time they were last seen on a valid domain through the `VALID_DOMAIN_LAST_VISITED` cookie.
        - They were not last seen on a valid domain, so immediately log them out.
        - Use the last time they were seen on a valid domain to get a new session timeout. Log the patron out in that new time.

When the patron is logged out, the `PAT_LOGGED_IN`, `VALID_DOMAIN_LAST_VISITED`, and `nyplIdentityPatron` cookies are deleted.

## How has this been tested?

Locally and waiting ⏲️  for the function to log me out from Encore and the Catalog while being on `local.nypl.org`.

- The loaded iframe is set to `display: none` so it should not be a problem.

### Checklist:

### Front End Review:


EdwinGuzman:
@nonword if you get a chance, please review this. The Vercel preview site won't work to test this since the cookies are not available on that domain.

nonword (approved):
The core logic looks essentially identical to the original to me

src/components/Header/utils/encoreCatalogLogOutTimer.ts
```diff
@@ -0,0 +1,129 @@
...
+  /**
+   * The timer to delete logged in related cookies and call the method to
+   * completely log out from Encore and Catalog.
+   */
+  this.logOutFromEncoreAndCatalogIn = (time: number) => {
+    let timeUntilLogOut = time > 0 ? time : 0;
+
+    // When in test mode, set the timeout no longer than 2 seconds.
+    if (isTestMode && timeUntilLogOut > 8000) {
+      timeUntilLogOut = 1000;
```
bigfishdesign13, line 84: The comment above refers to setting a value to `2 seconds`, but here is it set to 1 second.  Why the descrepancy?

bigfishdesign13 (approved):
This looks good, but I did not really test it.  I am marking as `approved`, but let me know if you'd like to see more thorough testing before merging.

src/components/Header/utils/encoreCatalogLogOutTimer.ts
EdwinGuzman: True, I updated it because I didn't think the test should have that much of a time out. I'll update the comment

src/components/Header/Header.tsx
```diff
@@ -63,6 +70,12 @@ export const Header = chakra(({ gaOptions = {}, isProduction = true }) => {
     }
   };
 
+  // Once the `Header` component is mounted, start a timer that will
+  // log the user out of Encore and the NYPL Catalog after 30 minutes.
```
oliviawongnyc, line 74: Was it always 30 minutes? I have a recollection of it being an hour, but could be wrong!

src/components/Header/utils/encoreCatalogLogOutTimer.ts
```diff
@@ -0,0 +1,129 @@
...
+      Cookies.remove("nyplIdentityPatron");
+
+      this.loadLogOutIframe(isTestMode);
+    }, timeUntilLogOut);
+  };
+
+  /**
+   * The function that loads and inserts a temporary `iframe` element into the
+   * body of the current page. The `iframe` has the log out endpoint to
+   * completely log out the patrong from Encore and Catalog. The `iframe`
```
oliviawongnyc: Nit: Typo in word, "patron".

oliviawongnyc (commented):
I read this over several times and think I understand it, but definitely found it hard to visualize. Might it be useful to include a section on this logic in the Header storybook page?

src/components/Header/Header.tsx
EdwinGuzman: For this timer, it always was 30 minutes. Maybe there's something else that's an hour?

EdwinGuzman:
> Might it be useful to include a section on this logic in the Header storybook page? 

Yes. That page is actually empty. I'll create a ticket that will be just to fill out that page, how to use it, its logic, etc. But, I'll ask you to fill out a few sections like the cookie login logic and the focus-trap implementation in the drop downs.

PR #1078, opened 2022-07-11
DSD-1016: Enabling forwarding refs to all components

EdwinGuzman:
Fixes JIRA ticket [DSd-1016](https://jira.nypl.org/browse/DSD-1016)

- Enables forwarding refs to all components through React's `forwardRef` function.
- Adds new tests for this functionality.

## How has this been tested?

Locally in Storybook. I used Chakra's own `Tooltip` to test the DS components and see if that component works as a baseline. Given that it works, other packages/components should work (including our own `Tooltip` component once it's complete). I specifically checked the components recommended that this will be used from the `Tooptip` spec doc:

- `Heading`: notice that this goes full width so the tooltip displays right in the middle of the entire component.
Screen Shot 2022-07-08 at 11 32 54 AM

- `Icon`
Screen Shot 2022-07-08 at 11 39 04 AM

- `Link`
Screen Shot 2022-07-08 at 11 42 24 AM

- `Logo`
Screen Shot 2022-07-08 at 11 43 56 AM

- `ProgressIndicator`
Screen Shot 2022-07-08 at 11 45 20 AM

- `StatusBadge`
Screen Shot 2022-07-08 at 11 46 17 AM

- `Text`
Screen Shot 2022-07-08 at 11 47 14 AM

- `Radio` (Form example) - the placement can be fixed later
Screen Shot 2022-07-08 at 11 48 31 AM

- N/A

### Checklist:

### Front End Review:


EdwinGuzman:
Note: in this PR, the `import * as React from "react";` when using `React.forwardRef` was updated to `import React, { forwardRef } from "react";`. According to the React docs, the latter is the preferred way and when we move over to react 18, we can remove the empty `import React from "react";` import (this is because Babel can do the jsx transform for us now but this is a separate thread).

More info under [Removing Unused React Imports](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#removing-unused-react-imports)

> Change all default React imports (i.e. import React from "react") to destructured named imports (ex. import { useState } from "react") which is the preferred style going into the future.

bigfishdesign13 (approved):
A lot to review, but everything appears to be in good shape. I am also relying on the testing you provided as validation.

PR #1079, opened 2022-07-11
DSD-1048: Implementing padding-inline/margin-inline for RTL support

EdwinGuzman:
Fixes JIRA ticket [DSD-1048](https://jira.nypl.org/browse/DSD-1048)

- Implements `marginStart`, `marginEnd`, `paddingStart`, and `paddingEnd` instead of `marginLeft`, `marginRight`, `paddingLeft`, and `paddingRight`, respectively. This is to support RTL directionality.
- The "Bidirectionality" documentation page was also updated to include this DS implementation.
- Internally, Chakra converts the `...Start` and `...End` properties to the "inline" equivalent. For example, `marginStart` is converted to `margin-inline-start`. For consistency and clarity, the "inline" word is not used in this PR conversion. The above makes it easier to implement this because of the difference between "inline" and "block".

For example, this is a manual test of `margin-inline-start` and manual-block-start`:
Screen Shot 2022-07-11 at 11 30 21 AM

Screen Shot 2022-07-11 at 11 30 13 AM

## How has this been tested?

Locally for all components affected in Storybook.

- This is better for directionality.

### Checklist:

### Front End Review:


src/theme/components/template.ts
```diff
@@ -1,8 +1,8 @@
 // From @mixin breakout:
 // Used for elements that should be edge-to-edge.
 const breakout = {
-  marginLeft: "calc(-50vw + 50%)",
-  marginRight: "calc(-50vw + 50%)",
+  marginEnd: "calc(-50vw + 50%)",
```
bigfishdesign13, line 4: I was curious if it woudl make sense to leave these as `marginLeft` and `marginRight`, but the UI appears to be working fine even with this update. :)

bigfishdesign13 (approved):
All looks good.

PR #1080, opened 2022-07-12
DSD-1052: Initial Footer component

EdwinGuzman:
Fixes JIRA ticket [DSD-1052](https://jira.nypl.org/browse/DSD-1052)

- Adds the initial version of the DS `Footer` component.
- This replicates the current production footer that exists on NYPL.
- Concerns that will be new tickets:
  - The DS social media icons don't match with production icons. There is already a ticket to get the new icons into the DS.
  - The media query breakpoints are different. Production uses these four: "320px", "768px", "1024px", "1500px". The DS uses these four: "320px", "600px", "960px", "1280px", "1536px". While not vastly different, the visual changes are noticeable. We should discuss this for a separate PR but do we want to replicate the legacy breakpoints and import them into the DS, or ignore them and use the current ones in the DS? 
- This is the React component. A solution still needs to be created for the static HTML/CSS version. This is TBD in a separate ticket.

## How has this been tested?

Locally.

- N/A

### Checklist:

### Front End Review:


src/components/Footer/Footer.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+    expect(icons[1]).toHaveAttribute("title", "NYPL on Twitter");
+    expect(icons[2]).toHaveAttribute("title", "NYPL on Instagram");
+    expect(icons[3]).toHaveAttribute("title", "NYPL on Youtube");
+  });
+
+  it("renders a footer image of the NYPL building", () => {
+    const logo = screen.getAllByRole("img")[4];
+
+    expect(logo).toBeInTheDocument();
+    expect(logo).toHaveAttribute(
```
bigfishdesign13: Why did you do this test by `src` and the logo by `title`?  Could you instead test for the `alt` text here?

src/theme/components/fieldset.ts
```diff
@@ -7,7 +7,7 @@ interface FieldSetProps {
 
 const Fieldset = {
   baseStyle: ({ isLegendHidden }: FieldSetProps) => {
-    const screenreaderStyles = isLegendHidden ? screenreaderOnly : {};
+    const screenreaderStyles = isLegendHidden ? screenreaderOnly() : {};
```
bigfishdesign13, line 10: Why this change?

src/theme/components/footer.ts
```diff
@@ -0,0 +1,132 @@
+import { displayScreenreaderOnly, screenreaderOnly } from "./globalMixins";
+
+export const footerGray = "#54514A";
```
bigfishdesign13, line 3: Great.  This gray sits almost half way between `ui.gray.dark` and `ui.gray.x-dark`.  If it was closer to one or the other, I would push to make the jump to our gray, but not when it's this far off.  Too bad.

src/theme/components/footer.ts
```diff
@@ -0,0 +1,132 @@
...
+      fontWeight: "400",
+      margin: "0 0.2rem",
+      textDecoration: "none",
+      _hover: {
+        color: "ui.white",
+        textDecoration: "underline",
+      },
+      _focus: {
+        backgroundColor: footerGray,
+        outline: "solid 2px var(--nypl-colors-ui-white)",
```
bigfishdesign13: For border styles, we've split this into two attributes so we could use the theme object: `border` and `borderColor`.  Maybe use the same approach here?

src/theme/components/globalMixins.ts
```diff
@@ -2,13 +2,28 @@
...
  */
-const screenreaderOnly = {
+
+/**
+ * If the element that should be visually hidden should only be hidden on
+ * small screens, then pass in true for the `isMobileOnly` argument. Otherwise,
+ * this will set an element to always be visually hidden regardless of the
+ * screen size.
+ */
+const screenreaderOnly = (isMobileOnly = false) => ({
```
bigfishdesign13, line 12: Ignore my other question about this.  I see now why you made that other change.

bigfishdesign13 (commented):
Overall, this look quite good.  I have a few questions sprinnkled throughout.

One thing that I noticed as I was comparing with the footer on NYPL.org is that we use font-smoothing, whereas they do not.  Before we roll out the `Header` and `Footer` components, we may need to make a push to have that attribute added to other projects.  It would be weird to have a page with mixed font treatments.

Why are the breakpoints so different between Reservoir and DXP?  Are the breakpoints you are referring to the same as what Bill is using in Scout?  If so, how does he deal with the discrepancies?

EdwinGuzman:
> One thing that I noticed as I was comparing with the footer on NYPL.org is that we use font-smoothing, whereas they do not. Before we roll out the Header and Footer components, we may need to make a push to have that attribute added to other projects. It would be weird to have a page with mixed font treatments.

I don't have a keen eye to notice this right away... can you make a ticket for this?

> Why are the breakpoints so different between Reservoir and DXP? Are the breakpoints you are referring to the same as what Bill is using in Scout? If so, how does he deal with the discrepancies?

The breakpoints from the Header/Footer are from the older design system based on what was at that point, the most common breakpoints for all devices. These have change since now devices have vastly different sizes. It looks like Scout is using the DS breakpoints: https://github.com/NYPL/dxp-react-search/blob/development/src/styles/base/_breakpoints.scss

Instead of importing the Header/Footer breakpoints, we are building the two new components with the DS breakpoints. That's the only big discrepancy between the old and new (besides the font treatment).

src/components/Footer/Footer.test.tsx
EdwinGuzman: It didn't work at first but I think it'll work with the latest updates from `development`. I'll try it and see if it works, otherwise, the reason I did the different tests is because the DOM didn't pick it up.

EdwinGuzman:
Updated. Got the image by alt text now in the unit test.

bigfishdesign13 (approved):
All good.  However, it looks like there are a lot more modified files now.  The merge from development must have done that?

EdwinGuzman:
I had to update the base branch from `development`, now it's down to 11 changed files.

PR #1094, opened 2022-07-25
DSD-1068: Updating how rest props propagate in the Link component

EdwinGuzman:
Fixes JIRA ticket [DSD-1068](https://jira.nypl.org/browse/DSD-1068)

- Propagates the `rest` props, usually style props, from the `Link` component into its span element when used with third-party libraries such as React Router.

## How has this been tested?

Locally and in Turbine.

- N/A

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/Link/Link.tsx
+++ b/src/components/Link/Link.tsx
@@ -107,7 +107,9 @@ export const Link = chakra(
     let variant = "link";
 
     if (typeof children === "string" && !href) {
-      throw new Error("`Link` needs the `href` prop.");
+      throw new Error(
+        "NYPL Reservoir Link: The `Link` component needs the `href` prop if its child element is a string."
+      );
     }
 
     if (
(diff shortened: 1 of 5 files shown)


src/components/Link/Link.tsx
```diff
@@ -107,7 +107,9 @@ export const Link = chakra(
     let variant = "link";
 
     if (typeof children === "string" && !href) {
-      throw new Error("`Link` needs the `href` prop.");
+      throw new Error(
+        "NYPL Reservoir Link: The `Link` component needs the `href` prop if its child element is a string."
```
bigfishdesign13, line 111: Interesting that we missed these during the push to v1.

bigfishdesign13 (approved):
Looks good.  Are there other components that should have this type of propagation added?

src/components/Link/Link.tsx
EdwinGuzman: I think because these are actual errors and not warnings.

EdwinGuzman:
> Are there other components that should have this type of propagation added?

Probably. At the minimum, all components have this feature set for their wrapper element. In this case, there is a separate way of rendering and we didn't test that use case.

I think more of these "issues" or a need for more flexibility will come up as we reach and try to push the limits of what we can do with these components.

PR #1095, opened 2022-07-25
DSD-1069: Updating select options props in SearchBar

EdwinGuzman:
Fixes JIRA ticket [DSD-1069](https://jira.nypl.org/browse/DSD-1069)

- Passes "text" and "value" properties for the `SearchBar`'s internal `Select` component through the `optionsData` prop.
- This is needed for `<option>` HTML elements that the `Select` must render that have different visual text labels and internal "value" attribute values.

## How has this been tested?

Locally.

- N/A

### Checklist:

### Front End Review:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 ## Prerelease
 
+### Updates
+
+- Updates how options are passed to the `SearchBar` component for its internal `Select` component.
+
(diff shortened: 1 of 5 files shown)


bigfishdesign13 (approved):
This looks good, but one question.  When we were talking about this the other day, in my head, this update was going to be made directly to the `Select` component.  Having said that, what do you think about adding a prop to `Select` to ingest a data object array and dynamically build up the `option` elements?  Similar to how `List` handles the `li` elements.

PR #1098, opened 2022-07-28
DSD-46: Using grid instead of flexbox for the template components

EdwinGuzman:
Fixes JIRA ticket [DSD-46](https://jira.nypl.org/browse/DSD-46)

- Uses grid instead of flexbox for template components.
- The biggest change is the width of the sidebar updating from 255px to 288px. It is visually the same width as before -- because of the gap between columns.

## How has this been tested?

Locally in Storybook and also through `npm link` in the Turbine app.

- N/A

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/Template/Template.tsx
+++ b/src/components/Template/Template.tsx
@@ -294,14 +294,16 @@ export const TemplateAppContainer = chakra(
     );
     return (
       <Template ref={ref} {...rest}>
-        {renderSkipNavigation ? <SkipNavigation /> : null}
-        {aboveHeaderElem}
-        {(header || breakoutElem) && (
-          <TemplateHeader renderHeaderElement={renderHeaderElement}>
-            {header}
-            {breakoutElem}
-          </TemplateHeader>
-        )}
+        <Box>
+          {renderSkipNavigation ? <SkipNavigation /> : null}
+          {aboveHeaderElem}
+          {(header || breakoutElem) && (
+            <TemplateHeader renderHeaderElement={renderHeaderElement}>
+              {header}
+              {breakoutElem}
+            </TemplateHeader>
+          )}
+        </Box>
         {/* Note that setting `sidebar` as a prop here affects the
        TemplateContentSidebar and TemplateContentPrimary components. */}
         <TemplateContent id={contentId} sidebar={sidebar}>
--- a/src/components/Template/Template.tsx
+++ b/src/components/Template/Template.tsx
@@ -90,7 +90,7 @@ const TemplateHeader = ({
   children,
   renderHeaderElement = true,
 }: React.PropsWithChildren<TemplateHeaderProps>) => {
-  const styles = useStyleConfig("TemplateHeader", {});
+  const styles = useStyleConfig("TemplateBreakout", {});
   let headerElement = <Box __css={styles}>{children}</Box>;
 
   // The user wants to render the `header` HTML element.
@@ -121,8 +121,7 @@ const TemplateHeader = ({
  * for the `Breadcrumbs` and `Hero` components, and other banner-like components.
  */
 const TemplateBreakout = (props: React.PropsWithChildren<TemplateProps>) => {
-  const styles = useStyleConfig("TemplateBreakout", {});
-  return <Box __css={styles}>{props.children}</Box>;
+  return <Box>{props.children}</Box>;
 };
 
 /**
@@ -140,7 +139,7 @@ const TemplateContent = (
 ) => {
   const { children, id = "mainContent", sidebar = "none" } = props;
   const styles = useStyleConfig("TemplateContent", {
-    variant: sidebar !== "none" ? "sidebar" : null,
+    variant: sidebar,
   });
   // Manually pass in the `sidebar` prop to the `TemplateContentPrimary` and
   // `TemplateContentSidebar` components.
@@ -227,6 +226,7 @@ const TemplateFooter = ({
   children,
   renderFooterElement = true,
 }: React.PropsWithChildren<TemplateFooterProps>) => {
+  const styles = useStyleConfig("TemplateBreakout", {});
   let footerElement = <>{children}</>;
 
   // The user wants to render the `footer` HTML element.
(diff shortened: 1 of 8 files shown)


src/components/Template/Template.tsx
```diff
@@ -90,7 +90,7 @@ const TemplateHeader = ({
   children,
   renderHeaderElement = true,
 }: React.PropsWithChildren<TemplateHeaderProps>) => {
-  const styles = useStyleConfig("TemplateHeader", {});
+  const styles = useStyleConfig("TemplateBreakout", {});
```
bigfishdesign13, line 95: Did you find commonalities between `TemplateHeader` and `TemplateBreakout` and determined that one could be removed?

src/theme/components/template.ts
```diff
@@ -1,111 +1,98 @@
...
+  // the column for the sidebar is max 288px width.
   variants: {
-    sidebar: {
-      display: "flex",
-      flexFlow: {
-        base: "column nowrap",
-        md: "row wrap",
-      },
+    left: {
+      gridTemplateColumns: { md: "288px 1fr" },
```
bigfishdesign13: Should this be 287px?  255 + 32

Regardless, can we use `gap` and allow the width of the sidebar to be set to `255px`?  Including the gap value as part of hardcoded width value assumes that the grid gap will never change.

bigfishdesign13 (commented):
It looks like it is working well, but I do have some questions.  I will do a final review after those questions are answered.

src/components/Template/Template.tsx
EdwinGuzman: Correct. The only difference was the name of the object.

src/theme/components/template.ts
EdwinGuzman: When I set it to 255px, the width is smaller than what I expected. This was when I compared the flexbox sidebar with the grid sidebar in Turbine. I'll try setting the gap value but I'm not sure it'll render as expected.

EdwinGuzman: The gap works.

EdwinGuzman:
Updated using 255px for the sidebar and using a gap value of "grid.l".

src/theme/components/template.ts
```diff
@@ -1,111 +1,91 @@
...
     },
   },
 };
 const TemplateContentTop = {
   baseStyle: {
-    flex: "1 0 100%",
-    marginBottom: "page.hstack", // --nypl-space-l
+    gridColumn: { base: "1", md: "1 / span 2" },
+    height: "100%",
+    marginBottom: "page.hstack",
```
bigfishdesign13: Now that we are using `gap` on the grid, this can be removed.

src/theme/components/template.ts
```diff
@@ -1,111 +1,91 @@
...
   },
 };
 const TemplateContentPrimary = {
   baseStyle: {
-    flex: "1 1",
-    marginBottom: "page.hstack", // --nypl-space-l
+    marginBottom: "page.hstack",
```
bigfishdesign13: Now that we are using `gap` on the grid, this can be removed.

src/theme/components/template.ts
```diff
@@ -1,111 +1,91 @@
...
       minWidth: { md: 0 },
     },
   },
 };
 const TemplateContentSidebar = {
   baseStyle: {
-    flex: { base: "0 0 auto", md: "0 0 255px" },
-    order: { md: "page.vstack" },
-    marginBottom: "page.hstack", // --nypl-space-l
+    marginBottom: "page.hstack",
```
bigfishdesign13: Now that we are using `gap` on the grid, this can be removed.

bigfishdesign13 (changes requested):
Now that we are using the `gap` attribute on the parent `grid` element, there are some additional style attributes that need to be removed from children within the `grid` .

EdwinGuzman:
Updated. I noticed there's a new Figma design for this component. In there, there's a "Content Bottom" section but that's not in the DS. Should a new ticket be created to match the Figma file?

bigfishdesign13:
> Updated. I noticed there's a new Figma design for this component. In there, there's a "Content Bottom" section but that's not in the DS. Should a new ticket be created to match the Figma file?

Weird.  I never realized the descrepency.  Adding the `Content Bottom` region should be a separate ticket.

src/theme/components/template.ts
```diff
@@ -1,111 +1,83 @@
...
-  paddingEnd: "s",
-  paddingStart: "s",
   width: "100%",
+  // This could be "1 / 4" and it would mean the same. This is
+  // "future-proof" the grid column assignment to the last column.
+  gridColumn: "1 / -1",
 };
 
 const Template = {
   baseStyle: {
```
bigfishdesign13, line 6: Add the following attribute to this element:
```
rowGap: “grid.l"
```

bigfishdesign13 (changes requested):
With the addition of the `rowGap` attibute on the parent grid element, all of the elements that are considered part of the header need to be wrapped in either a `div` or `Box`.  Wrapping those elements will remove unwanted gaps coming from the `rowGap` attribute.  See lines 297 thru 304 in `Template.tsx`.

src/theme/components/template.ts
```diff
@@ -1,111 +1,84 @@
...
-  marginX: "auto",
-  maxWidth: "1280px",
-  paddingTop: 0,
-  paddingBottom: 0,
-  paddingEnd: "s",
-  paddingStart: "s",
   width: "100%",
+  // This could be "1 / 4" and it would mean the same. This is
+  // "future-proof" the grid column assignment to the last column.
+  gridColumn: "1 / -1",
```
bigfishdesign13: This is no longer working for all the header elements.  The styles probably need to be added to the new `<Box>` wrapper.  If that is the issues, that fix will work in a prop controlled implementation of `Template`, but this may be a problem for using the template children.  How do we ensure that the breakout styles are applied to that wrapper elements?

bigfishdesign13 (commented):
We lost some of the styles for the breakout elements.

src/theme/components/template.ts
EdwinGuzman: Made a separate PR to work on this. Let me know what you think. #1109

EdwinGuzman:
I think most of the fixes were covered in the external PR, but another review would be great.

bigfishdesign13 (approved):
Everything looks good in this PR.

PR #1102, opened 2022-08-01
DSD-1065: Getting and displaying static Footer html and css

EdwinGuzman:
Fixes JIRA ticket [DSD-1065](https://jira.nypl.org/browse/DSD-1065)

- Displays the `Footer` component's HTML and CSS in Storybook documentation.
- Still need to figure out if this is the best location to keep the HTML/CSS in or if it should go into the Header App.
- This is also not the best way to maintain this static code because it requires manual updates. Have not been able to find an easy way to extract CSS from emotion or css-in-js tools. If we had created this in just SCSS or CSS, this wouldn't have been a problem.

## How has this been tested?

Locally.

- N/A

### Checklist:

### Front End Review:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24338,30 +24338,47 @@
       "dev": true
     },
     "node_modules/node-notifier": {
-      "version": "6.0.0",
-      "resolved": "https://registry.npmjs.org/node-notifier/-/node-notifier-6.0.0.tgz",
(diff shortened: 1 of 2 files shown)


EdwinGuzman:
Not really "blocked" but using it to make a note that this is a good approach that should be implemented later on.

EdwinGuzman:
Not needed right now. Closing!

PR #1109, opened 2022-08-05
DSD-46: [Addendum] Adding TemplateContentBottom and updating the Template DOM structure

EdwinGuzman:
This is in addition to PR #1098 .

- Fixes an issue found in the original PR but adds more to the `Template` components. Specifically, it adds a `TemplateContentBottom` component and also updates how the `TemplateBreakout` component is used. It also moves the `SkipNavigation` component _outside_ of the `Template` component but still renders it as part of the `TemplateAppContainer` component. All of this is more of a proposal so that's why it's a separate PR. It's not fully complete but if you think it's fine, I'll merge it and then update the main PR with more comments.

## How has this been tested?

Locally and in Turbine. All template pages render as expected.

- N/A

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/Template/Template.stories.mdx
+++ b/src/components/Template/Template.stories.mdx
@@ -28,14 +28,15 @@ import SkipNavigation from "../SkipNavigation/SkipNavigation";
 import {
   Template,
   TemplateAboveHeader,
-  TemplateHeader,
+  TemplateAppContainer,
   TemplateBreakout,
   TemplateContent,
-  TemplateContentTop,
+  TemplateContentBottom,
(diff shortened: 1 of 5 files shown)


bigfishdesign13 (approved):
This looks good and functions properly.  On a related note, can you think of any apps in NYPL that would choose to use the template children method rather than use `TemplateAppContainer`?

EdwinGuzman:
I can't think of any. I would say perhaps it would be used for more flexibility but that's also not really true... The `TemplateAppContainer` component is pretty flexible.

PR #1110, opened 2022-08-08
DSD-1091: Passing props in the Image component to the correct location

EdwinGuzman:
Fixes JIRA ticket [DSD-1091](https://jira.nypl.org/browse/DSD-1091)

- Correctly passes down HTML `img` attributes in the `Image` component to the `img` element.
- This came up in MLN where error checking was being performed in the `Image` component by checking its width and height. An `onLoad` callback function was passed but it was being passed to the wrapper `div` element and not the `img` element itself. Therefore, the width and height captured was wrong since it was from the `div` element and not the `img`.
- This PR passes down the `...rest` props to the `img` element.
- More work needs to be done to better delineate styling props for the wrappers and props for the `img` element itself, since we include different style props for the potential wrappers and figure elements. But, this can wait for another PR.

## How has this been tested?

Locally in Storybook.
It's difficult to unit test because images are not loaded in jest.

- N/A

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/Image/Image.stories.mdx
+++ b/src/components/Image/Image.stories.mdx
@@ -11,6 +11,7 @@ import { withDesign } from "storybook-addon-designs";
 import Heading from "../Heading/Heading";
 import Image from "./Image";
 import SimpleGrid from "../Grid/SimpleGrid";
+import Text from "../Text/Text";
 import { getCategory } from "../../utils/componentCategories";
 
 export const imageBlockStyles = {
@@ -66,7 +67,7 @@ export const imageBlockStyles = {
 | Component Version | DS Version |
 | ----------------- | ---------- |
 | Added             | `0.0.6`    |
-| Latest            | `1.0.7`    |
+| Latest            | `1.0.9`    |
 
 ## Table of Contents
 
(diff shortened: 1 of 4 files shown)


src/components/Image/Image.stories.mdx
```diff
@@ -66,7 +67,7 @@ export const imageBlockStyles = {
 | Component Version | DS Version |
 | ----------------- | ---------- |
 | Added             | `0.0.6`    |
-| Latest            | `1.0.7`    |
+| Latest            | `1.0.9`    |
```
bigfishdesign13, line 70: Why `1.0.9` and not `1.0.8`?

src/components/Image/Image.tsx
```diff
@@ -177,6 +179,7 @@ export const Image = chakra(
         loading={isLazy ? "lazy" : undefined}
         {...srcProp}
         __css={{ ...styles.img, ...additionalImageStyles }}
+        {...rest}
```
bigfishdesign13, line 182: This is a very logical chage.  As a dev, I would assume any props that I was passing in would end up on the `img` itself.  One step at a time, right?

bigfishdesign13 (approved):
I have one question about the Latest Version number, but this looks good.

PR #1112, opened 2022-08-08
DSD-1075: MultiSelect accessibility fixes

EdwinGuzman:
Fixes JIRA ticket [DSD-1075](https://jira.nypl.org/browse/DSD-1075)

### Issue 1

This fixes the accessibility issue `"Interactive controls must not be nested"` found by axe when items are selected: the selected count button that renders inside the `MultiSelectMenuButton` is the issue. When using VoiceOver on a Mac, I was able to focus on both the menu button and the selected count button and VoiceOver read _both_ buttons correctly. So, on an n=1 test case, it worked for me. I did not test on other machines with other screenreaders so I do think it's still an issue.

The proposed update is going from

```jsx
<button ...>
  <span role="button">
    <span>2</span><svg>
  </span>
  <span>title</title>
  <svg>
</button>
```

to 

```jsx
<>
  <button ...>
    <span>title</title>
    <svg>
  </button>
  <span role="button">
    <span>2</span><svg>
  </span>
</>
```

The selected count button now renders _after_ the main button but is still visually displayed above the main menu button. This also helps prevent click events from the selected count button from bubbling up to the main button (as noted in two unit tests).

An `aria-label` was also added so that screen readers now read the button with context. Before, it use to say something like "2 MultiSelect Listbox" when it read the entire button which can be confusing. It also just read "2" for the selected count button itself.

Screen Shot 2022-08-08 at 2 15 13 PM

With the update:

Screen Shot 2022-08-08 at 12 43 53 PM

### Issue 2

The second issue is `"Elements should not have tabindex greater than zero”`. This occurs for the "dialog" variant when it is opened and is caused by the `react-focus-lock` usage (which is used by Chakra's `focus-lock` implementation). I think in this case we can proceed and say that we _know_ this is an issue and a limitation, but also part of expected behavior we want to implement. While is not directly an issue we implemented, it's caused by something we use that implements a behavior we want. Typically, someone would write a VPAT document to list this issue but I'm not sure who would do it in this case or that is the right approach.

For now, let's list this down in the Storybook doc as something we know is an issue but doesn't affect component or app usage when it is rendered. It is only and issue when it is opened but something we can live with since we _want_ to trap focus within the component and the rest of the page is irrelevant.

## How has this been tested?

Locally in Storybook.

- Yeah, fixes one and explicitly lists down an issue we will have to proceed with until a better solution appears in the near (?) future.

### Checklist:

### Front End Review:

### DIFF
--- a/src/theme/components/multiSelectMenuButton.ts
+++ b/src/theme/components/multiSelectMenuButton.ts
@@ -14,6 +14,7 @@ const MultiSelectMenuButton = {
       textAlign: "left",
       textOverflow: "ellipsis",
       whiteSpace: "nowrap",
+      transition: "margin 150ms ease-out",
     },
     menuButton: {
       alignItems: "center",
--- a/src/theme/components/multiSelectMenuButton.ts
+++ b/src/theme/components/multiSelectMenuButton.ts
@@ -5,10 +5,11 @@ const MultiSelectMenuButton = {
     "selectedItemsCountButton",
     "toggleIcon",
   ],
-  baseStyle: ({ isOpen = false }) => ({
+  baseStyle: ({ hasSelectedItems = false, isOpen = false }) => ({
     buttonLabel: {
       justifyContent: "flex-start",
       overflow: "hidden",
+      marginLeft: hasSelectedItems ? "50px" : "0",
       marginRight: "l",
       textAlign: "left",
       textOverflow: "ellipsis",
@@ -20,7 +21,7 @@ const MultiSelectMenuButton = {
       borderBottomLeftRadius: isOpen ? "0" : "button.default",
       borderBottomRightRadius: isOpen ? "0" : "button.default",
       borderColor: isOpen ? "ui.gray.dark" : "ui.gray.medium",
-      justifyContent: "flex-start",
+      justifyContent: "space-between",
       width: "100%",
       _hover: {
         backgroundColor: isOpen ? "ui.gray.x-light-cool" : "ui.white",
@@ -40,7 +41,10 @@ const MultiSelectMenuButton = {
       flexShrink: 0,
       fontSize: "text.tag",
       justifyContent: "flex-end",
+      left: "15px",
       marginRight: "xs",
+      position: "absolute",
+      top: "10px",
       width: "46px",
       _hover: {
         borderColor: "ui.gray.dark",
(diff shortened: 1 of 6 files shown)


src/theme/components/multiSelectMenuButton.ts
```diff
@@ -5,10 +5,11 @@ const MultiSelectMenuButton = {
     "selectedItemsCountButton",
     "toggleIcon",
   ],
-  baseStyle: ({ isOpen = false }) => ({
+  baseStyle: ({ hasSelectedItems = false, isOpen = false }) => ({
     buttonLabel: {
       justifyContent: "flex-start",
       overflow: "hidden",
+      marginLeft: hasSelectedItems ? "50px" : "0",
```
bigfishdesign13, line 12: Now that the selected items button is not displacing the button label text, we are losing the animation on the label text.  I thought the `transition` attribute was built into the CSS for all elements, but it looks like that is not the case and we’ll need to add the following for `buttonLabel`.

`transition: "margin 150ms ease-out”`

But that should only be applied when the first checkbox is selected.  When clearing the checkboxes, the selected items button disappears immediately, so the marign on the label text should not transition from 50 back to 0.  It should jump immediately from one to the next.  I am not sure which variable would indicate that action.

bigfishdesign13 (changes requested):
This looks good, except for one style change.

I just took a look at the [Multiselect](https://carbondesignsystem.com/components/dropdown/usage/#multiselect) variant of Carbon’s `Dropdown` component and they have it setup with the same structure that is throwing the error for our component.

EdwinGuzman:
> I just took a look at the [Multiselect](https://carbondesignsystem.com/components/dropdown/usage/#multiselect) variant of Carbon’s Dropdown component and they have it setup with the same structure that is throwing the error for our component.

Yes, and the error exists (axe in devtools):

Screen Shot 2022-08-09 at 5 11 06 PM

EdwinGuzman:
@bigfishdesign13 added the animation. Not much needs to be done because the `margin-left` style is added when there is at least one item selected, so it doesn't have to further check or name sure it's not added again. Likewise, when it's removed, the animation still kicks in.

src/theme/components/multiSelectMenuButton.ts
```diff
@@ -14,6 +14,7 @@ const MultiSelectMenuButton = {
       textAlign: "left",
       textOverflow: "ellipsis",
       whiteSpace: "nowrap",
+      transition: "margin 150ms ease-out",
```
bigfishdesign13, line 17: Nit: alphabetize the attributes

bigfishdesign13 (approved):
Smooth.

PR #1118, opened 2022-08-17
DSD-336: Tooltip Component

jackiequach:
Fixes JIRA ticket [DSD-336](https://jira.nypl.org/browse/DSD-336)

- Adds Tooltip component

## How has this been tested?

Unit tests and in Storybook

- Aria-label is needed when Image/Icon is passed as content

### Checklist:

### Front End Review:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,13 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 ## Prerelease
 
+### Adds
+
+- Adds `Tooltip` component.
+
 ### Fixes
 
 - Fixes the styling for the `Image` component in the `Hero` "secondary" variant.
-- Adds `Tooltip` component.
 
 ## 1.0.8 (August 11, 2022)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 ### Fixes
 
 - Fixes the styling for the `Image` component in the `Hero` "secondary" variant.
+- Adds `Tooltip` component.
 
 ## 1.0.8 (August 11, 2022)
 
--- a/src/components/Tooltip/Tooltip.stories.mdx
+++ b/src/components/Tooltip/Tooltip.stories.mdx
@@ -5,16 +5,17 @@ import {
   Meta,
   Story,
 } from "@storybook/addon-docs";
-import ReactDOMServer from "react-dom/server";
 import { withDesign } from "storybook-addon-designs";
 
 import Tooltip from "./Tooltip";
 import Button from "../Button/Button";
 import Radio from "../Radio/Radio";
+import RadioGroup from "../RadioGroup/RadioGroup";
 import Icon from "../Icons/Icon";
 import Link from "../Link/Link";
 import Label from "../Label/Label";
 import Image from "../Image/Image";
+import { VStack } from "@chakra-ui/react";
 import { getCategory } from "../../utils/componentCategories";
 import DSProvider from "../../theme/provider";
 
@@ -30,6 +31,7 @@ import DSProvider from "../../theme/provider";
     jest: ["Tooltip.test.tsx"],
   }}
   argTypes={{
+    children: { control: false },
     className: { control: false },
     id: { control: false },
     isDisabled: {
@@ -42,8 +44,8 @@ import DSProvider from "../../theme/provider";
 
 | Component Version | DS Version |
 | ----------------- | ---------- |
-| Added             | `1.0.9`    |
-| Latest            | `1.0.9`    |
+| Added             | `1.1.0`    |
+| Latest            | `1.1.0`    |
 
 ## Table of Contents
 
@@ -65,20 +67,16 @@ component when a user hovers or focuses on the component.
   <Story
     name="Tooltip with Controls"
     args={{
+      children: undefined,
+      className: undefined,
       content:
         "This is a tooltip with information describing the hovered object.",
-      className: undefined,
       id: "tooltip-id",
       isDisabled: false,
     }}
   >
     {(args) => (
-      <Tooltip
-        content={args.content}
-        className={args.className}
-        id={args.id}
-        isDisabled={args.isDisabled}
-      >
+      <Tooltip {...args}>
         <Button id="tooltip-button">Hover me</Button>
       </Tooltip>
     )}
@@ -89,11 +87,15 @@ component when a user hovers or focuses on the component.
 
 ## Accessibility
 
+The DS `Tooltip` component is implemented with Chakra UI and Chakra handles the
+appropriate aria-describedby attribute for the element that the tooltip is associated with.
+
 The `Tooltip` component is accessible via keyboard. The color contrast between
 foreground color and background color is 4.5:1. If text size is 200%, the
-button scales with text so there is no overlap. If a tooltip is associated
-with a form control (text, radio, checkbox, etc), the text for the
-tooltip should be included in the form control label.
+button scales with text so there is no overlap.
+
+If a tooltip is associated with a form control (text, radio, checkbox, etc),
+the text for the tooltip should be included in the form control label.
 
 Resources:
 
@@ -105,38 +107,38 @@ Resources:
 
 ### Tooltips on form options
 
-The `Tooltip` component is included in the `labelText` prop of the radios.
+When using the `Tooltip` component in forms containing `Checkbox`, `Radio`, `Toggle`.
 
 <Canvas>
   <DSProvider>
-    <Label htmlFor="radio1">What would you like to request?</Label>
-    <Radio
-      type="radio"
-      id="radio1"
-      labelText={
-        <Tooltip content="General library orientation or a customized research lesson">
-          Class Visit
-        </Tooltip>
-      }
-    />
-    <Radio
-      type="radio"
-      id="radio2"
-      labelText={
-        <Tooltip content="Guided exploration of the library and its resources">
-          Group Tour
-        </Tooltip>
-      }
-    />
-    <Radio
-      type="radio"
-      id="radio3"
-      labelText={
-        <Tooltip content="Invite library staff to share resources at your community event">
-          Offsite Community Outreach
-        </Tooltip>
-      }
-    />
+    <RadioGroup
+      labelText="What would you like to request?"
+      id="radio-group"
+      name="tooltip-example"
+    >
+      <Tooltip
+        content="General library orientation or a customized research lesson"
+        shouldWrapChildren
+      >
+        <Radio id="radio1" labelText="Class Visit" value="Class Visit" />
+      </Tooltip>
+      <Tooltip
+        content="Guided exploration of the library and its resources"
+        shouldWrapChildren
+      >
+        <Radio id="radio2" labelText="Group Tour" value="Group Tour" />
+      </Tooltip>
+      <Tooltip
+        content="Invite library staff to share resources at your community event"
+        shouldWrapChildren
+      >
+        <Radio
+          id="radio3"
+          labelText="Offsite Community Outreach"
+          value="Offsite Community Outreach"
+        />
+      </Tooltip>
+    </RadioGroup>
   </DSProvider>
 </Canvas>
 
@@ -181,5 +183,43 @@ The `Tooltip` component is included in the `labelText` prop of the radios.
     >
       Hover to see image
     </Tooltip>
+  </DSProvider>
+</Canvas>
+
+### Disabling Tooltip
+
+A tooltip can be disabled by passing the `isDisabled` prop.
+
+<Canvas>
+  <DSProvider>
+    <Tooltip
+      content="This is a tooltip with information describing the hovered object."
+      isDisabled
+    >
+      <Button id="tooltip-button">Hover me</Button>
+    </Tooltip>
+  </DSProvider>
+</Canvas>
+
+By default, the `Tooltip` is not shown for a disabled `Button`. To show the `Tooltip`
+on a disabled `Button`, pass the `shouldWrapChildren` prop.
+
+<Canvas>
+  <DSProvider>
+    <VStack spacing="input.group.button.vstack" width="150px">
+      <Tooltip content="This is a tooltip with information describing the hovered object.">
+        <Button id="tooltip-button" isDisabled>
+          Disabled Tooltip
+        </Button>
+      </Tooltip>
+      <Tooltip
+        content="This is a tooltip with information describing the hovered object."
+        shouldWrapChildren
+      >
+        <Button id="tooltip-button" isDisabled>
+          Enabled Tooltip
+        </Button>
+      </Tooltip>
+    </VStack>
   </DSProvider>
 </Canvas>
--- a/src/components/Tooltip/Tooltip.stories.mdx
+++ b/src/components/Tooltip/Tooltip.stories.mdx
@@ -1,0 +1,185 @@
+import {
+  ArgsTable,
+  Canvas,
+  Description,
+  Meta,
+  Story,
+} from "@storybook/addon-docs";
+import ReactDOMServer from "react-dom/server";
+import { withDesign } from "storybook-addon-designs";
+
+import Tooltip from "./Tooltip";
+import Button from "../Button/Button";
+import Radio from "../Radio/Radio";
+import Icon from "../Icons/Icon";
+import Link from "../Link/Link";
+import Label from "../Label/Label";
+import Image from "../Image/Image";
+import { getCategory } from "../../utils/componentCategories";
+import DSProvider from "../../theme/provider";
+
+<Meta
+  title={getCategory("Tooltip")}
+  component={Tooltip}
+  decorators={[withDesign]}
+  parameters={{
+    design: {
+      type: "figma",
+      url: "https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Main?node-id=36835%3A25702",
+    },
+    jest: ["Tooltip.test.tsx"],
+  }}
+  argTypes={{
+    className: { control: false },
+    id: { control: false },
+    isDisabled: {
+      table: { defaultValue: { summary: false } },
+    },
+  }}
+/>
+
+# Tooltip
+
+| Component Version | DS Version |
+| ----------------- | ---------- |
+| Added             | `1.0.9`    |
+| Latest            | `1.0.9`    |
+
+## Table of Contents
+
+- [Overview](#overview)
+- [Component Props](#component-props)
+- [Accessibility](#accessibility)
+- [Examples](#examples)
+
+## Overview
+
+<Description of={Tooltip} />
+
+The `Tooltip` component shows contextual help or information about a specific
+component when a user hovers or focuses on the component.
+
+## Component Props
+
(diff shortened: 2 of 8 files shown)


src/components/Tooltip/Tooltip.stories.mdx
```diff
@@ -0,0 +1,185 @@
...
+  decorators={[withDesign]}
+  parameters={{
+    design: {
+      type: "figma",
+      url: "https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Main?node-id=36835%3A25702",
+    },
+    jest: ["Tooltip.test.tsx"],
+  }}
+  argTypes={{
+    className: { control: false },
```
bigfishdesign13, line 35: Add `children: { control: false }`

src/components/Tooltip/Tooltip.stories.mdx
```diff
@@ -0,0 +1,185 @@
...
+      table: { defaultValue: { summary: false } },
+    },
+  }}
+/>
+
+# Tooltip
+
+| Component Version | DS Version |
+| ----------------- | ---------- |
+| Added             | `1.0.9`    |
```
bigfishdesign13: Change these both to `1.1.0`.

src/components/Tooltip/Tooltip.stories.mdx
```diff
@@ -0,0 +1,185 @@
...
+The `Tooltip` component shows contextual help or information about a specific
+component when a user hovers or focuses on the component.
+
+## Component Props
+
+<Canvas withToolbar>
+  <Story
+    name="Tooltip with Controls"
+    args={{
+      content:
```
bigfishdesign13, line 75: Alphabetize the prop names and add a line for `children: undefined`.  Adding the entry to `children` ensures that the props in the arg table on the Story page are properly alphabetized.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 ### Fixes
 
 - Fixes the styling for the `Image` component in the `Hero` "secondary" variant.
+- Adds `Tooltip` component.
```
EdwinGuzman: Add this under a new heading `### Adds` above the `### Fixes` heading.

src/components/Tooltip/Tooltip.stories.mdx
```diff
@@ -0,0 +1,185 @@
+import {
+  ArgsTable,
+  Canvas,
+  Description,
+  Meta,
+  Story,
+} from "@storybook/addon-docs";
+import ReactDOMServer from "react-dom/server";
```
EdwinGuzman: It doesn't seem like this is being used in this file.

src/components/Tooltip/Tooltip.stories.mdx
```diff
@@ -0,0 +1,185 @@
...
+    },
+  }}
+/>
+
+# Tooltip
+
+| Component Version | DS Version |
+| ----------------- | ---------- |
+| Added             | `1.0.9`    |
+| Latest            | `1.0.9`    |
```
EdwinGuzman: Let's bump this up to `1.1.0` since this will be included with other new components.

src/components/Tooltip/Tooltip.stories.mdx
```diff
@@ -0,0 +1,185 @@
...
+
+## Examples
+
+### Tooltips on form options
+
+The `Tooltip` component is included in the `labelText` prop of the radios.
+
+<Canvas>
+  <DSProvider>
+    <Label htmlFor="radio1">What would you like to request?</Label>
```
bigfishdesign13: Use the `RadioGroup` component to build this example.  Wrap the `Tooltip` around the entire `Radio` component and not just the radio label.

src/components/Tooltip/Tooltip.stories.mdx
EdwinGuzman, line 35: Add `children: { table: { disable: true } },`, or maybe even just `{ control: false}`. It shouldn't be an option that a user can control in the Controls prop section.

src/components/Tooltip/Tooltip.stories.mdx
```diff
@@ -0,0 +1,185 @@
...
+      id: "tooltip-id",
+      isDisabled: false,
+    }}
+  >
+    {(args) => (
+      <Tooltip
+        content={args.content}
+        className={args.className}
+        id={args.id}
+        isDisabled={args.isDisabled}
```
EdwinGuzman: This is fine. You can also do `<Tooltip {...args}>` as both are equivalent.

src/components/Tooltip/Tooltip.stories.mdx
EdwinGuzman: A couple of notes here:
- The `Radio` component already renders a `<Label>` internally so this is not needed.
- To create the equivalent, wrap all three `Radio`s in a `RadioGroup` component: `<RadioGroup labelText="What would you like to request?" name="tooltip-example">`

src/theme/components/tooltip.ts
```diff
@@ -0,0 +1,14 @@
+import { cssVar } from "@chakra-ui/theme-tools";
+
+const $arrowBg = cssVar("popper-arrow-bg");
+
+const Tooltip = {
+  baseStyle: {
```
bigfishdesign13, line 6: The styles needs some more adjustments in order to match the Figma designs.  Please update the `baseStyle` with the following:

```
[$arrowBg.variable]: "colors.ui.gray.xx-dark",
bg: "ui.gray.xx-dark",
borderRadius: "4px",
color: "ui.white",
fontSize: "text.tag",
marginBottom: "xxs",
maxWidth: "240px",
px: "s",
py: "xs”,
```

src/components/Tooltip/Tooltip.stories.mdx
```diff
@@ -0,0 +1,185 @@
...
+
+### Tooltips on form options
+
+The `Tooltip` component is included in the `labelText` prop of the radios.
+
+<Canvas>
+  <DSProvider>
+    <Label htmlFor="radio1">What would you like to request?</Label>
+    <Radio
+      type="radio"
```
EdwinGuzman: The `type` prop isn't part of the DS `Radio`.

src/components/Tooltip/Tooltip.stories.mdx
```diff
@@ -0,0 +1,185 @@
...
+        </Tooltip>
+      }
+    />
+    <Radio
+      type="radio"
+      id="radio2"
+      labelText={
+        <Tooltip content="Guided exploration of the library and its resources">
+          Group Tour
+        </Tooltip>
```
EdwinGuzman: Just curious, does it work if the `Tooltip` wraps the entire `Radio`?

```
<Tooltip content="Guided exploration of the library and its resources">
   <Radio
      type="radio"
      id="radio2"
      labelText="Group Tour"
   />
</Tooltip>
```

src/components/Tooltip/Tooltip.stories.mdx
```diff
@@ -0,0 +1,185 @@
...
+with a form control (text, radio, checkbox, etc), the text for the
+tooltip should be included in the form control label.
+
+Resources:
+
+- [W3C WAI ARIA Tooltip](https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/)
+- [MDN ARIA: tooltip role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role)
+- [Chakra UI Tooltip](https://chakra-ui.com/docs/components/overlay/tooltip)
+
+## Examples
```
EdwinGuzman, line 110: Can you add one example describing what happens when `isDisabled` is true?

src/components/Tooltip/Tooltip.test.tsx
```diff
@@ -0,0 +1,159 @@
...
+    render(<TooltipOnButton content={tooltipContent} />);
+    fireEvent.mouseOver(screen.getByText(buttonLabel));
+    const tooltip = await screen.findByRole("tooltip");
+    expect(await axe(tooltip)).toHaveNoViolations();
+  });
+
+  it("passes axe accessibility test with Image content", async () => {
+    render(
+      <Tooltip
+        content={<Image src="https://placeimg.com/300/300/animals?x=1" />}
```
EdwinGuzman: I would imagine that this fails because there is no alt text.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,45 @@
...
+export const Tooltip = chakra(
+  forwardRef<HTMLDivElement, TooltipProps>((props, ref?) => {
+    const { children, className, content, id, isDisabled, ...rest } = props;
+
+    const styles = useMultiStyleConfig("Tooltip", {});
+
+    return (
+      <ChakraTooltip
+        hasArrow
+        aria-label={typeof content !== "string" ? "Tooltip" : undefined}
```
EdwinGuzman, line 62: I think this might need to be `content` when it is a string instead of `undefined`. `"Tooltip"` is a bit too generic but I don't have any better suggestions right now.

src/theme/components/tooltip.ts
```diff
@@ -0,0 +1,14 @@
...
+    [$arrowBg.variable]: "colors.ui.gray.xx-dark",
```
EdwinGuzman: This is interesting. How does it work?

src/components/Tooltip/Tooltip.stories.mdx
```diff
@@ -0,0 +1,185 @@
...
+      </Tooltip>
+    )}
+  </Story>
+</Canvas>
+
+<ArgsTable story="Tooltip with Controls" />
+
+## Accessibility
+
+The `Tooltip` component is accessible via keyboard. The color contrast between
```
bigfishdesign13, line 97: Add a new paragraph at `If a tooltip is associated…`

Add the following paragraph to the Accessibility section…

The DS `Tooltip` component is implemented with Chakra UI and Chakra handles the appropriate `aria-describedby` attribute for the element that the tooltip is associated with.

bigfishdesign13 (changes requested):
The component is working very nicely.  Just need to clean up the styles and the a11y docs.

EdwinGuzman (changes requested):
The tests and implementation are great. There are minor update that need to be made, mostly for the documentation.

There is one concern: the type for `content` is `string | React.ReactNode` and this works fine for using an `Icon` or `Image` component. But technically, you can pass in _any_ React node so any other component will render which is not what we want.

Let's add a failsafe for this. Keep the type as `string | React.ReactNode` but in the `Tooltip` function, go through the `content` prop when it is _not_ a string and check the type of it. I think you can do something like

```
React.Children.map(content as React.ReactNode, (contentChild, key) => {
  if (contentChild.type !== Icon || contentChild.type !== Image) {
  ....
  }
})
```

You'd have to import the `Icon` and `Image` component. You might also need to do `contentChild.props.mdxType !== "Icon"` (notice now it's in quotes). This is to verify the type in the MDX storybook file. However you check the type, you can then log a warning saying something like `"NYPL Reservoir Tooltip: Pass in a string, Icon, or Image into the 'content' prop.". We can figure out the wording when you get to it.

Of course, there might be a different way to do this so if there's a better way (possibly through the prop type), go for it.

src/theme/components/tooltip.ts
jackiequach: It overwrites the existing CSS variable defined in the [Chakra Tooltip theme](https://github.com/chakra-ui/chakra-ui/blob/main/packages/theme/src/components/tooltip.ts). The variable is used to set the background of the Chakra component. This needs to be done since the popper arrow is not part of the tooltip body. 

I'm going to update this to use "tooltip-bg" instead of "popper-arrow-bg" since it can be used to set the bg once.

You can read more about it here: https://chakra-ui.com/docs/styled-system/css-variables

EdwinGuzman: Makes sense. This Chakra component doesn't break the styles apart like other components, so this is new to me.

src/components/Tooltip/Tooltip.tsx
jackiequach: Correct me if I'm wrong here: aria-labels are used to describe elements that are not already defined through text, so the tooltip content should serve as definition. Otherwise, the aria-label will be redundant with the given string.

src/components/Tooltip/Tooltip.stories.mdx
jackiequach: When `shouldWrapChildren` is passed to the ChakraTooltip, the width of the wrapping `span`  is defaulted to take the full width due to the `Radio`s being wrapped in a `Stack`  in `RadioGroup`. To get around this,  I used `ComponentWrapper` with `width: "fit-content"`. Let me know if you think a different approach would be better.

src/components/Tooltip/Tooltip.stories.mdx
```diff
@@ -0,0 +1,225 @@
...
+    <Tooltip
+      content="This is a tooltip with information describing the hovered object."
+      isDisabled
+    >
+      <Button id="tooltip-button">Hover me</Button>
+    </Tooltip>
+  </DSProvider>
+</Canvas>
+
+By default, the `Tooltip` is not shown for a disabled `Button`. To show the `Tooltip`
```
EdwinGuzman, line 222: This is a great example. Is this something that Chakra is doing by default and so our own components behave the same way?

src/components/Tooltip/Tooltip.tsx
EdwinGuzman: Yes, you're right. Let's say you're viewing a page with a screenreader and there are multiple `Tooltip`s on the page that render icons or images. In that case, all the tooltips will say "Tooltip" and will be confusing for users. I'm not saying this is wrong, it can just be improved but I don't know how. We can consider this as an issue at a later time, so let's leave this for now.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,77 @@
...
+      ...rest
+    } = props;
+
+    if (typeof content !== "string") {
+      React.Children.map(
+        content as React.ReactNode,
+        (contentChild: React.ReactElement) => {
+          if (contentChild.type !== Icon || contentChild.type !== Image) {
+            console.warn(
+              "NYPL Reservoir Tooltip: Pass in a string, Icon, or Image into the 'content' prop."
```
EdwinGuzman: Can you add a test for this? Try passing in a random component into the `content` prop. Here's an example of a test: https://github.com/NYPL/nypl-design-system/blob/development/src/components/Checkbox/Checkbox.test.tsx#L247

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,77 @@
...
+            );
+          }
+        }
+      );
+    }
+    let newChildren: React.ReactNode;
+    newChildren = shouldWrapChildren ? (
```
EdwinGuzman: You can simplify this as `let newChildren: React.ReactNode = shouldWrapChildren ? ....`

src/theme/components/componentWrapper.ts
```diff
@@ -5,6 +5,7 @@ interface ComponentWrapperProps {
 const ComponentWrapper = {
   parts: ["helperErrorText"],
   baseStyle: ({ hasChildren }: ComponentWrapperProps) => ({
+    width: "fit-content",
```
EdwinGuzman: @bigfishdesign13 I suggested this and haven't noticed any issues in any component that uses `ComponentWrapper`. Can you double check?

src/components/Tooltip/Tooltip.stories.mdx
```diff
@@ -0,0 +1,225 @@
...
+    },
+    jest: ["Tooltip.test.tsx"],
+  }}
+  argTypes={{
+    children: { control: false },
+    className: { control: false },
+    id: { control: false },
+    isDisabled: {
+      table: { defaultValue: { summary: false } },
+    },
```
EdwinGuzman, line 39: Add something similar for `shouldWrapChildren`.

src/components/Tooltip/Tooltip.stories.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+- [W3C WAI ARIA Tooltip](https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/)
+- [MDN ARIA: tooltip role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role)
+- [Chakra UI Tooltip](https://chakra-ui.com/docs/components/overlay/tooltip)
+
+## Examples
+
+### Tooltips on form options
+
+When using the `Tooltip` component in forms containing `Checkbox`, `Radio`, `Toggle`.
```
EdwinGuzman: I think this needs one or two sentences explaining that the `shouldWrapChildren` prop should be added for form components and _why_.

EdwinGuzman (changes requested):
Good update! A few minor comments.

src/components/Tooltip/Tooltip.tsx
EdwinGuzman: Oh, maybe the warning should be clearer and say "a DS Icon, or DS Image". It's possible that a user may think an `Icon` or `Image` component from another library is okay to pass. What do you think?

src/components/Tooltip/Tooltip.stories.mdx
jackiequach: Yes this is due to Chakra's default behavior: https://chakra-ui.com/docs/components/tooltip/usage#disabling-tooltip

src/theme/components/componentWrapper.ts
bigfishdesign13: @EdwinGuzman Yes, this is causing a problem with any component that is supposed to expand to full width by default.  This would include:

- `Checkbox` and `CheckboxGroup`
- `Radio` and `RadioGroup`
- `SearchBar`
- `Select`
- `Slider`
- `TextInput`
- `Toggle`
- `VideoPlayer`

src/components/Tooltip/Tooltip.tsx
jackiequach: I agree that would be clearer to the user. Updated.

src/theme/components/componentWrapper.ts
EdwinGuzman: Okay, how about we do this:
1) @jackiequach removes this `width: "fit-content"` rule in this PR.
2) We create a new PR where we add a `width` prop to the `ComponentWrapper` so that the `Tooltip` can render correctly.

Now that I'm thinking about this, maybe you can remove this rule from this component _but_ pass it as a style prop. So in `Tooltip.tsx`:
```
    const newChildren = shouldWrapChildren ? (
      <ComponentWrapper width="fit-content">{children}</ComponentWrapper>
    ) : (
      children
    );
```

This should work and makes sense to do it _only_ for this use-case with the tooltip. This might be a better approach than creating a new PR as mentioned in steps 1 and 2. @jackiequach let us know if this works, otherwise, we can work on the two steps above.

jackiequach: Passing as a style prop works. Updated.

EdwinGuzman (approved):

bigfishdesign13 (approved):
Looks good!

PR #1119, opened 2022-08-22
DSD-411: Tagset component

EdwinGuzman:
Fixes JIRA ticket [DSD-411](https://jira.nypl.org/browse/DSD-411)

- Adds the `TagSet` component with the "explore" and "filter" variants.

## How has this been tested?

Locally in Storybook.

- N/A

### Checklist:

### Front End Review:


src/components/TagSet/TagSet.tsx
```diff
@@ -0,0 +1,101 @@
...
+  if (labelText.length > 17 && typeof labelText === "string") {
+    return <Tooltip content={labelText}>{children}</Tooltip>;
+  }
+  return <>{children}</>;
+};
+
+/**
+ * The `TagSet` component renders a group of individual tags which have two
+ * variants: "explore" and "filter". The "explore" tags are meant to be used for
+ * exploratory linkable elements. Whereas the "filter" tags are used to display
```
oliviawongnyc: Nit: if using "whereas," the two statements should be combined into one sentence, separated by a comma:

"...exploratory linkable elements, whereas the "filter" tags...."

If you prefer two sentences, you can drop the "whereas."

src/components/TagSet/TagSet.tsx
```diff
@@ -0,0 +1,101 @@
...
+  TagSetExploreProps,
+  TagSetExploreDataProps,
+} from "./TagSetExplore";
+import TagSetFilter, {
+  TagSetFilterDataProps,
+  TagSetFilterProps,
+} from "./TagSetFilter";
+
+export interface BaseTagSetProps {
+  /** Additional class for the componen.t */
```
oliviawongnyc: Period in wrong spot.

src/components/TagSet/TagSet.stories.mdx
```diff
@@ -0,0 +1,274 @@
...
+- Icons will not be rendered when `isDismissible` is true. In this case, the "X"
+  close icon will be rendered instead.
+
+<Canvas withToolbar>
+  <Story
+    name="TagSet Filter with Controls"
+    args={{
+      className: undefined,
+      id: "tagSet-id-filter",
+      isDismissible: false,
```
oliviawongnyc: I would personally make `isDismissable` true for this example. It will help someone new to TagSet wrap their mind around what it can do more quickly.

src/components/TagSet/TagSet.stories.mdx
```diff
@@ -0,0 +1,274 @@
...
+  {
+    iconName: "check",
+    label: <a href="#">The Beatles Sgt. Pepper's Lonely Hearts Club Band</a>,
+  },
+];
+```
+
+## Filter Component Props
+
+A "filter" `TagSet` variant is used to display keywords that a user selected to
```
oliviawongnyc: Nit: this first sentence can be simplified.

"The "filter" TagSet variant is used to display the keywords the user selected to filter content on the page."

src/components/TagSet/TagSet.stories.mdx
```diff
@@ -0,0 +1,274 @@
...
+  </Story>
+</Canvas>
+
+<ArgsTable story="TagSet Filter with Controls" />
+
+### Filter `tagSetData` Labels
+
+The `tagSetData` prop is an array of objects that contain `iconName` and `label`
+properties. The `iconName` is the name of the icon to display through the
+Reservoir `Icon` component and are optional. The `label` _MUST_ be a string for
```
oliviawongnyc: "_is_ optional?"

src/components/TagSet/TagSet.stories.mdx
```diff
@@ -0,0 +1,274 @@
...
+  </Story>
+</Canvas>
+
+<ArgsTable story="TagSet Explore with Controls" />
+
+### Explore `tagSetData` Labels
+
+The `tagSetData` prop is an array of objects that contain `iconName` and `label`
+properties. The `iconName` is the name of the icon to display through the
+Reservoir `Icon` component and are optional. The `label` _MUST_ be a React
```
oliviawongnyc: "_is_ optional?"

src/components/TagSet/TagSet.stories.mdx
```diff
@@ -0,0 +1,274 @@
...
+  {
+    iconName: "check",
+    label: "The Beatles Sgt. Pepper's Lonely Hearts Club Band",
+  },
+];
+```
+
+## Accessibility
+
+All proper tags in the "explore" and "filter" `TagSet` variants are accessible.
```
oliviawongnyc: "All properly implemented tags...?"

src/components/TagSet/TagSet.stories.mdx
```diff
@@ -0,0 +1,274 @@
...
+</Canvas>
+
+<ArgsTable story="TagSet Explore with Controls" />
+
+### Explore `tagSetData` Labels
+
+The `tagSetData` prop is an array of objects that contain `iconName` and `label`
+properties. The `iconName` is the name of the icon to display through the
+Reservoir `Icon` component and are optional. The `label` _MUST_ be a React
+element; typically, links are expected to be used for the "explore" variant.
```
oliviawongnyc, line 131: Is there any other React element that might be used? If not, is there a way to type the label so that it is not only a React element, but a React link element?

src/components/TagSet/TagSet.test.tsx
```diff
@@ -0,0 +1,335 @@
...
+        onClick={onClick}
+        tagSetData={filterTagSetData.withIcon}
+        type="filter"
+      />
+    );
+
+    expect(screen.getAllByTestId("ts-icon")).toHaveLength(7);
+  });
+
+  it("renders close icons when `isDismissible` is true", () => {
```
oliviawongnyc, line 216: Should there be a warning that the icons will not render if isDismissable is true?

src/components/TagSet/TagSet.tsx
```diff
@@ -0,0 +1,101 @@
...
+export function isFilterType(type: TagSetProps["type"]): type is "filter" {
+  return type === "filter";
+}
+
+/**
+ * This helper component wrapper renders a DS `Tooltip` component if the text is
+ * long or a React fragment. This assumes that the `label` prop is a rather
+ * simple single root JSX element, such as `<Link ...>....</Link>`.
+ */
+export const ToolTipWrapper = ({
```
oliviawongnyc: Cool feature!

oliviawongnyc (changes requested):
Looks great! Couple comments within. Also, I think the Clear Filters button should be mentioned in the Storybook doc.

src/components/TagSet/TagSet.tsx
```diff
@@ -0,0 +1,101 @@
...
+  label,
+  children,
+}: React.PropsWithChildren<{ label: any }>) => {
+  const labelText = typeof label === "string" ? label : label.props.children;
+  if (labelText.length > 17 && typeof labelText === "string") {
```
jackiequach: How is 17 determined? Should this be made a constant to avoid magic numbers?

src/components/TagSet/TagSet.tsx
jackiequach: Should be TooltipWrapper since tooltip is a single word

jackiequach (commented):
Looks good! I agree with Olivia's comment about including more info on the Clear Filters button in the docs

src/components/TagSet/TagSet.stories.mdx
EdwinGuzman: It could be a react router Link, nextjs link, a regular anchor, the DS link component, and more.... so it's difficult to type exactly what can be passed.

src/components/TagSet/TagSet.test.tsx
EdwinGuzman: Added.

src/components/TagSet/TagSet.tsx
EdwinGuzman: It was determined kinda randomly but it's more-or-less the character length before the tag reaches 200px width. It IS a magic number and I will fix it.

EdwinGuzman:
Thanks, updated.

@bigfishdesign13 based on the "style adjustment" commit, part of "g" gets cut off. I think other characters (I forgot the formal word for this letter characteristic) that also hang will be cut off slightly.
Screen Shot 2022-08-24 at 2 46 29 PM

bigfishdesign13:
> Thanks, updated.
> 
> @bigfishdesign13 based on the "style adjustment" commit, part of "g" gets cut off. I think other characters (I forgot the formal word for this letter characteristic) that also hang will be cut off slightly. Screen Shot 2022-08-24 at 2 46 29 PM

OK, I see what is causing that.  I will include a fix when I do the functional review today.

src/components/TagSet/TagSet.stories.mdx
```diff
@@ -0,0 +1,276 @@
...
+      type: "explore",
+    }}
+  >
+    {(args) => <TagSet {...args} />}
+  </Story>
+</Canvas>
+
+<ArgsTable story="TagSet Explore with Controls" />
+
+### Explore `tagSetData` Labels
```
bigfishdesign13: This title is a little misleading because the docs go on to talk about all of the properties in the `tagSetData` object.

src/components/TagSet/TagSet.stories.mdx
```diff
@@ -0,0 +1,276 @@
...
+    iconName: "check",
+    label: <a href="#">The Beatles Sgt. Pepper's Lonely Hearts Club Band</a>,
+  },
+];
+```
+
+## Filter Component Props
+
+A "filter" `TagSet` variant is used to display keywords the user selected to
+filter content on a page. For example, in a search results page, the filters a
```
bigfishdesign13: …the filter criteria a user selects...

src/components/TagSet/TagSet.stories.mdx
```diff
@@ -0,0 +1,276 @@
...
+      type: "filter",
+    }}
+  >
+    {(args) => <TagSet {...args} />}
+  </Story>
+</Canvas>
+
+<ArgsTable story="TagSet Filter with Controls" />
+
+### Filter `tagSetData` Labels
```
bigfishdesign13: Ditto my comment from the other heading like this.

src/components/TagSet/TagSet.stories.mdx
```diff
@@ -0,0 +1,276 @@
...
+          { label: <a href="#">Orange</a> },
+          { label: <a href="#">Yellow</a> },
+          { label: <a href="#">Green</a> },
+          { label: <a href="#">Blue</a> },
+          { label: <a href="#">Indigo</a> },
+          { label: <a href="#">Violet</a> },
+        ]}
+        type="explore"
+      />
+      <TagSet
```
bigfishdesign13, line 286: What do you think about updating these tags to show a list of DS components and then using real links to the component pages in Storybook?

src/theme/components/tagSet.ts
```diff
@@ -0,0 +1,99 @@
...
+    height: { base: "32px", md: "22px" },
+    minHeight: "22px",
+    maxWidth: "200px",
+    py: "xxxs",
+    paddingLeft: "s",
+    paddingRight: isDismissible ? { base: "8px", md: "6px" } : "s",
+    fontSize: "text.tag",
+    "> span": {
+      display: "inline-block",
+      lineHeight: "1",
```
bigfishdesign13: Remove this attribute.  This is what is causing the character descenders to get cut off.  And I see now that I forgot to add this attribute to the `TagSetExplore` object when I made the style updates the other day, so you will not need to remove it from that object.

src/components/TagSet/TagSetFilter.tsx
```diff
@@ -0,0 +1,111 @@
...
+                  color="ui.gray.x-dark"
+                  width="12px"
+                />
+              ) : null}
+            </Button>
+          </TooltipWrapper>
+        );
+      })}
+
+      {filters.length > 0 && isDismissible ? (
```
bigfishdesign13: The `Clear Filters` button should be visible when there is more than one tag.  If there is only one tag, then the button will not show.

src/components/TagSet/TagSet.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const maxCharLengthToShow = 17;
+  const labelText = typeof label === "string" ? label : label.props.children;
+  if (labelText.length > maxCharLengthToShow && typeof labelText === "string") {
+    return <Tooltip content={labelText}>{children}</Tooltip>;
+  }
+  return <>{children}</>;
+};
+
+/**
+ * The `TagSet` component renders a group of individual tags which have two
```
bigfishdesign13, line 47: Add another paragraph with details about a tag’s max width:

"The width of a single tag will never be greater than 200px.  If necessary, a tag’s label text will be truncated to keep a tag’s width at or below 200px.  The full label text will be revealed with a DS `Tooltip` component."

src/components/TagSet/TagSet.stories.mdx
```diff
@@ -0,0 +1,276 @@
...
+      onClick: { table: { disable: true } },
+      tagSetData: { control: false },
+    }}
+    args={{
+      className: undefined,
+      id: "tagSet-id-explore",
+      isDismissible: undefined,
+      onClick: undefined,
+      tagSetData: [
+        { iconName: "arrow", label: <a href="#">Abbey Road</a> },
```
bigfishdesign13: While these are only examples and the icons are somewhat irrelevant, let’s replace the `arrow` icon with `utilityAccountFilled`.  The down arrow is too loaded with meaning.  Replace here and in the code examples.

src/theme/components/tagSet.ts
```diff
@@ -0,0 +1,99 @@
...
+      color: "ui.black",
+      height: { base: "32px", md: "22px" },
+      fontSize: "text.tag",
+      minHeight: "22px",
+    },
+  }),
+};
+
+const TagSetExplore = {
+  baseStyle: {
```
bigfishdesign13, line 41: Add the following to the `baseStyle` object to handle the hover transition.
```
transition: "0.3s ease”,
```

src/theme/components/tagSet.ts
```diff
@@ -0,0 +1,99 @@
...
+    "> span": {
+      display: "inline-block",
+      overflow: "hidden",
+      textOverflow: "ellipsis",
+      whiteSpace: "nowrap",
+    },
+    svg: {
+      fill: "ui.link.primary",
+    },
+    _hover: {
```
bigfishdesign13, line 75: Add the following to the `_hover` object to handle the hover color of the ellipsis.
```
"> span": {
  color: "ui.white",
},
```

bigfishdesign13 (changes requested):
Looks and works great.  It just needs some fine tuning.

EdwinGuzman:
Updated.

src/components/TagSet/TagSet.tsx
bigfishdesign13: This seems significant enough to have a separate paragraph.  It has been added to the docs, but it is part of the first paragraph right now.

src/components/TagSet/TagSet.tsx
bigfishdesign13: @EdwinGuzman Sorry, I missed this the first time around.  Did you consider using a method similar to this?  https://thewebdev.info/2021/04/17/how-to-check-for-text-overflow-ellipsis-in-an-html-element/

EdwinGuzman:
> Sorry, I missed this the first time around. Did you consider using a method similar to this? https://thewebdev.info/2021/04/17/how-to-check-for-text-overflow-ellipsis-in-an-html-element/

I tried it and while it works, in the long run it won't work for our use case. I made a separate test PR #1123 so you can see the difference.

There is a `useIsOverflowElement` custom hook to check if an element's `scrollWidth` is larger than `offsetWidth`. This works well and there are no problem testing it either. The problem occurs with how we will use the `TagSet` component.

Case 1: We have a define set of tags.
- If there is a list that never changes, then nothing is wrong. We can use the element's ref since it will always be there.

Case 2: The set of tags increases or decreases.
- We expect this case when filters are selected or when a tag is dismissed. Because the list of data changes, the hook functions either get called too many times or not enough as _they_ expect it to be called. This came up when testing a case where a tag was added to the list and the component was re-rendered.  Then this error came up:
Screen Shot 2022-08-26 at 4 22 41 PM

- I tested this separately in Storybook and the same issue came up but because it's an error, the entire page broke.

Even without the hook, this same problem occurs and I don't think there's another way to not use `useRef` to get an element's `offsetWidth` or `scrollWidth`. I spent too much time on this and I don't like the random "17 character logic check, but I'm not sure if there's another easy solution. I can try something else next week.

src/components/TagSet/TagSet.tsx
bigfishdesign13: @EdwinGuzman Let’s stay with the character count solution, but maybe bump it up to “20" so it doesn’t seem so random.  Three more chars won’t increase the width so much that is feels too big.

src/components/TagSet/TagSet.tsx
```diff
@@ -0,0 +1,105 @@
...
+ * variants: "explore" and "filter". The "explore" tags are meant to be used for
+ * exploratory linkable elements, whereas the "filter" tags are used to display
+ * the filter values that were selected through another UI. Only "filter" tags
+ * can be dismissible. The width of a single tag will never be greater than
```
bigfishdesign13: Create a new paragraph starting at `The width of a single tag…`.

bigfishdesign13 (approved):
Wonderful.  Just break the component description into two paragraphs before merging.

EdwinGuzman:
@oliviawongnyc can you please re-review this?

oliviawongnyc (approved):
Looks good to me!

PR #1129, opened 2022-09-01
DSD-1133: Passing defaultValue down to hidden TextInput

EdwinGuzman:
Fixes JIRA ticket [DSD-1133](https://jira.nypl.org/browse/DSD-1133)

- Passes down the `defaultValue` prop to the `TextInput` component for the "hidden" type. The `value` prop is already passed down but we want the `defaultValue` prop because it doesn't change and we don't need to control this value. It is pretty much set once for hidden input elements.
- This issue was found while working with hidden inputs in the Library Card App.

## How has this been tested?

Locally and through npm link with the Library Card App.

- N/A

### Checklist:

### Front End Review:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 ## Prerelease
 
+### Fixes
+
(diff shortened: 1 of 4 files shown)


bigfishdesign13 (approved):
Yes.

PR #1134, opened 2022-09-07
DSD-1107: dark mode support for Form Elements component

bigfishdesign13:
Fixes JIRA ticket [DSD-1107](https://jira.nypl.org/browse/DSD-1107)

- Adds `dark` color mode support for the `Button`, `Checkbox`, `DatePicker`, `Fieldset`, `Label`, `Radio`, `SearchBar`, `Select`, `Slider`, `TextInput` and `Toggle` components.
- Most of the updates for `DatePicker` were executed in the `_DatePicker.scss` file and it became tricky to get all the necessary styles in place.

## How has this been tested?

- local Storybook

- The dark mode updates are implemented using the NYPL WCAG compliant `dark` color palette.

### Checklist:

### Front End Review:


src/theme/components/checkbox.ts
```diff
@@ -68,6 +68,58 @@ const baseStyleControl = {
     borderColor: "ui.error.primary",
     color: "ui.error.primary",
   },
+
+  _dark: {
+    borderColor: "dar.ui.border.default",
```
oliviawongnyc: typo here: "dar**k**.ui.border.default"

src/theme/components/slider.ts
```diff
@@ -24,37 +25,43 @@ const CustomSlider = {
...
 
     return {
       container: {
         display: "flex",
         alignItems: "center",
       },
-      leftValue: { ...staticValues, color: isDisabled ? "ui.gray.dark" : null },
+      leftValue: {
+        ...staticValues,
+        color: isDisabled ? "ui.disabled.primary" : null,
```
oliviawongnyc, line 49: Could you just use baseColor here and remove the _dark selector entirely since it switches automatically based on your logic above?

```
color: baseColor
```

src/theme/components/slider.ts
```diff
@@ -24,37 +25,43 @@ const CustomSlider = {
...
+      },
       rightValue: {
         ...staticValues,
         marginStart: "s",
         // If the text input boxes are shown, then there already is
         // a margin, so we can set this static value to "0".
         marginEnd: showBoxes ? "0" : "s",
-        color: isDisabled ? "ui.gray.dark" : null,
+        color: isDisabled ? "ui.disabled.primary" : null,
+        _dark: { color: isDisabled ? "dark.ui.disabled.primary" : null },
```
oliviawongnyc, line 59: Same comment as above.

src/theme/components/toggle.ts
```diff
@@ -25,7 +25,7 @@ const Switch = {
   baseStyle: ({ size }: { size: string }) => {
     return {
       alignItems: "start",
-      opacity: 0.4,
+      // opacity: 0.4,
```
oliviawongnyc: Still needed or can be gotten rid of?

src/theme/components/toggle.ts
```diff
@@ -56,6 +56,33 @@ const Switch = {
...
+            borderColor: "dark.ui.error.primary",
+            "> span": {
+              bg: "dark.ui.error.primary",
+            },
+          },
+          _disabled: {
+            bg: "dark.ui.disabled.secondary",
+            borderColor: "dark.ui.disabled.secondary",
+            opacity: 1,
+            // These are styles for the thumb. _isDisabled was not working within the `thumb` part.
```
oliviawongnyc, line 76: Strange! Thanks for adding the comment.

src/theme/components/searchBar.ts
```diff
@@ -12,9 +12,13 @@ const SearchBar = {
     },
     select: {
       flexShrink: "0",
-      marginBottom: "0",
+      marginBottom: { base: "-1px", md: "0" },
```
oliviawongnyc, line 15: This has to do with responsiveness?

src/theme/components/radio.ts
```diff
@@ -66,6 +66,48 @@ const baseStyleControl = {
   //   borderColor: "ui.gray.medium",
   //   bg: "ui.white",
   // },
+  _dark: {
+    borderColor: "dar.ui.border.default",
```
oliviawongnyc: Missing the "k" in dark.

src/theme/components/radio.ts
```diff
@@ -66,6 +66,48 @@ const baseStyleControl = {
   //   borderColor: "ui.gray.medium",
   //   bg: "ui.white",
   // },
+  _dark: {
```
oliviawongnyc, line 69: You could do something similar here that you did in the other component, where you created a conditional that checked whether isDark is true and appended a prefix of "dark" onto the colors. I see you have ui.focus/dark.ui.focus, ui.error.primary/dark.ui.error.primary, etc.

src/theme/components/checkbox.ts
```diff
@@ -68,6 +68,58 @@ const baseStyleControl = {
     borderColor: "ui.error.primary",
     color: "ui.error.primary",
   },
+
+  _dark: {
```
oliviawongnyc, line 72: Same comment here as I made on radio's theme file: I think you could add a conditional that checked whether isDark is true and appended a prefix of "dark" onto the colors. Let me know your thoughts.

src/theme/components/button.ts
```diff
@@ -106,6 +156,9 @@ const callout = {
   _active: {
     bg: "brand.secondary",
   },
+  _dark: {
+    bg: "dark.brand.primary",
```
oliviawongnyc, line 160: I feel like I don't see this callout button changing in storybook on dark mode, but maybe just me?

oliviawongnyc (commented):
Whew! This was a ton of work, Marty! Awesome job. A couple small comments/questions here and there. May want Edwin's eyes on this too as there is a lot here.

src/theme/components/checkbox.ts
bigfishdesign13: Good catch.

src/theme/components/slider.ts
bigfishdesign13: No, because the expected styles here do not follow the same pattern as `baseColor`.

src/theme/components/slider.ts
bigfishdesign13: Smae as above.

src/theme/components/toggle.ts
bigfishdesign13: Stranded comments…my bane.  Good catch.

src/theme/components/searchBar.ts
bigfishdesign13: Yes.  For mobile, the elements are stacked and I needed the bottom margin only applied in that state.  Outside of the mobile state, the elements are inline.

src/theme/components/radio.ts
bigfishdesign13: The conditional statement that you are referring to from `slider.ts` was already in there when I started adding the dark mode styles, so continued to use it.  That said, on the whole, I would prefer to keep the dark mode styles separated in the explicit `_dark` object.  I think maintenance will be easier in the future.

src/theme/components/checkbox.ts
bigfishdesign13: Ditto.

src/theme/components/button.ts
bigfishdesign13: Correct.  There is no difference.  We left the callout button the same hot color in both `light` and `dark` modes.

oliviawongnyc (approved):

EdwinGuzman (changes requested):
- The hover state for the `Checkbox` and `Radio` needs to be added (based on the Figma designs).
- For the `DatePicker`, it's (1) very difficult to see what dates are disabled and (2) what the date range selected is:
Screen Shot 2022-09-19 at 10 11 08 AM
- Some accessibility issues:
Screen Shot 2022-09-19 at 10 14 44 AM
Screen Shot 2022-09-19 at 10 14 56 AM

src/components/Slider/Slider.tsx
```diff
@@ -116,6 +117,7 @@ export const Slider = chakra(
         : defaultValue;
       const [currentValue, setCurrentValue] =
         React.useState<typeof defaultValue>(finalDevaultValue);
+      const isDark = useColorModeValue(false, true);
```
EdwinGuzman: A couple of things. `useColorModeValue` works here but it's not used as it's meant to be used. Two color values should be passed rather than true/false values. Let's not make it confusing and instead use the `useColorMode` hook and get the `colorMode` value (which can be "light" or "dark") instead.

Secondly, why is this being done here and the `isDark` boolean value passed to `useMultiStyleConfig`? Is it difficult to use the `_dark` key?

bigfishdesign13: Based on the the Chakra docs, I think `useColorModeValue` can be used to set the value of any prop.  But I am not too worried about this.  I changed to use `colorMode`.

> Secondly, why is this being done here and the isDark boolean value passed to useMultiStyleConfig? Is it difficult to use the _dark key?

Yes.  There were style values being set as variables, so I continued to work with that rather than trying to rewrite a lot of the styles so I could use `_dark`.

bigfishdesign13:
@EdwinGuzman I corrected the `a11y` issues.  For `DatePicker`, the coloring for dark mode should align with how light mode is handled.  Meaning, if some of the color contrast seems too low in dark mode, the same low contrast scenarios also exist in light mode.  That said, I did not go out of my way to make any specific color changes other than adding dark mode support.

EdwinGuzman (approved):

PR #1138, opened 2022-09-16
DSD-1139: Secondary Button styles

bigfishdesign13:
Fixes JIRA ticket [DSD-1139](https://jira.nypl.org/browse/DSD-1139)

- Updates the colors for the `secondary` type of the `Button` component.

## How has this been tested?

- local Storybook

- The colors changes are using our WCAG compliant color palette.

### Checklist:

### Front End Review:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 ### Updates
 
-- Updates the colors for the `secondary` type of the `Button` component.
+- Updates the colors for the `secondary` and `iconOnly` variants of the `Button` component.
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 - Adds the `text` variant in the `Button` component.
 
+### Updates
+
+- Updates the colors for the `secondary` type of the `Button` component.
+
 ### Fixes
 
 - Fixes a bug where the `defaultValue` for a `TextInput` component was not being passed correctly to the Chakra input element.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 - Adds the `text` variant in the `Button` component.
 
+### Updates
+
+- Updates the colors for the `secondary` type of the `Button` component.
```
EdwinGuzman: And `iconOnly` type?

src/theme/components/button.ts
```diff
@@ -100,11 +103,12 @@ const pill = {
 const iconOnly = {
   bg: "ui.white",
   border: "1px solid",
-  borderColor: "ui.gray.light-cool",
-  color: "inherit",
+  borderColor: "ui.link.primary",
+  color: "ui.link.primary",
```
oliviawongnyc, line 107: Does inherit no longer work or is this an update for clarity?

bigfishdesign13: The color for this variant is changing altogether, so `inherit` will not work in this case.

CHANGELOG.md
bigfishdesign13: This is interesting.  We have `iconOnly` as a unique variant on the same level as `primary`, `secondary`, etc.  However, `iconOnly` should be a state of a button type.  For example, we should be able to have an icon only `callout` button.  But right now, `iconOnly` is a variant that matches the `secondary` styles and that is the only type of icon onlu button.  That is why I wrote this as I did.

For now, I will update the CHANGELOG entry and I will also create a new ticket to refactor how `iconOnly` is handled.

EdwinGuzman (approved):

PR #1141, opened 2022-09-19
DSD-1137: Updating spacing between icon and text in the Notification component

EdwinGuzman:
Fixes JIRA ticket [DSD-1137](https://jira.nypl.org/browse/DSD-1137)

- Updates the spacing between the icon and text in the `Notification` component by reducing it from 16px (s) to 8px (xs).

## How has this been tested?

Locally in Storybook.

- N/A

### Checklist:

### Front End Review:

### DIFF
--- a/src/theme/components/notification.ts
+++ b/src/theme/components/notification.ts
@@ -66,7 +66,7 @@ const Notification = {
       },
       icon: {
         flexShrink: "0",
-        marginEnd: "s",
+        marginEnd: "xs",
       },
     };
   },
@@ -85,7 +85,7 @@ const NotificationContent = {
       color: notificationType === "warning" ? "brand.primary" : "currentColor",
       marginTop: icon ? "xxxs" : "0",
       paddingStart: alignText
-        ? "calc(var(--nypl-space-m) + var(--nypl-space-s))"
+        ? "calc(var(--nypl-space-s) + var(--nypl-space-s))"
         : null,
       width: "100%",
       // Links should always be black and underlined.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 - Fixes a bug where the `defaultValue` for a `TextInput` component was not being passed correctly to the Chakra input element.
 
+### Updates
+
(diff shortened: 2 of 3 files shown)


src/theme/components/notification.ts
```diff
@@ -66,7 +66,7 @@ const Notification = {
       },
       icon: {
         flexShrink: "0",
-        marginEnd: "s",
+        marginEnd: "xs",
```
Toxiapo, line 69: I am curious what's the difference between using paddingStart and marginEnd over paddingLeft, marginRight?

EdwinGuzman: `paddingStart`, `paddingEnd`, `marginStart`, and `marginEnd` are all specific to Chakra, whereas `paddingLeft` (padding-left), `paddingRight` (padding-right), `marginLeft` (margin-left), and `marginRight` (margin-right) are all CSS properties (the ones with hyphens are CSS syntax). They are closely related, so `paddingStart` is similar to `paddingLeft` and `marginEnd` is similar to `marginRight`. The reason we are using the "Start" and "End" suffixes is that they help with bidirectionality when the browser's language is set to a right-to-left language.

In English, the "Start" is the left side of any element on the page. In a language such as Arabic, where it is read right-to-left, the "Start" is now considered to be the right side of any element on the page. We want to support right-to-left languages so we use "Start" and "End" instead of "left" and "right". Chakra takes care of adding the value we want in the appropriate place.

Toxiapo: Ahh that's interesting, thank you @EdwinGuzman! So in essence, when dealing with text-related elements, we should always use `start` and `end`, but with spacing between layouts should be free to use any?

EdwinGuzman: No, use `Start` and `End` for everything.

oliviawongnyc (commented):
I'm still not sure how this will fit in with the PR I made for the Notification component. I changed the same values in my PR that you did in this one.

EdwinGuzman:
This PR will most likely go out soon whereas the dark mode update will go out later (since it'll probably be 2.0.0). So this feels like a quick design fix and the dark mode update is more stable.

Toxiapo (approved):

bigfishdesign13 (approved):
Yes.

PR #1143, opened 2022-09-19
DSD-578: AlphabetFilter Component

jackiequach:
Fixes JIRA ticket [DSD-578](https://jira.nypl.org/browse/DSD-578)

- Adds AlphabetFilter component

## How has this been tested?

Unit tests and in StoryBook

-

### Checklist:

### Front End Review:


src/utils/componentCategories.ts
```diff
@@ -116,6 +116,7 @@ const categories = {
   navigation: {
     title: "Components/Navigation",
     components: [
+      "AlphabetFilter",
```
EdwinGuzman, line 124: This seems reasonable but it could also go under the "Form Elements" heading. @bigfishdesign13 is the category that this component should go under listed anywhere? @Toxiapo and I couldn't find it in the spec doc.

src/theme/components/alphabetFilter.ts
```diff
@@ -0,0 +1,27 @@
...
+      _disabled: {
+        color: "ui.gray.medium",
+      },
+    },
+    mobileStack: {
+      alignItems: "left",
+    },
+    filterRow: {
+      display: "flex",
+      marginTop: "0!important",
```
EdwinGuzman: There should be a space between the "0" and the "!important". We try not to use "!important", why is this necessary here?

src/components/AlphabetFilter/AlphabetFilter.tsx
```diff
@@ -0,0 +1,197 @@
...
+    const getButtonElement = (item) => {
+      const isSelectedLetter =
+        selectedLetter === item.value && item.value !== "showAll";
+      const isShowAll = item.value === "showAll";
+      const isButtonDisabled =
+        isDisabled ||
+        (activeLetters && !activeLetters.includes(item.value) && !isShowAll);
+
+      const buttonStyles = isSelectedLetter
+        ? { ...styles, border: "1px solid gray" }
```
EdwinGuzman: By doing this, `...styles` is returning all of the styles including the "mobileStack and "filterRow" keys. Did you mean to do something like `...styles.button` to only get the specific button styles? If not, that might make this clearer.

src/components/AlphabetFilter/AlphabetFilter.tsx
```diff
@@ -0,0 +1,197 @@
...
+  /** ID that other components can cross reference for accessibility purposes. */
+  id?: string;
+  /** Adds the `disabled` prop to the AlphabetFilter when true. */
+  isDisabled?: boolean;
+  /** The callback function called when a letter button or the Show All button is clicked. */
+  onClick: (filterValue: string) => void;
+}
+
+export const AlphabetFilter = chakra(
+  forwardRef<HTMLDivElement, AlphabetFilterProps>((props, ref?) => {
```
EdwinGuzman, line 28: This is fine as it is setup, but the `ref` won't always be useful when we try to get the value from it. This is because the `ref` is being passed to the wrapper div, but the wrapper div doesn't capture any of the data that gets updated. There are also too many buttons and those can't use a single ref. This is not necessarily a problem, but if a developer expects to get the data using a `ref`, they won't get anything from the `ref.current.value` attribute.

I'd leave this as is until it does become a problem and we can figure out if we need to return data through the ref, somehow.

src/components/AlphabetFilter/AlphabetFilter.tsx
```diff
@@ -0,0 +1,197 @@
...
+      );
+    };
+
+    const getFilterLetters = () => {
+      if (isLargerThanMobile) {
+        return filterButtons.map((item) => {
+          return getButtonElement(item);
+        });
+      }
+      // split letters into rows of 10 + Show All
```
EdwinGuzman: @bigfishdesign13  The length of "10" is not defined in the spec doc but we (@Toxiapo and I) are assuming that this is because the Figma design is broken up so that there are 10 letters on each row. Is this intentional or should it be dynamic based on the width of the viewport, e.g. should a slightly larger mobile phone display 12 letters rather than 10 per row (should it wrap). Can you make this clearer in the spec doc as well?

src/components/AlphabetFilter/AlphabetFilter.tsx
```diff
@@ -0,0 +1,197 @@
...
+      const mobileRows = filterButtons.reduce((rows, key, index) => {
```
EdwinGuzman: This makes sense when reading through the code but can you add some comments to make it quicker to read through?

EdwinGuzman: Also making it clear that the returned value is a 2d array.

src/components/AlphabetFilter/AlphabetFilter.tsx
```diff
@@ -0,0 +1,197 @@
...
+            (index % 10 === 0
+              ? rows.push([key])
+              : rows[rows.length - 1].push(key)) && rows
+          );
+        } else {
+          return rows.push([key]) && rows;
+        }
+      }, []);
+      return (
+        <VStack sx={styles.mobileStack}>
```
EdwinGuzman: Do you think that maybe a `ul` list may be better than just buttons for accessibility? Perhaps a list of lists.

src/components/AlphabetFilter/AlphabetFilter.tsx
```diff
@@ -0,0 +1,197 @@
...
+      { text: "Y", value: "y" },
+      { text: "Z", value: "z" },
+      { text: "Show All", value: "showAll" },
+    ];
+
+    const refCurrentLetter = useRef(currentLetter);
+    const [selectedLetter, setSelectedLetter] = useState<string>(currentLetter);
+
+    // If the parent passes down a new currentLetter, and an onClick
+    // function exists, then set the internal state – selectedLetter –
```
EdwinGuzman: `onClick` is a required prop so when might it not be defined?

src/components/AlphabetFilter/AlphabetFilter.tsx
```diff
@@ -0,0 +1,197 @@
...
+          }
+        : styles;
+
+      return (
+        <Button
+          aria-current={isSelectedLetter ? "page" : null}
+          aria-label={
+            item.text === "Show All" ? item.text : "Page " + item.text
+          }
+          buttonType="link"
```
EdwinGuzman: Eventually this should be updated to the new "text" type.

src/components/AlphabetFilter/AlphabetFilter.tsx
```diff
@@ -0,0 +1,197 @@
...
+            );
+          })}
+        </VStack>
+      );
+    };
+
+    return (
+      <Box
+        as="nav"
+        role="navigation"
```
EdwinGuzman: You don't necessarily need this since it renders as a `nav` element, but it's fine to keep.

EdwinGuzman: It doesn't add it to the `nav` element when it renders but it is implicit. Just keep it here.

src/components/AlphabetFilter/AlphabetFilter.tsx
EdwinGuzman: Also, this does depend on @bigfishdesign13 's answer to whether it should be a row of 10 max characters per row. Otherwise, it could theoretically be a single list of buttons that wrap around.

src/components/AlphabetFilter/AlphabetFilter.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+    // D button
+    userEvent.click(buttons[5]);
+    expect(currentLetter).toEqual("e");
+
+    // Show All button
+    userEvent.click(buttons[buttons.length - 1]);
+    expect(currentLetter).toEqual("showAll");
+  });
+});
```
EdwinGuzman, line 151: Can you add a new block for snapshot tests with a similar structure such as the tests for the `Button`: https://github.com/NYPL/nypl-design-system/blob/development/src/components/Button/Button.test.tsx#L106

EdwinGuzman (commented):
The comments are from an initial pair review with @Toxiapo. I still need to review the Storybook file and the test file which I'll do soon.

The component looks good and works well but there are questions we have that I think you and @bigfishdesign13 can help answer. Not hitting "request changes" until I finish reviewing the other two files.

src/components/AlphabetFilter/AlphabetFilter.stories.mdx
```diff
@@ -0,0 +1,141 @@
...
+- [Overview](#overview)
+- [Component Props](#component-props)
+- [Accessibility](#accessibility)
+- [Examples](#examples)
+
+## Overview
+
+<Description of={AlphabetFilter} />
+
+The `AlphabetFilter` component provides a method for alphabetical filtering on a list of items.
```
oliviawongnyc: Since the `AlphabetFilter` doesn't actually do the filtering, what about something like: 

"The `AlphabetFilter` renders a list of letter buttons (and one pound sign button) that can be used to select one or more alphabetical filters. Displaying updated results based on the user's selection is the responsibility of the consuming app."

src/components/AlphabetFilter/AlphabetFilter.stories.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+### Consuming App Responsibilities
+
+When the `onClick` handler is executed, it will be up to the consuming app to execute filtering
+updates in real-time on the client-side or to perform a full page refresh.
+
+When the `Show All` button is clicked, all filtering should be removed from the list of items that
```
oliviawongnyc: Instead of saying, "When the `Show All` button is clicked....", how about "If the filterValue passed to the onClick is `showAll` (meaning the `Show All` button was clicked), all filtering should be...".

src/components/AlphabetFilter/AlphabetFilter.stories.mdx
```diff
@@ -0,0 +1,141 @@
...
+    {(args) => <AlphabetFilter {...args} />}
+  </Story>
+</Canvas>
+
+<ArgsTable story="AlphabetFilter with Controls" />
+
+## Accessibility
+
+The `AlphabetFilter` component renders as an HTML `<nav>` element with a WAI-ARIA role
+of `role="navigation"`. Each letter button has `aria-label="Page {#}"` applied, where
```
oliviawongnyc: Nit: "Each letter button has an attribute of `aria-label="Page {#}"`, where {#} is equal to the button's letter value."

src/components/AlphabetFilter/AlphabetFilter.stories.mdx
```diff
@@ -0,0 +1,141 @@
...
+{#} is equal to a button's letter value. The Show All button has `aria-label="Show All"`
```
oliviawongnyc: Put `Show All` in back ticks for consistency with the above text.

src/components/AlphabetFilter/AlphabetFilter.stories.mdx
```diff
@@ -0,0 +1,141 @@
...
+applied. When a letter button is set to a current state, `aria-current="page"` is applied
+to the button and removed from all other letter buttons.
+
+The `AlphabetFilter` component is accessible via keyboard. The color contrast between
+foreground color and background color is 4.5:1. If text size is 200%, the
+button scales with text so there is no overlap.
```
EdwinGuzman, line 101: Can you add this link as:
```
Resources:
- [MDN aria-current](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current)
```

src/components/AlphabetFilter/AlphabetFilter.stories.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+If filter results are refreshed on the client-side without a full page refresh, the
+updated filter results must be the child of an element that has the `role=”alert”` attribute
+applied or as a child of a Live Region container with the `aria-live="polite"` attribute applied.
+
+## Examples
+
+### AlphabetFilter with activeLetters set
+
+Pass an array of letters to narrow filter options.
```
EdwinGuzman: Add "... filter options to the `activeLetters` prop.".

src/components/AlphabetFilter/AlphabetFilter.stories.mdx
```diff
@@ -0,0 +1,141 @@
...
+        "r",
+        "s",
+      ]}
+      onClick={undefined}
+    />
+  </DSProvider>
+</Canvas>
+
+### Set letter selection programmatically
+
```
EdwinGuzman, line 149: Add more context such as, "Use the `currentLetter` prop to pass a single letter value to set the selected letter page when the component first renders, or to programmatically set the current letter."

This example shows how the initial selected letter works but maybe we need another example where the `currentLetter` is updated dynamically by the user. Check out the `Select` examples at the end https://nypl.github.io/nypl-design-system/reservoir/v1/?path=/docs/components-form-elements-select--select-with-controls where there are controlled and uncontrolled examples. This will also help show how the `onClick` function should be used by the consuming app.

Let me know if this makes sense. It is more involved on our side, but it'll give a better example to users on one particular implementation.

src/components/AlphabetFilter/AlphabetFilter.stories.mdx
EdwinGuzman: Also add more context to say that the letters not passed will _always_ be disabled.

src/components/AlphabetFilter/AlphabetFilter.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+        currentLetter={currentLetter}
+        onClick={onClick}
+      />
+    );
+
+    const buttons = screen.getAllByRole("button");
+
+    // D button
+    userEvent.click(buttons[5]);
+    expect(currentLetter).toEqual("e");
```
EdwinGuzman: The comment says "D" but the value is "e".

src/components/AlphabetFilter/AlphabetFilter.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+        id="alphabet-filter-id"
+        headingText="Heading"
+        descriptionText="This is a description."
+      />
+    );
+    expect(await axe(container)).toHaveNoViolations();
+  });
+});
+
+describe("AlphabetFilter", () => {
```
EdwinGuzman, line 29: This is missing a test for the `activeLetters` prop.

src/components/AlphabetFilter/AlphabetFilter.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+  it("should pass a ref to the nav wrapper element", () => {
+    const ref = React.createRef<HTMLDivElement>();
+    const { container } = render(
+      <AlphabetFilter ref={ref} id="alphabet-filter-id" onClick={undefined} />
+    );
+
+    expect(container.querySelector("nav")).toBe(ref.current);
+  });
+
+  it("should update active letter when letter or show all button is clicked", () => {
```
EdwinGuzman, line 129: This is a good test and this is what I'm thinking would be ideal to also show in the Storybook documentation as an example. Maybe in the Storybook page, you can simply log the returned letter value to the browser's console. 

This is done in other Storybook component pages. For an example, check the last example in the `Slider` page: https://nypl.github.io/nypl-design-system/reservoir/v1/?path=/docs/components-form-elements-slider--range-slider-with-controls where we also explicitly tell the user to "Open up the browser's developer console to see the values of the Slider after updating it in the isRangeSlider state."

EdwinGuzman (changes requested):
Added a few more comments. They're mostly for adding more documentation and a few more tests but the component itself is looking great (pending on Marty's answers to the mobile question).

src/components/AlphabetFilter/AlphabetFilter.stories.mdx
```diff
@@ -0,0 +1,141 @@
...
+</Canvas>
+
+<ArgsTable story="AlphabetFilter with Controls" />
+
+## Accessibility
+
+The `AlphabetFilter` component renders as an HTML `<nav>` element with a WAI-ARIA role
+of `role="navigation"`. Each letter button has `aria-label="Page {#}"` applied, where
+{#} is equal to a button's letter value. The Show All button has `aria-label="Show All"`
+applied. When a letter button is set to a current state, `aria-current="page"` is applied
```
oliviawongnyc: What does, "set to a current state" mean? Does this mean when a letter is clicked or if a letter is programmatically selected? Or both?

src/components/AlphabetFilter/AlphabetFilter.stories.mdx
oliviawongnyc: Why would you want letters disabled? If you don't have any results that match those filters?

src/components/AlphabetFilter/AlphabetFilter.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+        descriptionText="This is a description."
+      />
+    );
+    expect(await axe(container)).toHaveNoViolations();
+  });
+});
+
+describe("AlphabetFilter", () => {
+  it("should render a nav with 28 buttons", async () => {
+    render(<AlphabetFilter onClick={undefined} id="alphabet-filter-id" />);
```
oliviawongnyc: Since `onClick` is required, I am surprised to know that setting it to `undefined` works in these tests. I would have mocked it at the top with, `const onClick = jest.fn()`, but I am not sure if that is preferable.

src/utils/componentCategories.ts
bigfishdesign13: In the Visual Design Library (the Figma Main file), `AlphabetFilter` is listed under `Form Elements`, but because this component is similar to `Pagination`, we talked about moving it to `Navigation` (which I prefer).  That said, leave this as is.  I will update the Visual Design Library and I will also start to include a component's `Category` in the spec doc.

src/components/AlphabetFilter/AlphabetFilter.tsx
bigfishdesign13: To bring this component into better alignment with how we handle buttons for mobile, I updated the component in the Visual Design Library (the Figma Main file).  For the mobile view, the buttons now used fixed dimensions of `44 x 44`, which actually reduces the number of buttons that would fit in the design.  With that, let’s make it so the buttons are in one flex row and allow them to wrap onto new lines as needed based on the width of the device.

I will update the spec doc to reflect this.

bigfishdesign13: The spec doc has been updated…
https://docs.google.com/document/d/13IOduQYuHW4t_1EV05dJVfzt_d-FTNkyczZmHCgJEO4/edit#heading=h.kuqwpe6igp9d

src/components/AlphabetFilter/AlphabetFilter.tsx
bigfishdesign13: This wrapper and the interior text elements could be handled using the `ComponentWrapper` component.

src/components/AlphabetFilter/AlphabetFilter.tsx
```diff
@@ -0,0 +1,197 @@
...
+        id={id}
+        className={className}
+        ref={ref}
+        __css={styles}
+        {...rest}
+      >
+        {headingText && <Heading size="tertiary">{headingText}</Heading>}
```
bigfishdesign13: Sorry, there was some confusion here due to the Figma.  This should be `size=“secondary”` or the `size` prop should be removed altogether.  Using `ComponentWrapper` will take care of this.

bigfishdesign13 (changes requested):
Overall, this looks really good, but needs some adjustments based on Edwin’s comments.  I answered Edwin’s questions, which should also address some minor formatting issues.

oliviawongnyc (commented):
Just added a couple of thoughts, mostly on documentation phrasing!

src/theme/components/alphabetFilter.ts
jackiequach: Removed since VStack is no longer used.

src/components/AlphabetFilter/AlphabetFilter.tsx
jackiequach: Updated to Flex with specified dimensions

src/components/AlphabetFilter/AlphabetFilter.tsx
jackiequach: Removed onClick statement since it will always be defined.

src/components/AlphabetFilter/AlphabetFilter.stories.mdx
jackiequach: Added context for both comments.

src/components/AlphabetFilter/AlphabetFilter.stories.mdx
jackiequach: Both. Updated docs with this info.

src/components/AlphabetFilter/AlphabetFilter.test.tsx
jackiequach: I think this is a good idea. I implemented it.

bigfishdesign13:
@jackiequach Is this PR ready to be reviewed again?

EdwinGuzman (approved):
The update is great. Can you just add this to the accessibility resource links? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav#examples

It's just to give a bit more context about containing the heading, description, and a non-list element  within the `nav` element.

src/components/AlphabetFilter/AlphabetFilter.tsx
bigfishdesign13: The `text` variant for `Button` was part of the last release, so this change should be made before this is merged into `development`.

src/theme/components/alphabetFilter.ts
```diff
@@ -0,0 +1,19 @@
+import { defaultElementSizes } from "./global";
+
+const AlphabetFilter = {
+  baseStyle: {
+    button: {
+      minWidth: { base: defaultElementSizes.mobileFieldHeight, md: "1.5rem" },
+      height: { base: defaultElementSizes.mobileFieldHeight, md: "1.5rem" },
```
bigfishdesign13: For the desktop styles, let’s not use hard coded height and width values.  With a variable-width font, the fixed width on the letter buttons does not feel quite right visually.  For `md`, set the height and width values to `auto` and then use padding and margin to handle the spacing.

src/theme/components/alphabetFilter.ts
```diff
@@ -0,0 +1,19 @@
...
+      fontWeight: "bold",
+      lineHeight: "125%",
```
bigfishdesign13: Change this to `1`.

src/theme/components/alphabetFilter.ts
```diff
@@ -0,0 +1,19 @@
+import { defaultElementSizes } from "./global";
+
+const AlphabetFilter = {
+  baseStyle: {
+    button: {
+      minWidth: { base: defaultElementSizes.mobileFieldHeight, md: "1.5rem" },
+      height: { base: defaultElementSizes.mobileFieldHeight, md: "1.5rem" },
+      fontWeight: "bold",
```
bigfishdesign13, line 10: Also add `fontSize: "heading.callout"`.

src/components/AlphabetFilter/AlphabetFilter.tsx
```diff
@@ -0,0 +1,164 @@
...
+    const getButtonElement = (item) => {
+      const isSelectedLetter =
+        selectedLetter === item.value && item.value !== "showAll";
+      const isShowAll = item.value === "showAll";
+      const isButtonDisabled =
+        isDisabled ||
+        (activeLetters && !activeLetters.includes(item.value) && !isShowAll);
+
+      const buttonStyles = isSelectedLetter
+        ? { ...styles.button, border: "1px solid gray" }
```
bigfishdesign13: The border color should be `ui.border.default`.  You’ll need to pull that out of the `border` attribute and add it using a `borderColor` attribute.

src/components/AlphabetFilter/AlphabetFilter.tsx
```diff
@@ -0,0 +1,164 @@
...
+        : isShowAll
+        ? {
+            ...styles.button,
+            fontWeight: "normal !important",
```
bigfishdesign13: Why is the `!important` necessary?  It shoudl be that the latter style attributes override anything that comes beforehand.  Is that not the case here?

src/components/AlphabetFilter/AlphabetFilter.tsx
```diff
@@ -0,0 +1,164 @@
...
+      <ComponentWrapper
+        as="nav"
+        role="navigation"
+        id={id}
+        className={className}
+        ref={ref}
+        __css={styles}
+        {...rest}
+      >
+        {headingText && <Heading>{headingText}</Heading>}
```
bigfishdesign13: `ComponentWrapper` includes `descriptionText` and `headingText` props that will handle all of this formatting.  For an example of how we have used `ComponentWrapper`, take a look at `VideoPlayer`.

src/components/AlphabetFilter/AlphabetFilter.tsx
bigfishdesign13: This is an on-the-fly change.  Also add `fontSize: { base: null, md: “text.caption” }`.  I already updated the Figma designs to reflect this.

bigfishdesign13 (changes requested):
The functionality is great.  Some of the styles need to be adjusted.

src/components/AlphabetFilter/AlphabetFilter.tsx
jackiequach: @bigfishdesign13 We would need to either change `descriptionText` type to accept `JSX.Element` in `ComponentWrapper`, or only allow `descriptionText` to be typeof `string` in `AlphabetFilter`

bigfishdesign13: Hmm, yes, this is true.

@EdwinGuzman, do you recall if we had a specific reason for setting the `descriptionText` prop in `ComponentWrapper` as `string` and not as `string | JSX.Element`?

EdwinGuzman: No, I don't. It can be both.

src/components/AlphabetFilter/AlphabetFilter.tsx
jackiequach: The selector that theme file used was higher priority than the style here. Moved these styles to the theme file under button > _last to simplify

EdwinGuzman (approved):

bigfishdesign13 (approved):
Looks good!

PR #1147, opened 2022-09-20
DSD-1142: Accordion always render content

EdwinGuzman:
Fixes JIRA ticket [DSD-1142](https://jira.nypl.org/browse/DSD-1142)

- Adds a prop to always render content inside of an `Accordion`.
- This came up in MLN where checkboxes were checked, the `Accordion` closed, and then when it was opened again, the checked status was gone. For form-components, we want to always render the content so the state can persist when the `Accordion` is closed and opened.

## How has this been tested?

Locally in Storybook.

- N/A

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/Accordion/Accordion.tsx
+++ b/src/components/Accordion/Accordion.tsx
@@ -52,8 +52,8 @@ const getIcon = (isExpanded = false, index: number, id: string) => {
 const getElementsFromData = (
   data: AccordionDataProps[] = [],
   id: string,
-  panelMaxHeight: string,
-  isAlwaysRendered: boolean = false
+  isAlwaysRendered: boolean = false,
+  panelMaxHeight: string
 ) => {
   const colorMap = {
     default: "ui.white",
@@ -167,8 +167,8 @@ export const Accordion = chakra(
         {getElementsFromData(
           accordionData,
           id,
-          panelMaxHeight,
-          isAlwaysRendered
+          isAlwaysRendered,
+          panelMaxHeight
         )}
       </ChakraAccordion>
     );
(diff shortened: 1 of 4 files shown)


src/components/Accordion/Accordion.tsx
```diff
@@ -49,7 +52,8 @@ const getIcon = (isExpanded = false, index: number, id: string) => {
 const getElementsFromData = (
   data: AccordionDataProps[] = [],
   id: string,
-  panelMaxHeight: string
+  panelMaxHeight: string,
+  isAlwaysRendered: boolean = false
```
bigfishdesign13: Nit: alphabetize the list items

bigfishdesign13 (commented):
This all looks good, but I am wondering if the overall functionality should be changed to render the panel only when opened, but to then NOT be removed it from the DOM when closed.  That was actually what I thought the component was doing already.  That would remove the need for this new prop.  The original problem that was solved by only rendering when opened was that there was too much being rendered into the page on the initial page load.  What I am suggesting would also solve that problem.

Thoughts?

src/components/Accordion/Accordion.tsx
EdwinGuzman: Oh, these are arguments for the function and I don't think they need to be alphabetized, but I'll do it for this sake.

EdwinGuzman:
Your suggestion makes sense. We noticed a visible lag when we had a lot of accordions rendered on a page, but we haven't tested a complicated component that has many components with an `Accordion`. Maybe it's being too cautious but I don't want to _possibly_ cause rendering issues in those type of apps. It also makes the component slightly more complicated because it has to keep track for when it was already opened.

bigfishdesign13 (approved):
Looks good.

PR #1152, opened 2022-09-23
DSD-585: Add AudioPlayer Component

Toxiapo:
Fixes JIRA ticket[ DSD-585](https://jira.nypl.org/browse/DSD-585)

- Add AudioPlayer Component

## How has this been tested?

- Unit test and storybook

- N/A

### Checklist:

### Front End Review:


src/components/AudioPlayer/AudioPlayer.tsx
```diff
@@ -0,0 +1,192 @@
+import { Box, chakra, useMultiStyleConfig } from "@chakra-ui/react";
+import React, { forwardRef } from "react";
+
+import ComponentWrapper from "../ComponentWrapper/ComponentWrapper";
+
+/* Define all third party services, will be used for type check
+ * and also used to create the AudioType
```
EdwinGuzman: Nit: add periods at the end of every comment sentence.

src/components/AudioPlayer/AudioPlayer.tsx
```diff
@@ -0,0 +1,192 @@
...
+  /** Optional string to set the text for the audio player description */
+  descriptionText?: string;
+  /** Optional string to set a code snippet provided by Libsyn, SoundCloud or Spotify; the
+   * `AudioPlayer` component will accept the `embedCode` prop or the `filePath` prop
+   */
+  embedCode?: string;
+  /** Optional string to set the audio file, the path can be relative or absolute
+   * referring to a locally hosted file, or a fully qualified URL pointing to a locally hosted file
+   * or to another domain altogether.
+   * IMPORTANT: This prop will not be used for this version of the AudioPlayer component
```
EdwinGuzman: It would be good to clarify what "this version" means. It could be `1.1.0` or `1.1.x` but then we don't know when the `filePath` will be working. For now, maybe saying the inverse that "this prop won't be used until a future version" will be better. Or maybe we should omit this prop to avoid confusion, @bigfishdesign13 ?

src/theme/components/audioPlayer.ts
```diff
@@ -0,0 +1,13 @@
+const AudioPlayer = {
+  baseStyle: {
+    invalid: {
+      backgroundColor: "ui.bg.default",
+      padding: "s",
+    },
+    helperErrorText: {
+      fontSize: "text.caption",
```
EdwinGuzman: This should already be the default font size. Was it not correct before and you had to update it?

src/components/AudioPlayer/AudioPlayer.tsx
```diff
@@ -0,0 +1,192 @@
...
+    }
+  )
+);
+
+/**
+ * Parse the embedCode string to a DOM object for proper formatted HTML element.
+ * We are only interested in the iframe element
+ */
+export function parseIframeEmbedCode(
+  embedCode: string
```
EdwinGuzman: From the `AudioPlayerTypes`, the `embedCode` prop is optional and it can be undefined. So here, it should also be set to a potential undefined value `string | undefined`.

src/components/AudioPlayer/AudioPlayer.tsx
```diff
@@ -0,0 +1,192 @@
...
+      // Only set the iframe title if it doesn't already have it in the iframe
+      if (iframeDoc && !iframeDoc.title) {
+        iframeDoc.title = iframeTitle ? iframeTitle : "Embedded audio player";
+      }
+
+      // A friendly message states that we don't support filePath prop
+      const isLocalAudioFile = filePath || audioType === "file";
+      if (isLocalAudioFile) {
+        console.warn(
+          "The current AudioPlayer does not support locally hosted audio file."
```
EdwinGuzman: Set this as the pattern for other warning messages, `'NYPL Reservoir AudioPlayer: The current AudioPlayer does not support a locally hosted audio file.'`

src/components/AudioPlayer/AudioPlayer.tsx
```diff
@@ -0,0 +1,192 @@
...
+          'NYPL Reservoir AudioPlayer: The `embedCode` prop will be ignored when the audioType prop is set to "file".'
+        );
+      }
+      const isLocalWithoutPath = audioType === "file" && !filePath;
+      if (isLocalWithoutPath) {
+        // Question: Should we remove this warning since we don't support filepath?
+        console.warn(
+          'NYPL Reservoir AudioPlayer: The `filePath` prop is required when the `audioType` prop is set to "file."'
+        );
+      }
```
EdwinGuzman: Yeah, I'm not sure if we really need this now and maybe the first message above is enough.

src/components/AudioPlayer/AudioPlayer.tsx
```diff
@@ -0,0 +1,192 @@
...
+            }}
+          />
+        ) : null;
+      }
+
+      const embedElement = buildFinalElement();
+
+      return (
+        <Box
+          className={className}
```
EdwinGuzman: This should be
```
{`audioplayer ${className}`}
```
to align with the prop comment.

src/components/AudioPlayer/AudioPlayer.tsx
```diff
@@ -0,0 +1,192 @@
...
+          data-testid="audio-player-component"
+          id={id}
+          ref={ref}
+          __css={styles}
+          {...rest}
+        >
+          <ComponentWrapper
```
EdwinGuzman: I'm wondering if you need the wrapper `Box` at all. Maybe you can pass all the props to `ComponentWrapper` and have that be the top level element. Does that work? If not, that's okay, just curious.

src/components/AudioPlayer/AudioPlayer.test.tsx
```diff
@@ -0,0 +1,218 @@
...
+const spotifyCode = `<iframe style="border-radius:12px" src="https://open.spotify.com/embed/playlist/37i9dQZF1DX0uqkwkR49kK?utm_source=generator" width="100%" height="380" frameborder="0" allowfullscreen="" allow="autoplay; clipboard-write; encrypted-media; fullscreen; picture-in-picture" ></iframe>`;
```
EdwinGuzman: Remove "autoplay" for these two examples.

src/components/AudioPlayer/AudioPlayer.test.tsx
```diff
@@ -0,0 +1,218 @@
...
+
+describe("AudioPlayer Accessibility", () => {
+  it("Passes axe accessibility test", async () => {
+    const { container } = render(
+      <AudioPlayer embedCode={libsynCode} audioType="libsyn" />
+    );
+    expect(await axe(container)).toHaveNoViolations();
+  });
```
EdwinGuzman: Create one test for every 3rd party lib.

src/components/AudioPlayer/AudioPlayer.stories.mdx
```diff
@@ -0,0 +1,189 @@
...
+    id: { control: false },
+  }}
+/>
+
+# Audio Player
+
+| Component Version | DS Version |
+| ----------------- | ---------- |
+| Added             | `1.1.0`    |
+| Latest            | `1.1.0`    |
```
EdwinGuzman: I think this might be `1.2.0`.

src/components/AudioPlayer/AudioPlayer.stories.mdx
```diff
@@ -0,0 +1,189 @@
...
+<ArgsTable story="AudioPlayer" />
+
+## Accessibility
+
+The `VideoPlayer` component renders an `iframe` element from `Libsyn`, `SoundCloud`, or `Spotify`.
+`iframe`s can be accessible if a `title` attribute is provided.
+The title must already be provided in the embed code value that is passed to the
+`embedCode` prop. A default one is provided but a custom one is recommended.
+
+To guarantee that the audio player will be accessible.
```
EdwinGuzman: This sentence is dangling and missing a second part.

src/components/AudioPlayer/AudioPlayer.stories.mdx
```diff
@@ -0,0 +1,189 @@
...
+Besides setting the `title` in the embedded `iframe`, there are _some_ ways we
+can reduce accessibility issues.
+
+- The player should not be set to auto-play when the page first loads.
```
EdwinGuzman, line 94: Remove the "autoplay" text in the embed code examples. This is great to have but the examples still have that code.

src/components/AudioPlayer/AudioPlayer.stories.mdx
```diff
@@ -0,0 +1,189 @@
...
+- Consider using the description or helper text to add a link to a full transcript of the audio file. Refer to the link below for additional information.
+
+Resources:
+
+- [W3C Making Audio and Video Media Accessible](https://www.w3.org/WAI/media/av/)
+- [WebAIM Creating Accessible Frames and Iframes](https://webaim.org/techniques/frames/)
+- [MDN iframe: The Inline Frame element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe)
+- [Deque University Be Sure to Provide Titles for Iframes](https://dequeuniversity.com/tips/provide-iframe-titles)
+
+## Example Embed Code Snippets
```
EdwinGuzman, line 106: For each 3rd party code example, you can add the actual player. This is similar to how you have the example in the `## Errored` heading.

EdwinGuzman (changes requested):
This is great. The component itself is good but there are some minor updates. 

I think the `filePath` prop can be removed for now since it might cause confusion why it's there but not useful yet.

For another PR, the `parseIframeEmbedCode` function can go into the global utility file, be used in the `VideoPlayer`, and have separate tests for it. But this should be a different ticket and PR.

src/theme/components/audioPlayer.ts
Toxiapo: Yep, the default size for the component is 12px, the Figma design shows 14px for the AudioPlayer component, but it could be that we should be using the default? @bigfishdesign13

src/components/AudioPlayer/AudioPlayer.tsx
bigfishdesign13: If the code for this prop has already been added, I would suggest commenting the code and adding a `TODO` note with the code.  It can be vague and say that support for the prop will be added in a future release.

src/theme/components/audioPlayer.ts
bigfishdesign13: The deisgns were wrong, but have now been fixed.  That was an older design and the font size for the helper text has changed since the design were originally put together.  That being the case, you do not need to add any styles to override the font sizes coming from the `ComponentWrapper` component.  The styles that were added here can be removed.

src/components/AudioPlayer/AudioPlayer.tsx
bigfishdesign13: The same structure is used in `VideoPlayer`.  If we find that the parent `Box` element can be removed, then the same change should be made to `VideoPlayer`.

src/components/AudioPlayer/AudioPlayer.stories.mdx
bigfishdesign13: Yes, the next point release will be `1.2.0`, so let’s use that here.  I would like to include this component in that release.

src/components/AudioPlayer/AudioPlayer.tsx
EdwinGuzman: Yes, but that shouldn't be done in this PR if we decide to make that update happen. The other difference is that in this component, `ComponentWrapper` is always rendered, but in `VideoPlayer`, the `ComponentWrapper` is conditionally rendered. That's why I'm thinking that the `Box` can be removed and just always render the `ComponentWrapper`.

Here, if there's an error, the heading and description will render along with the error message. In the `VideoPlayer`, if there's an error, only the error message will render and not the heading or description. I think maybe we can rethink the `VideoPlayer` but, again, that should be a different PR.

Toxiapo: I would also prefer just to use a single `div` for the component if possible. I will remove the Box wrapper; if there are no errors, I will keep it that way for this PR.

src/components/AudioPlayer/AudioPlayer.tsx
Toxiapo: Just to align with the update from standups. I will remove these warnings as much as possible and comment out the `filePath` props for this PR.

src/components/AudioPlayer/AudioPlayer.tsx
Toxiapo: I will remove this from this PR for now.

src/components/AudioPlayer/AudioPlayer.stories.mdx
Toxiapo: Updated. Thanks!

src/theme/components/audioPlayer.ts
Toxiapo: Updated, thanks!

Toxiapo:
@EdwinGuzman @bigfishdesign13 This PR is ready to be reviewed again.

Should I create the ticket for the `parseIframeEmbedCode` update and potentially remove the `Box` wrapper on the `VideoPlayer` component? or maybe @bigfishdesign13 can? I am not familiar with how the process works. But either way, I am happy to work on it in future sprints.

src/components/AudioPlayer/AudioPlayer.stories.mdx
```diff
@@ -0,0 +1,166 @@
...
+
+The examples below are shown with precise HTML formatting. However, often the
+embed code snippets provided by 3rd party services are delivered as one
+long string.
+
+### Libsyn
```
EdwinGuzman, line 112: The example iframe code you had before was fine to keep. I should've been clearer to say that the example canvas should be added in _addition_.

You can also add something like "The following is an example iframe code for Libsyn that would go into the `embedCode` prop."

```html
<iframe
  title="Libsyn Player"
  style="border: none"
  src="//html5-player.libsyn.com/embed/episode/id/18268511/height/90/theme/custom/thumbnail/yes/direction/backward/render-playlist/no/custom-color/87A93A/"
  height="90"
  width="100%"
  scrolling="no"
  allowfullscreen
  webkitallowfullscreen
  mozallowfullscreen
  oallowfullscreen
  msallowfullscreen
></iframe>
```

EdwinGuzman: You can add the other iframe examples for Spotify and Soundcloud.

EdwinGuzman (approved):
Looks good to me. Adding the iframe example code would be nice to have.

I'll create the tickets and we can discuss them in the next sprint planning.

src/components/AudioPlayer/AudioPlayer.stories.mdx
```diff
@@ -0,0 +1,166 @@
...
+
+- [Overview](#overview)
+- [Component Props](#component-props)
+- [Accessibility](#accessibility)
+- [Example Embed Code Snippets](#example-embed-code-snippets)
+- [Errored](#errored)
+
+## Overview
+
+<Description of={AudioPlayer} />
```
bigfishdesign13: I would remove this.  Right now, the description coming in from the `.tsx` file is not offering anything more than what is described in the text below.

src/components/AudioPlayer/AudioPlayer.stories.mdx
```diff
@@ -0,0 +1,166 @@
...
+
+The `AudioPlayer` component is used to embed a streaming audio player from 3rd party services such as `Libsyn`, `SoundCloud`, and `Spotify`
+on a page. The component requires the `embedCode` prop or `filePath` prop. The component will not function properly if you try to set
+both of them.
+
+**This version will not be used for the playback of locally hosted audio files**.
```
bigfishdesign13: Update to the following:

The current version of the `AudioPlayer` component does not support the playback of locally hosted audio files.  That functionality will be added in a future version of the component.

src/components/AudioPlayer/AudioPlayer.stories.mdx
```diff
@@ -0,0 +1,166 @@
...
+      />
+    )}
+  </Story>
+</Canvas>
+
+<ArgsTable story="AudioPlayer" />
+
+## Accessibility
+
+The `VideoPlayer` component renders an `iframe` element from `Libsyn`, `SoundCloud`, or `Spotify`.
```
bigfishdesign13: Update this parapgrah to the following:

The `VideoPlayer` component renders an iframe element from `Libsyn`, `SoundCloud`, or `Spotify`. The accessibility of the `iframe` element can be improved if a `title` attribute is provided and most 3rd party audio services will include a `title` attribute in the embed code they provide.  If the `title` attribute is omitted from the embed code, a default `title` will be added by the `AudioPlayer` component, but it is recommended to include a custom `title` that is related to the audio file content.

bigfishdesign13 (changes requested):
The functionality is great.  I would just like to see some changes to the documentation.

src/components/AudioPlayer/AudioPlayer.stories.mdx
Toxiapo: That makes sense. Updated.

src/components/AudioPlayer/AudioPlayer.stories.mdx
Toxiapo: Updated. Thank you!

bigfishdesign13:
@Toxiapo Is this PR now ready for a final review?

Toxiapo:
@bigfishdesign13 yes it is!

bigfishdesign13 (approved):
Looks good.

PR #1160, opened 2022-10-03
DSD-1109: Dark mode nav category

oliviawongnyc:
Fixes JIRA ticket [DSD-1109](https://jira.nypl.org/browse/DSD-1109).

- Adds `dark mode` color mode support for the `Breadcrumbs`, `Link Types`, and `Pagination` components.
- Adds `brand` as a `breadcrumbsType` to the `Breadcrumbs` component.

## This PR does _not_:

- Include `dark mode` color mode support for the `AlphabetFilter,` because it has not yet been merged into develop.
- Include the styling for "Mid page (before numbers shift)" in the `Pagination` component as seen [here](https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Main?node-id=10766%3A1031). This is because I don't believe it is currently present in light mode either. @bigfishdesign13 to check on this and let me know if it should be added in both scenarios.

## How has this been tested?

- Locally in Storybook.

- None.

### Checklist:

### Front End Review:

### DIFF
--- a/src/theme/components/breadcrumb.ts
+++ b/src/theme/components/breadcrumb.ts
@@ -20,8 +20,17 @@ const booksAndMore = {
     bg: "dark.section.books-and-more.secondary",
   },
 };
+const brand = {
+  bg: "brand.secondary",
+  _dark: {
+    bg: "dark.brand.secondary",
+  },
+};
 const education = {
   bg: "section.education.secondary",
+  _dark: {
+    bg: "dark.section.education.secondary",
+  },
 };
 const locations = {
   bg: "section.locations.primary",
@@ -103,6 +112,7 @@ const Breadcrumb = {
   variants: {
     blogs,
     booksAndMore,
+    brand,
     education,
     locations,
     research,
--- a/src/theme/components/breadcrumb.ts
+++ b/src/theme/components/breadcrumb.ts
@@ -2,6 +2,9 @@
 const blogs = {
   bg: "section.blogs.secondary",
   color: "ui.black",
+  _dark: {
+    bg: "dark.section.blogs.secondary",
+  },
   a: {
     _hover: {
       color: "ui.gray.x-dark",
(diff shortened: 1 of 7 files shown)


EdwinGuzman:
The "Mid Page (before numbers shift) is available in both light and dark mode and the update looks fine: 
Screen Shot 2022-10-03 at 11 27 16 AM

In the figma designs, the "10" (which in the screenshot above would be "100") is colored white. I think that's a mistake, similar to how in the light mode version, the "10" is colored black.

EdwinGuzman (approved):

src/theme/components/link.ts
```diff
@@ -1,6 +1,9 @@
 export const baseLinkStyles = {
   color: "ui.link.primary",
   textDecoration: "underline",
+  _dark: {
```
bigfishdesign13, line 4: Also add dark mode colors for `_hover`.

src/theme/components/breadcrumb.ts
```diff
@@ -13,18 +16,39 @@ const blogs = {
...
 const research = {
   bg: "section.research.secondary",
+  _dark: {
+    bg: "dark.section.research.secondary",
+  },
 };
 const whatsOn = {
   bg: "section.whats-on.secondary",
+  _dark: {
+    bg: "dark.section.whats-on.secondary",
```
bigfishdesign13, line 50: The function that is used to produce this value is returning an bad value because the root hex value for this color has 3 chars.  All the others have 6 chars and that is what the function is setup to handle.  To fix this, update the `hexToRGB` function (found in utils.ts) with the following:
```
export const hexToRGB = (hex: string, alpha: number) => {
  const shortHex = hex.length === 4;
  const rSlice = shortHex ? hex.slice(1, 2).repeat(2) : hex.slice(1, 3),
    gSlice = shortHex ? hex.slice(2, 3).repeat(2) : hex.slice(3, 5),
    bSlice = shortHex ? hex.slice(3, 4).repeat(2) : hex.slice(5, 7);
  const r = parseInt(rSlice, 16),
    g = parseInt(gSlice, 16),
    b = parseInt(bSlice, 16);
  const rgb = `${r}, ${g}, ${b}`;
  return alpha ? `rgba(${rgb},${alpha})` : `rgb(${rgb})`;
};
```

bigfishdesign13 (changes requested):
This needs one small update and one medium update.

Regarding the styling for "Mid page (before numbers shift)" in the `Pagination` component…the light mode Figma designs contained an error that was simply compied over to the dark mode designs.  In the `Pagination` component, only one number should be highlighted at a time and what you have implemented is correct.

src/theme/components/link.ts
oliviawongnyc: In the Figma file, the color in the hover state is the same as the normal state, but I'm guessing it should be "dark.ui.link.secondary" so I updated to that. I also made added dark mode for the button variant based on the dark mode of the button component. Let me know if it looks right to you when I push!

bigfishdesign13: Looks great.

bigfishdesign13 (approved):
Looks good!

PR #1178, opened 2022-10-26
DSD-1138: Adding isClearable prop to TextInput component

EdwinGuzman:
Fixes JIRA ticket [DSD-1138](https://jira.nypl.org/browse/DSD-1138)

- Adds an `isClearable` prop to the `TextInput` component. This adds a "close" button that renders above the input field on its right side.
- When set to `true`, the button will be set to render once the user types in any value.
- When the close button is clicked, the value in the input field will be cleared and the button is removed.
- Had to add internal state to the `TextInput` to keep track of the input value and also make it a "controlled" component since it will always need an `onChange` function to keep track of the state changes.

## How has this been tested?

Locally.

- We cannot have any empty buttons so there is a hidden span with text. This allows the `"text"` `buttonType` to render appropriately. Otherwise, the other option is to add an `aria-label` to the button. The problem with the `aria-label` approach is that the `"icon-only"` type takes precedence and it's harder to override the button styles. To make it look closer to the Figma design, I added the span with hidden text and used the `"text"` type. @bigfishdesign13 let me know what you think and what the text inside the span should be after asking Clare.

### Checklist:

### Front End Review:


EdwinGuzman:
Ready.

src/components/TextInput/TextInput.stories.mdx
```diff
@@ -51,14 +53,16 @@ import DSProvider from "../../theme/provider";
...
-| Latest            | `1.2.0`    |
+| Latest            | `1.2.2`    |
 
 ## Table of Contents
 
 - [Overview](#overview)
 - [Component Props](#component-props)
 - [Accessibility](#accessibility)
 - [Labelling Variations](#labelling-variations)
+- [Visible States](#visible-states)
```
bigfishdesign13: In the Figma design, we use `Browser States`.

src/theme/components/textInput.ts
```diff
@@ -38,14 +39,26 @@ const input = {
...
+  baseStyle: ({ showLabel }) => ({
+    position: "relative",
     input,
     textarea: {
       ...input,
       lineheight: "1.5",
       minHeight: "xxl",
     },
-  },
+    clearButton: {
```
bigfishdesign13, line 50: The default padding is not great, so add the following:
```
px: "xs"
```

bigfishdesign13 (changes requested):
Overall, this works really well.  Just need to adjust the padding for the new button.

src/theme/components/textInput.ts
EdwinGuzman: This is odd but updating this does not affect the padding. Check it out. I think it's because of the flex display inside of the button.

bigfishdesign13 (approved):
Looks good!

PR #1184, opened 2022-10-28
DSD-1152: Close Chakra's Tooltip

EdwinGuzman:
Fixes JIRA ticket [DSD-1152](https://jira.nypl.org/browse/DSD-1152)

- Adds more props to Chakra's `Tooltip` component so that it closes more forcefully and does not stay open, especially on DS `Image`s.

## How has this been tested?

Locally and in Turbine. After speaking with Deepika on this, the `Tooltip` does _eventually_ close if it's hovered on and off again. So it doesn't stay on indefinitely. The recommendation would be to not use the `shouldWrapChildren` prop for DS `Image`s so they render correctly.

- N/A

### Checklist:

### Front End Review:


src/components/Tooltip/Tooltip.tsx
```diff
@@ -58,12 +58,16 @@ export const Tooltip = chakra(
 
     return (
       <ChakraTooltip
-        hasArrow
         aria-label={typeof content !== "string" ? "Tooltip" : undefined}
-        label={content}
+        closeDelay={750}
```
bigfishdesign13, line 62: Does the addition of these new props act as a backup to ensure that the tooltip does close eventually?  Meaning, does the bug still exist and we are just adding something to gracefully get around the bug?  Or does this resolve the bug altogether?

bigfishdesign13 (commented):
Please see my question.  I am just curious before I approve.

src/components/Tooltip/Tooltip.tsx
EdwinGuzman: Yep, pretty much works as a backup to ensure it closes. That being said, it does resolve the bug but in a roundabout way.

bigfishdesign13 (approved):
Release the hounds!

PR #1185, opened 2022-10-31
DSD-1162: StyledList component

EdwinGuzman:
Fixes JIRA ticket [DSD-1162](https://jira.nypl.org/browse/DSD-1162)

- Adds the `StyledList` component.

## How has this been tested?

Locally.

- N/A

### Checklist:

### Front End Review:


src/theme/components/styledList.ts
```diff
@@ -0,0 +1,23 @@
...
+const StyledList = {
+  baseStyle: ({ textSize = "default" }) => ({
+    fontSize: `text.${textSize}`,
+    ...baseListStyles(false, true),
+    ...baseUnorderedStyles(true),
+  }),
+  variants: {
+    capped: {
+      li: {
+        borderLeft: "2px solid",
```
bigfishdesign13: Both the left and right borders should be 1px.

src/theme/components/styledList.ts
```diff
@@ -0,0 +1,23 @@
...
+        borderRight: "2px solid",
```
bigfishdesign13: Sorry, it wasn't clearly apparent in the Figma designs, but the styling for `desktop` and `mobile` are slightly different.  For `desktop`, the right border and padding should be removed.  The border on both ends is only needed for the `mobile` view.  

For a better example, see the link below: 
https://www.figma.com/file/CRPxcR5Olmz50IrFZxvLKE/DRB---Search-Highlighting-(SFR-1519)?node-id=1279%3A12441

bigfishdesign13 (changes requested):
Just needs some polishing.

src/theme/components/styledList.ts
EdwinGuzman: Right. This could use it's own figma card like the other components.

EdwinGuzman: Especially when more styles are added.

bigfishdesign13: Yes, the designs in the Visual Design Library (the Figma Main file) are coming.

src/theme/components/styledList.ts
```diff
@@ -0,0 +1,26 @@
...
+      li: {
+        borderLeft: "1px solid",
+        borderRight: {
+          base: "1px solid",
+          md: "none",
+        },
+        borderColor: "ui.border.default",
+        my: "xs",
+        px: "xs",
+        width: "fit-content",
```
bigfishdesign13: Sorry, I missed this earlier.  This can be removed altogether.  The mobile UI should span the full width of the window and for desktop it does not impact anything.

bigfishdesign13 (changes requested):
😳

src/theme/components/styledList.ts
bigfishdesign13: That's supposed to be an embarrassed emoji, but I don't think it's all that convincing.

bigfishdesign13 (approved):

src/components/StyledList/StyledList.stories.mdx
```diff
@@ -0,0 +1,200 @@
...
+
+## Table of Contents
+
+- [Overview](#overview)
+- [Component Props](#component-props)
+- [Accessibility](#accessibility)
+- [StyledList Children](#styledlist-children)
+- [Font Size](#font-size)
+
+## Overview
```
oliviawongnyc, line 51: Did you intentionally leave out the "Overview"?

src/components/StyledList/StyledList.stories.mdx
```diff
@@ -0,0 +1,200 @@
...
+
```
oliviawongnyc, line 52: Should there be a heading, "Component Props", here?

src/components/StyledList/StyledList.tsx
```diff
@@ -0,0 +1,65 @@
...
+        id,
+        listItems = [],
+        style = "capped",
+        textSize = "default",
+      } = props;
+      const styles = useMultiStyleConfig("StyledList", {
+        textSize,
+        variant: style,
+      });
+      let finalChildren;
```
oliviawongnyc, line 44: Just curious: why do you declare this variable here as opposed to just declaring it when you assign its value below?

src/theme/components/styledList.ts
```diff
@@ -0,0 +1,25 @@
...
+const StyledList = {
+  baseStyle: ({ textSize = "default" }) => ({
+    fontSize: `text.${textSize}`,
+    ...baseListStyles(false, true),
+    ...baseUnorderedStyles(true),
+  }),
+  variants: {
+    capped: {
+      li: {
+        borderLeft: "1px solid",
```
oliviawongnyc: Should we use border-inline-start rather than left per your multilingual talk yesterday?

https://developer.mozilla.org/en-US/docs/Web/CSS/border-inline-start

src/theme/components/styledList.ts
```diff
@@ -0,0 +1,25 @@
...
+        borderRight: {
```
oliviawongnyc: What is this borderRight doing?

src/components/StyledList/StyledList.stories.mdx
```diff
@@ -0,0 +1,200 @@
...
+If `<li>` elements cannot be passed as children to the `StyledList` component,
+an array with data can be passed into the `listItems` prop. Note: either
+children elements or the `listItems` prop should be passed but not both.
+
+For data in the `listItems` prop, the `StyledList` component will render the
+`li` elements. There is no need to add them in the `listItems` array.
+
+```jsx
+<!-- With `itemList` data prop -->
+const listItems = [
```
oliviawongnyc: Is this meant to be "`listItems` data prop"?

src/components/StyledList/StyledList.stories.mdx
```diff
@@ -0,0 +1,200 @@
...
+  </DSProvider>
+</Canvas>
+
+### With listItems prop
+
+If `<li>` elements cannot be passed as children to the `StyledList` component,
+an array with data can be passed into the `listItems` prop. Note: either
+children elements or the `listItems` prop should be passed but not both.
+
+For data in the `listItems` prop, the `StyledList` component will render the
```
oliviawongnyc: Maybe add a reminder here that the data can be a string or a JSX Element.

Data in the `listItems` prop array can be JSX Elements or strings. The `StyledList` component will render the data as `li` elements, so there is no need to add them in the `listItems` array.

src/components/StyledList/StyledList.stories.mdx
```diff
@@ -0,0 +1,200 @@
...
+`li` elements. There is no need to add them in the `listItems` array.
```
oliviawongnyc: If a user does add `li` items to the listItems array, what will happen to the component? What do nested `li`s look like? Should we make sure this doesn't happen in some way?

oliviawongnyc (commented):
Looks good! Just some small musings and questions.

src/components/StyledList/StyledList.stories.mdx
EdwinGuzman: wow no, good catch

src/theme/components/styledList.ts
EdwinGuzman: There was something that did not work when I tried that initially. Maybe I'll have to do a workaround but you're right that we shouldn't use "left".

src/components/StyledList/StyledList.tsx
EdwinGuzman: I didn't want the component to go through the assignment of `children` or run through the `map` function on line 54 if there were any issues. Those issues are caught before and the component just returns null -- extremely minor and negligible optimization. The js compiler does variable hoisting anyway so this pattern is similar to what it does.

EdwinGuzman: I also prefer declaring all variables at the beginning rather than in the middle of the component. Just a preference nitpicky thing though.

src/theme/components/styledList.ts
EdwinGuzman: It wasn't obvious to me but the end border should not display on desktop, only mobile.

src/components/StyledList/StyledList.stories.mdx
EdwinGuzman: Two borders will render! This isn't ideal. We can iterate on this later. This would be a good discussion for a larger audit of what other components we can do the same functionality where we _remove_ unnecessary HTML elements.

EdwinGuzman:
@oliviawongnyc updated

PR #1202, opened 2022-11-29
DSD-234: FilterBar

isastettler:
Fixes JIRA ticket DSD-234

- Adds FilterBar component
- Adds useFilterBar hook
- Adds tests for FilterBar and useFilterBar
- Adds storybook documentation for FilterBar component
- Adds storybook documentation for useFilterBar
- Updates useMultiSelect hook to work in Scout App
- Updates tests of MultiSelect and useMultiSelect accordingly
- Updates storybook for MultiSelect and useMultiSelect

-

### Checklist:

### Front End Review:


src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -50,21 +50,25 @@ export const ButtonGroup = chakra(
...
-          if (child.type !== Button) {
-            // Special case for Storybook MDX documentation.
-            if (child.props.mdxType && child.props.mdxType === "Button") {
-              noop();
-            } else {
-              console.warn(
-                "NYPL Reservoir ButtonGroup: Only Button components can be children of ButtonGroup."
-              );
-              return;
+          // @TODO: discuss with DS team - this fixes the issue of a null hcild passed (see FilterBar conditional buttons)
```
EdwinGuzman: This is fine, thanks. You can remove the comment.

src/components/FilterBar/FilterBar.tsx
```diff
@@ -0,0 +1,241 @@
...
+import ButtonGroup from "../ButtonGroup/ButtonGroup";
+import Heading from "../Heading/Heading";
+import { LayoutTypes } from "../../helpers/types";
+import MultiSelect from "../MultiSelect/MultiSelect";
+import MultiSelectGroup from "../MultiSelectGroup/MultiSelectGroup";
+import { SelectedItems } from "../MultiSelect/MultiSelect";
+import useNYPLBreapoints from "../../hooks/useNYPLBreakpoints";
+
+interface FilterBarCommonProps {
+  children: React.ReactNode;
```
EdwinGuzman, line 27: These values, along with the ones in the `type`s, should have a comment with a description. It seems like you already have those in the Storybook doc, but it should be here as well.

src/components/FilterBar/FilterBar.stories.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+| Component Version | DS Version |
+| ----------------- | ---------- |
+| Added             | `1.4.0`    |
+| Latest            | `1.4.0`    |
+
+## Table of Contents
+
+- [Overview](#overview)
+- [Component Props](#component-props)
```
EdwinGuzman, line 72: There should be an `Accessibility` section after `Component Props`. You can write a very brief description and say that the accessibility depends on the `FilterBar`s children. You can then link to the components that it will render, such as `MultiSelect#accessibility`, etc.

src/components/FilterBar/FilterBar.stories.mdx
```diff
@@ -0,0 +1,121 @@
...
+- [Layout Patterns](#layout-patterns)
```
EdwinGuzman, line 74: Add a new heading here for the `useFilterBar hook`  section.

src/components/FilterBar/FilterBar.stories.tsx
```diff
@@ -0,0 +1,495 @@
...
+        showClearAll={true}
+      >
+        <MultiSelectGroup
+          showLabel={false}
+          id="clear-all"
+          labelText="Row with Clear All button"
+        >
+          {multiSelectItems &&
+            multiSelectItems.map((multiSelect, i) => {
+              if (i > 1) return;
```
EdwinGuzman: Minor lint issue.

src/components/FilterBar/FilterBar.stories.tsx
```diff
@@ -0,0 +1,495 @@
...
+        showSubmitAll={true}
+      >
+        <MultiSelectGroup
+          showLabel={false}
+          id="submit-button"
+          labelText="Submit Button"
+        >
+          {multiSelectItems &&
+            multiSelectItems.map((multiSelect, i) => {
+              if (i > 1) return;
```
EdwinGuzman: Minor lint issue.

src/components/FilterBar/FilterBar.stories.tsx
```diff
@@ -0,0 +1,495 @@
...
+                  );
+                }}
+                onClear={() => {
+                  onClear(`${multiSelect.id}-3`);
+                }}
+                onApply={() => {}}
+              />
+            ))}
+        </MultiSelectGroup>
+      </FilterBar>
```
EdwinGuzman, line 769: This is is not really an issue but because every `FilterBar` uses the same props when they are rendered, then when you select a checkbox in one MultiSelect (like in Colors), then it selects the same value for all other MultiSelects.
You can keep this as is but add a sentence or two explaining that this functionality only occurs for the Storybook example.

src/components/FilterBar/FilterBar.test.tsx
```diff
@@ -0,0 +1,237 @@
+import { axe } from "jest-axe";
+import { render, screen } from "@testing-library/react";
+// import renderer from "react-test-renderer";
+// import userEvent from "@testing-library/user-event";
```
EdwinGuzman: Remove these.

src/components/FilterBar/FilterBar.test.tsx
```diff
@@ -0,0 +1,237 @@
...
+import * as React from "react";
```
EdwinGuzman: Nitpick: don't import everything, do `import React from "react";`.

src/components/FilterBar/FilterBar.test.tsx
```diff
@@ -0,0 +1,237 @@
...
+                onClear(multiSelect.id);
+              }}
+              onApply={() => {}}
+            />
+          ))}
+      </MultiSelectGroup>
+    </FilterBar>
+  );
+};
+const MulitSelectTestGroup = (multiSelectItems) => (
```
EdwinGuzman: Nitpick: typo for `MulitSelectTestGroup`.

src/components/FilterBar/FilterBar.test.tsx
```diff
@@ -0,0 +1,237 @@
...
+  </MultiSelectGroup>
+);
+describe("FilterBar Accessibility", () => {
+  it("should have no axe violations", async () => {
+    const { container } = render(<FilterBarTestComponent />);
+    expect(await axe(container)).toHaveNoViolations();
+  });
+});
+describe("FilterBar", () => {
+  it("should clear all selectedItems when 'Clear All' button is clicked", () => {
```
EdwinGuzman: There should be a test for the "Apply Filters" button that appears next to the "Clear filters" button.

src/components/FilterBar/FilterBar.tsx
```diff
@@ -0,0 +1,241 @@
...
+      } = props;
+
+      const { isLargerThanMedium } = useNYPLBreapoints();
+      const finalLayout = isLargerThanMedium ? layout : "column";
+      const styles = useMultiStyleConfig("FilterBar", { layout: finalLayout });
+      const getSelectedItemsCount = () =>
+        Object.entries(selectedItems).length
+          ? `(${Object.entries(selectedItems).length})`
+          : "";
+      //@TODO: resolved using Typescript descriminated unions
```
EdwinGuzman: Try creating a "type guard". See the `isFilterType` function here: https://github.com/NYPL/nypl-design-system/blob/development/src/components/TagSet/TagSet.tsx#L71

src/components/FilterBar/FilterBar.tsx
```diff
@@ -0,0 +1,241 @@
...
+                direction={layout}
+                width="full"
+              >
+                <WrapItem>{newChildren}</WrapItem>
+                <WrapItem alignItems={layout === "row" ? "end" : ""}>
+                  <ButtonGroup>
+                    {showSubmitAll && (
+                      <Button
+                        buttonType="primary"
+                        id={`${id}-clear-all-button`}
```
EdwinGuzman: This should have a different id. It's currently the same as line number 172.

src/hooks/useFilterBar.stories.mdx
```diff
@@ -0,0 +1,271 @@
...
+  setSelectedItems,
+  selectedItems,
+  onOpen,
+  onClose,
+  onToggle,
+  isModalOpen,
+} = useFilterBar();
+```
+
+## useFilterBar Hook simple implementation with no buttons or modul
```
EdwinGuzman: typo: `modal`.

src/hooks/useFilterBar.stories.mdx
```diff
@@ -0,0 +1,271 @@
...
+  const { onChange, onMixedStateChange, onClear, onClearAll, selectedItems } =
+    useFilterBar();
+
+  return (
+    <FilterBar
+      id="filterbar"
+      headingText="FilterBar"
+      layout="row"
+      selectedItems={selectedItems}
+      showClearAll={true}
```
EdwinGuzman: Nitpick: pass just the prop name `showClearAll`, `showSubmitAll`, etc instead of setting `={true}` which is the same.

src/hooks/useFilterBar.stories.mdx
```diff
@@ -0,0 +1,271 @@
...
+      onClear={onClearAll}
+      showSubmitAll={true}
+      onSubmit={() => {}}
+    >
+      <MultiSelectGroup
+        id="multiselect-group"
+        key="multiselect-group-key"
+        labelText="MultiSelect Group"
+      >
+        {multiSelectItems &&
```
EdwinGuzman, line 162: I know this is an example, but it would be good to define what `multiSelectItems` is as part of the example.

src/hooks/useFilterBar.stories.mdx
```diff
@@ -0,0 +1,271 @@
...
+      { id: "rat", name: "Rat" },
+    ],
+  },
+];
+```
+
+The required data structure to set the selectedItems object:
+
+```jsx
+// The selectedItems structur for one MultiSelect
```
EdwinGuzman: typo: `structure`.

src/hooks/useFilterBar.stories.mdx
```diff
@@ -0,0 +1,271 @@
...
+  FilterBar,
+  MultiSelectGroup,
+  MultiSelect,
+  useFilterBar,
+} from "@nypl/design-system-react-components";
+
+export default function FilterBarWithHookExample() {
+  const { onChange, onClear, onClearAll, setSelectedItems, selectedItems } =
+    useFilterBar();
+  setSelectedItems(initialFilter);
```
EdwinGuzman: This is nice but just define an example value for `initialFilter`.

src/hooks/useFilterBar.ts
```diff
@@ -0,0 +1,31 @@
+import { SelectedItems } from "./../components/MultiSelect/MultiSelect";
+import { useState } from "react";
+
+import useMultiSelect from "./useMultiSelect";
+
+export default function useFilterBar(initialState?: SelectedItems) {
```
EdwinGuzman, line 15: Just add a comment above the function describing it.

src/hooks/useMultiSelect.stories.mdx
```diff
@@ -0,0 +1,255 @@
...
+  MultiSelectGroup,
+  useMultiSelect,
+} from "@nypl/design-system-react-components";
+export default function MultiSelectGroupWithHookExample() {
+  const { onChange, onMixedStateChange, onClear, selectedItems, setSelectedItems } =
+    useMultiSelect();
+
+  React.useEffect(() => {
+    setSelectedItems(initialSelectedItems)
+  }, [])
```
EdwinGuzman, line 226: The example in `useFilterBar.ts` does not use the `useEffect`. Make it consistent for both examples.

EdwinGuzman (changes requested):
Works well overall! Just minor updates. Perhaps creating the type guard might be more involved.

src/hooks/useFilterBar.stories.mdx
isastettler: Hey @EdwinGuzman just to clarify: You suggest to remove the detailed example and just leave the data structure example code (line 199 to line 208) for the `initialFilter`? Just want to make sure I understand correctly.

EdwinGuzman: Don't remove anything. You can copy what's on line 200-207 and paste it in this example inside `FilterBarWithHookExample` or outside of it, but just to be located in the entire example you have.

src/components/ButtonGroup/ButtonGroup.tsx
isastettler: Removed.

src/components/FilterBar/FilterBar.tsx
isastettler: Added.

src/components/FilterBar/FilterBar.stories.mdx
isastettler: Added. Happy to make any edits to the text.

src/components/FilterBar/FilterBar.stories.mdx
isastettler: Done.

src/components/FilterBar/FilterBar.stories.tsx
isastettler: I just realized I did not address this yet. What would you like me to change?

src/components/FilterBar/FilterBar.stories.tsx
isastettler: Same ^^.

src/components/FilterBar/FilterBar.test.tsx
isastettler: Removed.

src/components/FilterBar/FilterBar.test.tsx
isastettler: Done.

src/components/FilterBar/FilterBar.test.tsx
isastettler: Ups. Fixed.

src/components/FilterBar/FilterBar.test.tsx
isastettler: Added a bunch of tests...had some struggles mocking the viewport size. Now it is working but I am very aware that there might be a better way to do it. If you have an idea, pls point me in the right direction.

src/components/FilterBar/FilterBar.tsx
isastettler: I was not able to think about it in a way that would apply here. Happy to give it another try if you have a hint for me. However I added a check that my brain was able to produce.

src/components/FilterBar/FilterBar.tsx
isastettler: Fixed.

src/hooks/useFilterBar.stories.mdx
isastettler: Fixed.

src/hooks/useFilterBar.stories.mdx
isastettler: Ah yes. Done.

src/hooks/useFilterBar.stories.mdx
isastettler: To avoid adding all those lines I added it just ones on top of the page. But if you think it is better to add it to each example - or some of the examples - let me know.

src/hooks/useFilterBar.stories.mdx
isastettler: Fixed.

src/hooks/useFilterBar.stories.mdx
isastettler: Done.

src/hooks/useFilterBar.ts
isastettler: Added.

src/hooks/useMultiSelect.stories.mdx
isastettler: Added useEffect since in Scout we have to use it it this way in order to work.

src/components/FilterBar/FilterBar.stories.tsx
isastettler: I added a fix for it. I think it's the more elegant solution ;)

src/components/FilterBar/FilterBar.tsx
```diff
@@ -25,41 +25,62 @@ import useNYPLBreapoints from "../../hooks/useNYPLBreakpoints";
...
 }
 
-type ClearAllProps =
-  | {
-      showClearAll?: false;
-      onClear?: never;
-    }
-  | { showClearAll: true; onClear: () => void };
+// /** Types related to a optional global clear button */
+// type ClearAllProps =
```
EdwinGuzman: You can remove this.

src/components/FilterBar/FilterBar.test.tsx
```diff
@@ -138,15 +150,66 @@ const MulitSelectTestGroup = (multiSelectItems) => (
...
+    render(<FilterBarTestComponent showClearAll id="filter-bar-test-1" />);
+    expect(
+      screen.getByRole("button", { name: /clear filters/i })
+    ).toBeInTheDocument();
+  });
+  it("should clear the selectedItems when 'Clear Filters' button is clicked", () => {
+    const { rerender } = render(
+      <FilterBarTestComponent showClearAll id="filter-bar-test-2" />
+    );
+    rerender(<FilterBarTestComponent showClearAll id="filter-bar-test-2" />);
```
EdwinGuzman: You shouldn't need this because the props don't change from the initial render to the rerender. Why was this added?

EdwinGuzman (commented):
The update looks good. The tests seem alright, but I'm just confused why the `rerender` is called immediately after the initial `render` even though the props don't change.

src/components/FilterBar/FilterBar.test.tsx
isastettler: Yes so that's where I could not find a better solution. If you  add `console.log(id, isLargerThanMedium)` in FilterBar.tsx you will see that for each test the isLargerThanMedium is returning false before the rerender (exept for the first `it()` block. I tried `await`, `await waitFor()` without success, I tried to call window resize in different places and played around with the setup but this was the only thing I was able to get to work. If you have a hunch I'd love to know where to look further.

src/components/FilterBar/FilterBar.tsx
isastettler: Removed.

src/components/FilterBar/FilterBar.stories.tsx
isastettler: Done.

src/components/FilterBar/FilterBar.stories.tsx
isastettler: Done.

src/components/FilterBar/FilterBar.test.tsx
EdwinGuzman: Okay, there are two things going on. I tried removing the `rerender` calls and the tests do fail.

1) When I removed the `rerender` calls and only ran the test on line 232 `it("should render a 'Apply Filters' button when showSubmitAll prop is passed"...` , it passes. When _all_ tests are run, it fails. The order affects whether tests run successfully and that's not ideal. Move the "desktop" tests towards the end of the file (before the snapshot tests).
2) call `window.resizeTo(....)` _after_ the `render` function.

Try that. I managed to make all tests pass except `"should call onSubmit function when 'Apply Filters' button is clicked"`. The `mockonSubmit` function is not called so that's odd but i think that can be figured out.

isastettler: I changed the order of tests and was able to delete `rerender()` for all tests except the `"should call onSubmit function when 'Apply Filters' button is clicked"`. I never ran into the issue you mentioned but it kept rendering the mobile view for me.

src/components/FilterBar/FilterBar.tsx
```diff
@@ -0,0 +1,251 @@
...
+    };
+
+export type FilterBarProps = FilterBarCommonProps &
+  SubmitAllProps &
+  HandleModalProps;
+
+/**
+ * `FilterBar` is a wrapper component for filter components.
+ * The DS considers the following components as filter components:
+ * - `MultiSelectGroup`
```
bigfishdesign13, line 82: Is this currently the only form "group" component that is supported?

src/components/FilterBar/FilterBar.stories.mdx
```diff
@@ -0,0 +1,134 @@
...
+
+## `useFilterBar` hook
+
+`useFilterBar` is an optional hook for
+cases where managing the state of the component in the consuming app is less
+of a concern and general ease of use is prefered.
+
+The hook returns an object containing all the props and state needed to handle the selectedItems.
+That includes the functions `onChange`, `onMixedStateChange`, `onClear`, `onClearAll` for handling any changes to the selection of items,
+a `setSelectedItems` function to control or manipulate the (initial) state of selectedItems and the current state of the selection: `selectedItems`.
```
bigfishdesign13: nit: adjust this text to the following...

...a `setSelectedItems` function to control or manipulate the (initial) state of `selectedItems`, and  a `selectedItems` object to house the current state of the selections.

src/components/FilterBar/FilterBar.stories.mdx
```diff
@@ -0,0 +1,134 @@
...
+On screen width under 600px width the `layout` is automatically set to "column".
+Note: when `layout = "column"` is passed, the MultiSelect components will have `isBlockElement` set to `true` by default.
+
+<Canvas>
+  <DSProvider>
+    <story.FilterBarLayoutStory />
+  </DSProvider>
+</Canvas>
+
+## `useFilterBar` hook
```
bigfishdesign13, line 126: Add a to this section a direct link to the `useFilterBar` page under the `Hooks` heading.

src/components/FilterBar/FilterBar.stories.mdx
```diff
@@ -0,0 +1,134 @@
...
+See the accessibility details of valid FilterBar child components:
+
+- [MultiSelect](https://nypl.github.io/nypl-design-system/reservoir/v1/?path=/docs/components-form-elements-multiselect--multi-select-list-box#accessibility)
+- [MultiSelectGroup](https://nypl.github.io/nypl-design-system/reservoir/v1/?path=/docs/components-form-elements-multiselectgroup--multi-select-group#accessibility)
+
+## Layout Patterns
+
+The display of the FilterBar child components can be set to a "row" or a "column" layout.
+On screen width under 600px width the `layout` is automatically set to "column".
+Note: when `layout = "column"` is passed, the MultiSelect components will have `isBlockElement` set to `true` by default.
```
bigfishdesign13: Also add info that the width of the child elements will be set to "full."

bigfishdesign13 (changes requested):
This is looking very good.  I added some change requests directly to the code, but the bulk of my requests are listed below.

#### Column Layout:
- The child `MultiSelect` components should be set to `width: "full"`
- The `ButtonGroup` should be set to `layout: "column"` and `buttonWidth: "full"` (similar to how the action buttons are setup for the mobile view)

#### Story Page
- In examples with where either `showClearAll` or `showSubmitAll` are `true`, the child `MultiSelect` components should be set to `variant: "listbox"`
- Update some of the `MultiSelect` components so there are enough `Checkbox` options to cause the scrollbar to kick in
- Add at least one example with two `MultiSelectGroup` components

#### Mobile View
- In the modal, the position of the heading (the title and close X) and the footer (the action buttons) should be locked in place
- In the modal, only the content area (the form elements) should scroll if the height extends beyond the available space
- Refer to this example: https://nypl.github.io/nypl-design-system/reservoir/v1/?path=/docs/components-overlays-switchers-modal--content-window-scrolling

src/components/FilterBar/FilterBar.tsx
isastettler: yes

src/components/FilterBar/FilterBar.stories.mdx
isastettler: Updated.

src/components/FilterBar/FilterBar.stories.mdx
isastettler: Added.

src/components/FilterBar/FilterBar.stories.mdx
isastettler: I moved the link that was at  the bottom of this section to the top. Pls let me know if I misunderstood the task.

isastettler:
> This is looking very good. I added some change requests directly to the code, but the bulk of my requests are listed below.
> 
> #### Column Layout:
> * The child `MultiSelect` components should be set to `width: "full"`
> * The `ButtonGroup` should be set to `layout: "column"` and `buttonWidth: "full"` (similar to how the action buttons are setup for the mobile view)

I have changed the default width for `layout: "column` to be` "full"` for the child components as well as the buttons and added an additional `filterWidth` prop that can control the width if needed. Bill has mentioned that on DXP we need to be able to set the width of the child components and I believe Edwin mentioned that too (I could be wrong though with Edwin). 

> 
> #### Story Page
> * In examples with where either `showClearAll` or `showSubmitAll` are `true`, the child `MultiSelect` components should be set to `variant: "listbox"`

Here I am not sure if you would like this just on storybook or in the component. In the second scenario I think it would require the Submit button to be rendered though - or do I get this wrong? 

> * Update some of the `MultiSelect` components so there are enough `Checkbox` options to cause the scrollbar to kick in
> * Add at least one example with two `MultiSelectGroup` components

Updated the examples. Let me know if this works.

> 
> #### Mobile View
> * In the modal, the position of the heading (the title and close X) and the footer (the action buttons) should be locked in place
> * In the modal, only the content area (the form elements) should scroll if the height extends beyond the available space
> * Refer to this example: https://nypl.github.io/nypl-design-system/reservoir/v1/?path=/docs/components-overlays-switchers-modal--content-window-scrolling

This should be working as expected now.

EdwinGuzman:
The update looks good to me but I'll defer to @bigfishdesign13 on the final display review.

> Bill has mentioned that on DXP we need to be able to set the width of the child components and I believe Edwin mentioned that too (I could be wrong though with Edwin).

I don't remember this! I think if what Marty is asking for is a DS pattern and in DXP it needs to be different, it would be good to bring up why this is the case and what exactly the situation is.

src/components/FilterBar/FilterBar.stories.mdx
bigfishdesign13: Oops.  I must have missed that initially.  I'm sorry.  Had the link not existed, the bottom of the section is where I would have suggested to put it.  However, I clearly misssed it when it was there, so leave it in the no position below the section heading.

src/components/FilterBar/FilterBar.tsx
bigfishdesign13: OK, that's what I initially thought.  We'll just need to add support for additional form components before other projects can use the `FilterBar` component.

bigfishdesign13:
> Here I am not sure if you would like this just on storybook or in the component. In the second scenario I think it would require the Submit button to be rendered though - or do I get this wrong?

Sorry, I could have been a little clear.  The suggested pattern is to only use the global `Apply` and `Clear` buttons with the MultiSelect `listBox` variant.  The MultiSelect `dialog` variant has that functionality built into the component, so those global buttons are not necessary.  I would update the examples to match the suggested patterns.

EdwinGuzman:
I haven't checked, but is that pattern in the spec doc? We're adding a bit more responsibility on the user to get the pattern right. I'm not sure how we can enforce this right now, but for the time being, would it make sense to have a section in the Storybook doc that says something like "It's suggested that the global Apply and Clear buttons be used with `listBox` MultiSelect variants." ? This can be included near the examples.

bigfishdesign13:
> I haven't checked, but is that pattern in the spec doc? We're adding a bit more responsibility on the user to get the pattern right. I'm not sure how we can enforce this right now, but for the time being, would it make sense to have a section in the Storybook doc that says something like "It's suggested that the global Apply and Clear buttons be used with `listBox` MultiSelect variants." ? This can be included near the examples.

@EdwinGuzman 
No, this idea is somethign I neglected to include in the spec doc.  But yes, I do think it would be a good idea to include details about the expected NYPL patterns.  I don't want to hold this up, so we can create a ticket to add that documentation after the component is released.

bigfishdesign13 (changes requested):
This is looking better.  There are some additional styling issues that I would like to cleanup, but I will create a new ticket for those updates and I will pick it up after the release.  

There is, however, a functional bug that I found in the mobile view.  While a `MultiSelect` components is open, you can not open any `MultiSelect` components below the one that is open.  When you try to open one of the others, the screen bonces back up to the one that is already open.  If all of the `MultiSelect` components are closed, then there are no problems when opening any of the `Multiselect` components.

We can move forward to release this iteration once this bug has been resolved.

bigfishdesign13 (approved):
Here we go!

PR #1212, opened 2022-12-08
DSD-1220: secondary button BG color

bigfishdesign13:
Fixes JIRA ticket [DSD-1220](https://jira.nypl.org/browse/DSD-1220)

- Updates the background color for the `hover` state in the `secondary` variant of the `Button` component.

## How has this been tested?

- local Storybook

- n/a

### Checklist:

### Front End Review:


src/utils/utils.ts
```diff
@@ -82,3 +82,16 @@ export const getAriaAttrs = ({
 
   return ariaAttributes;
 };
+
+/** Convert a hex color value to an rgb or rgba value */
+export const hexToRGB = (hex: string, alpha: number) => {
```
EdwinGuzman, line 87: Am I imagining this or do we have a similar function somewhere else?

src/theme/foundations/colors.ts
```diff
@@ -56,8 +59,9 @@ const colors: Colors = {
     },
     focus: "#4181F1",
     link: {
-      primary: "#0576D3",
-      secondary: "#004B98",
+      primary: linkPrimary,
+      secondary: linkSecondary,
+      wash: hexToRGB(linkPrimary, 0.05),
```
EdwinGuzman: Why is this called "wash"? Just curious.

EdwinGuzman (approved):
Just non-blocking questions.

src/utils/utils.ts
bigfishdesign13: Yes, this same function was added to the dark mode feature branch.  I needed it here to get the `rgba` value, so I brought it over.

src/theme/foundations/colors.ts
bigfishdesign13: This color value is a transparency and I was trying to be semantic, but this could be better.  I will establish a more logical pattern that we can use for future colors with various transparency values.

EdwinGuzman (approved):

PR #1213, opened 2022-12-08
Bugfix: Node issues

EdwinGuzman:

- Removes the node_options environment variable flag only for release.
- Vercel will fail but that's fine. We don't need a Vercel instance of the `release` branch.


bigfishdesign13 (approved):
Yes.

PR #1215, opened 2022-12-15
Release v1.3.1

bigfishdesign13:
### Adds

- Adds the optional `isClearableCallback` prop to the `TextInput` component. This prop works with the `isClearable` prop and the passed function is invoked when the "clear" button is clicked. This is most helpful for controlled `TextInput` component implementations.

### Updates

- Updates the background color for the `hover` state in the `secondary` variant of the `Button` component.

### Fixes

- Fixes the z-index value of the SVG icon in the `Select` component so it does not render above other elements on a consuming application's page.


EdwinGuzman (approved):

EdwinGuzman:
After merging, we have to make a manual update to `package.json` to remove the node_options flag in the `build-storybook:v1` script.

PR #1222, opened 2022-12-23
DSD-1234: adjustments for FilterBar spacing and alignment

bigfishdesign13:
Fixes JIRA ticket [DSD-1234](https://jira.nypl.org/browse/DSD-1234)

- This PR includes fine-tuning for the spacing, alignment and color styles in the `FilterBar`, `MultiSelect` and `MultiSelectGroup` components.
- The CHANGELOG was not updated because the updates in this PR are a continuation of the initial dev for the `FilterBar`, `MultiSelect` and `MultiSelectGroup` components.
- **Note:** The `UI Container` examples currently fail color contrast testing. This color issue is known and will be addressed in upcoming typography updates.

## How has this been tested?

- local build of Storybook
- Figma color testing

- Color choices were validated in Storybook and Figma.

### Checklist:

### Front End Review:


src/components/FilterBar/FilterBar.tsx
```diff
@@ -8,9 +8,10 @@ import {
   ModalFooter,
   ModalHeader,
   ModalOverlay,
+  Stack,
   useMultiStyleConfig,
-  Wrap,
-  WrapItem,
+  // Wrap,
```
EdwinGuzman: You can remove.

src/components/FilterBar/FilterBar.stories.mdx
```diff
@@ -117,12 +117,22 @@ The passed in width will take precedence over the default behavior.
...
   <DSProvider>
     <story.FilterBarLayoutStory />
   </DSProvider>
 </Canvas>
 
+### Examples With Containers
+
+<Canvas>
+  <DSProvider>
+    <story.FilterBarLayoutContainerStory />
```
EdwinGuzman: I think it would be good to show how the background element was created. It seems to be a simple `Box` with some styles. This pattern is part of the UI in a page in the Research Catalog so it can be a reusable pattern with code that can/should be copied and pasted (or perhaps an actual wrapper ds component?).

EdwinGuzman (commented):
Looks good but I think the container story should have more context, description, and code example.

src/components/FilterBar/FilterBar.stories.mdx
bigfishdesign13: I was considering making the wrapper (and the background treatment) part of the component itself.  Maybe a `wrapper` prop with the following values: 

- `none`: this would match what is currently implemented
- `default`: this would match the examples that I created
- `fullWidth`: this would allow the wrapper to span the full window width and keep the content within 1280px (similar to the Breadcrumbs component)

`fullWidth` would really only be applicable for the `row` layout, so maybe that's not the best choice.  Perhaps we just implement an `isWrapped` prop and allow the consuming app to place the `FilterBar` component accordingly in the `Template` layout.

What do you think?

EdwinGuzman: Yeah, that sounds good although that might overcomplicate this component? Is there a need for other components to have a similar looking background pattern? If it's just the FilterBar, then a `isWrapped` prop is fine. I was thinking if other components need a similar background pattern, then it could be a new component.

bigfishdesign13: OK, I understand and I like the idea of handling the UI pattern with a DS component.  That said, I think it would make the most sense to extend the functionlaity of the existing `ComponentWrapper` component to do this.

For now, I will update the examples here to show the current code that I have used.  I will also take a closer look at `ComponentWrapper` and see what it woudl take to extend the existing functionality and then create the necessary tickets.

EdwinGuzman (approved):

PR #1223, opened 2022-12-27
DSD-1242: remove aria-live from HelpErerrorText

bigfishdesign13:
Fixes JIRA ticket [DSD-1242](https://jira.nypl.org/browse/DSD-1242)

- Deprecates the `ariaLive` prop in the `HelperErrorText` component.
- The Jira ticket requests that the `ariaLive` prop be removed, but that would produce a breaking change.  In light of that, the `ariaLive` prop has been deprecated and the logic in the `HelperErrorText` component was updated to omit the `aria-live` attribute unless the `ariaLive` prop was explicitly passed.

## How has this been tested?

- local build of Storybook

- Due to how the `HelperErrorText` component is used within the DS, the `ariaLive` prop was found to be unnecessary.

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/HelperErrorText/HelperErrorText.stories.mdx
+++ b/src/components/HelperErrorText/HelperErrorText.stories.mdx
@@ -30,9 +30,6 @@ import DSProvider from "../../theme/provider";
       control: false,
       table: { defaultValue: { summary: true } },
     },
-    ariaLive: {
-      table: { defaultValue: { summary: "undefined" } },
-    },
     children: { table: { disable: true } },
     className: { control: false },
     id: { control: false },
@@ -68,7 +65,7 @@ import DSProvider from "../../theme/provider";
     name="HelperErrorText with Controls"
     args={{
       ariaAtomic: undefined,
-      ariaLive: "polite",
+      ariaLive: undefined,
       className: undefined,
       id: "helperErrorText-id",
       isInvalid: false,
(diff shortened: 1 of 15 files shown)


src/components/HelperErrorText/HelperErrorText.stories.mdx
```diff
@@ -30,7 +31,7 @@ import DSProvider from "../../theme/provider";
       table: { defaultValue: { summary: true } },
     },
     ariaLive: {
-      table: { defaultValue: { summary: "polite" } },
+      table: { defaultValue: { summary: "undefined" } },
```
EdwinGuzman: Should `undefined` be a string or just the type itself?

src/components/HelperErrorText/HelperErrorText.stories.mdx
```diff
@@ -87,10 +88,27 @@ import DSProvider from "../../theme/provider";
...
 
-In the "invalid" state when `isInvalid` is set to true, the text passed to the
-`invalidText` prop will be presented to screen readers be default. If you need
-to turn this off, set the `ariaLive` prop to `"off"`. This is used in DS "form"
-components for invalid form submissions or incorrect values entered.
+In the Reservoir Design System (DS), the `HelperErrorText` component is always
+associated with a related element by using the `aria-describedby` attribute in
+the associated element. In this way, the content of the `HelperErrorText` is
+made available to screenreaders.
+
```
EdwinGuzman, line 100: Maybe write more to explain the example and how the input field is associated with the helper text "This is error text :(".

EdwinGuzman (changes requested):
Just a minor comment.

src/components/HelperErrorText/HelperErrorText.tsx
```diff
@@ -44,11 +40,10 @@ export const HelperErrorText = chakra(
       ref?
     ) => {
       // Only announce the text in the invalid state.
-      const announceAriaLive = isInvalid;
       const styles = useStyleConfig("HelperErrorText", { isInvalid });
       const props = {
         "aria-atomic": ariaAtomic,
-        "aria-live": announceAriaLive ? ariaLive : "off",
+        "aria-live": ariaLive ? ariaLive : undefined,
```
oliviawongnyc: Do we need this ternary if the default parameter is already undefined? I think this could just be `"aria-live": ariaLive`.

src/components/HelperErrorText/HelperErrorText.stories.mdx
bigfishdesign13: I removed it altogether.

src/components/HelperErrorText/HelperErrorText.tsx
bigfishdesign13: Good catch.  That change still allows the component to function properly.

EdwinGuzman (approved):

PR #1224, opened 2022-12-27
DSD-1230: SR-only descriptive text for Link component

bigfishdesign13:
Fixes JIRA ticket [DSD-1230](https://jira.nypl.org/browse/DSD-1230)

- Updates the `Link` component to include descriptive text for screen readers in the component's `"external"` variant.

## How has this been tested?

- local build of Storybook

- The addition of the new descriptive text is for screen readers only.

### Checklist:

### Front End Review:

### DIFF
--- a/src/theme/components/link.ts
+++ b/src/theme/components/link.ts
@@ -6,6 +6,7 @@ import {
   primary,
   secondary,
 } from "./button";
+import { screenreaderOnly } from "./globalMixins";
 
 export const baseLinkStyles = {
   color: "ui.link.primary",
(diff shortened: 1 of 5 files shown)


src/theme/components/link.ts
```diff
@@ -101,13 +102,16 @@ const variants = {
...
+    /** This is needed for custom anchor elements or link components
+     * that are passed as children to the `Link` component. */
     a: {
       _hover: {
         color: "ui.link.secondary",
       },
     },
+    /** The span element will handle descriptive text added to aid
+     * screen readers. */
+    span: screenreaderOnly(),
```
EdwinGuzman: This seems fine but do you think we'll get some styles where a link may need a word that is bold or italicized and so the DOM will look like `<Link><span>check this link to<b>something</b> to read more</span></Link>`. Or something similar where spans are used inside Links? It may cause unintended consequences. Maybe target it with a specific id or classname.

EdwinGuzman (changes requested):
Just a minor concern.

oliviawongnyc (commented):
Interesting! Didn't know about the screenReaderOnly function. Will hold off on approving until Edwin's comment is addressed.

src/theme/components/link.ts
bigfishdesign13: Yes, to make this truely scalable, I will make some adjustments to target something unique.

EdwinGuzman (approved):

PR #1232, opened 2023-01-11
DSD-1279: improved a11y for CheckboxGroup, MultiSelectGroup and RadioGroup

bigfishdesign13:
Fixes JIRA ticket [DSD-1279](https://jira.nypl.org/browse/DSD-1279)

- Removes the `arial-label` attribute from the `CheckboxGroup`, `MultiSelectGroup` and `RadioGroup`  components for improved accessibility. Using the attribute is redundant with  the existing "screen reader only" `<legend>` element.

## How has this been tested?

- local build of Storybook

- Using the `aria-label` attribute on the noted components is redundant with  the existing "screen reader only" `<legend>` element that is also implemented in those components.

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/CheckboxGroup/CheckboxGroup.test.tsx
+++ b/src/components/CheckboxGroup/CheckboxGroup.test.tsx
@@ -60,7 +60,12 @@ describe("CheckboxGroup Accessibility", () => {
   });
   it("passes axe accessibility with the legend hidden", async () => {
     const { container } = render(
-      <CheckboxGroup id="checkboxGroup" labelText="Test Label" name="test2">
+      <CheckboxGroup
+        id="checkboxGroup"
+        labelText="Test Label"
+        name="test2"
+        showLabel={false}
+      >
         <Checkbox id="checkbox2" value="2" labelText="Checkbox 2" />
         <Checkbox id="checkbox3" value="3" labelText="Checkbox 3" />
         <Checkbox id="checkbox4" value="4" labelText="Checkbox 4" />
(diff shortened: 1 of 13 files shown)


src/components/CheckboxGroup/CheckboxGroup.test.tsx
```diff
@@ -58,6 +58,16 @@ describe("CheckboxGroup Accessibility", () => {
     );
     expect(await axe(container)).toHaveNoViolations();
   });
+  it("passes axe accessibility with the legend hidden", async () => {
```
EdwinGuzman, line 61: This is essentially the same test as the first one. Maybe we need more comments in the first one instead of repeating the test to explain what we are looking for in terms of a11y testing.

src/components/CheckboxGroup/CheckboxGroup.test.tsx
```diff
@@ -76,18 +86,15 @@ describe("Checkbox", () => {
...
         <Checkbox id="checkbox2" value="2" labelText="Checkbox 2" />
         <Checkbox id="checkbox3" value="3" labelText="Checkbox 3" />
         <Checkbox id="checkbox4" value="4" labelText="Checkbox 4" />
       </CheckboxGroup>
     );
-    expect(screen.getByTestId("checkbox-group")).not.toHaveAttribute(
-      "aria-label",
-      "Test Label"
-    );
+    expect(screen.getByText("Test Label")).toBeVisible();
```
EdwinGuzman: It might make sense to use the `container` and `querySelector("legend")` to find the specific DOM element and then test that it is on the DOM and that it has the correct text.

EdwinGuzman (commented):
This looks fine but maybe some things can be fine-tuned, not just for `CheckboxGroup` but also for `MultiSelectGroup` and `RadioGroup`.

src/components/CheckboxGroup/CheckboxGroup.test.tsx
isastettler: I agree that it might be better to check differently. `.toBeVisible()` was to me a little confusing at first.

isastettler (commented):
Looking good.

src/components/CheckboxGroup/CheckboxGroup.test.tsx
bigfishdesign13: Well, of course, I am an idiot and I did not set `showLabel` to `false` for this test.  How about now?

EdwinGuzman (approved):

PR #1237, opened 2023-01-17
DSD-1283: arial-label for Total Items Selected button

bigfishdesign13:
Fixes JIRA ticket [DSD-1283](https://jira.nypl.org/browse/DSD-1283)

- Updates the arial-label for the `Total Items Selected` button
- There is no CHANGELOG update associated with this change because this is continued development of the new `MultiSelect` component.
- **NOTE:** Some of the tests are failing, but I am not sure why.  The tests are failing due to the changes made on line 59 in `MultiSelectMenuButton.tsx`.  Specifically, the tests are fialing when the `multiSelectLabel` variable is used.  When that variable is added to the value for `selectedItemsAriaLabel`, the tests fail.  When that variable is not used on line 59, all tests pass.

## How has this been tested?

- local build of Storybook

- n/a

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/MultiSelect/MultiSelectMenuButton.tsx
+++ b/src/components/MultiSelect/MultiSelectMenuButton.tsx
@@ -56,7 +56,7 @@ const MultiSelectMenuButton = forwardRef<
   if (selectedItems[multiSelectId]?.items.length > 0) {
     getSelectedItemsCount = `${selectedItems[multiSelectId].items.length}`;
     const itemPlural = getSelectedItemsCount === "1" ? "" : "s";
-    selectedItemsAriaLabel = `${getSelectedItemsCount} item${itemPlural} selected`;
+    selectedItemsAriaLabel = `remove ${getSelectedItemsCount} item${itemPlural} selected from ${multiSelectLabel}`;
   }
   const styles = useMultiStyleConfig("MultiSelectMenuButton", {
     isOpen,
@@ -112,6 +112,7 @@ const MultiSelectMenuButton = forwardRef<
             marginLeft="xs"
             name="close"
             size="xsmall"
+            title="Remove selected items"
           />
         </Box>
       )}
--- a/src/components/MultiSelect/MultiSelect.test.tsx
+++ b/src/components/MultiSelect/MultiSelect.test.tsx
@@ -394,7 +394,7 @@ describe("MultiSelect Dialog", () => {
     // Check for the count of selectedItems
     expect(countButton).toHaveTextContent("3");
     // Close menu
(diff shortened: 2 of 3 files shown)


bigfishdesign13:
@EdwinGuzman and @isastettler 
Please see my comment in the PR description about the problems I am experiencing with the tests.  Thanks.

src/components/MultiSelect/MultiSelectMenuButton.tsx
```diff
@@ -112,6 +112,7 @@ const MultiSelectMenuButton = forwardRef<
             marginLeft="xs"
             name="close"
             size="xsmall"
+            title="Remove selected items"
```
EdwinGuzman, line 115: The entire span acts as a button and has an aria-label, but this SVG still needs its own title, right?

EdwinGuzman:
In the `"should render a count button with the correct count, should clear the selectedItems on click "` test, update:
```
userEvent.click(screen.getByRole("button", { name: /MultiSelect Label/i }));
```
to 
```
userEvent.click(screen.getByRole("button", { name: "MultiSelect Label" }));
```

The `/MultiSelect Label/i` regex is picking up the "MultiSelect Label" text inside of the span that has text such as `"remove 1 item selected from MultiSelect Label"`.

And this is because we're making the span act as a button with `role="button"`.

src/components/MultiSelect/MultiSelectMenuButton.tsx
bigfishdesign13: Yes, but also no.  Rather than leaving the default title, which is not very descriptive, I added something in line with the actual functionality.  However, we have do have `aria-hidden="true"` applied, so the title does not get seen by assistive technology.

bigfishdesign13:
@EdwinGuzman 
I did try that type of change when I was first working on this, but an error persisted where the test could not fine an element with the `role` of `button` and the `name` of `MultiSelect Label`.  When I analyzed the response more closely, it was showing that the button that the test is looking for has a `name` value of `Colors MultiSelect Label`.  One of the internal options has a label of `Colors` and it seems that RTL is concatenating those two strings into one `name` value.

To get the tests to work, I changed the string value that the test is checking to match the name that is being read by RTL: `Colors MultiSelect Label`.  Is that the best practice?  Probably not, but the tests are now all passing.

EdwinGuzman:
I think ideally we should have generic enough labels that we can query for them using RTL, but in this case this is a big component so the error encountered makes sense. I think this is fine for now and we can reassess during/after making more a11y updates as well.

EdwinGuzman (approved):

PR #1245, opened 2023-02-07
DSD: Add Focus Managment On MultiSelect component

isastettler:
Fixes JIRA ticket 1289 and 1290

- Adds focus management for after Apply or Close button is pressed in the Dialog MultiSelect
- Adds focus management for after the count button is pressed on the MultiSelectMenuButton
- Adds test for the new functionality
- Fixes a style bug on MultiSelectGroup mobile

- This PR should address the accessibility issue of focus loss after the MultiSelect component closes or the count button has been clicked. Now the focus should return to the MultiSelectMenuButton after any of these actions have been performed.

### Checklist:

### Front End Review:


src/components/MultiSelect/MultiSelect.test.tsx
```diff
@@ -23,9 +27,22 @@ const items = [
...
+    onChange,
+    onMixedStateChange,
+    selectedItems,
+    setSelectedItems,
+    onClear,
+  } = useMultiSelect();
+
+  React.useEffect(() => {
+    setSelectedItems(initialSelectedItems);
+  }, []);
```
bigfishdesign13, line 45: The following error is being flagged `GitHub Actions / Lint`:
```
React Hook React.useEffect has missing dependencies: 'initialSelectedItems' and 'setSelectedItems'. Either include them or remove the dependency array
```

bigfishdesign13 (commented):
Other than the issue that I noted, this looks good and functions well.  Before I approve, I just want to get your thoughts on that issue that I noted.

src/components/MultiSelect/MultiSelect.test.tsx
```diff
@@ -42,8 +59,17 @@ const MultiSelectTestDialogComponent = ({ multiSelectId }) => {
...
+const MultiSelectTestListboxComponent = ({
+  multiSelectId,
+  initialSelectedItems = {},
+}) => {
+  const { onChange, selectedItems, setSelectedItems, onClear } =
+    useMultiSelect();
+
+  React.useEffect(() => {
+    setSelectedItems(initialSelectedItems);
+  }, []);
```
EdwinGuzman, line 73: Similar to the above, dependencies need to be added to the dependency array.

isastettler: Hey @EdwinGuzman  and @bigfishdesign13 I saw that. However when I added the requested dependencies the MultiSelect tests were hanging. Same when I remove the array all together. I am not sure what I am missing.

EdwinGuzman (changes requested):
This works fine for the cases pointed out but when you go to the `MultiSelect` or `MultiSelectGroup` pages, the component gets focus right away. I don't think this is desired. For the group, the last component gets the focus. This might be confusing on a consuming app where the user goes to the page and the focus automatically goes to the `MultiSelect`. 

The `useEffect` where `isOpen` is being tracked may need to be modified or perhaps you can call it as a callback when `onMenuToggle` is called?

src/components/MultiSelect/MultiSelect.test.tsx
EdwinGuzman: Interesting. Do you only need to call this once when the component mounts? You can add the linting comment so that this warning is ignored. Try that.

isastettler: Yes this will only set an initial selectedItems object. It should only run at first render. I will add the lint ignore now. Thank you.

isastettler:
Good catch on the focus logic @EdwinGuzman! I decided on a check for the previous isOpen value to catch the first render. Since the onApply is not connected to the MenuButton logic I think this is the easiest way to get it to work when the "Apply" button is clicked.

EdwinGuzman:
Great!

bigfishdesign13 (commented):
The focus works well now.  Sorry I missed that the first time around.  Regarding the linting errors, the first error messages have been resolved, but I am seeing a new one now:
```
Check warning on line 586 in src/components/MultiSelect/MultiSelect.test.tsx

GitHub Actions / Lint

Unnecessarily computed property ["multiselect-dialog-test-id"] found
```

src/components/MultiSelect/MultiSelectMenuButton.tsx
```diff
@@ -50,6 +55,8 @@ const MultiSelectMenuButton = forwardRef<
   const iconType = isOpen ? "minus" : "plus";
   const growAnimation = `${grow} 150ms ease-out`;
 
+  const [prevIsOpen, setPrevIsOpen] = React.useState(isOpen);
```
EdwinGuzman, line 58: I haven't tested this but can you try testing to see if the focus works when `isDefaultOpen` is set to `true` manually? When this is false, the focus check works just fine. I'm curious if when it is open by default, does it still know where to focus when it is closed.

isastettler: Hey @EdwinGuzman. So two things: Manually tested it looks fine. It will have focus on the open MultiSelect (wether it is in a single MultiSelect or a Group) 
Second: should that prop maybe be something that can be changed in the controls in storybook? And in that case maybe isBlockElement too?

Screen Shot 2023-02-07 at 2 38 12 PM
Screen Shot 2023-02-07 at 2 42 16 PM

EdwinGuzman: Great, thanks for checking.

That prop in Storybook doesn't affect the component because they need to be set _before_ the component renders. Otherwise, updating it does nothing. This is hard to test in Storybook and I haven't found a good way to do so with Controls but if you do, you can add it. I usually add an example in the docs; in this case it would be a `MultiSelect` that is already opened on the page when you go to it. That's not necessary for this update and we can create a new ticket for that if you want.

EdwinGuzman (approved):

bigfishdesign13:
@isastettler 
Before this is merged into `development`, let me have Clare do an a11y review in the Vercel preview.  I'll let you know how it goes.

bigfishdesign13:
> @isastettler
> Before this is merged into `development`, let me have Clare do an a11y review in the Vercel preview. I'll let you know how it goes.

@isastettler 
Clare also gives a 👍 on the updates.  Please go ahead and merge this PR into `development`.

PR #1257, opened 2023-03-07
DSD-1243: isUnderlined prop

oliviawongnyc:
Fixes JIRA ticket xxx

- Adds an `isUnderlined` prop to the `Link` component. If set to `true`, the link text will be underlined in all browser states. If `false`, the link text will only be underlined during hover.

## How has this been tested?

- Locally, in Storybook.

- No concerns. Updated Storybook's `Link` story to include a brief note on the new prop and how underlines are not a required accessibility measure for links.

### Checklist:

### Front End Review:


src/components/Link/Link.stories.mdx
```diff
@@ -81,6 +82,13 @@ interface action is required, a `Button` component should be used instead. The
 The `"external"` variant of the `Link` component includes `"screen reader only"`
 descriptive text to clarify that the link will open in a new tab/window.
 
+### isUnderlined Prop
+
+As mentioned in the [Links Accessibility Guide](/story/accessibility-guide-links--page),
+there is no explicit rule that dictates using an underline to indicate a link. The optional
+`isUnderlined` prop is true by default, but can be set to false to remove the underline in
+all browser states except for hover.
```
EdwinGuzman: Might be better to be explicit about which browser states are affected instead of "all".

src/theme/components/link.ts
```diff
@@ -101,8 +100,9 @@ const variants = {
 };
 const Link = {
   parts: ["srOnly"],
-  baseStyle: {
+  baseStyle: ({ isUnderlined }) => ({
```
EdwinGuzman: Add the default `= true` here.

EdwinGuzman (approved):
Looks good just minor non-block comments.

bigfishdesign13:
@oliviawongnyc, I got the same block of Typecheck errors earlier today when I updated a PR.  It didn't make sense because all tests were passing locally.  I pushed an empty commit to force the PR to rebuild and everything was fine that second time around.  In the end, I don't know why the errors came up nor why the tests passed during the second build.  

@EdwinGuzman, any insight on this?

EdwinGuzman:
I have not encountered that before...

PR #1265, opened 2023-03-15
DSD-1348: dark mode colors for Header

bigfishdesign13:
Fixes JIRA ticket [DSD-1348](https://jira.nypl.org/browse/DSD-1348)

- Updates overall dark mode colors for the `Header` component.
- NOTE: Nothing was added to the CHANGELOG because this is not worth noting.

## How has this been tested?

- local Storybook

- n/a

### Checklist:

### Front End Review:


src/theme/components/header/header.ts
```diff
@@ -8,6 +8,7 @@ export const headerRed = "#ED1C24";
 export const headerRedDarkMode = "#DC6F73";
 export const headerRedDonate = "#E32B31";
 export const headerYellow = "#FEE34A";
+export const headerYellowDark = "#403B2D";
```
EdwinGuzman, line 11: I added these colors because there were custom from the previous Header implementation and we wanted to have a DS version that was on parity. I think for this dark yellow color that we should use the DS design token instead of a variable like the rest. Perhaps the `headerRedDarkMode` variable should also really be a DS design token.

bigfishdesign13: The notification in the `Header` component is not the DS `Notification` component and therefore does not use the layout and treatments of our component.  I am trying to deal with the background color.  The `Notification` component does that, but here we still get overwhelmed by the background color.  We do not have a decent color for this in our dark mode palette other than the grays.  And if we just use a gray, then we lose the impact of any color.  The addition of `headerYellowDark` is a compromise until the `Header` is reworked.

EdwinGuzman: Gotcha. I forgot that the dark mode colors for the Header are _also_ custom.

EdwinGuzman (approved):

EdwinGuzman (approved):

PR #1270, opened 2023-03-17
DSD-1339: Fixing keyboard controls state update issue in the range Slider variant

EdwinGuzman:
Fixes JIRA ticket [DSD-1339](https://jira.nypl.org/browse/DSD-1339)

- Fixes an issue where keyboard controls used in the range `Slider` loops updating the internal state. This happened because both `onChange` and `onChangeEnd` props in Chakra's `RangeSlider` component get called. This is _not_ the problem. The problem occurs because we passed the `setCurrentValue` function to both prop functions and when both get called, the internal state tries to update. This is what caused the state update loop that causes the page to slow down or crash.
- The `setCurrentValue` function is removed from the `onChangeEnd` prop function but is kept in the `onChange` prop function. For the cases where a user wants to only use the `onChangeEnd` function, this change does not affect the original implementation. This is because Chakra will still call the `onChange` function, it just won't call our own internal `onChange` function since `onChangeEnd` was passed.

## How has this been tested?

In Storybook. Any of the range `Slider`s can be used to test. Just focus on either the first or second slider thumb and use keyboard controls to update the value.

- N/A the Slider still works as expected, or better, with keyboard controls.

### Checklist:

### Front End Review:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 ## Prerelease
 
+### Fixes
+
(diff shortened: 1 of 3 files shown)


bigfishdesign13 (approved):
Very interesting, but that makes sense.  I tested the component using keyboard controls and I am no longer seeing a CPU spike.

PR #1278, opened 2023-03-23
Release v1.5.1

bigfishdesign13:
### Updates

- Updates the `Button` component to use a transparent background for the default state of the "secondary" variant.
- Updates the `Hero` component so that the height of the image in the "campaign" and "fiftyFifty" variants will grow and shrink based on the text content.
- Updates the `SearchBar` component to include `defaultValue` as a prop in `textInputProps`.

### Removes

- Removes the `Color Mode` docs page from Storybook due to the addition of the `Supporting Dark Mode` docs page.

### Fixes

- Removes the `Slider` component's internal state update function (`setCurrentValue`) from Chakra's own `RangeSlider` component's `onChangeEnd` prop function. This fixes an issue where both Chakra's `onChange` and `onChangeEnd` prop functions were being called regardless of whether they were passed to the DS `Slider` component or not. Since Chakra's `onChange` function will always be called, the internal state function is kept. This ensures that there is no multiple calls or loop of state update calls when using keyboard controls to change the range slider's value.


EdwinGuzman:
Fixed merge conflicts and added a minor update to the Slider version.

EdwinGuzman (approved):

PR #1279, opened 2023-03-23
DSD-1338 POC: testing new tooltip to wrap around form elements

EdwinGuzman:
This is a quick test so unit tests will fail . Instead of doing 
```
<Tooltip ...>
  <Radio />
</Tooltip>
```
now the `Tooltip` component is rendered inside of the `Radio` component when it should render through the `tooltipLabelText` prop. When rendered, the `aria-describedby` will be added to the `input` html element instead of the wrapper div. This should fix a11y issues.

Only test the following section on this link https://nypl-design-system-git-tooltip-poc-update-nypl.vercel.app/?path=/docs/components-overlays-switchers-tooltip--tooltip-with-controls

Screen Shot 2023-03-23 at 6 30 48 PM

If it goes well, we can think through a better update for all component. This change will change how the `Tooltip` and any form component are used together so it is technically a breaking change. I'm a bit wary about releasing this soon but we should aim to bundle this with the next major bump (which is typography?).

### DIFF
--- a/src/components/Checkbox/Checkbox.tsx
+++ b/src/components/Checkbox/Checkbox.tsx
@@ -53,6 +53,7 @@ export interface CheckboxProps extends CheckboxIconProps {
   showLabel?: boolean;
   /** Populates the value of the input */
   value?: string;
+  tooltipLabelText?: any;
 }
 
 function CheckboxIcon(props: CheckboxIconProps) {
@@ -89,6 +90,7 @@ export const Checkbox = chakra(
       showHelperInvalidText = true,
       showLabel = true,
       value,
+      tooltipLabelText,
       ...rest
     } = props;
     const styles = useMultiStyleConfig("Checkbox", {});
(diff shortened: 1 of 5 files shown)


EdwinGuzman:
Added another pattern for rending the tooltip. We should discuss to see which one makes more sense.

1. Add `Tooltip` directly into the form component. For example, it is imported and implemented in `Radio`. This would add more complexity to the form component and will have to be done for every single one.
```
// Radio.tsx

<ComponentWrapper>
  {tooltipLabelText ? (
    <Tooltip content={tooltipLabelText}>
      <ChakraRadio />
    </Tooltip>
  : <ChakraRadio />}
</ComponentWrapper>
```

```
How to use:
<Radio ... tooltipLabelText="text" />
```

1. Add `Tooltip` directly into the `ComponentWrapper` component. This only needs to be done once and will affect all form components. At the same time, it opens it up to use the same pattern in non-form components such as `AudioPlayer`.
```
// ComponentWrapper.tsx

<Box>
  // ..
  {tooltipLabelText ? (
    <Tooltip content={tooltipLabelText}>
      {children}
    </Tooltip>
  : <>{children}</>}
</Box>
```

```
How to use:
<Radio ... tooltipLabelText="text" />
```

PR #1284, opened 2023-03-29
DSD-1390: Vite update

EdwinGuzman:
Fixes JIRA ticket [DSD-1390](https://jira.nypl.org/browse/DSD-1390)

This updates replaces TSDX with Vite for the build tool. This allows the Reservoir DS library to be easily maintained and allow individual npm packages to be updated for security patches and general bug fixes. At the same time, the build is now much faster. Specific updates and changes:

- Node 16 is the recommended version to use and it is set in the `.nvmrc` file.
- Github Actions are updated to use Node 16. The Typecheck workflow is removed since that's covered by the build script. Using Node 16 also allows the Vercel deployments to work without the the added `NODE_OPTIONS` flag in the npm build script.
- Vite is used to build the library and start a development build.
- Rollup plugins are removed in favor of Vite plugins.
- Gulp and the `resources.scss` file are removed. Other related `.scss` files were also removed since they are not used (in the library or in consuming applications).
- ESLint packages are updated.
- Jest is updated along with its config file. `jest-environment-jsom` and `ts-jest` are used now. Snapshot test output files have _slightly_ different rendered HTML syntax.
- The build is _still_ outputting CommonJS and ES module dist files, but the final name is slightly different. This does not affect how the DS is imported into any application.
- Storybook is updated to v6.5.16, which is a minor update from v6.5.13.
- Due to moving to ESM as the default for the repo, types are exported separated from components in the main `/src/index.ts` file. This doesn't affect the final build but just how components, interfaces, and types are exported.

## How to test

Make sure you have node 16 locally.

- Checkout the branch
- Run `nvm use`
- Delete `node_modules` and `package-lock.json`
- Run `npm i`
- Run all the npm scripts in `package.json`, such as `npm start`, `npm run build`, `npm test`, etc. They should all run and tests should pass.
  - When running `npm run build-storybook:v1`, make sure to then delete the `reservoir` folder as that should not be committed but also does not belong in the `.gitignore` file.
- Check the `dist` folder and make sure there are `design-system-react-components.cjs`, `design-system-react-components.js`, and `styles.css` files. Verify that there are type definitions in the `dist/src` directory.
- Verify the local storybook instance runs in the browser with `npm run storybook`.

## How has this been tested?

Locally.

If approved, we can merge this into the feature branch and create a release candidate. That npm version can be used to test in Turbine and other applications.

- N/A, no functional changes to components are made.

### Checklist:

### Front End Review:


gulpfile.js
hokei, line 1: why are we still using this gulpfile?  Do we still need to be merging sass?

EdwinGuzman: don't think we'll need it but just haven't gotten to that part yet.

EdwinGuzman: I don't think anyone uses it and I just want to remove it. Thoughts @bigfishdesign13 ?
This is for the "resources.scss" file that is optional for apps to import.

bigfishdesign13: @EdwinGuzman I feel this is a leftover artifact from when the DS was first built.  We have trasitioned away from CSS/SCSS and, like you noted, I don't think anyone is using this.  Would it worth asking on the `dev` channel if any projects are using this?  Or is there another method to determine if it's being used?

EdwinGuzman: Yeah, can ask on the `dev` channel. Also, we don't use it in Turbine which is our test/example app so if we don't use it ourselves, other apps probably don't either.

EdwinGuzman:
Ready for review. Note this is pointing to a feature branch.

bigfishdesign13:
Running `npm test` shows errors in the following files:
- `src/components/FilterBar/FilterBar.test.tsx` 
  - Unable to find an accessible element with the role "button"
  - Exceeded timeout of 5000 ms for a test
- `src/components/AudioPlayer/AudioPlayer.test.tsx`
  - accessibility

**NOTE:** I just ran `npm test` again and no errors were produced.

EdwinGuzman:
That's fun... I'm finding the same issue but it succeeds more than it fails.

bigfishdesign13 (approved):
All of the script commands work as expected.

isastettler (approved):
Looks good to me.

jackiequach (approved):
All npm scripts run as expected and local storybook instance works.

PR #1293, opened 2023-04-06
DSD-1397: Remove the Header and Footer components

EdwinGuzman:
Fixes JIRA ticket [DSD-1397](https://jira.nypl.org/browse/DSD-1397)

- This removes the `Header` and `Footer` components from the DS.
- They will be moved into the new `nypl-header-app` repo.

Questions:
- Something we haven't figured out yet is where the new documentation for the Header will go. It makes sense to go into the new `nypl-header-app` repo but we will lose the Storybook documentation style. It can probably just be part of a `.md` documentation file.
- No one is using the DS `Header` or `Footer` so this is technically not a breaking change. Should we "deprecate" these components before we remove them? Or should we just remove it entirely and not worry about a deprecation note?

## How has this been tested?

Locally.

- N/A

### Checklist:

### Front End Review:


bigfishdesign13:
> Something we haven't figured out yet is where the new documentation for the Header will go. It makes sense to go into the new nypl-header-app repo but we will lose the Storybook documentation style. It can probably just be part of a .md documentation file.

Reservoir Storybook is where, I hope, devs have become accustom to viewing the docs for shared components, so I think we should consider keeping the current `Header` docs in Reservoir Storybook, either listed under `Components`, as it is now, or moved under `Development Guide` or a new section altogether.

> No one is using the DS Header or Footer so this is technically not a breaking change. Should we "deprecate" these components before we remove them? Or should we just remove it entirely and not worry about a deprecation note?

I think we can skip the deprecation notices and simply remove them outright.

src/components/AccessibilityGuide/SkipNavigation.stories.mdx
```diff
@@ -20,8 +20,6 @@ the top of the page and is visually hidden until a user focuses on the link.
 In the case of NYPL applications, the skip navigation contains two links; the
 first link points to the main content of the page and the second link points to
 accessibility information on [NYPL.org/accessibility](https://nypl.org/accessibility).
-For most digitial applications on the NYPL.org platform, the [NYPL Header](https://github.com/NYPL/dgx-header-component)
```
bigfishdesign13, line 23: Can this refrence be kept, but reworded to refer to the new app that Jackie is working on?

bigfishdesign13 (commented):
I have one question here, but please also take a look at the standalone comment I left on the PR.

src/components/AccessibilityGuide/SkipNavigation.stories.mdx
EdwinGuzman: Yes, we can update it when it's ready.

EdwinGuzman:
> Reservoir Storybook is where, I hope, devs have become accustom to viewing the docs for shared components, so I think we should consider keeping the current Header docs in Reservoir Storybook, either listed under Components, as it is now, or moved under Development Guide or a new section altogether.

Yes, let's try this. I think under `Development Guide` should work. Right now we cannot do this because the current docs reference the _component_ but we'll have to re-write it so it references the embed script and how to use it. This should be a new ticket.

PR #1295, opened 2023-04-11
[DSD-1396] Allow external clearing of TextInput's value.

dgcohen:
Fixes JIRA ticket [DSD-1396](https://jira.nypl.org/browse/DSD-1396)

- It removes the check for "value" in the TextInput's useEffect. This check prevents the TextInput from being cleared externally with an empty string, which is falsey and prohibits an update to finalValue in state.
- It also allows finalValue to be undefined, effectively allowing TextInput to be uncontrolled if desired (as it is used in the FeedbackBox component for example).
- It removes finalValue from the useEffect's dependency array, which causes the useEffect to unnecessarily run on changes to finalValue.
- It adds optional chaining to a length check on final value in case it is undefined.

## How has this been tested?

I've manually tested the component and updated the test snapshots.

### DIFF
--- a/src/components/TextInput/TextInput.stories.mdx
+++ b/src/components/TextInput/TextInput.stories.mdx
@@ -53,7 +53,7 @@ import DSProvider from "../../theme/provider";
 | Component Version | DS Version |
 | ----------------- | ---------- |
 | Added             | `0.22.0`   |
-| Latest            | `1.5.0`    |
+| Latest            | `Prerelease`    |
 
 ## Table of Contents
 
--- a/src/components/TextInput/TextInput.stories.mdx
+++ b/src/components/TextInput/TextInput.stories.mdx
@@ -273,7 +273,7 @@ export const App = () => {
       isClearable
       isClearableCallback={() => setValue("")}
       labelText="What is your favorite color?"
-      onChange={undefined}
+      onChange={(e) => setValue(e.target.value)}
       placeholder="i.e. blue, green, etc."
       value={value}
     />
--- a/src/components/TextInput/TextInput.stories.mdx
+++ b/src/components/TextInput/TextInput.stories.mdx
@@ -273,7 +273,7 @@ export const App = () => {
       isClearable
       isClearableCallback={() => setValue("")}
       labelText="What is your favorite color?"
-      onChange={(e) => setValue(e.target.value)}
+      onChange={undefined}
       placeholder="i.e. blue, green, etc."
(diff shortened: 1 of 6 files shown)


EdwinGuzman (commented):
Before taking a closer look at the update, a few things are missing from the PR:
- in `TextInput.stories.mdx`, update the current version in the markdown table to "Prerelease",
- add or update any relevant tests, in this case a test that wraps the component and passes in an empty value will be great to have to verify this update works,
- `TextInput` is internally used in `DatePicker`, `FeedbackBox`, `SearchBar`, and `Slider`. The respective `.stories.mdx` files should be updated with the "prerelease" text for the current version. All those unit tests pass already but we need to verify they still work as intended. You can check each storybook documentation page to check.
- you can `.skip` the `Toggle` unit tests that are failing for now.

src/components/TextInput/TextInput.tsx
```diff
@@ -152,7 +152,7 @@ export const TextInput = chakra(
         value,
         ...rest
       } = props;
-      const [finalValue, setFinalValue] = useState<string>(value || "");
+      const [finalValue, setFinalValue] = useState<string | undefined>(value);
```
EdwinGuzman: One thing about this update, I was reading this yesterday but did not finish the entire article: https://www.joshwcomeau.com/react/data-binding/

I was curious about the part here https://www.joshwcomeau.com/react/data-binding/#gotchas-4,
> When working with text inputs, be sure to use an empty string ('') as the initial state:

The explanation goes pretty in-depth.

I want to audit and take another pass at _all_ input components in the DS so that's why I was reading it but it is immediately relevant here.

bigfishdesign13 (approved):
This looks good.  Just make sure to set `Latest Version` on the `TextInput` story page to "Prerelease."

src/components/TextInput/TextInput.stories.mdx
```diff
@@ -273,7 +273,7 @@ export const App = () => {
       isClearable
       isClearableCallback={() => setValue("")}
       labelText="What is your favorite color?"
-      onChange={(e) => setValue(e.target.value)}
+      onChange={undefined}
```
EdwinGuzman: Why was this changed to undefined?

dgcohen: Committed this by accident. Fixed.

EdwinGuzman (approved):

PR #1308, opened 2023-04-28
DSD-1422: Storybook7 Breadcrumbs

EdwinGuzman:
Fixes JIRA ticket [DSD-1422](https://jira.nypl.org/browse/DSD-1422)

- Migrates the `Breadcrumbs` component documentation to Storybook 7.
- Also minor update to the `Accordion` doc for placement in the sidebar.

## How has this been tested?

Locally and Vercel preview, and compared against production: https://nypl.github.io/nypl-design-system/reservoir/v1/?path=/docs/components-navigation-breadcrumbs--breadcrumbs-with-controls

- N/A

### Checklist:

### Front End Review:


src/components/Breadcrumbs/Breadcrumbs.mdx
```diff
@@ -0,0 +1,81 @@
...
+- [Component Props](#component-props)
+- [Accessibility](#accessibility)
+- [Long Titles](#long-titles)
+- [Color Variations](#color-variations)
+
+## Overview
+
+<Description of={BreadcrumbsStories} />
+
+<Source
```
bigfishdesign13, line 33: @EdwinGuzman 
I see that you used the same page structure for `Accordion`, adding a code example right after the Overview.  If we are going to use that same structure for all pages, then I think we should create a new page section for the examples (ex: Code Example).  

However, with all the code examples that we include on some of the `docs` pages, I am not sure about adding something like this to every page.

What was your reason for adding this to both pages?  And what do you think about adding a new page section to all of the `docs` pages?

src/components/Breadcrumbs/Breadcrumbs.stories.tsx
```diff
@@ -0,0 +1,103 @@
...
+  },
+};
+
+// The following are additional Breadcrumbs example Stories.
+export const LongText: Story = {
+  render: () => <Breadcrumbs breadcrumbsData={breadcrumbsLongTextData} />,
+};
+export const ColorVariations: Story = {
+  render: () => (
+    <VStack align="stretch" spacing="s">
```
bigfishdesign13: The spacing in this `VStack` could be better.  Maybe wrap each `Heading/Breadcrumbs` pair in a `div` and then bump up the spacing to `l`?

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -6,18 +6,21 @@ import {
...
 
-export type BreadcrumbsTypes =
-  | "blogs"
-  | "booksAndMore"
-  | "brand"
-  | "education"
-  | "locations"
-  | "research"
-  | "whatsOn";
+export const breadcrumbTypeArray = [
```
bigfishdesign13, line 14: `BreadcrumbsTypes` is exported in `index.ts`.  That should be updated.  And I am curious how this type of change will affect the use of that export.

bigfishdesign13 (changes requested):
It works, but I have some questions and changes.

src/components/Breadcrumbs/Breadcrumbs.mdx
EdwinGuzman: This and `Accordion` both have a `[name]Data` prop. The code example was for the data structure for the prop and not the component itself. Other components won't need this.

src/components/Breadcrumbs/Breadcrumbs.tsx
EdwinGuzman: It shouldn't affect the export. The type is still exported and is the same as before, but how it was created was updated.

bigfishdesign13 (approved):
Looks good!

PR #1310, opened 2023-05-01
DSD-1420: AudioPlayer in Storybook 7

bigfishdesign13:
Fixes JIRA ticket [DSD-1420](https://jira.nypl.org/browse/DSD-1420)

- Updates the story for the `AudioPlayer` component to work in Storybook 7.

## How has this been tested?

- local Storybook

- n/a

### Checklist:

### Front End Review:


src/components/AudioPlayer/AudioPlayer.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+    className: { control: false },
+    id: { control: false },
+  },
+};
+
+export default meta;
+type Story = StoryObj<typeof AudioPlayer>;
+
+/**
+ * Main Story for the Accordion component. This must contains the `args`
```
EdwinGuzman: Update `Accordion` with `AudioPlayer`.

src/components/AudioPlayer/AudioPlayer.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { withDesign } from "storybook-addon-designs";
+
+import AudioPlayer from "./AudioPlayer";
+
+const meta: Meta<typeof AudioPlayer> = {
+  title: "Components/Media & Icons/AudioPlayer",
+  component: AudioPlayer,
+  decorators: [withDesign],
+  argTypes: {
+    audioType: {},
```
EdwinGuzman: Can you make this control a radio type with the four string types? That's how the current Controls displays this prop and the set of four values.

src/components/AudioPlayer/AudioPlayer.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+  parameters: {
+    design: {
+      type: "figma",
+      url: "https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Main?node-id=36835%3A25338",
+    },
+    jest: ["AudioPlayer.test.tsx"],
+  },
+};
+
+// The following are additional Accordion example Stories.
```
EdwinGuzman: Updating `Accordion` text.

EdwinGuzman (changes requested):
Minor updates.

EdwinGuzman (approved):
The four option update doesn't work but it doesn't work in production so technically this is still the same as before.

bigfishdesign13:
> The four option update doesn't work but it doesn't work in production so technically this is still the same as before.

@EdwinGuzman On its own, no, it looks like this prop is not working.  But it is.  There is validation that ensures that the embed code matches the 3rd party service specified in the `audioType` prop.  If you change this prop and then update the embed code with the corresponding code snippet, the audio player renders as expected.  I was confused by this at first myself.

PR #1311, opened 2023-05-02
DSD-1472: dark mode text color for Select

bigfishdesign13:
Fixes JIRA ticket [DSD-1472](https://jira.nypl.org/browse/DSD-1472)

- Updates the `Select` component to use a set text color value.

## How has this been tested?

- local Storybook

- n/a

### Checklist:

### Front End Review:


EdwinGuzman (approved):
I can't tell the difference but otherwise the rest is good.

bigfishdesign13:
> I can't tell the difference but otherwise the rest is good.

@EdwinGuzman This adds CSS that will resolve a color issue that only seems to be showing on MLN right now, so it makes sense that you could not see any difference rendered in Storybook.

PR #1319, opened 2023-05-04
Prep for Release v1.5.4

bigfishdesign13:

- Prep for Release `v1.5.4`


EdwinGuzman (approved):
The `Select` version should also be updated.

bigfishdesign13:
> The `Select` version should also be updated.

I had already changed the version number when I made the update.  It's good to go.

PR #1324, opened 2023-05-04
DSD-1437: Storybook 7 Hero documentation

EdwinGuzman:
Fixes JIRA ticket [DSD-1437](https://jira.nypl.org/browse/DSD-1437)

- Updates the `Hero` component documentation to Storybook 7.
- Compare with production documentation: https://nypl.github.io/nypl-design-system/reservoir/v1/?path=/docs/components-basic-elements-hero--hero-with-controls

## How has this been tested?

Locally.

- N/A

### Checklist:

### Front End Review:


bigfishdesign13 (commented):
Looks good, but I am wondering if we can/should enable the controls for all of the stories and only enable the props that are applicable for each `heroType` variant?

EdwinGuzman:
That's a good idea. How about we keep it as is for now since that's our current implementation and we have controls only in every "With Controls" story.

bigfishdesign13:
> That's a good idea. How about we keep it as is for now since that's our current implementation and we have controls only in every "With Controls" story.

But can you really turn them off in the other stories?  Right now, they are visible, but they don't affect how the component renders.

bigfishdesign13 (approved):

PR #1345, opened 2023-05-18
DSD-1457: Updating Storybook documentation for StructuredContent component

EdwinGuzman:
Fixes JIRA ticket [DSD-1457](https://jira.nypl.org/browse/DSD-1457)

- Updates the Storybook documentation for the `StructuredContent` component to version 7.
- Compare with prod: https://nypl.github.io/nypl-design-system/reservoir/v1/?path=/docs/components-page-layout-structuredcontent--structured-content-with-controls

## How has this been tested?

Locally.

- N/A

### Checklist:

### Front End Review:


src/components/StructuredContent/StructuredContent.stories.tsx
```diff
@@ -0,0 +1,512 @@
...
+        }
+      />
+    </>
+  ),
+};
+export const ExampleWithThreeComponents: Story = {
+  render: () => (
+    <>
+      <StructuredContent
+        bodyContent="This is the body text"
```
bigfishdesign13: The original example didn't include `bodyContent`, so it was able to rely on the spacing from the callout to add space between the first and second `<StructuredContent>` components.  With `bodyContent` in there, the spacing is now gone.  Howeverr, the spacing issue can be corected by wrapping the `bodyContent` text in a `<p>` element.

bigfishdesign13 (changes requested):
One small adjustmenet.

src/components/StructuredContent/StructuredContent.stories.tsx
EdwinGuzman: The nice thing about this migration is that it is catching errors that did not show up in MDX. Yes, the original example did not include `bodyContent` but it is actually a required prop. So in this `.tsx` file, this came up as an error in the IDE. I can add the `<p>` element to fix this example, but I don't think it's safe to assume that devs will be able to wrap content in a `<p>` element if spacing is off. So perhaps this is a bug that _we_ should fix?

bigfishdesign13: Yes, I will create a ticket and assign it to myself.

bigfishdesign13: Actually, no, we should not add default spacing after the component.  We don't do that for other components and we expect consuming apps to use soemthing like `VStack` to add the necessary spacing.

bigfishdesign13 (approved):
🥇

PR #1352, opened 2023-05-22
DSD-1443: Updating Storybook documentation for List and StyledList components

EdwinGuzman:
Fixes JIRA ticket [DSD-1443](https://jira.nypl.org/browse/DSD-1443)

- Updates the Storybook documentation for `List` and `StyledList` components to version 7.
- Compare with prod: https://nypl.github.io/nypl-design-system/reservoir/v1/?path=/docs/components-typography-styles-list--list-with-controls
- `StyledList`: https://nypl.github.io/nypl-design-system/reservoir/v1/?path=/docs/components-typography-styles-styledlist--styled-list-with-controls

## How has this been tested?

Locally.

- N/A

### Checklist:

### Front End Review:


src/components/StyledList/StyledList.stories.tsx
```diff
@@ -0,0 +1,130 @@
...
+    id: "nypl-styledlist",
+    listItems: itemGroups,
+    style: "capped",
+    textSize: "default",
+  },
+  render: (args) => <StyledList {...args} />,
+  parameters: {
+    design: {
+      type: "figma",
+      url: "https://www.figma.com/file/CRPxcR5Olmz50IrFZxvLKE/DRB---Search-Highlighting-(SFR-1519)?node-id=1223%3A11775",
```
bigfishdesign13: Replace this URL with the following:
```
https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Main?type=design&node-id=71684-76060&t=XbyIY1tCtH4Qkkfo-4
```

src/components/StyledList/StyledList.mdx
```diff
@@ -0,0 +1,100 @@
...
+  <li>
+    ALTERNATIVE TITLES: <i>"...Society of <b>New York</b> delivered at..."</i> + 1 more
+  </li>
+  <li>SUBJECTS: <i>"...<b>City of New York</b>, in the"</i> + 2 more</li>
+</StyledList>
+`}
+  language="jsx"
+/>
+
+<Canvas of={StyledListStories.LiChildren} />
```
bigfishdesign13: What do you think about removing this and the `listItems` prop stories?  Don't remove the docs.  Just remove the rendered story.  And then move the `Font Size` section above the `StyledList Children` section.

bigfishdesign13 (changes requested):
Maybe a few changes.  See my comments.

src/components/StyledList/StyledList.mdx
EdwinGuzman: Why this change? This means removing the rendered lists but still keep the code examples? Storybook doesn't have an easy way to not display a story in the sidebar when a story is exported in the `.stories.tsx` file.

But did move `Font Size` higher up.

bigfishdesign13: Those two stories on their own, without the associated docs, don't really mean much, so I think it would best to just we removed them from the sidebar.  But if there is no way to render the components on the docs page without also listing them in the sidebar, then leave it as is.  Just update the order of the stories in the stories file.  Right now the sidebar is not showing the same order as the Docs page.

bigfishdesign13 (approved):
Looks good.  Thanks.

PR #1361, opened 2023-05-31
DSD-1490: Storybook 7 doc links

EdwinGuzman:
Fixes JIRA ticket [DSD-1490](https://jira.nypl.org/browse/DSD-1490)

- Adds additional component code to render headings and links in MDX somewhat better. Clicking on the anchor link element won't add the hash to the URL and clicking on any link in a TOC also won't add the hash to the URL. But now it will jump to the correct location.
- Updating the hash is a general issue. I can't seem to get it to work and the documentation says one thing but that doesn't work. This is good for now and we can further update soon after.

## How has this been tested?

Locally.

- N/A

### Checklist:

### Front End Review:


bigfishdesign13:
The TOC links always open a new window.  And they open the story without the full Storybook interface.  Is that expected?

EdwinGuzman:
Yeah, I think that's expected. Storybook loads the main content inside an iframe so links open in the iframe and remove the sidebar. Something we can do is map the storybook `a` with something like ` <a href={href} target="_self">{children}</a>`.

The `target="_self"` will allow the anchored link to stay within the iframe and on the main storybook doc page. However, any regular links to an external page open with in the iframe:
Screen Shot 2023-06-01 at 11 13 11 AM

Another approach is to update all the TOC from 

```
- [Overview](#overview)
- [Component Props](#component-props)
- ...
```

to (with additional styles and also the `Link` component does not accept a `target` prop...)

```
- <a href="#overview" target="_self">Overview</a>
- <a href="#component-props" target="_self">Component Props</a>
- <a href="#accessibility" target="_self">Accessibility</a>
- ...
```

Screen Shot 2023-06-01 at 11 16 21 AM


I do not mind redo-ing all the TOC to add explicit links. It's annoying but will make the user experience better. The problem still exists, however, where the hash is not added to the URL. That I can't figure out yet.

bigfishdesign13:
I think we need to make that custom change to the TOC links and add the `target` attribute.  Otherwise, the UX, as you noted, will be poor and very confusing.  Is there any way I can help with that?

When you say the `Link` component does not accept a `target` prop, which `Link` component are you referring to?  Ours or somethign offered by Storybook?  I recall that there is a Storybook link add-on, but I didn't think we were using it.

EdwinGuzman:
Ok, np. I can make the TOC update in this PR by EOD.

The documentation here: https://storybook.js.org/docs/react/writing-docs/mdx#linking-to-other-stories-and-pages does not seem to work for me which makes this very confusing. There are some open issues regarding others encountering this as well. The workarounds do not work for me.

Our own `Link` component does not seem to take the `target` prop. That can be a minor separate PR and ticket. Maybe it can just be bundled into this PR? I don't mind making that update. It could update the suggested TOC approach to

```
- <Link href="#overview" target="_self">Overview</Link>
```

This will also help with styling now. So how about I add that to the `Link` component?

bigfishdesign13:
> Ok, np. I can make the TOC update in this PR by EOD.
> 
> The documentation here: https://storybook.js.org/docs/react/writing-docs/mdx#linking-to-other-stories-and-pages does not seem to work for me which makes this very confusing. There are some open issues regarding others encountering this as well. The workarounds do not work for me.
> 
> Our own `Link` component does not seem to take the `target` prop. That can be a minor separate PR and ticket. Maybe it can just be bundled into this PR? I don't mind making that update. It could update the suggested TOC approach to
> 
> ```
> - <Link href="#overview" target="_self">Overview</Link>
> ```
> 
> This will also help with styling now. So how about I add that to the `Link` component?

Yes, that would be great.  Thanks.

EdwinGuzman:
Updated.

1. TOC now looks like this:

```
- {<Link href="#overview" target="_self">Overview</Link>}
- {<Link href="#component-props" target="_self">Component Props</Link>}
- {<Link href="#accessibility" target="_self">Accessibility</Link>}
```

It was the easiest way to make it look right.

2. Internal links have been updated. Now links are relative rather than absolute. To do this, the URL has to be prefixed with `../` along with the id of the story or doc url.

Before:
```
[HelperErrorText Storybook documentation](https://nypl.github.io/nypl-design-system/reservoir/v1/?path=/docs/components-content-display-helpererrortext--helper-error-text-with-controls)
```

Now:
```
[HelperErrorText Storybook documentation](../?path=/docs/components-content-display-helpererrortext--docs)
```

bigfishdesign13 (approved):
I like relative paths over absolute paths, so I welcome this change.  It looks like you got everything, but we can patch anything that was missed if and when they surface.

PR #1362, opened 2023-06-08
Release v1.6.0

EdwinGuzman:
Fixes JIRA ticket [DSD-1390](https://jira.nypl.org/browse/DSD-1390)

### Vite Foundation Update

This updates replaces TSDX with Vite for the build tool. This allows the Reservoir DS library to be easily maintained and allow individual npm packages to be updated for security patches and general bug fixes. At the same time, the local and production build process are now much faster. Specific updates and changes:

- Node 16 is the recommended version to use and it is set in the `.nvmrc` file.
- Github Actions are updated to use Node 16. The Typecheck workflow is removed since that's covered by the build script. Using Node 16 also allows the Vercel deployments to work without the the added `NODE_OPTIONS` flag in the npm build script.
- Vite is used to build the library and start a development build.
- Rollup plugins are removed in favor of Vite plugins.
- Gulp and the `resources.scss` file are removed. Other related `.scss` files were also removed since they are not used (in the library or in consuming applications).
- ESLint packages are updated.
- Jest is updated along with its config file. `jest-environment-jsom` and `ts-jest` are used now. Snapshot test output files have _slightly_ different rendered HTML syntax.
- The build is _still_ outputting CommonJS and ES module dist files, but the final name is slightly different. This does not affect how the DS is imported into any application.
- Due to moving to ESM as the default for the repo, types are exported separated from components in the main `/src/index.ts` file. This doesn't affect the final build but just how components, interfaces, and types are exported.
- Storybook is updated to 7.0.18 which is a major version.
- All Storybook component documentation have been updated to use the Storybook 7 CSF format, along with MDX for the custom documentation pages.

### Adds

- Adds the `target` prop to the `Link` component.

### Removals

- Removes the `Header` and `Footer` components from the DS. These components will be moved and maintained in the [nypl-header-app](https://github.com/NYPL/nypl-header-app).


bigfishdesign13 (approved):
1.6.0!

PR #1368, opened 2023-06-15
DSD-1301: Updates placeholder image source

oliviawongnyc:
Fixes JIRA ticket [DSD-1301](https://jira.nypl.org/browse/DSD-1301).

- Updates placeholder images to come from placekitten, rather than placeimg, which is being deprecated.

## How has this been tested?

- Locally in Storybook
- With unit tests

- None.

### Checklist:

### Front End Review:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 ### Updates
 
 - Updates the `Button` component to increase the internal spacing between the button text and an icon.
+- Updates placeholder images to come from placekitten, rather than placeimg, which is being deprecated.
 
 ## 1.6.0 (June 8, 2023)
 
--- a/src/components/Card/Card.stories.tsx
+++ b/src/components/Card/Card.stories.tsx
@@ -74,7 +74,7 @@ export const WithControls: Story = {
     "imageProps.component": undefined,
     "imageProps.isAtEnd": false,
     "imageProps.size": "default",
-    "imageProps.src": "https://placeimg.com/400/300/animals",
+    "imageProps.src": "http://placekitten.com/400/300",
     isAlignedRightActions: false,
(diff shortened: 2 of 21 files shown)


EdwinGuzman (approved):
🐱 🐈 😸

bigfishdesign13 (commented):
I see that the URL connection protocols were changed form `http` to `https`.  I assume that was because the examples on the placekitten.com homepage use `http`.  But what I find interesting is that when you click on those examples, the images that come up use `https`.

Our site is now throwing a `Mixed Content` warning becuase our domain is using `https` and we are loading images that use `http`.  If the `https` protocol is available from placekitten.com, we should use that protocol to load the images.

@EdwinGuzman Thoughts?

oliviawongnyc:
> I see that the URL connection protocols were changed form `http` to `https`. I assume that was because the examples on the placekitten.com homepage use `http`. But what I find interesting is that when you click on those examples, the images that come up use `https`.
> 
> Our site is now throwing a `Mixed Content` warning becuase our domain is using `https` and we are loading images that use `http`. If the `https` protocol is available from placekitten.com, we should use that protocol to load the images.
> 
> @EdwinGuzman Thoughts?

@bigfishdesign13, I think that makes sense, but will wait for Edwin to respond!

EdwinGuzman:
Yeah, that's true. I didn't notice that at first. The URLs can be updated to be protocol agnostic from `http://placekitten.com/400/300` to `//placekitten.com/400/300`. Try that, I might've gotten the syntax wrong. That should pick up whatever protocol is being used on the current page that is serving the image.

bigfishdesign13:
@EdwinGuzman Yes, what you have there is the correct way to write a `protocol-relative URL`.  Let's use that.

EdwinGuzman (approved):

bigfishdesign13 (approved):
Looks good!

PR #1381, opened 2023-07-03
DSD-1265/1269: Adding screenreaderOnlyText prop to the Link and Button components

EdwinGuzman:
Fixes JIRA ticket [DSD-1265](https://jira.nypl.org/browse/DSD-1265) and [DSD-1269](https://jira.nypl.org/browse/DSD-1265)

- Adds `screenreaderOnlyText` prop to the `Link` and `Button` component. This adds the passed text to the component's DOM and visually hides but screenreaders can still read it.

## How has this been tested?

Locally.

- Tested this with VO but should be further reviewed.

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -1,7 +1,8 @@
 import {
+  Box,
   Button as ChakraButton,
   chakra,
-  useStyleConfig,
+  useMultiStyleConfig,
 } from "@chakra-ui/react";
 import React, { forwardRef } from "react";
 
@@ -37,6 +38,8 @@ interface ButtonProps {
   mouseDown?: boolean;
   /** The action to perform on the `<button>`'s onClick function. */
   onClick?: (event: React.MouseEvent | React.KeyboardEvent) => void;
+  /** Visibly hidden text that will only be read by screenreaders. */
+  screenreaderOnlyText?: string;
   /** The size of the `Button`. */
   size?: ButtonSizes;
   /** The HTML button type attribute. */
@@ -57,6 +60,7 @@ export const Button = chakra(
         isDisabled = false,
         mouseDown = false,
         onClick,
+        screenreaderOnlyText,
         size = "medium",
         type = "button",
         ...rest
@@ -65,7 +69,7 @@ export const Button = chakra(
       let childCount = 0;
       let hasIcon = false;
       let variant: string | ButtonTypes = buttonType;
-      let styles = {};
+      let styles: any = {};
 
       if (!id) {
         console.warn(
@@ -89,7 +93,7 @@ export const Button = chakra(
         variant = "iconOnly";
       }
 
-      styles = useStyleConfig("Button", { variant, buttonSize: size });
+      styles = useMultiStyleConfig("Button", { variant, buttonSize: size });
 
       return (
         <ChakraButton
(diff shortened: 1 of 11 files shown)


src/components/Link/Link.mdx
```diff
@@ -47,6 +47,11 @@ the relation names "nofollow", "noopener" and "noreferrer".
 
 <Canvas of={LinkStories.Accessibility} />
 
+The `Link` component also has a `screenreaderOnlyText` prop that can be used to
+add additional text to the link that is only visible to screen readers. This can
+be used to provide additional context to the link when the text is short, such
+as visible "Read more..." text.
```
bigfishdesign13, line 53: This should also be added to the story for `Button`.

src/components/Button/Button.tsx
```diff
@@ -103,6 +107,11 @@ export const Button = chakra(
           __css={styles}
           {...rest}
         >
+          {screenreaderOnlyText && (
+            <Box as="span" __css={styles.screenreaderOnly}>
+              {screenreaderOnlyText}
+            </Box>
+          )}
```
bigfishdesign13, line 115: The screen reader only text should be placed after the standard button content.  In fact, it would be best if it was immediatelt folloowing the text and before an icon.  However, I am not sure that is possible with the way we currently compose the button text and icon.

src/components/Link/Link.tsx
```diff
@@ -190,20 +197,27 @@ export const Link = chakra(
...
-              ref,
-              rel,
-              target: internalTarget,
-            },
-            [childrenToClone.props.children]
+        <>
+          {screenreaderOnlyText && (
+            <Box as="span" __css={styles.screenreaderOnly}>
+              {screenreaderOnlyText}
+            </Box>
```
bigfishdesign13: Same thing here regarding the positioning.

bigfishdesign13 (changes requested):
I have some questions about the repositioning of the sr only element.

src/components/Link/Link.mdx
EdwinGuzman: A similar paragraph or the example? There is a similar paragraph already but I'll add the example.

src/components/Button/Button.tsx
EdwinGuzman: I'll move it after the entire `children`. It'd be more work to insert this element between the text and the icon, but possible. The real issue, I think, is that the order is not always "text -> icon". A dev can pass in "icon -> text" so inserting in between those elements will not always provide the ideal placement.

src/components/Link/Link.tsx
EdwinGuzman: Yes. Similar issue happens here that is mentioned for the `Button` component above. The children are not always "text -> Icon".

src/components/Link/Link.mdx
bigfishdesign13: I was referring to the paragraph, but I must have missed it.  Sorry.  And I see you adjusted it slightly in this update.

src/components/Button/Button.tsx
bigfishdesign13: Yes, understood.  We'll go with these latest updates.

bigfishdesign13 (approved):
Yes.

bigfishdesign13:
Just make sure to update the `Latest Version` on the story pages.

EdwinGuzman:
I forgot to update the latest version on all my recent PRs 😨

PR #1387, opened 2023-07-21
DSD-1531: better ARIA for Heading

bigfishdesign13:
Fixes JIRA ticket [DSD-1531](https://jira.nypl.org/browse/DSD-1531)

- Adds the `role` prop to the `Text` component.
- Updates the ARIA attributes for the `overline` and `subtitle` elements in the `Heading` component.

## How has this been tested?

- local Storybook

- The addition of the new ARIA attributes will improve a11y for screen readers.

### Checklist:

### Front End Review:


src/components/Heading/Heading.tsx
```diff
@@ -256,7 +256,12 @@ export const Heading = chakra(
       const finalContent = (
         <>
           {overline && (
-            <Text mb="xxs" size={overlineSize}>
+            <Text
+              aria-roledescription="subtitle"
```
EdwinGuzman: Both the overline and subtitle have the same `aria-roledescription` of "subtitle"?

src/components/Text/Text.tsx
```diff
@@ -30,6 +30,8 @@ export interface TextProps {
   isLowercase?: boolean;
   /** Optional prop used to remove default spacing */
   noSpace?: boolean;
+  /** Optional prop used to explicitly set the ARIA role */
+  role?: string;
```
EdwinGuzman, line 34: `role` is an HTML attribute that should be passed by default by Chakra. It is most likely in the `...rest` spread prop value. I'm assuming we want to explicitly make this a prop for devs to know that it _should_ be updated?

EdwinGuzman (changes requested):
Questions and update component version to prerelease.

src/components/Heading/Heading.tsx
bigfishdesign13: That was my initial intent. However, the value for `aria-roledescription` is not bound to specific values (like role), so yes, I will change this to align with the `overline` element.

src/components/Text/Text.tsx
bigfishdesign13: I added this explicitly because VS Code was saying the `role` prop was not there by default and it was throwing a typescript error.  I would rather not add this as an actual component prop, but this was the only way I knew to get around the error.  Is there another option?

bigfishdesign13: @EdwinGuzman Can you suggest another way of doing this?  Normally you don't need to set a role, as it should be an assumed property for `<p>` elements, but for some reason that fundamental `role` is not getting picked up by the a11y tests.  It only passes when it is explicitly set.

EdwinGuzman: I think there might be two issues here. The first is that we need to set the type correctly so Typescript knows that it's a native HTML attribute. The other is that `"paragraph"` is not in the list of role values. That is probably why it's not getting picked up by a11y test. I am looking here https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles

I'm taking a closer look at the typescript issue but maybe it's fine to keep for now as an explicit prop.

bigfishdesign13: OK, thanks.  For some clarity, adding `role="paragraph"` was something that Clare had suggested and I know she did do some research on that.

EdwinGuzman (approved):

PR #1392, opened 2023-08-10
DSD-1308: standalone link variant

bigfishdesign13:
Fixes JIRA ticket [DSD-1308](https://jira.nypl.org/browse/DSD-1308)

- Adds the `"standalone"` variant to the `Link` component.
- Adds `NYPL Patterns` to the `Links Accessibility Guide`.

## How has this been tested?

- local Storybook

- Style changes and recommended patterns have been updated based on UX exploration and input from a11y consultant.

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/AccessibilityGuide/Links.mdx
+++ b/src/components/AccessibilityGuide/Links.mdx
@@ -50,7 +50,74 @@ format are generally considered to be fine without an underline.
 
 ## NYPL Patterns
 
-TBD
+Text links can be presented in three general scenarios: in a heading, mixed
+within a block of plain text or as an isolated navigation element. In all
+instances, the aim is to reduce the visual impact of the link underline while
+still meeting accessibility requirements and best practices.
+
+In cases where an underline for a text link is necessary, the 1px width and
+dotted underline styles will allow text links to be visually highlighted without
+being overly intrusive. This, in turn, will help text- and link-heavy pages to
+be less overwhelming and cluttered.
+
+### Headings
+
+When a link is applied to a full heading, the link text SHOULD be rendered using
+the `link primary` color and the default state of the link text SHOULD NOT be
+underlined. The link text hover state SHOULD be rendered using the `link secondary` color and MUST be underlined. Applying a link to a full heading is
+the recommended pattern.
+
+When a link is applied to a subset of words within a heading, the link text MUST
+be rendered using the `link primary` color and the default state of the link
+text MUST be underlined. The link text hover state SHOULD be rendered using the
+`link secondary` color and MUST be underlined. While this is acceptable,
+applying a link to a potion of a heading is not the recommended pattern.
+
+### Mixed Text
+
+When a body text link sits within a block of plain (unlinked) text &mdash; a
+paragraph, a single setence, or even just a few words &mdash; the link text
+SHOULD be rendered using the `link primary` color and the default state of the
+link text MUST be underlined. The link text hover state SHOULD be rendered using
+the `link secondary` color and MUST be underlined.
+
+### Standalone Links
+
+When a body text link sits by itself within the page content area &mdash; a Read
+More link at the end of a short description, a View In Catelog link on a book
+detail page &mdash; the link text SHOULD be rendered using the `link primary`
+color, the default state of the link text SHOULD NOT be underlined, and the link
+text SHOULD be displayed with a "forward" directional arrow rendered using the
+`link primary` color. The link text hover state SHOULD be rendered using
+the `link secondary` color and MUST be underlined.
+
+### Tips for Writing Standalone Link Text
+
+Avoid using a URL as your text link - URLs, especially long ones, may cause
+reading difficulty for both humans and screen readers.
+
+Keep link text as short as possible - Instead of "you can contact us,"" use "contact
+us.""
+
+Link at least one full word (two is better) - Avoid using a single letter or
+a symbol for a link as it lacks context and is difficult to click on.
+
+### Acceptable Style Deviations
+
+There are times when deviations from the standard link styles are allowed.
+
+Links in main navigation areas or in a common list format can be presented
+without an underline or some other link indicator (ex. text with icon). It is
+also acceptable to use a color other than the standard link color, but it is
+generally recommended to use black or white.
+
+Links on colored backgrounds will almost always run into problems with color
+contrast. In those cases, the link color should be set to either black or
+white, depending on the background color.
+
+In situations when the link color is not used and the link color is changed to
+white or black, an underline must be applied whenever a link it mixed with plain
+text. This mirrors the patterns of the standard link styles.
 
 ## Resources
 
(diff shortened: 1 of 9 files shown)


src/components/Link/Link.mdx
```diff
@@ -36,9 +36,10 @@ import * as LinkStories from "./Link.stories";
 
 ## Accessibility
 
-The `Link` component should be used for navigation. If an `onClick` user
-interface action is required, a `Button` component should be used instead. The
-`Link` component renders an `<a>` element with the `href` attribute.
+The `Link` component renders an `<a>` element with the `href` attribute, so the
+`Link` component should be used for navigation. If an `onClick` user interface
+action is required, a `Button` component should be used instead. In simple
+terms, a link goes somewhere and a button does something.
```
EdwinGuzman: I think we should mention that it still is possible to pass an `onClick` to the link component because there are some very specific use cases and it is also a native HTML pattern that is allowed (but the HTML attribute is `onclick` all lowercase). We just really urge against it unless they know what they are doing.

src/components/Link/Link.tsx
```diff
@@ -118,6 +122,30 @@ function getExternalExtraElements(
   );
 }
 
+function getStandaloneIcon(children: JSX.Element, linkId: string) {
+  const iconId = `${linkId}-icon`;
```
EdwinGuzman: The `Link`'s id prop is optional so this could turn into `"undefined-icon"`. We need a default or a better way to create this id.

src/components/Link/Link.tsx
```diff
@@ -118,6 +122,30 @@ function getExternalExtraElements(
...
+  const extraElements = (
+    <>
```
EdwinGuzman: You don't need to wrap this in a fragment. Unless if it throws an error?

src/components/Link/Link.tsx
```diff
@@ -118,6 +122,30 @@ function getExternalExtraElements(
...
+      <Icon
+        align={"right"}
+        className="more-link"
```
EdwinGuzman: Not a big deal but I think we use this only for tests, right? Maybe this should be `data-testid` for our unit tests.

src/components/Link/Link.tsx
```diff
@@ -18,6 +18,7 @@ export const linkTypesArray = [
   "default",
   "external",
   "forwards",
+  "standalone",
```
EdwinGuzman, line 22: In the Link storybook doc, there's no example of this new variant... but also there are no real example of the other variants so this can also be updated in a separate PR.

src/components/AccessibilityGuide/Links.mdx
```diff
@@ -50,7 +50,74 @@ format are generally considered to be fine without an underline.
...
+When a link is applied to a full heading, the link text SHOULD be rendered using
+the `link primary` color and the default state of the link text SHOULD NOT be
+underlined. The link text hover state SHOULD be rendered using the `link secondary` color and MUST be underlined. Applying a link to a full heading is
+the recommended pattern.
+
+When a link is applied to a subset of words within a heading, the link text MUST
+be rendered using the `link primary` color and the default state of the link
+text MUST be underlined. The link text hover state SHOULD be rendered using the
+`link secondary` color and MUST be underlined. While this is acceptable,
+applying a link to a potion of a heading is not the recommended pattern.
```
EdwinGuzman: Can this be bold to highlight that it's not recommended?

src/components/AccessibilityGuide/Links.mdx
```diff
@@ -50,7 +50,74 @@ format are generally considered to be fine without an underline.
...
+
+### Mixed Text
+
+When a body text link sits within a block of plain (unlinked) text &mdash; a
+paragraph, a single setence, or even just a few words &mdash; the link text
```
EdwinGuzman: minor typo `setence`.

src/components/AccessibilityGuide/Links.mdx
```diff
@@ -50,7 +50,74 @@ format are generally considered to be fine without an underline.
 
 ## NYPL Patterns
 
-TBD
+Text links can be presented in three general scenarios: in a heading, mixed
+within a block of plain text or as an isolated navigation element. In all
```
EdwinGuzman: nitpick: comma after "text" and before "or".

src/components/AccessibilityGuide/Links.mdx
```diff
@@ -50,7 +50,74 @@ format are generally considered to be fine without an underline.
...
+When a body text link sits within a block of plain (unlinked) text &mdash; a
+paragraph, a single setence, or even just a few words &mdash; the link text
+SHOULD be rendered using the `link primary` color and the default state of the
+link text MUST be underlined. The link text hover state SHOULD be rendered using
+the `link secondary` color and MUST be underlined.
+
+### Standalone Links
+
+When a body text link sits by itself within the page content area &mdash; a Read
+More link at the end of a short description, a View In Catelog link on a book
```
EdwinGuzman: minor typo `Catelog`

src/components/AccessibilityGuide/Links.mdx
```diff
@@ -50,7 +50,74 @@ format are generally considered to be fine without an underline.
...
+`link primary` color. The link text hover state SHOULD be rendered using
+the `link secondary` color and MUST be underlined.
+
+### Tips for Writing Standalone Link Text
+
+Avoid using a URL as your text link - URLs, especially long ones, may cause
+reading difficulty for both humans and screen readers.
+
+Keep link text as short as possible - Instead of "you can contact us,"" use "contact
+us.""
```
EdwinGuzman: extra `"` at the end of both examples.

src/components/AccessibilityGuide/Links.mdx
```diff
@@ -50,7 +50,74 @@ format are generally considered to be fine without an underline.
...
+
+Link at least one full word (two is better) - Avoid using a single letter or
```
EdwinGuzman: I think these can be two sentences instead of having the hyphen.

EdwinGuzman (changes requested):
Looks good overall but some visual examples would be nice.

src/components/Link/Link.tsx
bigfishdesign13: I think that is leftover from the original `Link` component.  There is currently no mention of `more-link` anywhere outside of the `Link` files.  And even with that, it only seems necessary in the tests.  I will remove from the Story page, but for now I am going to leave it in the tests.

src/components/Link/Link.tsx
bigfishdesign13: I added a new `All Link Types` section.

src/utils/utils.ts
```diff
@@ -125,3 +125,9 @@ export const truncateText = (text: string, truncateTextLength: number = 60) => {
   const updatedText = text.substring(0, truncateTextLength - 1);
   return `${updatedText.substring(0, updatedText.lastIndexOf(" "))}...`;
 };
+
+/** Prepare a string for use in an ID or class attribute */
+export const sanitizeStringForAttribute = (str: string) => {
```
EdwinGuzman, line 130: I'm surprised we don't already have this.

EdwinGuzman (approved):
Looks good!

PR #1409, opened 2023-08-28
DSD-1559: link patterns in consuming components

bigfishdesign13:
Fixes JIRA ticket [DSD-1559](https://jira.nypl.org/browse/DSD-1559)

- Updates the `Breadcrumbs`, `Card`, `Heading`, `Pagination`, and `SkipNavigation` components to implement the `Typo2023` link patterns.
- NOTE: The changes in `SkipNavigation` are not yet apparent because there is a bug with the `isUnderlined` prop in the `Link` component.

## How has this been tested?

- local Storybook

- n/a

### Checklist:

### Front End Review:


src/components/Heading/Heading.tsx
```diff
@@ -123,6 +123,7 @@ export const Heading = chakra(
         isUppercase,
         isLowercase,
         noSpace,
+        url,
```
EdwinGuzman, line 126: I forgot but are we marking all components as "prerelease" for the 2.0 update or should we make it a point that some components did not have an update at all? I imagine even if it wasn't a direct update, all components should be bumped up to 2.0, right?

src/components/SkipNavigation/SkipNavigation.tsx
```diff
@@ -37,10 +37,15 @@ export const SkipNavigation = chakra(
         >
           <List inline noStyling type="ul">
             <li>
-              <Link href={target}>Skip to Main Content</Link>
+              <Link href={target} isUnderlined={false}>
```
EdwinGuzman, line 40: I still see these as underlined:
Screen Shot 2023-08-30 at 2 50 42 PM

EdwinGuzman (commented):
Some questions

src/components/SkipNavigation/SkipNavigation.tsx
EdwinGuzman: Made this comment before reading the description! ignore.

EdwinGuzman (approved):

src/components/Heading/Heading.tsx
bigfishdesign13: This is a question I have been asking myself as well.  And truthfully, I don't know the answer.  There definitely will be some components that are not changed as part of this update and therefore don't need to have the `Latest Version` bumped.  However, this is going to be a major version update, so shouldn't all the components within the release be marked as `v2.0`?

PR #1422, opened 2023-09-22
RENO-3681: NewsletterSignup Component

aarnold101:
Newest branch [Preview](https://nypl-design-system-oixxe9n85-nypl.vercel.app/?path=/docs/welcome--docs)


src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,227 @@
...
+ *
+ * The component can show four different views, depending on the state of the
+ * email submission: form, loading, confirmation, and error.
+ */
+export const NewsletterSignup = chakra(
+  forwardRef<any, NewsletterSignupProps>(
+    (
+      {
+        className,
+        confirmationText = "Thank you! You have successfully subscribed to our email updates! You can update your email subscription preferences at any time using the links at the bottom of the email.",
```
EdwinGuzman: Nitpick: can these longer sentences be constant variables assigned outside the component? Such as `const defaultConfirmationText = "Thank you.....";`

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,227 @@
...
+        isInvalidEmail = false,
+        newsletterSignupType = "whatsOn",
+        onSubmit,
+        title = "Sign Up for Our Newsletter!",
+        view = "form",
+        ...rest
+      },
+      ref?
+    ) => {
+      const viewType = view;
```
EdwinGuzman: Since we're not keeping track of the internal state anymore, you can use the `view` prop directly instead of assigning it to a new variable that also doesn't get updated.

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,227 @@
...
+      const [buttonClicked, setButtonClicked] = useState(false);
+      const [email, setEmail] = useState("");
+
+      // Hook into NYPL breakpoint
+      const { isLargerThanMobile } = useNYPLBreakpoints();
+      const focusRef = useRef<HTMLDivElement>(); // @todo Is this needed? It is a holdover from FeedbackBox
```
EdwinGuzman: This is need to focus on the view when it is updated. Since the component will no longer keep track of its view, this is technically not needed.

If the consuming app is now in full control of the view, it's hard for the component itself to trigger the updated focus. So perhaps we need to revisit this to better plan out how the consuming app will trigger a focus. 

I'm thinking you can pass a new prop like `focusRef` that is a React ref and in the consuming app you can call `focusRef.current.focus()` whenever the `view` prop is updated.  For example, if a fetch request is made in the consuming app and it is successful, you'd update the `view` prop to "confirmation" and then call `focusRef.current.focus()`. 

Another option can be to call `focusRef.current.focus()` whenever `view` is "confirmation" or "error" in the component. You have to stagger the call, or give it enough time, so that the confirmation or error view can render. Otherwise, you may be attempting to focus on an element that doesn't exist. If it already does exist and you focus to it but it doesn't render with the updated view in time, the screenreader may not pick up the changes.

Do both options make sense? We can chat about this tomorrow with @bigfishdesign13

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,227 @@
...
+      const styles = useStyleConfig("NewsletterSignup", {});
+      // The viewType may change as a result of something happening in the internalOnSubmit function.
+      const formView = viewType === "form";
+      const confirmationView = viewType === "confirmation";
+      const errorView = viewType === "error";
+      const iconColor = useColorModeValue(null, "dark.ui.typography.body");
+
+      const internalOnSubmit = (e) => {
+        e.preventDefault();
+        setButtonClicked(true);
```
EdwinGuzman: If we have a "loading" view, then the `buttonClicked` state is no longer needed. It is also currently not being updated to `false` and only to `true` so you'd only ever be able to submit the form once.

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,227 @@
...
+        >
+          {/* Open info div */}
+          <Stack direction={isLargerThanMobile ? "row" : "column"} id={"info"}>
+            <Box
+              bg={getSectionColors(newsletterSignupType, "primary")}
+              id={"color-box"}
+              width={isLargerThanMobile ? ".75rem" : null} // @todo Not sure how to adapt this into the theme.ts and have it work the same.
+              height={isLargerThanMobile ? null : ".75rem"} // @todo ibid.
+            >
+              &nbsp;
```
EdwinGuzman: _I think_ you might be able to get the same result by having an empty Box with a wide right (or left) border. I'm not sure if having an empty `&nbsp;` is best practice for a visual cue.

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,227 @@
...
+            </Box>
+            <VStack id={"pitch"}>
+              <Heading level={"h3"} text={title} />
```
EdwinGuzman: Just FYI and no need to update, but you can pass strings as props without the braces. So `level={"h3"}` and `level="h3"` both work the same.

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,227 @@
...
+          <VStack id={"action"}>
+            {/* Initial form Screen */}
+            {formView && (
+              <>
+                <Form id="newsletter-form" onSubmit={internalOnSubmit}>
+                  <FormField key={"formfield-input"}>
+                    <TextInput
+                      id={"email-input"}
+                      isRequired
+                      invalidText="Please enter a valid email address." // @todo This could be set to helperText value when `if (isInvalid === true)`, allowing devs to set text displayed when bad email.
```
EdwinGuzman: I'm not sure I understand the todo, but it might be a case where the error message will always be the same.

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,227 @@
...
+                      Submit
+                    </Button>
+                  </FormField>
+                </Form>
+              </>
+            )}
+
+            {/* Confirmation Screen */}
+            {confirmationView && (
+              <>
```
EdwinGuzman: You technically don't need the react fragment element `<>` here because the Box is a single top level element.

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,227 @@
...
+                    size="large"
+                  />
+                  <Text fontWeight="medium">{confirmationText}</Text>
+                </Box>
+              </>
+            )}
+
+            {/* Error Screen */}
+            {errorView && (
+              <>
```
EdwinGuzman: Same here and for the form view, you can remove the fragment element. If you have multiple sibling elements, then the fragment is necessary. It doesn't hurt to have but just letting you know you can safely remove it.

src/components/NewsletterSignup/NewsletterSignup.stories.tsx
```diff
@@ -0,0 +1,141 @@
...
+      values,
+      "isInvalidEmail: ",
+      isInvalidEmail,
+      "View: ",
+      view
+    );
+  };
+  return (
+    <NewsletterSignup
+      {...args} // @todo All the same values below are already contained in this ...args array. But the ones below get to "win." I dunno why. Cuz they come after?
```
EdwinGuzman: Okay, _now_ I understand your concern about this. It does seem ordering matters when using the spread operator and yes it's because it comes after. I think I misunderstood at first.

Yes, `args` does contain the same values as the ones below but the values in `args` comes from the Storybook Controls section which the user can update. In this case, these values seem reasonable to have to override. For `confirmationText`, you don't need to have state for that since it seems it only is really updated once.

src/components/NewsletterSignup/NewsletterSignup.test.tsx
```diff
@@ -0,0 +1,297 @@
...
+    );
+    expect(await axe(container)).toHaveNoViolations();
+  });
+});
+
+describe("NewsletterSignup Unit Tests", () => {
+  // If you want to see what's happening, insert below render()
+  // screen.debug();
+  let onSubmit = jest.fn();
+  describe("Renders the Minimum Required Elements", () => {
```
EdwinGuzman: You could add a check for the privacy policy link since that is always rendered.

src/components/NewsletterSignup/NewsletterSignup.test.tsx
```diff
@@ -0,0 +1,297 @@
...
+      expect(
+        screen.getByText(/Please enter a valid email address./i)
+      ).toBeInTheDocument();
+      expect(screen.getByRole("form")).toBeInTheDocument();
+      expect(screen.getByRole("textbox")).toBeInTheDocument();
+      expect(screen.getByRole("button")).toBeInTheDocument();
+    });
+  }); // Close feedback tests
+
+  // @todo These will likely need to be rewritten under the new specs
```
EdwinGuzman: 👍

EdwinGuzman (commented):
This is great so far. We should chat about the focus feature for this component because the spec has changed. Not a problem but we should get it right.

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,195 @@
...
+import Heading from "../Heading/Heading";
+import useNYPLBreakpoints from "../../hooks/useNYPLBreakpoints";
+import { getSectionColors } from "../../helpers/getSectionColors";
+import { SectionTypes } from "../../helpers/types";
+
+interface NewsletterSignupProps {
+  /** Additional class name to add. */
+  className?: string;
+  /** Used to add additional information to the default confirmation message in
+   * the confirmation view. */
```
oliviawongnyc: The way this description reads, it sounds like if you add a `confirmationText` prop, it will be displayed in _addition_ to the "Thank you! You have successfully subscribed...." text below. Is that right?

src/components/NewsletterSignup/NewsletterSignup.stories.tsx
```diff
@@ -0,0 +1,122 @@
...
+        },
+      },
+    },
+    descriptionText: {
+      control: "text",
+      table: {
+        defaultValue: {
+          summary:
+            "Stay connected with the latest research news from NYPL, including information about our events, programs, exhibitions, and collections.",
+        },
```
oliviawongnyc: Generally, the `defaultValue` should be reflected in the example component above the props table, and Storybook users should be able to edit this value and see the changes reflected in the example, but that is not what is happening. Is this intentional?

Screen Shot 2023-09-27 at 10 09 56 AM

src/components/NewsletterSignup/NewsletterSignup.stories.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * Main Story for the NewsletterSignup component. This must contains the `args`
+ * and `parameters` properties in this object.
+ */
+export const WithControls: Story = {
+  args: {
+    className: undefined,
+    confirmationText:
+      "Fantastic! You're all set. Check the console for the data you submitted.",
```
oliviawongnyc: Why is it that the `confirmationText` prop does not match what appears in the component?

Screen Shot 2023-09-27 at 10 13 10 AM

src/components/NewsletterSignup/NewsletterSignup.mdx
```diff
@@ -0,0 +1,126 @@
...
+- {<Link href="#accessibility" target="_self">Accessibility</Link>}
+
+## Overview
+
+<Description of={NewsletterSignupStories} />
+
+
+**Note**: The `NewsletterSignup` example
+below alternately renders the "confirmation" and "error" screens on each form
+submission. This is just to demonstrate the different states of the component.
```
oliviawongnyc: How does a Storybook user reset the form so they can see the "error" screen?

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,195 @@
...
+  confirmationText?: string | JSX.Element;
+  /** Used to add description text above the form input fields in
+   * the initial/form view. Accepts a string or a JSX */
```
oliviawongnyc: I wasn't sure what this meant when I read it. What about, "Used to add a form description, rendered just below the form's `title`."

src/components/NewsletterSignup/NewsletterSignup.mdx
```diff
@@ -0,0 +1,126 @@
...
+DS and Chakra components. The two main components are the DS `Button` component
+used to open Chakra's `Drawer` component.
+
+When the primary button is clicked, the dialog opens and focus is set to the
+first focusable element which is the "close" button that contains minus icon in
+the header of the dialog. While opened, focus is trapped within the dialog until
+it is closed either by clicking on the "close" or "Cancel" buttons, pressing the
+"Escape" key, or by clicking outside of the dialog. When the `NewsletterSignup`
+component is closed, focus is set back to the primary button that opened the
+dialog.
```
oliviawongnyc: Is there a way to show this opening and closing functionality in Storybook?

src/components/NewsletterSignup/NewsletterSignup.mdx
```diff
@@ -0,0 +1,126 @@
...
+
+The markup of the `NewsletterSignup` structurally matches the modal dialog pattern
+that is implemented by Chakra's `Modal` and `Drawer` components. The container
+has `role=”dialog”`, `aria-modal=”true”`, `tabindex={0}`, `aria-labelledby` that
+references the title within the dialog, and `aria-describedby` that references a
+descriptive piece of text within the dialog.
```
oliviawongnyc: What do you think of including what the `aria-labelledby` and `aria-describedby` would be in the example at the top? I'm not sure if I have the values right, but something like, "...`aria-labelledby` that
references the title within the dialog (for instance, "The Newsletter Everyone Is Talking About" in the example above), and `aria-describedby` that references a
descriptive piece of text within the dialog (ie, "This is it.")"?

src/components/NewsletterSignup/NewsletterSignup.stories.tsx
oliviawongnyc: Also I think it would be helpful to set the `descriptionText` and `formHelperText` to something that might actually be seen on one of these components to give devs a better sense of what these can/should be used for. What do you think?

src/components/NewsletterSignup/NewsletterSignup.tsx
nypl-wluisi: This is fixed now

src/components/NewsletterSignup/NewsletterSignup.tsx
nypl-wluisi: This is fixed

src/components/NewsletterSignup/NewsletterSignup.mdx
```diff
@@ -0,0 +1,126 @@
...
+
+Within the `NewsletterSignup` component, the radio input field is created from the DS
```
oliviawongnyc: Should this be, "Within the dialog..." or "Within the form..." or would that not be accurate?

src/components/NewsletterSignup/NewsletterSignup.tsx
nypl-wluisi: This is fixed

src/components/NewsletterSignup/NewsletterSignup.tsx
nypl-wluisi: Fixed

src/components/NewsletterSignup/NewsletterSignup.mdx
```diff
@@ -0,0 +1,126 @@
...
+`RadioGroup` and `Radio` components, and input fields are created from the DS
+`TextInput` component. Each of these components has their own accessibility
+features documented in their respective Storybook pages.
+
+When the form is submitted, focus is set to the confirmation message or the
+error message if an error occurs.
+
+Whereas the `NewsletterSignup`'s primary button element is placed within the DOM
+structure where it is rendered, the dialog DOM structure is appended to the end
+of the DOM tree and it is done by Chakra.
```
oliviawongnyc: Nit: I would write this, "...is appended to the end of the DOM tree. This is a Chakra convention." 

Just makes it seem a little more like we don't have control over this. It's just the way Chakra does things (assuming I am understanding this correctly).

src/components/NewsletterSignup/NewsletterSignup.test.tsx
```diff
@@ -0,0 +1,297 @@
...
+import { sectionDataMap } from "../../helpers/types";
+
+describe("NewsletterSignup Accessibility", () => {
+  const onSubmit = jest.fn();
+  it("Form state passes accessibility", async () => {
+    const { container } = render(
+      <NewsletterSignup
+        title="Testing"
+        onSubmit={onSubmit}
+        formHelperText={"Form helper."}
```
oliviawongnyc: Nit: do we need this `formHelperText`?

src/components/NewsletterSignup/NewsletterSignup.test.tsx
```diff
@@ -0,0 +1,297 @@
...
+    );
+    expect(await axe(container)).toHaveNoViolations();
+  });
+
+  it("Confirmation state passes accessibility", async () => {
+    const { container } = render(
+      <NewsletterSignup
+        title="Testing"
+        onSubmit={onSubmit}
+        view={"confirmation"}
```
oliviawongnyc: Nit: curly braces not required around strings

src/components/NewsletterSignup/NewsletterSignup.test.tsx
```diff
@@ -0,0 +1,297 @@
...
+      render(<NewsletterSignup title="Testing" onSubmit={onSubmit} />);
+      expect(
+        screen.getByRole("button", { name: "Submit" })
+      ).toBeInTheDocument();
+    });
+    it("Renders the title", () => {
+      render(<NewsletterSignup title="Testing" onSubmit={onSubmit} />);
+      expect(screen.getByText(/Testing/i)).toBeInTheDocument();
+    });
+  }); // Close minimum elements tests
```
oliviawongnyc: Often we will put all of these checks in one test. It could make your tests a bit faster since it won't be rendering the component over and over, and be fewer lines of code.

```
 it("Renders the form components", () => {
      render(<NewsletterSignup title="Testing" onSubmit={onSubmit} />);
      expect(screen.getByRole("form")).toBeInTheDocument();
      expect(screen.getByRole("textbox")).toBeInTheDocument();
      expect(
        screen.getByRole("button", { name: "Submit" })
      ).toBeInTheDocument();

     // ...additional checks
    });
```

src/components/NewsletterSignup/NewsletterSignup.test.tsx
```diff
@@ -0,0 +1,297 @@
...
+                title="Testing"
+                onSubmit={onSubmit}
+                newsletterSignupType={section.type}
+              />
+            )
+            .toJSON();
+          expect(view).toMatchSnapshot();
+        }
+      );
+    });
```
oliviawongnyc: Nice!

src/helpers/getSectionColors.ts
```diff
@@ -0,0 +1,20 @@
+/* Retrieves section color values.
+ *
+ * Requires a section type as defined by sectionDataMap. Accepts an optional color
+ * preference ("primary" or "secondary").
+ *
+ * @returns A string if one color is requested or an object containing both (default).
```
oliviawongnyc: Can be a little more specific here:

@returns A string if one color is requested (by passing a `colorVal`) or an object containing both (default).

src/helpers/getSectionColors.ts
```diff
@@ -0,0 +1,20 @@
...
+ */
+import { sectionDataMap, SectionTypes } from "./types";
+
+export function getSectionColors(
+  type: SectionTypes,
+  colorVal?: "primary" | "secondary"
+) {
+  let colorName = sectionDataMap.filter((section) => section.type === type)[0]
```
oliviawongnyc: Is the [0] is needed here?
And nit: I would call this variable `color`, which I think is a bit more inclusive of the object or the string value.

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,195 @@
...
+        <Stack
+          direction={isLargerThanMobile ? "row" : "column"}
+          ref={ref}
+          __css={styles}
+          {...rest}
+        >
+          <Stack direction={isLargerThanMobile ? "row" : "column"} id={"info"}>
+            <Box
+              bg={getSectionColors(newsletterSignupType, "primary")}
+              id={"color-box"}
```
oliviawongnyc: Not sure what the convention is here - should this be camel case?

src/components/NewsletterSignup/NewsletterSignup.tsx
oliviawongnyc, line 1: I think some of the `key`s and `id`s in this component are unnecessary and can be removed.

oliviawongnyc (commented):
Nice job!! This is a complicated component.

src/components/NewsletterSignup/NewsletterSignup.tsx
aarnold101: Done

src/components/NewsletterSignup/NewsletterSignup.tsx
aarnold101: This would seem to be solved.

src/components/NewsletterSignup/NewsletterSignup.tsx
aarnold101: I know, but my IDE (PHPStorm) inserts the brackets automatically and I can't be bothered to remove them or change the behavior.

src/components/NewsletterSignup/NewsletterSignup.tsx
oliviawongnyc: Noticed kebab-case in other places in the DS so ignore this!

src/components/NewsletterSignup/NewsletterSignup.tsx
aarnold101: Marty confirmed via Slack that the error msg should always be the same and not customizable.

src/components/NewsletterSignup/NewsletterSignup.stories.tsx
aarnold101: Understood.

src/components/NewsletterSignup/NewsletterSignup.tsx
aarnold101: You're right. It was poorly worded. Fixed.

src/components/NewsletterSignup/NewsletterSignup.stories.tsx
aarnold101: Yes it is intentional. My thinking is: why pass the default values as a demonstration since you would not do this IRL? So I opt to use the WithControls story as its own example of how one could use the component. I was picking up on some similar [practices](https://nypl.github.io/nypl-design-system/reservoir/v1/?path=/story/components-page-layout-table--with-controls) already in the DS.

src/components/NewsletterSignup/NewsletterSignup.stories.tsx
aarnold101: The first issue mentioned is now fixed. The second has also been addressed, I think.

src/components/NewsletterSignup/NewsletterSignup.mdx
aarnold101: If you reload the story via the toolbar you can start fresh. Or if you change the view to something other than "form" via the UI, and then change it back, you can start over. I am unsure if/how one can set the storybook arg to match the current value of the view prop when it is changed via the onSubmit function.

src/components/NewsletterSignup/NewsletterSignup.tsx
aarnold101: You are right again. Poorly worded by me. Updated.

src/components/NewsletterSignup/NewsletterSignup.mdx
aarnold101: This all leftover from Feedbackbox. Deleted.

src/components/NewsletterSignup/NewsletterSignup.mdx
aarnold101: Sorry, again this is all irrelevant holdover from the FeebackBox text.

src/components/NewsletterSignup/NewsletterSignup.mdx
aarnold101: ibid.

src/components/NewsletterSignup/NewsletterSignup.test.tsx
aarnold101: As a prop generally speaking, or in this test case? I couldn't answer the former. But since it is meant to appear if utilized, I wrote the test to make sure it works.

src/components/NewsletterSignup/NewsletterSignup.test.tsx
aarnold101: I know. It's my IDE's fault. It is probably configurable but I don't know where look for it in the preferences.

src/components/NewsletterSignup/NewsletterSignup.test.tsx
aarnold101: OK. Great. I just thought writing out each test explicitly might be more informative if any one part should fail, but I am all in favor of expediency.

src/helpers/getSectionColors.ts
aarnold101: OK. Rewritten.

src/components/NewsletterSignup/NewsletterSignup.mdx
aarnold101: The aria-related notes were a holdover from the FeedbackBox. I am uncertain if they are relevant to this component or not.

src/components/NewsletterSignup/NewsletterSignup.test.tsx
aarnold101: Thanks!

src/components/NewsletterSignup/NewsletterSignup.tsx
aarnold101: CSS class and id naming convention [seems to prefer hyphens](https://www.freecodecamp.org/news/css-naming-conventions-that-will-save-you-hours-of-debugging-35cea737d849/).

src/components/NewsletterSignup/NewsletterSignup.tsx
aarnold101: I took some of the `key`s out. But the `id`s are all being used by the CSS.

src/helpers/getSectionColors.ts
aarnold101: Removing the `[0]` breaks the function as it can't find the colorVals value.

As for the nit, I have found that using "color" as a prop name conflicts with Chakra's definition of the same prop. So I don't use it.

nypl-wluisi: Let's chat about the `getSectionColors()` function and the `sectionDataMap` lookup table next week, there's a much simpler way we can do this. It's probably better to do the map as an object of objects, i.e,

```js
const sectionDataMap = {
  blogs: {
    primary: "section.blogs.primary",
    secondary: "section.blogs.secondary",
  },
  booksAndMore: {
    primary: "section.books-and-more.primary",
    secondary: "section.books-and-more.secondary",
  },
}
```
Then you can find the item you want by the object key, i.e,

```js
sectionDataMap["blogs"]
sectionDataMap["booksAndMore"]
```

aarnold101: @bigfishdesign13 Asked if I could create a function that returned either the primary or the secondary color or both so that's why I did it that way. I suppose one could just do `sectionDataMap["blogs"]["primary"]` but that did not occur to me.

src/components/NewsletterSignup/NewsletterSignup.mdx
EdwinGuzman: In the `NewsletterSignupWithControls` function, you could set up a timer that resets the view to "form" after a few seconds so it "resets". That would be helpful so users don't have to click elsewhere to refresh the Storybook view.

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,193 @@
...
+const defaultDescriptionText =
+  "Stay connected with the latest research news from NYPL, including information about our events, programs, " +
+  "exhibitions, and collections.";
+
+/**
+ * The NewsletterSignup component provides a way for patrons to register for an
+ * email-based newsletter distribution list.
+ */
+export const NewsletterSignup = chakra(
+  forwardRef<any, NewsletterSignupProps>(
```
EdwinGuzman: This `any` can be better defined. It can be an `HTMLDivElement`, try that.

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,193 @@
...
+                </FormField>
+              </Form>
+            )}
+            {view === "confirmation" && (
+              <Box
+                className="feedback-body response"
+                margin="auto"
+                textAlign="center"
+                ref={focusRef}
+                tabIndex={-1}
```
EdwinGuzman: I think this should be `0` instead of `-1`. The -1 removes it from the tab flow and we want to keep it.

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,193 @@
...
+            </VStack>
+          </Stack>
+          <VStack id={"action"}>
+            {isFormView && (
+              <Form id="newsletter-form" onSubmit={onSubmit}>
+                <FormField key={"formfield-input"}>
+                  <TextInput
+                    id={"email-input"}
+                    isRequired
+                    invalidText="Please enter a valid email address." // This text is boilerplate and not meant to be customized.
```
EdwinGuzman: Nitpick: move the comment above this line.

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,193 @@
...
+              >
+                Privacy Policy
+              </Link>
+            </VStack>
+          </Stack>
+          <VStack id={"action"}>
+            {isFormView && (
+              <Form id="newsletter-form" onSubmit={onSubmit}>
+                <FormField key={"formfield-input"}>
+                  <TextInput
```
EdwinGuzman, line 145: When you submit, the button becomes disabled and the text input should as well. You can add `isDisabled={view === "submitting"}` to this component.

src/components/NewsletterSignup/NewsletterSignup.test.tsx
```diff
@@ -0,0 +1,332 @@
...
+import renderer from "react-test-renderer";
+import NewsletterSignup from "./NewsletterSignup";
+import { sectionDataMap } from "../../helpers/types";
+
+// If you want to see what's happening, insert below render()
+// screen.debug();
+
+describe("NewsletterSignup Accessibility", () => {
+  const onSubmit = jest.fn(),
+    onChange = jest.fn(),
```
EdwinGuzman: Nitpick: our convention is to declare every variable on a new line with a new `const`.

src/components/NewsletterSignup/NewsletterSignup.test.tsx
```diff
@@ -0,0 +1,332 @@
...
+    );
+    expect(await axe(container)).toHaveNoViolations();
+  });
+});
+
+describe("NewsletterSignup Unit Tests", () => {
+  /** Notes
+   *
+   * 1. The newsletterSignupType tests are covered in the snapshot tests below.
+   * 2. Because the functionality of the submit click is handled entirely by the consuming app, there seems to be no
```
EdwinGuzman: This is true, but you can make tests for verifying that the functions you passed are being called. So you passed onSubmit and onChange and when you click on the submit button or type into the input field, you can verify that the jest mock functions are called.

Here is an example: https://github.com/NYPL/nypl-design-system/blob/74eb7820726b4d8720c5e9f43f25a3752b97b451/src/components/Checkbox/Checkbox.test.tsx#L193

src/theme/components/newsletterSignup.ts
```diff
@@ -0,0 +1,69 @@
...
+      flex: "1 1 0",
+      bg: "ui.bg.default",
+    },
+    "div#color-box": {
+      alignSelf: "stretch",
+    },
+    "div#pitch": {
+      width: { base: "100%", md: "50%" }, // It's a two-column layout >md
+      padding: {
+        base: "var(--nypl-space-s) var(--nypl-space-l) var(--nypl-space-l) var(--nypl-space-l)", // @todo For some reason using "padding:" with more than one shorthand value, e.g. l, sm, xxl, doesn't work.
```
EdwinGuzman: It's an issue with general Chakra values for the full CSS property. We know that you have to use the entire CSS variable when adding more than one value and as you have for the `md` case, when it's a single value you can pass the shorthand directly.

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,193 @@
...
+import Link from "../Link/Link";
+import Text from "../Text/Text";
+import TextInput from "../TextInput/TextInput";
+import Heading from "../Heading/Heading";
+import useNYPLBreakpoints from "../../hooks/useNYPLBreakpoints";
+import { getSectionColors } from "../../helpers/getSectionColors";
+import { SectionTypes } from "../../helpers/types";
+
+interface NewsletterSignupProps {
+  /** Optional: Additional class name to add. */
```
isastettler: I noticed that this pattern of prefixing each prop description with "Optional" or "Required" is only on this component present. I feel like it might be better to keep it in line with the other storybook entries but I might be missing something?

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,193 @@
...
+  /** Optional: Value to determine the section color highlight */
+  newsletterSignupType?: SectionTypes;
+  /** Required: A handler function that will be called when the form is submitted. */
+  onSubmit: (event: React.FormEvent<any>) => void;
+  /** Required: A handler function that will be called when the text input changes. */
+  onChange: (event: React.ChangeEvent<HTMLInputElement>) => void;
+  /** Optional: Used to populate the `<h3>` header title. */
+  title?: string;
+  /** Required: The value of the email text input field. */
+  valueEmail: string;
```
isastettler: I am wondering if the `valueEmail` prop is even needed. Looking at the TextInput component it seems that the `value prop it is passed to is optional there. If there is a use case, it might be an optional prop?

isastettler (commented):
Fun to look over this component. Well done.

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,193 @@
...
+  forwardRef<any, NewsletterSignupProps>(
+    (
+      {
+        className,
+        confirmationText = defaultConfirmationText,
+        descriptionText = defaultDescriptionText,
+        formHelperText,
+        id,
+        isInvalidEmail = false,
+        newsletterSignupType = "whatsOn",
```
isastettler: I am also wondering if this should be set to "blogs" for default. Looks like on the figma mock up that is the "generic" styling.

src/components/NewsletterSignup/NewsletterSignup.tsx
isastettler: Resolved. I made changes to the style/html set up of the left/top of the component.

src/helpers/getSectionColors.ts
isastettler: I changed the look up table from an array to an object and adjusted the function accordingly. It would be possible to get rid of the function all together as the section colors are available with the passed `newsletterSignupType` prop as such: `sectionDataMap[newsletterSignupType]` and the color values with the `primary` or `secondary` key too. Let me know if that would be desirable.

src/components/NewsletterSignup/NewsletterSignup.tsx
isastettler: Updated.

src/components/NewsletterSignup/NewsletterSignup.tsx
isastettler: Updated.

src/components/NewsletterSignup/NewsletterSignup.tsx
isastettler: Changed.

src/components/NewsletterSignup/NewsletterSignup.tsx
isastettler: Updated.

isastettler: Resolved.

src/components/NewsletterSignup/NewsletterSignup.test.tsx
isastettler: Changed.

src/components/NewsletterSignup/NewsletterSignup.mdx
isastettler: I changed the storybook set up so that you can select the different views in the controls. 
I also set up Browser states (like the TextInput component) for the invalidEmail state and the disabled state (when submitting). I am not sure if that is needed and would love some feedback.

src/components/NewsletterSignup/NewsletterSignup.test.tsx
isastettler: Added test for onChange and onSubmit

src/components/NewsletterSignup/NewsletterSignup.test.tsx
isastettler: Andrew added this already.

src/components/NewsletterSignup/NewsletterSignup.mdx
```diff
@@ -0,0 +1,124 @@
...
+## Component Props
+
+<Canvas of={NewsletterSignupStories.WithControls} />
+
+<ArgTypes of={NewsletterSignupStories.WithControls} />
+
+## Accessibility
+
+The `NewsletterSignup` component is a complex component built from various Reservoir
+DS and Chakra components. The two main components are the DS `Button` component
```
EdwinGuzman: The last line should be updated to reflect "two main components". The other can be `TextInput`.

EdwinGuzman: Actually, this may be a remnant of the FeedbackBox component. The `Drawer` component is not used in this one.

src/components/NewsletterSignup/NewsletterSignup.mdx
```diff
@@ -0,0 +1,124 @@
...
+used to open Chakra's `Drawer` component.
+
+When the primary button is clicked, the dialog opens and focus is set to the
```
EdwinGuzman: There's no dialog in this component.

EdwinGuzman: The rest of the paragraph can also be updated.

src/components/NewsletterSignup/NewsletterSignup.mdx
```diff
@@ -0,0 +1,124 @@
...
+first focusable element which is the "close" button that contains minus icon in
+the header of the dialog. While opened, focus is trapped within the dialog until
+it is closed either by clicking on the "close" or "Cancel" buttons, pressing the
+"Escape" key, or by clicking outside of the dialog. When the `NewsletterSignup`
+component is closed, focus is set back to the primary button that opened the
+dialog.
+
+The markup of the `NewsletterSignup` structurally matches the modal dialog pattern
```
EdwinGuzman: This also needs some updates.

src/components/NewsletterSignup/NewsletterSignup.mdx
EdwinGuzman: Yeah, I think you could remove the entire paragraph.

You can replace it with information regarding the text input and relevant labels there.

src/components/NewsletterSignup/NewsletterSignup.mdx
```diff
@@ -0,0 +1,124 @@
...
+`RadioGroup` and `Radio` components, and input fields are created from the DS
+`TextInput` component. Each of these components has their own accessibility
+features documented in their respective Storybook pages.
+
+When the form is submitted, focus is set to the confirmation message or the
+error message if an error occurs.
+
+Whereas the `NewsletterSignup`'s primary button element is placed within the DOM
+structure where it is rendered, the dialog DOM structure is appended to the end
+of the DOM tree and it is done by Chakra.
```
EdwinGuzman: All of this can be removed. But do include information regarding the focus on the confirmation and error views.

src/components/NewsletterSignup/NewsletterSignup.mdx
```diff
@@ -0,0 +1,124 @@
...
+
+Resources:
+
+- [MDN ARIA: dialog role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role)
```
EdwinGuzman: Dialog, Drawer, and Radio links are not needed.

src/components/NewsletterSignup/NewsletterSignup.mdx
```diff
@@ -0,0 +1,124 @@
...
+## Form Submission Data
+
+Submitted form data can be retrieved when the `NewsletterSignup` component is
+submitted through the required `onSubmit` prop. This prop expects a function and
+it will be called when the form is submitted. Similar to other DS form-components
+that have function props, the data from the component will be returned in the
+function's argument. In this case, it will be a single object.
+
+The submitted form data will be passed as an object that the parent component
+can use. The returned object will always conatain the "email" field.
```
EdwinGuzman: Minor typo "conatain"

src/components/NewsletterSignup/NewsletterSignup.stories.tsx
```diff
@@ -0,0 +1,131 @@
...
+    design: {
+      type: "figma",
+      url: "https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Main?node-id=80849%3A174194&mode=dev",
+    },
+    jest: "NewsletterSignup.test.tsx",
+  },
+  render: (args) => <NewsletterSignup {...args} />,
+};
+
+export const BrowserStates: Story = {
```
EdwinGuzman: Can you include this in the `.mdx`  file?

Also, another story that explicitly displays the confirmation and error views would be helpful, _or_ include those in this story and maybe rename this to "Component states"?

src/components/NewsletterSignup/NewsletterSignup.tsx
EdwinGuzman: Sure, you can remove it. It's also implied with the `?` in the prop definition.

src/components/NewsletterSignup/NewsletterSignup.tsx
EdwinGuzman: That's true but "whatsOn" is in the spec doc. Thoughts, @bigfishdesign13 ?

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,206 @@
...
+          direction={isLargerThanMobile ? "row" : "column"}
+          ref={ref}
+          __css={styles}
+          {...rest}
+        >
+          <VStack id="pitch">
+            <Heading level="h3" text={title} />
+            <Text>{descriptionText}</Text>
+            <Link
+              // @TODO I would make this a prop -- WL.
```
EdwinGuzman: The url or the entire link component? I agree if it's the url because it _could_ change and the DS should not be responsible for the update. If this is a prop, the default can be `"https://www.nypl.org/help/about-nypl/legal-notices/privacy-policy"`.

@bigfishdesign13 this sounds like a good update to the spec doc and component.

EdwinGuzman: @nypl-wluisi , sorry forgot to tag initially.

EdwinGuzman (commented):
Functionally, it looks good. Added some comments. Did not review the `newsletterSignup.ts` theme file yet.

src/components/NewsletterSignup/NewsletterSignup.stories.tsx
```diff
@@ -0,0 +1,131 @@
...
+    confirmationText:
+      "You can update your email subscription preferences at any time using the links at the bottom of the email.",
+    descriptionText: undefined,
+    formHelperText: undefined,
+    id: undefined,
+    isInvalidEmail: false,
+    newsletterSignupType: undefined,
+    onChange: undefined,
+    onSubmit: undefined,
+    title: undefined,
```
EdwinGuzman: I noticed that when I try to update this prop in the controls, when it's empty the page breaks. I think it's because this should not be `undefined` by default.

src/components/NewsletterSignup/NewsletterSignup.mdx
isastettler: Fixed.

src/components/NewsletterSignup/NewsletterSignup.stories.tsx
isastettler: Added.

src/components/NewsletterSignup/NewsletterSignup.tsx
isastettler: I added an optional prop called `privacyPolicyLink` of type `string`. The default value is set to `"https://www.nypl.org/help/about-nypl/legal-notices/privacy-policy"`.

src/components/NewsletterSignup/NewsletterSignup.stories.tsx
isastettler: Good catch. But this is a tricky issue: I realized it has to to with the `Heading` component. If you go to the Heading Storybook entry it is having the same issue. I fixed it by adding a check in the Newsletter component itself even-though there is a default string always passed. I tried several things in the .stories.tsx file but could not get it to work. It might be possible that the fix needs to be added to the Heading component? I am not entirely sure. 🤔

src/components/NewsletterSignup/NewsletterSignup.mdx
isastettler: Yup I have to admit it only moved the text that was there up to fix the order of the doc without reading it. I rewrote it now and would love some more eyes on the new text. Thanks for pointing this out.

src/components/NewsletterSignup/NewsletterSignup.tsx
isastettler: One other thing I noticed is that the "blogs" value used for the border is the only one using the `secondary` color value..?

src/components/NewsletterSignup/NewsletterSignup.stories.tsx
EdwinGuzman: Yes, we should fix that in the `Heading` component but that's out of scope for this ticket. I'll create a follow up ticket for that.

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -103,7 +103,7 @@ export const NewsletterSignup = chakra(
           {...rest}
         >
           <VStack id="pitch">
-            <Heading level="h3" text={title} />
+            {title && <Heading level="h3" text={title} />}
```
EdwinGuzman: Yes, thanks. This works for now.

src/components/NewsletterSignup/NewsletterSignup.tsx
isastettler: Edwin explained the use case for this prop. Made it optional though.

src/components/NewsletterSignup/NewsletterSignup.tsx
isastettler: I added a fix in the theme file for the "blogs" issue. Still curious about the default.

src/components/NewsletterSignup/NewsletterSignup.test.tsx
```diff
@@ -147,6 +146,33 @@ describe("NewsletterSignup Unit Tests", () => {
     expect(screen.getByText(/Just trying to help/i)).toBeInTheDocument();
   });
 
+  it("Renders the Optional descriptionText and formHelperText HTML Values for the Form", () => {
+    const testNewsletterSignup = (
+      <NewsletterSignup
+        descriptionText={
+          <div>
+            Do not send <a href="https://chias.website/">cash</a>.
```
EdwinGuzman, line 149: Can we add a comment to above this to say that the link is safe and fun. At first glance it looks like an issue 😅

src/components/NewsletterSignup/NewsletterSignup.stories.tsx
```diff
@@ -100,6 +100,25 @@ export const WithControls: Story = {
...
+  render: () => (
+    <NewsletterSignup
+      id="jsx-description-text"
+      view="form"
+      onChange={() => {}}
+      onSubmit={() => {}}
+      descriptionText={
+        <div>
+          If the description text needs to have <i>special</i> styling or needs
+          to have a <a href="https://corgiorgy.com/">nested link</a>
```
EdwinGuzman: 😭

src/theme/components/newsletterSignup.ts
```diff
@@ -0,0 +1,77 @@
...
+  newsletterSignupType: SectionTypes;
+}
+
+const NewsLetterSignup = {
+  baseStyle: ({ newsletterSignupType }: NewsLetterStyleProps) => ({
+    alignItems: "center",
+    borderWidth: { base: "0px 1px 1px 1px", md: "1px 1px 1px 0px" },
+    maxWidth: "1280px",
+    width: "100%",
+    "div#pitch": {
```
EdwinGuzman: Try using fewer ids and this can be `pitch: { ....`. Then in the component file, you can use `styles.pitch` explicitly instead of relying on the id css rule.

src/components/NewsletterSignup/NewsletterSignup.mdx
```diff
@@ -0,0 +1,110 @@
...
+
+- {<Link href="#overview" target="_self">Overview</Link>}
+- {<Link href="#component-props" target="_self">Component Props</Link>}
+- {<Link href="#accessibility" target="_self">Accessibility</Link>}
+
+## Overview
+
+<Description of={NewsletterSignupStories} />
+
+**Note**: The `NewsletterSignup` example below will render different states depending on what email address you input.
```
aarnold101: I noticed that the example form no longer did anything and if you tried to use it, you got an error page saying there was no preview. So I restored my old function using the default values. I also noticed the .mdx had an outdated reference to behavior from the FeedbackBox component, so I replaced it. Lastly, I suppressed the source buttons on the canvases b/c what they reveal is not helpful and may be confusing.

src/theme/components/newsletterSignup.ts
isastettler: Changed.

src/components/NewsletterSignup/NewsletterSignup.test.tsx
isastettler: Added.

src/components/NewsletterSignup/NewsletterSignup.tsx
bigfishdesign13: This is an error on my part.  The default should not be `"whatsOn"`.  Rather, if `newsletterSignupType` is not set, the color should default to what is shown in the Figma designs (`"ui.border.default"` or `"dark.ui.border.default"`).

src/theme/components/newsletterSignup.ts
```diff
@@ -0,0 +1,78 @@
...
+        base: "var(--nypl-space-s) var(--nypl-space-l) var(--nypl-space-l) var(--nypl-space-l)",
+        md: "l",
+        lg: "var(--nypl-space-l) var(--nypl-space-xxl) var(--nypl-space-l) var(--nypl-space-xl)",
+      },
+      alignItems: "flex-start",
+      gap: "xs",
+      margin: "unset",
+    },
+    privacy: {
+      gap: "xxs",
```
bigfishdesign13: Remove this.

src/theme/components/newsletterSignup.ts
```diff
@@ -0,0 +1,78 @@
...
+      fontSize: "desktop.caption",
+      fontWeight: "caption",
+      margin: "unset",
```
bigfishdesign13: Neither of these shoudl be necessary. Was there a specific reason they were added?

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,238 @@
...
+                descriptionText
+              )
+            ) : null}
+
+            <Link
+              href={privacyPolicyLink}
+              type="external"
+              margin="unset"
+              __css={styles.privacy}
+            >
```
bigfishdesign13: Add `isUnderlined={false}`.

Also, I see the `margin="unset"` here again.  This can be removed, unless there is a reason I am not seeing.

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,238 @@
...
+      }, [view]);
+      return (
+        <Stack
+          direction={isLargerThanMobile ? "row" : "column"}
+          ref={ref}
+          __css={styles}
+          {...rest}
+        >
+          <VStack __css={styles.pitch} alignItems="flex-start">
+            {title && <Heading level="h3" text={title} margin="unset" />}
```
bigfishdesign13: I imagined that this would be set as `level="h2"` with `size="heading3"` because it's almost like a new section.  Can you talk about why it was set as an `h3`?

Cc @EdwinGuzman

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,238 @@
...
+                    size="large"
+                  />
+                  <Text
+                    fontSize="xl"
+                    marginStart="xs"
+                    marginBottom="unset"
+                    fontWeight="medium"
+                  >
+                    {confirmationHeading}
+                  </Text>
```
bigfishdesign13: Why was the `Heading` component not used here?

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,238 @@
...
+                  {/* This text is boilerplate and not meant to be customized. */}
+                  <Text
+                    color="ui.error.primary"
+                    fontSize="xl"
+                    marginStart="xs"
+                    marginBottom="unset"
+                    fontWeight="medium"
+                  >
+                    {errorHeading}
+                  </Text>
```
bigfishdesign13: Why was the `Heading` component not used here?

src/theme/components/newsletterSignup.ts
```diff
@@ -0,0 +1,78 @@
...
+      width: { base: "100%", md: "50%" },
+    },
+    form: {
+      width: "100%",
+    },
+    // Overwrites the defaut styling of the From component layout
+    "#newsletter-form-parent": {
+      // The button is 78px wide and must sit to the right of the input field >lg.
+      gridTemplateColumns: { base: null, xl: "1fr 78px" },
+      gap: { base: "s", xl: "xs" },
```
bigfishdesign13: Change `xl` to `lg`.  Why were these shifted to the larger breakpoint?

src/theme/components/newsletterSignup.ts
```diff
@@ -0,0 +1,78 @@
...
+    },
+    button: {
+      // The button must align w/ the input field, but using {align-items: center} doesn't quite work due to the input field not being the literal v-center.
+      marginTop: { base: null, xl: "31px" },
```
bigfishdesign13: Change `xl` to `lg`.

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,238 @@
...
+          direction={isLargerThanMobile ? "row" : "column"}
+          ref={ref}
+          __css={styles}
+          {...rest}
+        >
+          <VStack __css={styles.pitch} alignItems="flex-start">
+            {title && <Heading level="h3" text={title} margin="unset" />}
+            {descriptionText ? (
+              typeof descriptionText === "string" ? (
+                <Text margin="unset">{descriptionText}</Text>
```
bigfishdesign13: Change this to:
```
<Text noSpace size="body2">{descriptionText}</Text>
```

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,238 @@
...
+                  <Text
+                    fontSize="xl"
+                    marginStart="xs"
+                    marginBottom="unset"
+                    fontWeight="medium"
+                  >
+                    {confirmationHeading}
+                  </Text>
+                </Box>
+                <Text>{confirmationText}</Text>
```
bigfishdesign13: Change this to:
```
<Text noSpace size="body2">{confirmationText}</Text>
```

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -0,0 +1,238 @@
...
+                    marginStart="xs"
+                    marginBottom="unset"
+                    fontWeight="medium"
+                  >
+                    {errorHeading}
+                  </Text>
+                </Box>
+                {errorText ? (
+                  typeof errorText === "string" ? (
+                    <Text margin="unset">{errorText}</Text>
```
bigfishdesign13: Change this to:
```
<Text noSpace size="body2">{errorText}</Text>
```

src/components/NewsletterSignup/NewsletterSignup.stories.tsx
```diff
@@ -0,0 +1,216 @@
...
+          confirmationText="You can update your email subscription preferences at any time using the links at the bottom of the email."
+          errorHeading="An error has occurred."
+          errorText={
+            <div>
+              Please refresh this page and try again. If this error persists,
+              <a href="mailto:enews@nypl.org?subject=Please re-activate my e-mail address">
+                contact our e-mail team
+              </a>
+              .
+            </div>
```
bigfishdesign13: Using native HTML causes styling issues.  Can you rewite this using DS `Text` and `Link` components?
```
<Text noSpace size="body2">
  Please refresh this page and try again. If this error persists,{" "}
  <Link href="mailto:enews@nypl.org?subject=Please re-activate my e-mail address">
    contact our e-mail team
  </Link>
  .
</Text>
```

src/components/NewsletterSignup/NewsletterSignup.mdx
```diff
@@ -0,0 +1,107 @@
...
+<ArgTypes of={NewsletterSignupStories.WithControls} />
+
+## Accessibility
+
+The `NewsletterSignup` component is a complex component built from various Reservoir
+DS and Chakra components.
+
+Within the `NewsletterSignup` component, the DS `form` component wraps around two DS `FormField` components.
+Those`FormField` components hold a DS `TextInput` component of `type="email"` and a DS `Button` component of `type="submit"` respectively.
+Each of these components has their own accessibility features documented in their respective Storybook pages.
```
bigfishdesign13, line 44: I found a great extension for VSCode called `Rewrap` that helps with wrapping and formatting long lines of text.

https://marketplace.visualstudio.com/items?itemName=stkb.rewrap

src/components/NewsletterSignup/NewsletterSignup.stories.tsx
```diff
@@ -0,0 +1,216 @@
...
+import Heading from "../Heading/Heading";
+
+const meta: Meta<typeof NewsletterSignup> = {
+  title: "Components/Form Elements/NewsletterSignup",
+  component: NewsletterSignup,
+  decorators: [withDesign],
+  parameters: {
+    jest: ["NewsletterSignup.test.tsx"],
+  },
+  argTypes: {
```
bigfishdesign13, line 19: Add `privacyPolicyLink` to the argTypes and make sure to show the default value.

src/theme/components/newsletterSignup.ts
```diff
@@ -0,0 +1,78 @@
+import { getSectionColors } from "../../helpers/getSectionColors";
+import { SectionTypes } from "../../helpers/types";
+interface NewsLetterStyleProps {
+  newsletterSignupType: SectionTypes;
+}
+
+const NewsLetterSignup = {
```
bigfishdesign13: Nit: this should be `NewsletterSignup`

src/theme/components/newsletterSignup.ts
```diff
@@ -0,0 +1,78 @@
...
+      gap: { base: "s", xl: "xs" },
+    },
+    button: {
+      // The button must align w/ the input field, but using {align-items: center} doesn't quite work due to the input field not being the literal v-center.
+      marginTop: { base: null, xl: "31px" },
+    },
+  }),
+};
+
+export default NewsLetterSignup;
```
bigfishdesign13: Nit: this should be `NewsletterSignup`

src/helpers/getSectionColors.ts
```diff
@@ -0,0 +1,28 @@
...
+  type: SectionTypes,
+  colorVal?: "primary" | "secondary"
+): ColorSection | ColorVal {
+  let colorValOrSection = sectionDataMap[type];
+
+  if (colorVal) {
+    colorValOrSection = colorValOrSection[colorVal];
+  }
+  return colorValOrSection;
+}
```
bigfishdesign13, line 28: This function needs to account for both light mode and dark mode values.  Let's setup a call to talk about the best way to handle this.

Cc @EdwinGuzman

bigfishdesign13 (changes requested):
This looks very good.  It just needs a few style changes, some language updates, and possibly one functionality adjustment.

src/components/NewsletterSignup/NewsletterSignup.tsx
EdwinGuzman: That you mention this, I think the entire `title` prop _should_ be dictated by the consuming app. The DS has an expectation of how and where this component should go on a page, but if we set this to h3 or h2, we are now telling the app how to structure their app around this component. I don't think we should have that much power. If this becomes a small part on the footer of the page that doesn't follow the app's heading hierarchy, we're going to cause accessibility issues.

So perhaps we should:

- make `title: JSX.Element` as a prop
- set this line as `{title}`
- use this as `<NewsletterSignup title={<Heading level="h3" text={title} margin="unset" />} ... />`

Then apps will have more control.

src/theme/components/newsletterSignup.ts
```diff
@@ -5,12 +5,13 @@ interface NewsLetterStyleProps {
 }
 
 const NewsLetterSignup = {
+  parts: ["pitch", "privacy", "action"],
```
EdwinGuzman, line 8: 👍

src/components/NewsletterSignup/NewsletterSignup.mdx
```diff
@@ -56,7 +51,9 @@ Resources:
 Alternatively to a `descriptionText` of type `string`, a HTML Element or React component can be passed. When passing a JSX Element, the consuming app is responsible
 to assure its accessibility.
 
-<Canvas of={NewsletterSignupStories.DescriptionUsingJSXElements}  sourceState="none"/>
+_NOTE: This is applicable for all component props excepting HTML/JSX elements._
```
EdwinGuzman: typo: "except" instead of "excepting"

src/components/NewsletterSignup/NewsletterSignup.mdx
```diff
@@ -102,6 +102,51 @@ const onSubmit = async (values) => {
...
+_NOTE: open the browser console to see the values logged in the example below._
+
+The input value typed into the `TextInput` of the `NewsletterSignup` component can be accessed by the
+functions passed to the `onChange` and `onSubmit` prop.
+
+Both the `onChange` and `onSubmit` callback functions can retrieved the submitted value as `event.target.email.value`
+through the `event` object passed as the single argument.
+
+The following example logs the `event.target.email.value` to the console on each `onChange` call and upon
+cklicking the Submit button which triggers the `onSubmit` function and simulate a submission.
```
EdwinGuzman: minor typo "cklicking"

src/components/NewsletterSignup/NewsletterSignup.stories.tsx
```diff
@@ -112,6 +113,57 @@ export const WithControls: Story = {
   },
   render: (args) => <NewsletterSignup {...args} />,
 };
+let counter = 0;
+
+function NewsletterSignupOnSubmitExampleComponent() {
+  const [view, setView]: [
+    NewsletterSignupViewType,
+    React.Dispatch<React.SetStateAction<NewsletterSignupViewType>>
+  ] = React.useState("form");
```
EdwinGuzman, line 150: This is interesting. No need to remove, just pointing out I'd normally do `const [...] = useState<NewsletterSignupViewType>("form");` but this is nice to see.

src/helpers/getSectionColors.ts
EdwinGuzman: Can we make this a new ticket to work on for a later iteration? We don't need the dark mode variant right now and I don't want this to be a blocker for its initial release.

EdwinGuzman:
Most of the changes are from Marty and I just pointed out minor things. From what Marty pointed out:
- Let's talk about using the `Heading` component inside vs outside the component. We currently have other components that set specific heading values so this might lead to a larger discussion. For now, I think the `h3` fits slightly better than an `h2` because it may be more versatile and perhaps used in more places than an `h2`. Unless if we already know that in DXP it will be used one level below the `h1`, then let's make it `h3`. @isastettler you mentioned you already tested this a bit in SCOUT so you might know best on its expected implementation and what level it should be.
- Can we make the dark mode feature a new ticket and PR?

src/components/NewsletterSignup/NewsletterSignup.tsx
isastettler: As Edwin mentioned I started testing the component integration with Scout. I did  a quick inspection of the heading:
- Our heading prop was type string 
BUT
- It renders as a `h2` with `font-size=28px`, with the NewsletterSignup component test it renders a `h3` with `font-size=32px`.

I would agree with Edwin that it would make sense to allow a customized heading.

src/components/NewsletterSignup/NewsletterSignup.tsx
isastettler: I removed the underline, just want to point out that on the figma mock up the privacy policy link is underlined.
As for the `margin="unset`: I removed the ones in the theme file - this was left over from my efforts to move it from the .tsx file to the theme file. I commented the `margin="unset"` out for now but as you will see it will create big spaces around the different text areas of the left side that are not in line with. the figma file.

src/theme/components/newsletterSignup.ts
isastettler: Removed.

src/components/NewsletterSignup/NewsletterSignup.stories.tsx
isastettler: Added.

src/theme/components/newsletterSignup.ts
isastettler: Good catch. Fixed this.

src/components/NewsletterSignup/NewsletterSignup.mdx
isastettler: I meant accepting. Fixed it.

src/theme/components/newsletterSignup.ts
isastettler: Removed.

src/helpers/getSectionColors.ts
isastettler: For now I simply changed the default corder color to use the blogs color since that uses the `ui.default.border` color.
Once it is decided, it can be updated accordingly.

src/components/NewsletterSignup/NewsletterSignup.mdx
isastettler: Fixed.

src/components/NewsletterSignup/NewsletterSignup.tsx
bigfishdesign13: Yes, the Figma still shows the underline.  That also needs to be corrected.

For both the `Heading` and `Text` components you can use the `noSpace` prop to remove the default spacing built into the components.  That prop was added to the components to do exactly what you are doing with the `"unset"`.  I guess use whichever is easier for you.

src/components/NewsletterSignup/NewsletterSignup.tsx
bigfishdesign13: Yes, go ahead and update the title prop to accept the `Heading` component instead of a plain string.  Make sure to also update the component docs accordingly.  I will update the spec doc to reflect this change.

src/helpers/getSectionColors.ts
bigfishdesign13: Ok, thanks.  That works.  I see no reason to change that in the future.  Please add a comment to the code that I explains why the default is `"blog"`.

As for the dark mode support in the `getSectionColors` function, yes, let's create a new ticket to add that functionality in the next iteration of the component.

src/components/NewsletterSignup/NewsletterSignup.tsx
bigfishdesign13: But let's also change the default to render as an `<h2>` with `size="heading3"`.

src/components/NewsletterSignup/NewsletterSignup.stories.tsx
```diff
@@ -272,3 +284,15 @@ export const ComponentStates: Story = {
     </VStack>
   ),
 };
+
+/* To fix focus issue where the page focuses on the last NewsletterSignup
```
EdwinGuzman, line 298: Yes, this is a bit hacky but it works. Let's keep it for now knowing it could break later. Can you add to the comment that it's expected behavior for the different screen views to have focus, so this is to better view the docs when multiple newslettersignup components are on a single page but this should not happen in a real app.

src/components/NewsletterSignup/NewsletterSignup.tsx
isastettler: I used the `noSpace` prop.. works excellent! Thank you.

src/components/NewsletterSignup/NewsletterSignup.tsx
isastettler: I added this and added a line to the accessibility text too. Tests are updated and it's the args table too. I did remove the controls as it seems that storybook handle it.

src/components/NewsletterSignup/NewsletterSignup.stories.tsx
isastettler: I added some more comment. It is def hacky 😎

EdwinGuzman (approved):
The latest update looks good to me! I think you have two or so unresolved comments from Marty.

src/components/NewsletterSignup/newsletterSignupChangelogData.ts
```diff
@@ -0,0 +1,26 @@
...
+    affects: [],
+    notes: ["Adds the NewsletterSignup component to the DS library"],
+  },
+  {
+    date: "Prerelease",
+    version: "Prerelease",
+    type: "New Feature",
+    affects: [],
+    notes: ["Adds the NewsletterSignup component to the DS library"],
+  },
```
EdwinGuzman: You only need the first object since the second is redundant.

isastettler: Removed.

src/components/NewsletterSignup/NewsletterSignup.tsx
isastettler: This was the implementation I took over and kept as I assumed that was intentional.

src/components/NewsletterSignup/NewsletterSignup.tsx
isastettler: This was the implementation I took over and kept as I assumed that was intentional.

src/theme/components/newsletterSignup.ts
isastettler: Updated.

src/components/NewsletterSignup/NewsletterSignup.tsx
isastettler: Updated.

src/components/NewsletterSignup/NewsletterSignup.tsx
isastettler: Updated.

src/components/NewsletterSignup/NewsletterSignup.tsx
isastettler: Updated.

src/components/NewsletterSignup/NewsletterSignup.stories.tsx
isastettler: Updated.

src/helpers/getSectionColors.ts
bigfishdesign13: New ticket to update the function: https://jira.nypl.org/browse/DSD-1604

src/components/NewsletterSignup/NewsletterSignup.tsx
bigfishdesign13: Understood.  With our recent discussions about concerns with a11y and heading hierarchy, we should keep this as a `<Text`> component.  However, let's use a DS token for sizing.  

For all instances where this formatting is used, remove `fontSize="xl"` and replace it with `size="subtitle1"`.

src/components/NewsletterSignup/NewsletterSignup.tsx
bigfishdesign13: See my response above.

src/components/NewsletterSignup/NewsletterSignup.stories.tsx
bigfishdesign13: Please also add the default value: https://www.nypl.org/help/about-nypl/legal-notices/privacy-policy

isastettler: It was a typo that it did not show up before. This is fixed now.

src/components/NewsletterSignup/NewsletterSignup.tsx
isastettler: Updated.

bigfishdesign13:
Everything looks great.  Release the hounds!

bigfishdesign13 (approved):
Yes!

PR #1424, opened 2023-09-27
Updating documentation and links to better reference the v2 update

EdwinGuzman:
No JIRA ticket.

- Updates `/v1` Storybook documentation references to `/v2`.
- Adds better documentation about v2.x and v1.x and where each documentation site is hosted.

## How has this been tested?

Locally. Links will eventually work once we deploy.

- N/A

### Checklist:

### Front End Review:

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "prepare": "npm run build && husky install",
     "storybook": "storybook dev -p 6006",
     "build-storybook:v1": "storybook build -c .storybook -o ./reservoir/v1",
+    "build-storybook:v2": "storybook build -c .storybook -o ./reservoir/v2",
     "prebuild:storybook": "npm run test:generate-output",
     "build-storybook": "storybook build"
   },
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -39,6 +39,7 @@
     "prepare": "npm run build && husky install",
     "storybook": "storybook dev -p 6006",
     "build-storybook:v1": "storybook build -c .storybook -o ./reservoir/v1",
```
bigfishdesign13: What's the reason for keeping this?

bigfishdesign13 (approved):
Looks good, but I do have a question.  Is there a way to force anything that is trying to access the v1 folder to forward to the v2 folder?

package.json
EdwinGuzman: Hmm you're right, and it actually won't work. I'll remove.

EdwinGuzman:
> Looks good, but I do have a question. Is there a way to force anything that is trying to access the v1 folder to forward to the v2 folder?

I'm not sure Github Pages gives you that type of control but we should research it.

PR #1431, opened 2023-10-02
DSD-1281: Fix Hero backgroundColor/foregroundColor issue

jackiequach:
Fixes JIRA ticket [DSD-1281](https://jira.nypl.org/browse/DSD-1281)

- Fixes the issue with `backgroundColor` and `foregroundColor` props not prioritizing the passed design token values for the `Hero` component. Chakra prioritizes `bg` over `backgroundColor` which caused the default value to overwrite the passed value.
- `foregroundColor` for the campaign variant was being overwritten by the default styles passed through `__css`, this was resolved by adding `contentBoxStyling` to the object.

## How has this been tested?

Locally in Storybook

-

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/Hero/Hero.tsx
+++ b/src/components/Hero/Hero.tsx
@@ -165,8 +165,8 @@ export const Hero = chakra(
 
       if (!heroSecondaryTypes.includes(heroType)) {
         contentBoxStyling = {
-          ...(backgroundColor && { backgroundColor }),
           ...(foregroundColor && { color: foregroundColor }),
+          ...(backgroundColor && { backgroundColor }),
         };
       } else if (
         foregroundColor ||
@@ -230,6 +230,7 @@ export const Hero = chakra(
         >
           <Box
             data-testid="hero-content"
+            style={contentBoxStyling}
             __css={{ ...styles.content, ...contentBoxStyling }}
           >
             {childrenToRender}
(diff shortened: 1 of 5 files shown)


src/components/Hero/Hero.tsx
```diff
@@ -160,13 +160,13 @@ export const Hero = chakra(
           ? { bgColor: backdropBackgroundColor }
           : { backgroundColor };
       } else if (heroType === "tertiary" || heroType === "fiftyFifty") {
-        backgroundImageStyle = { backgroundColor };
+        backgroundImageStyle = backgroundColor ? { bg: backgroundColor } : {};
```
EdwinGuzman, line 163: `backgroundColor` is a css prop, but was the issue that `bg` has more precedence? Or perhaps a mix of this and that the `contentBoxStyling` came before the `styles.content` object?

jackiequach: It seems to be somewhat unknown, `bg` for content is overwritten by the `backgroundColor` added with `contentBoxStyling`, but `backgroundColor` does not overwrite `bg` in the parent component. By setting `backgroundImageStyle` to use `bg` instead, the spread operator overwrites the default value consistently.

EdwinGuzman (approved):

bigfishdesign13 (approved):
Looks good.  Just make sure to update the `Latest` version number on the `Hero` component story page.  Also, I will need some help writing the AC for this ticket.  Is there any way to validate this update through Storybook?

EdwinGuzman:
True, @jackiequach can you add an example in the storybook doc or update the existing examples with the better design token values instead of hex values (if any are used).

jackiequach:
Updated the last Campaign example with design token values.

EdwinGuzman (approved):
Looks good.

bigfishdesign13 (approved):
Looks good!

PR #1432, opened 2023-10-03
DSD-1581: better aria-label values for Slider

bigfishdesign13:
Fixes JIRA ticket [DSD-1581](https://jira.nypl.org/browse/DSD-1581)

- Updates the `Slider` component to use appropriate `aria-label` values for the slider thumbs and text input fields.

## How has this been tested?

- local Storybook

- The aria-label attribute on the slider thumbs will now always be present and the thumbs will no longer point to the main commponent label in aria-describedby.   Additionalmly, the aria-label attributes on the text input fields was updated to better align with the aria-label values on the slider thumbs.

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/Slider/Slider.tsx
+++ b/src/components/Slider/Slider.tsx
@@ -276,17 +276,15 @@ export const Slider = chakra(
       const getSliderType = () => {
         return isRangeSlider ? (
           <ChakraRangeSlider
-            // Both slider thumbs need values and should be in an array.
+            /**
+             * ChakraRangeSlider expects an array and applies the values to the
+             * `aria-label` attributes on the two slider thumbs respectively.
+             */
             aria-label={
-              showLabel
-                ? [
-                    `${labelText} - slider handle for start value`,
-                    `${labelText} - slider handle for end value`,
-                  ]
-                : [
-                    `${labelText} - slider handle for start value`,
-                    `${labelText} - slider handle for end value`,
-                  ]
+              [
+                `${labelText} - slider handle for start value`,
+                `${labelText} - slider handle for end value`,
+              ] as string[]
             }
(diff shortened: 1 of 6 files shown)


src/components/Slider/Slider.tsx
```diff
@@ -278,14 +278,15 @@ export const Slider = chakra(
...
             aria-label={
-              !showLabel
-                ? [`${labelText} - start value`, `${labelText} - end value`]
-                : undefined
-            }
-            // Both slider thumbs need values and should be in an array,
-            // even if it's the same label.
-            aria-labelledby={
-              showLabel ? [`${id}-label`, `${id}-label`] : undefined
+              showLabel
```
EdwinGuzman: If the array is defined when both `showLabel`  is true or false, then we can remove the tertiary logic check.

EdwinGuzman (approved):
Just a code question but otherwise works and not a blocker.

src/components/Slider/Slider.tsx
bigfishdesign13: Ugh, I meant to include a comment in reference to this array.  When I tried to remove the tertiary logic, the `aria-label` prop complained about the formatting.  I wanted to ask you abnout this.  Perhaps I was just missing some syntax that was throwing it off.

EdwinGuzman: I see. Chakra expects it but I think the linter doesn't (rightly) expect an array. Try
```
aria-label={[
              `${labelText} - slider handle for start value`,
              `${labelText} - slider handle for end value`,
            ] as string[]}
```

bigfishdesign13: Thanks.  That was perfect.

PR #1433, opened 2023-10-04
DSD-1577: helper text association

oliviawongnyc:
Fixes JIRA ticket DSD-1577.

- Updates `TextInput` so it no longer incorrectly overwrites the `aria-describedby` value to undefined when part of the `DatePicker` component.
- Updates `DatePicker` to pass a `secondaryHelperTextId` to its `TextInput` if needed so that the `aria-describedby` value can be associated with all relevant `helperText`s.

## Important notes:

- There were two possible ways to resolve this second issue. The way I did it was by switching two objects that are being spread in `TextInput`: `...rest` and `...ariaAttributes`. Now, `ariaAttributes` is second, so it is no longer overwritten by any values in `...rest`, which was causing the problem (`ariaAttributes` contains `aria-describedby: [helperTextId]` and `rest` contains `aria-describedby: undefined`.
- The second way to solve this, deals with the question why does `DatePicker` pass `aria-describedby: undefined` in the first place? The answer? I DON'T KNOW! Couldn't figure it out. But what I could have done was to just delete it in `DatePicker` by adding `delete (rest as any)['aria-describedby'];` to the `CustomTextInput` in `DatePicker`.
- I chose the first approach because we are carefully creating `ariaAttributes` in `TextInput` and presumably don't want any of its values overwritten whereas `rest` contains hand-me-downs from elsewhere and other components. Putting `ariaAttributes` after `rest` ensures we always have the values from `ariaAttributes`.

## How has this been tested?

- Locally in Storybook, as well as adding tests in the `DatePicker` test file.

- None. This work improves accessibility.

### Checklist:

### Front End Review:


src/utils/utils.ts
```diff
@@ -78,9 +80,10 @@ export const getAriaAttrs = ({
         ? `${labelText} - ${footnote}`
         : (labelText as string);
   } else if (footnote) {
-    ariaAttributes["aria-describedby"] = `${id}-helperText`;
+    ariaAttributes["aria-describedby"] = `${
+      secondaryHelperTextId ? secondaryHelperTextId + " " : ""
```
EdwinGuzman: This fixes the issue but _if_ we need to later add more ids to this string list, then this will become unmanageable with more props and logic checks because the name implies it's a single additional id. Perhaps this can be renamed to `additionalHelperTextIds` and specify that it can be a single id or a space-separated string of ids (similar to how aria-describedby is documented).

EdwinGuzman (commented):
Can you also update the storybook doc so that `secondaryHelperTextId` appears in. alphabetical order in the args table?

oliviawongnyc:
@EdwinGuzman, do you know why the snapshots are failing? I see a space was removed in the "react-datepicker__input-container" classname from the .snap files, but....why?!

EdwinGuzman:
...that's odd. I wouldn't expect that. The snapshot tests are failing even if you're updating them?

src/utils/utils.ts
bigfishdesign13: That's a good suggestion.  

@oliviawongnyc Is this something you can do?  This goes back to that idea of making the component as flexible as possible.

oliviawongnyc: This is done!

oliviawongnyc:
> ...that's odd. I wouldn't expect that. The snapshot tests are failing even if you're updating them?

@EdwinGuzman and passing locally... we'll see on this latest run what happens

EdwinGuzman (approved):
Looks good but don't merge until a11y test has been done.

PR #1434, opened 2023-10-04
DR-2525 FeaturedContent Component

7emansell:
Fixes JIRA ticket [DR-2525](https://jira.nypl.org/browse/DR-2525).

- Adds the `FeaturedContent` component
- Adds testing and documentation for the `FeaturedContent` component

## How has this been tested?

Local Storybook and Turbine.

N/A.

### Checklist:

### Front End Review:


src/components/FeaturedContent/FeaturedContent.test.tsx
```diff
@@ -0,0 +1,221 @@
+import { render, screen } from "@testing-library/react";
+import { axe } from "jest-axe";
+import renderer from "react-test-renderer";
+
+import FeaturedContent from "./FeaturedContent";
+import Button from "../Button/Button";
+import Heading from "../Heading/Heading";
+import Text from "../Text/Text";
+export const textContent = (
+  <>
```
EdwinGuzman: Nitpick: you don't technically need the react fragment in this case because this is all wrapped in a div.

src/components/FeaturedContent/FeaturedContent.test.tsx
```diff
@@ -0,0 +1,221 @@
...
+    <div>
+      <br />
+      <Text size="overline1"> Featured </Text>
+      <Heading level="h1"> Spotlight on the Public Domain </Heading>
```
EdwinGuzman: We _just_ added this recently and didn't bring it up but you can get a similar style if you use the `overline` prop in the `Heading` component.

```
<Heading level="h1" overline="Featured">Spotlight....
```

src/components/FeaturedContent/FeaturedContent.tsx
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * The FeaturedContentImage component is used internally in the `FeaturedContent` component. It
+ * renders an `Image` component but with overriding styles specific to the
+ * `FeaturedContent` component.
+ */
+function FeaturedContentImage(
+  props: React.ComponentProps<"img"> & FeaturedContentImageProps
+) {
+  const { alt, src } = props;
+  return <img alt={alt} src={src} style={{ display: "none" }} />;
```
EdwinGuzman: Not necessary to make this update but just pointing this out:

`style={{ display: "none" }}` is correct and you can also achieve the same style through Chakra's style props as `display="none"`.

src/components/FeaturedContent/FeaturedContent.tsx
```diff
@@ -0,0 +1,112 @@
...
+          alt: "",
+          position: "end",
+          width: "oneHalf",
+          src: "",
+        },
+      } = props;
+      const styles = useMultiStyleConfig("FeaturedContent", {
+        imagePosition: imageProps.position,
+        imageWidth: imageProps.width,
+        isFullWidth: isFullWidth,
```
EdwinGuzman: You can use the JS shorthand here for an object that has a key name the same as a variable name. So this could be 
```
{
  imagePosition: imageProps.position,
  imageWidth: imageProps.width,
  isFullWidth,
}
```

JS will know you are using the value from the `isFullWidth` variable and that the key in the object has the same name.

src/components/FeaturedContent/FeaturedContent.tsx
EdwinGuzman: Just curious, did you try using the DS `Image` component and found that didn't scale as well as this `img` element?

src/theme/components/featuredContent.ts
```diff
@@ -0,0 +1,63 @@
...
+const FeaturedContent = {
+  baseStyle: ({
+    imagePosition,
+    imageWidth,
+    isFullWidth,
+  }: FeaturedContentBaseStyleProps) => {
+    let imageAtEnd = imagePosition === "end";
+    let full = isFullWidth === true;
+    let wrapperWidth = "50%";
+    if (imageWidth === "oneQuarter") {
```
EdwinGuzman, line 19: This works fine for this use case because it's unlikely that new image width values will be added, so you _don't_ need to make an update.
If there's an update and we want 3 new image widths, this may grow unmanageable. Another pattern is to create an object to get the value.

```
const imageWidthsValues = {
  oneQuarter: "25%",
  oneThird: "33%",
  ...
};

let wrapperWidth = imageWidthsValues[imageWidth];
```
(the default should still be set somewhere).

EdwinGuzman (commented):
The component looks good! I pointed out a few minor updates that could be made but are technically not necessary or blockers for the component.
I did not review the `.stories.tsx` or `.mdx` files because I know you're going to work on that next.

The only question and potential update that is needed is using a DS `Image` component within the `FeaturedContentImage` component. If that works out, great. If not that's fine but let us know. There are additional wrapper elements we added for styling that may make it hard to work with. I'd be okay with using an html `img` element for this component if it becomes difficult to style.

EdwinGuzman:
Forgot to mention: you can update snapshot tests by running `npm test -- --updateSnapshot` and committing the updated snapshot files.

src/components/FeaturedContent/FeaturedContent.tsx
7emansell: Yes, so I originally tried with the`Image` component but like you noted, there's props on that that interfered with the variable width display in FeaturedContent. Using the `img` element allowed me to style it much more straightforwardly (though I can definitely see the merit of reusing components wherever possible).

EdwinGuzman: That makes sense. We've encountered similar friction in other components when we had custom values for props that expect a certain set of values (especially with color and background color).

In this case, `imageProps.width` is used for the wrapper and the theme object but not in the image element itself. I think the intersection type for `props` is causing the conflict for the `width` HTML attribute and react prop. Try removing `React.ComponentProps<"img">`, I'm not sure it's needed if this only needs src and alt. Let's also go through this together at some point.

7emansell: This does work the same without `React.ComponentProps<"img">`, so just took it out. It would definitely be helpful to step through each layer of this with someone familiar with the `Image` component.

src/components/FeaturedContent/FeaturedContent.tsx
```diff
@@ -43,7 +43,7 @@ export interface FeaturedContentProps {
  */
 function FeaturedContentImage(props: FeaturedContentImageProps) {
   const { alt, src } = props;
-  return <img alt={alt} src={src} style={{ display: "none" }} />;
+  return <Image alt={alt} src={src} style={{ display: "none" }} />;
```
7emansell: @EdwinGuzman Now using the `Image` component, though still needs to have `style={{ display="none }}` as far as I can tell. This isn't related to the `imageProps` warning failing to appear though, which is strange– even if I take the `style={{ display="none }}` off and have `img: screenReaderOnly()`, the test is only catching that `imageProps.alt` and `imageProps.src` aren't there, not that `imageProps` isn't on the component at all.

src/components/FeaturedContent/FeaturedContent.mdx
```diff
@@ -0,0 +1,77 @@
...
+import * as FeaturedContentStories from "./FeaturedContent.stories";
+import Link from "../Link/Link";
+
+<Meta of={FeaturedContentStories} />
+
+# Featured Content
+
+| Component Version | DS Version |
+| ----------------- | ---------- |
+| Added             | ``         |
```
EdwinGuzman: Our convention is to use `Prerelease` for updates for "Latest". Since this is new, it should also go in the "Added" row.

src/components/FeaturedContent/FeaturedContent.mdx
```diff
@@ -0,0 +1,77 @@
...
+
+The `FeaturedContent` component provides a method to visually emphasize a text block with an image, within a full page layout.
+
+## Component Props
+
+<Canvas of={FeaturedContentStories.WithControls} />
+
+<ArgTypes of={FeaturedContentStories.WithControls} />
+
+## Accessibility
```
EdwinGuzman, line 36: 👍

src/components/FeaturedContent/FeaturedContent.stories.tsx
```diff
@@ -0,0 +1,375 @@
...
+ * Main Story for the FeaturedContent component. This must contain the `args`
+ * and `parameters` properties in this object.
+ */
+
+export const WithControls: Story = {
+  args: {
+    className: undefined,
+    id: "FeaturedContent-id",
+    isFullWidth: false,
+    //textContent: (
```
EdwinGuzman: You can remove commented out code and in this case, I think `textContent: undefined,` is okay to have.

src/components/FeaturedContent/FeaturedContent.stories.tsx
```diff
@@ -0,0 +1,375 @@
...
+    "imageProps.alt": "Alt text",
+    "imageProps.position": "end",
+    "imageProps.width": "default",
+    "imageProps.src": "//placekitten.com/600/600",
+  },
+  render: (args) => (
+    <FeaturedContent
+      className={args.className}
+      id={args.id}
+      isFullWidth={args.isFullWidth}
```
EdwinGuzman: You _could_ make these three lines `{...args}` to get a similar result. It's a shorthand and any other prop defined after it will override whatever is in `{...args}`.

src/components/FeaturedContent/FeaturedContent.stories.tsx
```diff
@@ -0,0 +1,375 @@
...
+          <Button id="test"> Discover more </Button>
+        </div>
+      }
+      imageProps={{
+        alt: args["imageProps.alt"],
+        width: args["imageProps.width"],
+        position: args["imageProps.position"],
+        src: args["imageProps.src"],
+      }}
+    ></FeaturedContent>
```
EdwinGuzman: Nitpick: since this component takes no children, it can be a self-closing component `<FeaturedContent .... />`.

src/components/FeaturedContent/FeaturedContent.tsx
```diff
@@ -0,0 +1,112 @@
...
+      if (!textContent) {
+        console.warn(
+          `NYPL Reservoir FeaturedContent: The 'textContent' prop is required.`
+        );
+      }
+
+      return (
+        <Box data-testid="featuredcontent" __css={styles.wrapper}>
+          <Box
+            data-testid="bg-image"
```
EdwinGuzman: This is not really a problem and it _should_ not be one in the future, but `bg-image` is generic enough that a consuming app could use that in their own tests. Namespace this so it's specific to this component `featuredcontent-bg-image`.

src/theme/components/featuredContent.ts
```diff
@@ -0,0 +1,65 @@
...
+        padding: "20px",
+        flex: 1,
+      },
+      imgWrapper: {
+        backgroundPosition: "center",
+        backgroundSize: "cover",
+        height: { sm: "320px", md: "auto" },
+        width: { sm: "100%", md: wrapperWidth },
+      },
+      img: screenreaderOnly(),
```
EdwinGuzman, line 62: This might be because of the top level node styles. You can:
1 - update the top level wrapper element to use `__css={styles}` instead of `__css={styles.wrapper}`. Then in this file you can remove the `wrapper: { ... }` object but still keep the styles inside of it, promoting them to the higher level.
2 - move line 60 `img: screenreaderOnly(),` inside of `imageWrapper`. This gives it more specificity and the nests works like css/scss nesting.

The first approach of using the direct `styles` object in the component file and having top level styles in this theme file is more conventional.

src/components/FeaturedContent/FeaturedContent.tsx
EdwinGuzman: I added a comment in `featuredContent.ts` that should allow you to remove the display none here.

You're right. If we assign the `imageProps` default object with values, then we'll never get to a point where a missing `imageProps` value will log the warning. Let's remove that and only keep the warnings for missing `imageProps.alt` and `imageProps.src`.

EdwinGuzman (commented):
Some minor comments. This is almost ready:
- Try updating how the `styles` object is set in the component and then remove the `display: "none"` from the Image component
- remove the test and console warning for the undefined `imageProps` value
- and export `FeaturedContent` and the type `FeaturedContentProps` in the `src/index.ts` file.

src/theme/components/featuredContent.ts
7emansell: First approach works great, thank you for looking into this!!

src/theme/index.ts
```diff
@@ -104,6 +105,7 @@ const theme: any = {
     CustomSelect,
     DatePicker,
     FeedbackBox,
+    FeaturedContent,
```
7emansell, line 108: @EdwinGuzman I'm not seeing an example of something like `FeaturedContentProps` being imported into this file, did you mean what I have here?

7emansell:
> Some minor comments. This is almost ready:
> 
> * Try updating how the `styles` object is set in the component and then remove the `display: "none"` from the Image component
> * remove the test and console warning for the undefined `imageProps` value
> * and export `FeaturedContent` and the type `FeaturedContentProps` in the `src/index.ts` file.

7emansell:
Accidentally closed this sorry!

src/theme/index.ts
EdwinGuzman: Ah, this is okay for the theme file import, but I meant for the component and type import/export in `src/index.ts`. Both files are named `index.ts` so it's confusing. The one in the `src` directory is to export the component for use in a consuming app. This one in `src/theme` is for Chakra's use.

7emansell: Ohh sorry should have read more carefully!

src/components/FeaturedContent/FeaturedContent.stories.tsx
```diff
@@ -0,0 +1,351 @@
...
+    "imageProps.position": "end",
+    "imageProps.width": "default",
+    "imageProps.src": "//placekitten.com/600/600",
+  },
+  render: (args) => (
+    <FeaturedContent
+      {...args}
+      textContent={
+        <div>
+          <Heading level="h1" overline="Featured">
```
EdwinGuzman: Sorry, I missed this before. Update the `"h1"` to `"h2"` for all the examples. In an app, the "h1" is for the page title rather than the heading here, and it should only be used once on a page. Using "h2" is safer but it also depends on the app and it's heading hierarchy.

vite.config.ts.timestamp-1696957646854-ce882d3e67223.mjs
```diff
@@ -0,0 +1,59 @@
+// vite.config.ts
```
EdwinGuzman: This is odd and I'm not sure how it got in here (e.g. what command created this file). This shouldn't be needed so it's okay to remove.

7emansell: Mysterious

src/theme/components/featuredContent.ts
```diff
@@ -0,0 +1,66 @@
...
+        maxWidth: "1280px",
+        display: "flex",
+        alignItems: "stretch",
+        flexDirection: imageAtEnd
+          ? { sm: "column-reverse", md: "row-reverse" }
+          : { sm: "column", md: "row" },
+      },
+      text: {
+        display: "flex",
+        padding: "20px",
```
bigfishdesign13: Change this to `padding: "l",`

src/theme/components/featuredContent.ts
```diff
@@ -0,0 +1,66 @@
...
+      wrapperWidth = "25%";
+    } else if (imageWidth === "oneThird") {
+      wrapperWidth = "33%";
+    } else if (imageWidth === "twoThirds") {
+      wrapperWidth = "66%";
+    } else if (imageWidth === "threeQuarters") {
+      wrapperWidth = "75%";
+    }
+    return {
+      bgColor: "ui.gray.x-light-cool",
```
bigfishdesign13: Change this to `bgColor: "ui.bg.default",`

src/theme/components/featuredContent.ts
```diff
@@ -0,0 +1,66 @@
...
+      _dark: {
+        bgColor: "dark.ui.bg.default",
+      },
+      width: full ? "100vw" : "100%",
+      left: full ? "50%" : "auto",
+      right: full ? "50%" : "auto",
+      paddingX: full ? { "2xl": "5%" } : "auto",
```
bigfishdesign13: Let's talk about this.

src/theme/components/featuredContent.ts
```diff
@@ -0,0 +1,66 @@
...
+        display: "flex",
+        alignItems: "stretch",
+        flexDirection: imageAtEnd
+          ? { sm: "column-reverse", md: "row-reverse" }
+          : { sm: "column", md: "row" },
+      },
+      text: {
+        display: "flex",
+        padding: "20px",
+        flex: 1,
```
bigfishdesign13, line 52: Add `flexDirection: "column",`.  This will ensure that the text elements are always stacked even if the elements in the incoming `textContent` prop are not wrapped in a `div`.

Add `justifyContent: "center",`.  This will veritically center the text content.

src/theme/components/featuredContent.ts
```diff
@@ -0,0 +1,66 @@
...
+      left: full ? "50%" : "auto",
+      right: full ? "50%" : "auto",
+      paddingX: full ? { "2xl": "5%" } : "auto",
+      position: "relative",
+      marginLeft: full ? "-50vw" : "auto",
+      marginRight: full ? "-50vw" : "auto",
+      wrapper: {
+        ...wrapperStyles,
+        minHeight: "320px",
+        maxWidth: "1280px",
```
bigfishdesign13: Change this to `maxWidth: full ? "1280px" : "100%",`

We want the default component to expand to the full width of its parent container.  The 1280px restriction should only come into play when the component is set to be full width.

src/components/FeaturedContent/FeaturedContent.stories.tsx
```diff
@@ -0,0 +1,351 @@
...
+};
+
+export const textContentVariations: Story = {
+  render: () => (
+    <SimpleGrid columns={1}>
+      <FeaturedContent
+        isFullWidth={false}
+        textContent={
+          <div>
+            <p>
```
bigfishdesign13: It would be best to change this to a `Text` component and then use the following...
```
<Text noSpace>...</Text>
```

src/components/FeaturedContent/FeaturedContent.mdx
```diff
@@ -0,0 +1,77 @@
...
+
+The image width variations of the `FeaturedContent` component can be rendered through the `imageProps.width` prop.
+The options for width are `oneQuarter`, `oneThird`, `oneHalf`, `twoThirds`, and `threeQuarters`. The default width is `oneHalf`.
+
+<Canvas of={FeaturedContentStories.imageWidthVariations} />
+
+## Text Content
+
+`textContent` is a string or JSX element passed into `FeaturedContent`, so accessibility standards and spacing should be considered independently of the `FeaturedContent` styling.
+The above example with an overline, a heading, two lines of body text, and a button is the recommended usage, though other uses are possible. Keep in mind that the minimum height of `FeaturedContent` is 320px.
```
bigfishdesign13: Let's emphasize this recommendation even more.  Move this sentence to a new line and update with the following:

```
The above example with an overline, a heading, a short paragraph of body text, and a CTA button is the recommended usage. While other configurations are possible, keep in mind that the minimum height of the component is 320px and it is best to avoid excessive white space.
```

src/components/FeaturedContent/FeaturedContent.stories.tsx
```diff
@@ -0,0 +1,351 @@
...
+          position: "end",
+          src: "//placekitten.com/600/600",
+        }}
+      />
+      <FeaturedContent
+        isFullWidth={false}
+        textContent={
+          <div>
+            <Heading level="h2">Not recommended</Heading>
+            <p>This text is not tall enough for this component.</p>
```
bigfishdesign13: Change this to the following:
```
<Heading level="h2" size="heading5">Not recommended</Heading>
<Text noSpace>This does not fill the space and could use more.</Text>
```

bigfishdesign13 (changes requested):
This looks great.  I just had a few comments about some of the styles and some of the docs.  I also want to touch base about 5% padding that was added for very large screens.

bigfishdesign13 (approved):
This looks great!

EdwinGuzman (approved):

EdwinGuzman (approved):
The last update with the Changelog table looks good. Re-approving.

There's the merge conflict that should be fixed before merging. Once that's resolved, feel free to merge this in.

PR #1436, opened 2023-10-06
DSD-1603: component changelogs

bigfishdesign13:
Fixes JIRA ticket [DSD-1603](https://jira.nypl.org/browse/DSD-1603)

- Adds the `ComponentChangelogTable` component.
- Adds a changelog to the story pages for the `DatePicker`, `FeedbackBox`, `Hero`, and `Slider` components.

## How has this been tested?

- local Storybook

- n/a

### Checklist:

- [] I have added relevant accessibility documentation for this pull request.

### Front End Review:

### DIFF
--- a/src/components/DatePicker/datePickerChangelogData.ts
+++ b/src/components/DatePicker/datePickerChangelogData.ts
@@ -3,15 +3,17 @@
  * date: string (when adding new entry during development, set value as "Prerelease")
  * version: string (when adding new entry during development, set value as "Prerelease")
  * type: "Bug Fix" | "New Feature" | "Update";
- * affects: "Accessibility" | "Documentation" | "Functionality" | "Styles";
+ * affects: array["Accessibility" | "Documentation" | "Functionality" | "Styles"];
  * notes: array (will render as a bulleted list, add one array element for each list element)
  */
-export const changelogData = [
+import { ChangelogData } from "../../utils/ComponentChangelogTable";
+
+export const changelogData: ChangelogData[] = [
   {
     date: "Prerelease",
     version: "Prerelease",
     type: "Update",
-    affects: "Accessibility",
+    affects: ["Accessibility"],
     notes: [
       "Updates to pass a secondaryHelperTextId to its TextInput if needed so that the aria-describedby value can be associated with all relevant helperTexts.",
       "Updates so that focus remains on input after value is changed.",
@@ -21,7 +23,7 @@ export const changelogData = [
     date: "2023-9-28",
     version: "2.0.0",
     type: "Update",
-    affects: "Styles",
+    affects: ["Styles"],
     notes: [
       "Applied Typo2023 styles, including font size, font color, and text link patterns.",
     ],
--- a/src/components/DatePicker/datePickerChangelogData.ts
+++ b/src/components/DatePicker/datePickerChangelogData.ts
@@ -1,0 +1,29 @@
+/** This data is used to populate the ComponentChangelogTable component.
+ *
+ * date: string (when adding new entry during development, set value as "Prerelease")
+ * version: string (when adding new entry during development, set value as "Prerelease")
+ * type: "Bug Fix" | "New Feature" | "Update";
+ * affects: "Accessibility" | "Documentation" | "Functionality" | "Styles";
+ * notes: array (will render as a bulleted list, add one array element for each list element)
+ */
+export const changelogData = [
+  {
+    date: "Prerelease",
+    version: "Prerelease",
+    type: "Update",
+    affects: "Accessibility",
+    notes: [
+      "Updates to pass a secondaryHelperTextId to its TextInput if needed so that the aria-describedby value can be associated with all relevant helperTexts.",
+      "Updates so that focus remains on input after value is changed.",
+    ],
+  },
+  {
+    date: "2023-9-28",
+    version: "2.0.0",
+    type: "Update",
+    affects: "Styles",
+    notes: [
+      "Applied Typo2023 styles, including font size, font color, and text link patterns.",
+    ],
+  },
+];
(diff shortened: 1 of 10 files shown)


bigfishdesign13:
I flagged everyone on this PR so we're all on the same page going forward.  As we update components (or add them anew), we will add and update a changelog for each individual component.  Refer to the story pages that I've updated in this PR.

@aarnold101 and @7emansell – Once this is merged into `development`, you can pull it and then go ahead and add the appropriate code to your new components.

src/utils/ComponentChangelogTable.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+export interface ComponentChangelogTableProps {
+  changelogData: any[];
+}
+export interface TableRowProps {
+  date: string;
+  version: string;
+  type: "Bug Fix" | "New Feature" | "Update";
+  affects: "Accessibility" | "Documentation" | "Functionality" | "Styles";
+  notes: any[];
```
EdwinGuzman: Should be `string[]`.

src/utils/ComponentChangelogTable.tsx
```diff
@@ -0,0 +1,99 @@
...
+}
+
+export const ComponentChangelogTable = (
+  props: PropsWithChildren<ComponentChangelogTableProps>
```
EdwinGuzman: `PropsWithChildren` can be removed, this component takes no children node.

src/utils/ComponentChangelogTable.tsx
```diff
@@ -0,0 +1,99 @@
...
+) => {
+  const { changelogData } = props;
+  const tableData = [];
+
+  changelogData.forEach((tableRow: PropsWithChildren<TableRowProps>) => {
```
EdwinGuzman: You can make this a `.map` and assign it directly to `tableData`. That way you can `return rowData;` instead of pushing to a separate array.

src/utils/ComponentChangelogTable.tsx
```diff
@@ -0,0 +1,99 @@
+import { PropsWithChildren } from "react";
+
+import List from "../components/List/List";
+import Table from "../components/Table/Table";
+
+export interface ComponentChangelogTableProps {
+  changelogData: any[];
```
EdwinGuzman: `ChangelogData` is a better name than for what is currently `TableRowProps`. This could be

```
export interface ChangelogData {
  date: string;
  version: string;
  type: "Bug Fix" | "New Feature" | "Update";
  affects: "Accessibility" | "Documentation" | "Functionality" | "Styles";
  notes: any[];
}

export interface ComponentChangelogTableProps {
  changelogData: ChangelogData[];
}
```

Then line 23 can be updated to `changelogData.forEach((tableRow: ChangelogData) => {`.

Right now, the `changelogData` property in `props` doesn't have any type information since this is `any[]`.

src/utils/ComponentChangelogTable.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+export const ComponentChangelogTable = (
+  props: PropsWithChildren<ComponentChangelogTableProps>
+) => {
+  const { changelogData } = props;
+  const tableData = [];
+
+  changelogData.forEach((tableRow: PropsWithChildren<TableRowProps>) => {
+    const { date, version, type, affects, notes } = tableRow;
+    const dateFormatted = new Date(date).toLocaleDateString("en-us", {
```
EdwinGuzman: I would change how the date is displayed because the "Invalid Date" is confusing and also setting `"???"` as the default. Can the dev input also be `"Prerelease"` like the version?

A logic check can be set here, if the input is "Prerelease" then keep it and display it, otherwise, it's a date string so run it through `new Date(...)`. Then we can show "Prerelease" or the date instead of "Invalid Date".
 
Screen Shot 2023-10-09 at 3 43 08 PM

EdwinGuzman (changes requested):
Some updates but really like how it looks. As we add more updates in the changelog, the `.mdx` files will grow long and unmaintainable. I'm thinking we can have an additional file called something like `[component]-changelog.ts` that exports its own `changelogData` array for the `ComponentChangelogTable` component.

So for DatePicker, it would be `DatePicker-changelog.ts`. What do you think?

src/components/DatePicker/datePickerChangelogData.ts
```diff
@@ -0,0 +1,29 @@
+/** This data is used to populate the ComponentChangelogTable component.
+ *
+ * date: string (when adding new entry during development, set value as "Prerelease")
+ * version: string (when adding new entry during development, set value as "Prerelease")
+ * type: "Bug Fix" | "New Feature" | "Update";
+ * affects: "Accessibility" | "Documentation" | "Functionality" | "Styles";
+ * notes: array (will render as a bulleted list, add one array element for each list element)
+ */
+export const changelogData = [
```
EdwinGuzman: You can define the type to `ChangelogData` since that's already declared and could be used here to give better guidance for the `type` and `affects` values.

EdwinGuzman (approved):
The update looks good. The only minor comment is to type the `changelogData` arrays in each data file so it's easier for devs to know what each property expects.

bigfishdesign13:
@EdwinGuzman I think the changes in my last update were significant enough to necessitate one more review.

src/utils/ComponentChangelogTable.tsx
```diff
@@ -0,0 +1,113 @@
...
+      fontSize: "desktop.caption",
+      _last: {
+        borderColor: borderColor,
+      },
+    },
+    td: {
+      borderColor: borderColor,
+      fontSize: "desktop.caption",
+      // pb: "xs",
+      // pt: "xs",
```
EdwinGuzman: Okay to remove.

EdwinGuzman (approved):

7emansell (approved):
Looks great to me

PR #1444, opened 2023-10-20
DSD-1579: Adds aria-label to input for DatePicker

oliviawongnyc:
Fixes JIRA ticket [DSD-1579](https://jira.nypl.org/browse/DSD-1579)

- Updates `DatePicker`'s `TextInput` to always have an `aria-label` attribute that tells screen reader users how to access the calendar.

## How has this been tested?

- Locally, in Storybook
- RTL tests needed to be updated

- Accessibility improvement

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/DatePicker/DatePicker.test.tsx
+++ b/src/components/DatePicker/DatePicker.test.tsx
@@ -105,7 +105,10 @@ describe("DatePicker", () => {
       );
 
       expect(input).toBeInTheDocument();
-      expect(input).not.toHaveAttribute("aria-label");
+      expect(input).toHaveAttribute(
+        "aria-label",
+        "Press tab to access the calendar."
+      );
       // Date format based on component specification yyyy-mm-dd.
       expect(date).toEqual(`${year}-${month}-${day}`);
       expect(screen.getByDisplayValue(date)).toBeInTheDocument();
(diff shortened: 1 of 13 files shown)


src/components/DatePicker/DatePicker.test.tsx
```diff
@@ -596,20 +599,21 @@ describe("DatePicker", () => {
       // Helper text for the "To" input
       expect(screen.getByText(/Note for the 'to' field./i)).toBeInTheDocument();
 
-      const fromInput = screen.getByRole("textbox", { name: "From" });
-      const toInput = screen.getByRole("textbox", { name: "To" });
+      const inputs = screen.getAllByRole("textbox", {
+        name: "Press tab to access the calendar.",
```
EdwinGuzman: You mentioned this before about the order of precedence that RTL uses to get by role, can you include a comment that references that?

EdwinGuzman (approved):
Looks good, let's get Clare to review this before it's merged.

src/components/DatePicker/DatePicker.test.tsx
oliviawongnyc: I added a comment on this in utils/utils.ts. Let me know if you think that is sufficient.

src/components/DatePicker/DatePicker.test.tsx
```diff
@@ -596,20 +599,24 @@ describe("DatePicker", () => {
       // Helper text for the "To" input
       expect(screen.getByText(/Note for the 'to' field./i)).toBeInTheDocument();
 
-      const fromInput = screen.getByRole("textbox", { name: "From" });
-      const toInput = screen.getByRole("textbox", { name: "To" });
+      const inputFrom = screen.getByRole("textbox", {
+        name: "From - Note for the 'from' field., Press tab to access the calendar.",
```
EdwinGuzman: The extra comma after the period is odd. It's not really an issue.

src/utils/utils.ts
```diff
@@ -69,16 +71,32 @@ export const getAriaAttrs = ({
 }: GetAriaAttrsProps): AriaAttributes => {
   let ariaAttributes: AriaAttributes = {};
 
+  // If both a label and an aria-label are present, screen
+  // readers will only read the aria-label so we need to
+  // provide all necessary details in the aria-label.
+  if (additionalAriaLabel) {
+    ariaAttributes["aria-label"] = `${
+      labelText && footnote
```
EdwinGuzman: This update makes sense. Can you make 
```
labelText && footnote
        ? `${labelText} - ${footnote}`
        : (labelText as string)
```
into a variable since it appears twice in this file?

At some point, this work and Clare's explanation should go into a standalone accessibility document. Also to help devs understand when testing what the appropriate RTL function to use should be.

EdwinGuzman (approved):
Looks good! One minor comment/update but not a blocker.

src/components/DatePicker/DatePicker.test.tsx
oliviawongnyc: I agree it's a bit weird, but the reason I added a comma is that there is not necessary a period after the footnote. In this case, the `aria-label` might look like this, "From - Note for the 'from' field Press tab to access the calendar." Adding the comma ensures the screen reader pauses before reading the "Press tab..." piece. Is there another way to do this? Or do you feel it's not important that the screen reader pause between these two pieces of information?

EdwinGuzman: You could read the last character in the string and check for a period, but tbh it seems like more effort than is needed. The premeditated comma seems fine. It'd be good to see how others feel about it.

EdwinGuzman (approved):

PR #1448, opened 2023-10-24
DSD-1631/DSD-1635: upgrade to Node 18, React 18, and Chakra 2.8

EdwinGuzman:
JIRA Ticket [DSD-1631](https://jira.nypl.org/browse/DSD-1631)

Updates:
- Upgrades to node 18
- Upgrades to React 18
- Upgrades to Chakra 2.8
- Updates types for all components but is set to `React.FC<any>` for now.
- Components that take in children must be typed as `React.FC<React.PropsWithChildren<any>>`.
- `Tooltip` tests were updated from `fireEvent.mouseOver` to `fireEvent.pointerOver`.
- Temporarily removed some of the welcome docs link because the syntax was breaking the page.
- For breakpoints, the `createBreakpoints` function is deprecated and was removed in `src/theme/foundations/breakpoints.ts`.
- The custom `internalChakraProvider` is no longer needed. This was added for disabling custom global CSS. Since this is now a part of Charka, we can just use their implementation (which we copied when using Chakra 1.8 anyway).

Updates that were made for a few components but needs to be done for more:
- In the NYPL custom theme file, we cannot have the same name for a component style as an existing Chakra component. For example, Chakra has its own `Button` component in the theme file and it had to be renamed as `CustomButton` in order for the styles to be picked up. This may need to be done for more components but for now it was done for `CustomBreadcrumb`, `CustomButton`,  and `CustomIcon`.
- Chakra added better type definitions for their styles but in order to do so and take advantage of them, we need to use the new `createMultiStyleConfigHelpers`. It appears that we MUST make this update whenever we use the `useMultiStyleConfig` function in DS components but NOT for the `useStyleConfig`. See `src/theme/components/button.ts` and `src/theme/components/link.ts` for examples. More tickets will be created to update other components in separate PRs.


src/theme/index.ts
```diff
@@ -94,8 +94,8 @@ const theme: any = {
     Accordion,
     AlphabetFilter,
     AudioPlayer,
-    Breadcrumb,
-    Button,
+    CustomBreadcrumb: Breadcrumb,
+    CustomButton: Button,
```
bigfishdesign13, line 101: As you noted, there are a number of other components where we use the same name as the Chakra component.  Is the name update for these two components only necessary in this PR becuase the styles have also been updated?

src/components/Table/Table.tsx
```diff
@@ -178,9 +178,11 @@ export const Table = chakra(
           sx={styles}
           {...rest}
         >
-          {tableCaption}
-          {columnHeadersElems}
-          {tableBodyElems()}
+          <>
```
bigfishdesign13, line 187: Why was this necessary?

src/theme/components/link.ts
```diff
@@ -8,211 +10,233 @@ import {
...
-  textDecorationStyle: "dotted",
-  textDecorationThickness: "1px",
-  textUnderlineOffset: "2px",
-  _dark: {
-    color: "dark.ui.link.primary",
-  },
-  _hover: {
-    color: "ui.link.secondary",
+const { defineMultiStyleConfig, definePartsStyle } =
+  createMultiStyleConfigHelpers(["base", "screenreaderOnly"]);
```
bigfishdesign13, line 14: Add the comments for this function as you did in `button.ts`.

bigfishdesign13 (approved):
This looks good.  I tried to scan the changes in every file, but (as you well know) there are a lot of changes.

src/theme/index.ts
bigfishdesign13: Also, I wonder if it would be better to use `ReservoirButton` as the pattern.  What do you think?

EdwinGuzman: For `Breadcrumbs`, I didn't make updates in this PR but doing this made the styles propagate correctly. I think once we get to updating `breadcrumbs.ts`, that we'll find we don't need to do `CustomBreadcrumbs`.

If we absolutely _need_ to rename properties, the `Reservoir` prefix is a great pattern.

src/components/Table/Table.tsx
EdwinGuzman: There was a typescript issue. `Table` probably updated in Chakra.

src/theme/components/link.ts
EdwinGuzman: Can do this when the `Link` component goes through a better update.

PR #1457, opened 2023-11-02
DSD-1618: autoComplete prop for TextInput

bigfishdesign13:
Fixes JIRA ticket [DSD-1618](https://jira.nypl.org/browse/DSD-1618)

- Added the `autoComplete` prop to the `TextInput` component.

## How has this been tested?

- local Storybook

- The `autoComplete` prop will provide additional methods to improve a11y in consuming apps.

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/TextInput/TextInput.mdx
+++ b/src/components/TextInput/TextInput.mdx
@@ -80,11 +80,12 @@ Resources:
 
 ## Autocomplete
 
-The `autocomplete` attribute is another tool that can be used to improve
-accessibility. It is a native HTML attribute that improves the browser's ability
-to pre-populate form fields with user-preferred values and makes inputs easier
-and more efficient to complete for all users. For ease of use, the `TextInput`
-component provides a few methods for incorporating the `autocomplete` attribute.
+The native HTML `autocomplete` attribute is another tool that can be used to
+improve accessibility. The `autocomplete` attribute improves the browser's
+ability to pre-populate form fields with user-preferred values and makes inputs
+easier and more efficient to complete for all users. For ease of use, the
+`TextInput` component provides a few methods for incorporating the
+`autocomplete` attribute.
 
 If the `type` prop is set to `"email"`, `"tel"`, or `"url"`, the component will
 automatically add the `autocomplete` attribute with an appropriate value.
--- a/src/components/TextInput/TextInput.mdx
+++ b/src/components/TextInput/TextInput.mdx
@@ -9,16 +9,17 @@ import { changelogData } from "./textInputChangelogData";
 
 # TextInput
 
-| Component Version | DS Version |
-| ----------------- | ---------- |
-| Added             | `0.22.0`   |
-| Latest            | `2.1.0`    |
+| Component Version | DS Version   |
+| ----------------- | ------------ |
+| Added             | `0.22.0`     |
+| Latest            | `Prerelease` |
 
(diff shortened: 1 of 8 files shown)


src/components/TextInput/TextInput.mdx
```diff
@@ -70,18 +71,41 @@ When the `type` prop is set to `"textarea"`, the `<textarea>` element
...
 
 - [MDN input: The Input (Form Input) element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input)
 - [MDN textarea: The Textarea element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea)
-- [MDN HTML attribute: Autocomplete](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete)
 - [Chakra UI Input](https://chakra-ui.com/docs/components/form/input)
 - [Chakra UI Textarea](https://chakra-ui.com/docs/components/form/textarea)
 
+## Autocomplete
+
+The `autocomplete` attribute is another tool that can be used to improve
```
EdwinGuzman: Can you add "native HTML `autocomplete` attribute"? To differentiate it between the native html attribute and the React-based prop (which is the camelcase one).

src/components/TextInput/TextInput.tsx
```diff
@@ -247,7 +309,11 @@ export const TextInput = chakra(
           }
         : {
             "aria-required": isRequired,
-            autoComplete: hasAutocomplete ? type : null,
+            autoComplete: hasAutocomplete
```
EdwinGuzman, line 317: I originally thought this was more complicated than it should be but then there's the logic check on line 255 that makes this clearer. Can you add a comment to describe what's going on? The confusing thing for me was that this could be true (based on the logic on line 255) but that `autoComplete` could be false if nothing was passed. In that case it would be the `type` value.

EdwinGuzman (commented):
Just some suggestions but looks good overall.

src/components/TextInput/TextInput.mdx
bigfishdesign13: I did have that in the next sentence, but I moved it up to the first sentence to help clarify the point right away.

src/components/TextInput/TextInput.tsx
```diff
@@ -247,7 +309,16 @@ export const TextInput = chakra(
...
+            /** If the `autoComplete` prop is passed, that value will take
+             * precedence and will be used here. Otherwise, a value will be set
+             * based on the `type` prop. Lastly, if `autoComplete` is not passed
+             * and a default value is not set based on the `type` prop, then
+             * `autoComplete` will not be set for the input. */
+            autoComplete: hasAutocomplete
+              ? autoComplete
+                ? autoComplete
+                : type
+              : null,
```
bigfishdesign13, line 321: @EdwinGuzman Would this be better as `undefined`?

EdwinGuzman: I think either `null` or `undefined` work for attributes we don't want to render. I thought I read that react recommends `undefined` but I can't find any info on that. So this should be okay.

bigfishdesign13: Ok, I will leave it as is.

EdwinGuzman (approved):

PR #1458, opened 2023-11-07
DSD-1607: Updates zIndex for searchbar select icon

oliviawongnyc:
Fixes JIRA ticket [DSD-1607](https://jira.nypl.org/browse/DSD-1607)

- Adds a z-index on hover to the `SearchBar`'s select icon so it no longer disappears.
- Notes for reviewers: 
1. In the `Select` theme file, we are using the same technique for this issue, but we are adding it on focus rather than on hover. Is there a reason the select in the `Searchbar` should be different than the `Select` component itself?
2. Do we have a philosophy for our zIndex values? It seems like in many places we use 9999, in others we use 5 or 10.

## How has this been tested?

- Locally, in Storybook.

- I am wondering if the `SearchBar`'s select element is keyboard navigable. I can focus on it, but if I press "enter" it seems like it runs a search with an empty search value, rather than exposing the drop down. Unless there is another way keyboard users are meant to access it.

### Checklist:

### Front End Review:


EdwinGuzman (commented):
This looks good. I don't think the `SearchBar` has a storybook "Changelog" section. Can you add it? Here's an example of the data: https://github.com/NYPL/nypl-design-system/blob/development/src/components/Slider/sliderChangelogData.ts and how its implemented: https://github.com/NYPL/nypl-design-system/blob/development/src/components/Slider/Slider.mdx

> Is there a reason the select in the Searchbar should be different than the Select component itself?

There isn't.

> Do we have a philosophy for our zIndex values? It seems like in many places we use 9999, in others we use 5 or 10.

I vaguely remember using 9999 but that was too high. The icon "bled through" popups that were layered on top of the 
`Select` or `SearchBar`. Because of that, it had to have a lower z-index but not too low. Otherwise, I'd say if we are not 
using popups or elements that have dropdowns that go over other elements, then we should use 9999 which is more conventional.

EdwinGuzman (approved):

bigfishdesign13 (approved):
Looks good!

PR #1461, opened 2023-11-14
DSD-624: hasVisitedState prop

oliviawongnyc:
Fixes JIRA ticket [DSD-1624](https://jira.nypl.org/browse/DSD-1624).

- Adds `hasVisitedStyles` prop to `Link` which is used to include or omit the visited state styles.
- Removes `disabled` variant from `Link` theme file, as it isn't being used.

## How has this been tested?

- Locally on Storybook.

- None.

### Checklist:

### Front End Review:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 ### Updates
 
 - Adds `hasVisitedStyles` prop to `Link` which is used to include or omit the visited state styles.
+- Removes `disabled` variant from `Link` theme file, as it isn't being used.
 
 ## 2.1.2 (November 9, 2023)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 ## Prerelease
 
+### Updates
+
+- Adds `hasVisitedStyles` prop to `Link` which is used to include or omit the visited state styles.
+
 ## 2.1.2 (November 9, 2023)
 
 ### Adds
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,11 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 ## Prerelease
 
+### Updates
+
+- Adds `hasVisitedStyles` prop to `Link` which is used to include or omit the visited state styles.
```
EdwinGuzman: Can you add that this is set to true by default? We don't normally add that level of granularity in the changelog but this is an impact change that we should let users know.

src/components/Link/Link.tsx
```diff
@@ -152,6 +154,7 @@ export const Link = chakra(
     const {
       children,
       className,
+      hasVisitedState = false,
```
EdwinGuzman: This should be true by default according to the ticket, unless if there was an update?

EdwinGuzman (changes requested):
The default should be updated to true unless that changed and is not reflected in the ticket. Otherwise this is fine. I think you updated the variants from an object to individual variables to be able to pass the `hasVisitedState` prop, right?

src/components/Link/Link.tsx
oliviawongnyc: Oops. Great catch. I was testing. Thank you!

oliviawongnyc:
> The default should be updated to true unless that changed and is not reflected in the ticket. Otherwise this is fine. I think you updated the variants from an object to individual variables to be able to pass the `hasVisitedState` prop, right?

@EdwinGuzman, yes, that's right. I first tried to change the `variants` variable from an object to a function that took argument `hasVisitedState` and returned an object, but it didn't work.

EdwinGuzman (approved):

bigfishdesign13 (approved):
Looks good and works well, but I think we need a `Visited State` section in the docs.  I will create a new ticket for that.

PR #1465, opened 2023-11-21
DSD-1521: Add sizeBasedOn prop to Image

jackiequach:
Fixes JIRA ticket [DSD-1521](https://jira.nypl.org/browse/DSD-1521)

- Adds sizeBasedOn prop to Image component

## How has this been tested?

Locally on Storybook

N/A

### Checklist:

### Front End Review:


src/components/Image/imageChangelogData.ts
```diff
@@ -0,0 +1,19 @@
...
+ * version: string (when adding new entry during development, set value as "Prerelease")
+ * type: "Bug Fix" | "New Feature" | "Update";
+ * affects: array["Accessibility" | "Documentation" | "Functionality" | "Styles"];
+ * notes: array (will render as a bulleted list, add one array element for each list element)
+ */
+import { ChangelogData } from "../../utils/ComponentChangelogTable";
+
+export const changelogData: ChangelogData[] = [
+  {
+    date: "2023-11-09",
```
EdwinGuzman: Date and version should be updated to "Prerelease".

src/components/Image/Image.tsx
```diff
@@ -29,9 +29,11 @@ export const imageSizesArray = [
   "large",
 ] as const;
 export const imageTypesArray = ["default", "circle"] as const;
+export const imageSizeBasedOnArray = ["height", "width"] as const;
```
EdwinGuzman: You can try to reuse the `logoSizeBasedOnArray` but rename it to something more generic for image and logo and any other component that may use height and width as values.

src/components/Image/Image.stories.tsx
```diff
@@ -53,6 +59,66 @@ const meta: Meta<typeof Image> = {
...
+      size: imageValue.size,
+      displayValue: imageValue.display,
+      id: imageValue.id,
+    })
+  );
+  sizesBasedOnHeight.push(
+    imageRow({
+      size: imageValue.size,
+      displayValue: imageValue.display,
+      id: imageValue.id,
```
EdwinGuzman: This will give you duplicate id issues on the page so add something else to make it unique.

src/theme/components/image.ts
```diff
@@ -1,17 +1,32 @@
 interface CustomImageBaseStyle {
-  size: keyof typeof imageSizes;
+  ratio: keyof typeof imageRatios;
+  size: keyof typeof imageSizes | keyof typeof imageSizesBasedOnHeight;
+  sizeBasedOn: "height" | "width";
 }
 interface CustomImageWrapperBaseStyle {
```
EdwinGuzman: It looks like both interfaces have the same structure, so you can probably rename it and delete the other.

src/theme/components/image.ts
```diff
@@ -1,17 +1,32 @@
...
+  sizeBasedOn: "height" | "width";
 }
 
-const imageWrap = {
-  position: "relative",
-  width: "100%",
-  height: "auto",
-  overflow: "hidden",
+const sizes = {
+  xxxsmall: 32,
```
EdwinGuzman, line 8: Do you not need the specific "px" text in the value? Don't think you need to change anything but I wonder if the number 64 is the same as the string "64px" in css.

EdwinGuzman: Ah, I now see how you're using this. Never mind the comment!

src/components/Image/Image.mdx
```diff
@@ -62,7 +64,12 @@ then an `img` element will be rendered with or without wrapper style divs.
 
 Use the `size` prop to set the desired size with the `ImageSizes` enum.
 
+In addition, the `sizeBasedOn` prop can be passed to determine whether the size
+of the `Image` is updated based on either the `"height"` or `"width"`. By default,
+the `sizeBasedOn` prop is set to `"width"`.
+
 <Canvas of={ImageStories.Sizes} />
+<Canvas of={ImageStories.SizesBasedOnHeight} />
```
EdwinGuzman, line 77: Can you add a heading above each section for the size based on width and height. It doesn't have to be here and could go into the story component if that's easier.

EdwinGuzman (commented):
This looks good but I'm not sure if it's working when `sizeBasedOn` is set to height and the aspect ratios are updated. Or rather, is it only supposed to work when aspectRatio is original?

src/components/Image/Image.tsx
jackiequach: Added `dimensionTypeArray` to types.ts and changed the imports.

jackiequach:
I tested sizeBasedOn height with different aspect ratio on the Storybook [With Controls page](https://nypl-design-system-git-dsd-1521-image-sizebasedon-nypl.vercel.app/?path=/story/components-media-icons-image--with-controls&args=aspectRatio:oneByTwo;size:large;sizeBasedOn:height) and it works correctly. Let me know if I am misunderstanding.

EdwinGuzman:
The update looks good, thanks.

This is what I mean regarding the aspect ratio but I may also not understand the reqs. This is a somewhat exaggerated example. For `sizedBasedOn="height"` and `src="//placekitten.com/400/40"`, when `aspectRatio="original"` I get this which seems right:
Screen Shot 2023-12-01 at 2 31 56 PM

If I keep the `src` the same with the height of 40px but update the aspect ratio, it's no longer 40px. This is what I'm unsure of. If the aspect ratio is updated to "square", I get the following:

Screen Shot 2023-12-01 at 2 33 51 PM

jackiequach:
> If I keep the `src` the same with the height of 40px but update the aspect ratio, it's no longer 40px. This is what I'm unsure of. If the aspect ratio is updated to "square", I get the following:
> 
> Screen Shot 2023-12-01 at 2 33 51 PM

This matches the behavior of size based on width. The height of the image will be determined by the `size` prop.

EdwinGuzman:
Hmm yeah that makes more sense. Let's wait on Marty to review this but looks good otherwise.

EdwinGuzman (approved):

bigfishdesign13:
Overall, this is working well, but I am seeing one issue.  When `sizeBasedOn` is set to `"height"` and the height of the actual image is less than the height set by the size value, the image does not scale up to match the size height value.  It doesn't go any larger than the actual image height.  A larger image will scale down to match the size height value, but a smaller image is not being scaled up.  Conversely, when `sizeBasedOn` is set to `"width"`, a small image will scale to match the size width value.

I think this is what @EdwinGuzman was also seeing.

bigfishdesign13 (approved):
Very nice.  The last update works great.

PR #1467, opened 2023-11-29
DSD-1589: Adds placeholder props and allows empty state

oliviawongnyc:
Fixes JIRA ticket [DSD-1589](https://jira.nypl.org/browse/DSD-1589). 

NOTE: this is a draft PR because Edwin and I made some assumptions about how this should work and want to get sign off from others on the approach before finishing the PR (adding to the Changelogs, updating Storybook, etc).

- Allows users to pass an empty string for initialDate and initialDateTo, which results in empty inputs.
- Allows users to optionally pass a string placeholder and placeholderTo.

## How has this been tested?

- Locally, on Storybook.

- None.

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/DatePicker/DatePicker.stories.tsx
+++ b/src/components/DatePicker/DatePicker.stories.tsx
@@ -186,8 +186,17 @@ export const OtherStates: Story = {
   ),
 };
 
-export const InitialDate: Story = {
-  render: () => (
+// Examples that use the `initialDate` and `initialDateTo` props and
+// also the `onChange` prop to show an empty input returns null.
+function InitialDateExample() {
+  const onChange = (data) => {
+    console.log({
+      startDate: data.startDate,
+      endDate: data.endDate,
+    });
+  };
+
+  return (
     <VStack align="stretch" spacing="s">
       <DatePicker
         id="init-dates"
@@ -207,6 +216,7 @@ export const InitialDate: Story = {
         initialDate=""
         initialDateTo=""
         isDateRange
+        onChange={onChange}
       />
       <Heading level="h3" size="heading6">
         Passing no initialDate renders today's date
@@ -218,7 +228,11 @@ export const InitialDate: Story = {
         isDateRange
       />
     </VStack>
-  ),
+  );
+}
+
+export const InitialDate: Story = {
+  render: () => InitialDateExample(),
 };
 
 export const Placeholder: Story = {
--- a/src/components/DatePicker/DatePicker.stories.tsx
+++ b/src/components/DatePicker/DatePicker.stories.tsx
@@ -188,12 +188,50 @@ export const OtherStates: Story = {
 
 export const InitialDate: Story = {
   render: () => (
+    <VStack align="stretch" spacing="s">
+      <DatePicker
+        id="init-dates"
+        dateType="full"
+        labelText="Select the date you want to visit NYPL"
+        initialDate="12/1/21"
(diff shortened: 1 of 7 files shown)


src/components/DatePicker/DatePicker.tsx
```diff
@@ -275,8 +284,16 @@ export const DatePicker = chakra(
     } = props;
     const styles = useMultiStyleConfig("DatePicker", {});
     const finalStyles = isDateRange ? styles : {};
-    const initStartDate = initialDate ? new Date(initialDate) : new Date();
-    const initEndDate = initialDateTo ? new Date(initialDateTo) : new Date();
+    const initStartDate = initialDate
+      ? new Date(initialDate)
+      : initialDate === ""
+      ? null
```
EdwinGuzman, line 290: I'm surprised that there's no ts issue with the `FullDateType` interface when setting `let initFullDate: FullDateType = { startDate: initStartDate };` since `initialStartDate` could be empty.

EdwinGuzman:
This looks good. For the draft, can you add a working example in storybook with the initial empty value set? So we can see it working.

src/components/DatePicker/DatePicker.tsx
oliviawongnyc: I'm not exactly sure why either, but I added the possibility of null to this interface.

src/components/DatePicker/DatePicker.stories.tsx
```diff
@@ -187,13 +187,51 @@ export const OtherStates: Story = {
 };
 
 export const InitialDate: Story = {
+  render: () => (
```
EdwinGuzman: This is great, thanks!

bigfishdesign13:
I think the choices that you made for the prop functionality are fine.  But for fun, here is another possibility...

Omitting the `initialDate` or `initialDateTo` props or passing empty strings to the props would render empty fields.  To get today's date, pass a wildcard like `@today`.  This would be a new concept in our components, but it would then allow for other wildcards like `@yesterday`, `@nextweek`, etc.

EdwinGuzman:
@bigfishdesign13 that sounds like a good feature for another PR.

@oliviawongnyc I made a small update to the Storybook example. I'm using this now in a Research Catalog form and I wanted to make sure that an empty value does return null or an empty string. For example, If the user only wants to search from start year of 1950 until now, the end year can be empty and the `onChange` function will return the appropriate values (as far as I can tell). Based on this, this update looks good.

oliviawongnyc:
> @oliviawongnyc I made a small update to the Storybook example. I'm using this now in a Research Catalog form and I wanted to make sure that an empty value does return null or an empty string. For example, If the user only wants to search from start year of 1950 until now, the end year can be empty and the `onChange` function will return the appropriate values (as far as I can tell). Based on this, this update looks good.

@EdwinGuzman, I'm not entirely sure what you mean. What do you expect the values to be if a user empties one of the inputs?

EdwinGuzman:
@oliviawongnyc I expect it to be null or undefined.

oliviawongnyc:
> @oliviawongnyc I expect it to be null or undefined.

@EdwinGuzman, gotcha. It's null. Do you feel like this is good to go?

EdwinGuzman (commented):
This looks good to me but update the changelog for the repo and for the component, and add one test where the values passed are empty and no dates get rendered initially.

EdwinGuzman (approved):

PR #1468, opened 2023-11-30
React 18/Chakra 2.8 Checkbox update

EdwinGuzman:
## This PR does the following:

- Updates the DS `Checkbox` component with Chakra 2.8 updated API for its style.

For this specific component, I had to update it based on Chakra's own `Checkbox` styles. The "parts" had to match up in order for the DS component to appropriately pick up the styles.
- https://github.com/chakra-ui/chakra-ui/blob/main/packages/components/theme/src/components/checkbox.ts

The DOM structured for the "parts" can be compared here https://github.com/chakra-ui/chakra-ui/blob/main/packages/components/checkbox/src/checkbox.tsx

Here is the initial React 18/Chakra 2.8 update where the styles are not 100% correct: https://nypl-design-system-git-react18-and-chakra-nypl.vercel.app/?path=%2Fdocs%2Fcomponents-form-elements-checkbox--docs

The Vercel preview for this PR https://nypl-design-system-git-chakra2-checkbox-nypl.vercel.app/?path=%2Fdocs%2Fcomponents-form-elements-checkbox--docs should be reviewed against the production Checkbox styles and functionality: https://nypl.github.io/nypl-design-system/reservoir/v2/?path=/docs/components-form-elements-checkbox--docs

#### Chakra function types

DS components need better types. Initially, components did not have types `export const Checkbox = chakra(`. Now we need the following which is more verbose but gives better type definitions.

```
export const Checkbox: ChakraComponent<
  React.ForwardRefExoticComponent<
    CheckboxProps & React.RefAttributes<HTMLInputElement>
  >,
  CheckboxProps
>
```

#### Theme file 

In Chakra 1.x, a component's style "parts" could be defined as

```
const Checkbox = {
  parts:  "helperErrorText", "icon", "control", "label", "container",
  baseStyles: ...
}
```

Now we need to use the `createMultiStyleConfigHelpers` hook which returns two functions.

```
const { defineMultiStyleConfig, definePartsStyle } =
  createMultiStyleConfigHelpers([
    "helperErrorText",
    "icon",
    "control",
    "label",
    "container",
  ]);

const Checkbox = defineMultiStyleConfig({
  baseStyles: ...
}
```

Review `src/theme/components/checkbox.ts` for more!

## How has this been tested?

In Storybook

## Accessibility concerns or updates

- N/A

### Checklist:

- [ ] I have updated the Storybook documentation accordingly.
- [ ] I have added relevant accessibility documentation for this pull request.
- [ ] All new and existing tests passed.

[...]

PR #1471, opened 2023-12-06
DSD-1637: Update Content Display Chakra

isastettler:
Fixes JIRA ticket xxx

- Update Component type for HelperErrorText and StatusBadge
- Update component style file for helperErrorText and statusBadge

## How has this been tested?

Locally

N/A
-

### Checklist:

### Front End Review:

### DIFF
--- a/src/theme/components/helperErrorText.ts
+++ b/src/theme/components/helperErrorText.ts
@@ -1,20 +1,31 @@
-interface HelperErrorTextBaseStyle {
+import { createMultiStyleConfigHelpers } from "@chakra-ui/styled-system";
+import { defineStyle, StyleFunctionProps } from "@chakra-ui/system";
+
+interface HelperErrorTextBaseStyle extends StyleFunctionProps {
   isInvalid: boolean;
 }
 
-const helperErrorText = {
-  parts: ["innerChild"],
-  baseStyle: ({ isInvalid }: HelperErrorTextBaseStyle) => ({
+const { defineMultiStyleConfig, definePartsStyle } =
+  createMultiStyleConfigHelpers(["innerChild"]);
+
+const innerChildStyle = defineStyle({
+  marginTop: "xxs",
+  marginBottom: "0",
+});
+
+const baseStyle = definePartsStyle(
+  ({ isInvalid }: HelperErrorTextBaseStyle) => ({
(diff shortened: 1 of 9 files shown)


src/theme/components/helperErrorText.ts
```diff
@@ -1,20 +1,31 @@
-interface HelperErrorTextBaseStyle {
+import { createMultiStyleConfigHelpers } from "@chakra-ui/styled-system";
+import { defineStyle, StyleFunctionProps } from "@chakra-ui/system";
+
+interface HelperErrorTextBaseStyle extends StyleFunctionProps {
```
EdwinGuzman, line 4: This makes sense and I see why I didn't need this in the `Checkbox` PR, this is needed when you are passing props to the theme functions.

EdwinGuzman (changes requested):
The update itself looks great, thanks!

The only thing missing is changelog updates. Can you update the respective storybook files to include "prerelease" in the version table and also add `helperErrorTextChangelogData.ts` and `statusBadgeChangelogData.ts` files? They can similar to this: https://github.com/NYPL/nypl-design-system/pull/1468/files#diff-8bc92befba3589679a49736327283e9bf375f12582239b94874b15c9312ce479 
and then added to a changelog section in the `.mdx` file at the very end.

EdwinGuzman (approved):

src/theme/components/helperErrorText.ts
EdwinGuzman, line 4: Just FYI, I ran into an issue where this caused a problem:

Screen Shot 2023-12-06 at 4 55 17 PM

It's not a problem for this component or interface but if you run into the same problem, it can be fixed by making it a partial `Partial<StyleFunctionProps>`. We do need to extend the interface but we don't need to make all the properties required.

isastettler: To keep it consistent I make the change here too. It might be better in the long run.

EdwinGuzman: We talked about this offline but for history: we couldn't figure out why Chakra is being inconsistent and why not using `Partial<StyleFunctionProps>` in this case breaks. So for this case (it might be a case-by-case basis), it's okay to not use `Partial`.

PR #1480, opened 2023-12-19
DSD-1646: Updating NewsletterSignUp, Radio, RadioGroup, and Searchbar to Chakra 2.8 theme API

7emansell:
Fixes JIRA ticket [DSD-1646](https://jira.nypl.org/browse/DSD-1646)

- Updates `NewsletterSignUp`, `Radio`, `RadioGroup`, and `Searchbar` components to use the Chakra 2.8 updated API for their theme object.
- No functional changes, only theme file

## How has this been tested?

Locally and compared against the production Storybook instance.

N/A 

### Checklist:

### Front End Review:


7emansell:
Noting that the `Searchbar` doesn't fill the given width as it should (and does on prod) but that's how it appears on my local instance regardless of Chakra update? Also noting that the `NewsletterSignUp` looks off (button, heading) but those updates are coming on those components' respective tickets.

package.json
```diff
@@ -53,14 +53,15 @@
     "types-ramda": "0.29.4"
   },
   "dependencies": {
-    "@chakra-ui/focus-lock": "2.1.0",
-    "@chakra-ui/react": "2.8.1",
-    "@chakra-ui/system": "2.6.2",
+    "@chakra-ui/focus-lock": ">=2.1.0",
+    "@chakra-ui/react": ">=2.8.1",
```
EdwinGuzman: Did you make all the ">=" updates? Did you encounter any issues while installing packages?

EdwinGuzman (approved):
Looks good to me pending updates to the `Heading` and `Button` components which will be updated in other PRs. I think that's also what's happening in `SearchBar` where the `TextInput` component needs an update but that's a separate PR, so it makes sense that it's not full width just yet.

package.json
7emansell: Oops yes accidentally, I had something up with esbuild, they're locked now

package.json
```diff
@@ -60,6 +60,7 @@
     "@emotion/react": "11.11.1",
     "@emotion/styled": "11.11.0",
     "downshift": "6.1.7",
+    "esbuild": "^0.19.10",
```
EdwinGuzman: I had some esbuild issues as well. Can you try running this without `esbuild`? If we can remove this dependency, that'd be great, otherwise it's okay to keep.

7emansell: It's running fine without `esbuild` now so I removed it.. weird though

PR #1485, opened 2024-01-04
DSD-1660: default text colors in Hero

bigfishdesign13:
Fixes JIRA ticket [DSD-1660](https://jira.nypl.org/browse/DSD-1660)

- Updates the default styles for text elements in the `"primary"`, `"tertiary"`, and `"campaign"` variants of the `Hero` component.

## How has this been tested?

- local Storybook

- The default color choices are compliant with WCAG 2.1 standards.

### Checklist:

### Front End Review:


src/theme/components/hero.ts
```diff
@@ -189,12 +189,18 @@ const campaign = {
...
     position: { md: "relative" },
     zIndex: 2,
+    ".chakra-heading": {
+      color: "ui.typography.inverse.heading",
+    },
     _dark: {
       color: "dark.ui.typography.body",
+      ".chakra-heading": {
+        color: "dark.ui.typography.heading",
+      },
```
bigfishdesign13: @EdwinGuzman This is where I am adding styles to set the color of all headings within the campaign hero.  It sets the color, as expected, but it also removes the ability to set custom colors for the headings.  If you look in the stories file, you'll see that I have custom text values all over the place for the campaign heroes, but none of the custom colors are showing.  This solution would be a problem when, like in DC facelift (DCF), the background color is set to a light color and the text color then needs to be set to a dark color.

If we don't set the color in the component styles, then you are able to set custom colors for the heading.  And again, this is what they are doing successfully in DCF.  However, this falls apart when dark mode is enabled.  I know DCF currectly is not offering dark mode, but they will some day.  Additionally, DCF will not be the only site where this could be a problem.

If you have a chance, can you see if you cam find a way around this?  Or at least take a look and then let's talk about it.  I don't think we are going to solve the heading color issue with this type of solution, but I do have some ideas for how we can resolve this.

EdwinGuzman: This is quite the issue. I don't think DCF is doing it successfully because they have to add `!important` for the colors to show. I did the same by doing 

```
__css={{ color: "magenta !important", _dark: { color: "yellow" } }}
```

and this worked. My first thought was to do more of what the `Hero` component already does with styles similar to the `campaignBgStyles` or `backgroundImageStyle` css objects. So we would do more preprocessing in the component for styles to pass. Another option is to try to clone and set the proper styles in the component similar to this:

```
const finalHeading =
  heading && React.cloneElement(heading, { __css: headingStyles });
```

But none of these approaches work. We are assuming devs will pass DS `Heading` components and it's also too complicated for the `Hero` component to parse the `subHeaderText` prop to find and update any `Heading` components it finds.

What if we don't set any default styles or loosely set them so they are very easy to override? Then the onus really falls on the consuming app to implement the `Hero` component correctly. Theoretically, to me, I think this is okay because we can ask designers to check for color contrast and proper color values at the design stage before a dev implements it.

We should talk about this next week. I'm curious about your ideas to resolve this.

bigfishdesign13: @EdwinGuzman 
Ultimately, the color of the text, especially in the `Hero` component, shoudl be the default text colors for light and dark mode or the inverse text colors for light and dark mode.  I feel strongly about this.  If this is what we want, then we can set the styles as I've done, which will force the colors to the default light and dark mode text colors.  We can also add a new `isInverseTextColor` prop.  And if `isInverseTextColor` is true, the component will use the `inverse` text colors.  Then, as you mentioned, it is up to the designer to specify which color should be used: `default` or `inverse`.

EdwinGuzman: If we make this change, would it mean that any apps currently using the component will need an update when this update is deployed?

bigfishdesign13: > If we make this change, would it mean that any apps currently using the component will need an update when this update is deployed?

I don't think so.  Searching the code, I only see DC and MLN using the DS campaign hero.

https://github.com/search?q=org%3ANYPL+heroType%3D%22campaign%22+-repo%3ANYPL%2Fnypl-design-system+-repo%3ANYPL%2Fnypl-ds-test-app&type=code

bigfishdesign13:
@EdwinGuzman I started making the changes that we talked about.  For now, they are only aplied to the `campaign` variant and it may have to stay that way.  But take a look and let me know what you think.  If this update is only going to apply to the `campaign` variant, I will need to add some additional logic and warnings.

EdwinGuzman (commented):
Yes, this approach makes sense but the first example doesn't display well:
Screen Shot 2024-01-12 at 11 14 09 AM

EdwinGuzman:
The update looks good. Is this ready for another review or is this still being worked on?

bigfishdesign13:
> The update looks good. Is this ready for another review or is this still being worked on?

@EdwinGuzman I cleaned up a few more things.  One last review and then yes, this is ready to move forward.

src/components/Hero/Hero.stories.tsx
```diff
@@ -221,6 +237,9 @@ export const Tertiary: Story = {
         }
         heroType="tertiary"
         subHeaderText={otherSubHeaderText}
+        // foregroundColor="yellow"
```
EdwinGuzman: Can remove this.

src/components/Hero/Hero.stories.tsx
```diff
@@ -353,19 +372,154 @@ export const CampaignBackgroundColors: Story = {
...
+              level="h1"
+              id="campaign-hero-default-heading"
+              text="Hero Campaign"
+            />
+          }
+          imageProps={imageProps}
+          isDarkBackgroundImage
+          subHeaderText={otherSubHeaderText}
+        />
+        <Source
```
EdwinGuzman, line 681: We've been putting the `Source` component in mdx files but this seems to make more sense -- to be colocated with the code.

EdwinGuzman: The "Show code" button still shows up and can make it confusing. I think it can be disabled, can you look into that?

src/components/Hero/Hero.tsx
```diff
@@ -62,6 +62,7 @@ export interface HeroProps {
    * can only be used in conjunction with `backgroundImageSrc` for the "campaign"
    * `Hero` type. Note: not all `Hero` variations utilize this prop. */
   imageProps?: HeroImageProps;
+  isDarkText?: boolean;
```
EdwinGuzman, line 70: Give this a comment for the Storybook args table.

src/theme/components/hero.ts
```diff
@@ -220,12 +314,42 @@ const campaign = definePartsStyle({
...
+        color: getLinkColor("hover", "light", foregroundColor, isDarkText),
+      },
+      _visited: {
+        color: getLinkColor("visited", "light", foregroundColor, isDarkText),
+        svg: {
+          fill: getLinkColor("visited", "light", foregroundColor, isDarkText),
+        },
+      },
+    },
+    ".chakra-heading": {
```
EdwinGuzman, line 316: Is this done because it's hard to override or add the value directly to the component?

src/theme/components/hero.ts
```diff
@@ -138,16 +196,21 @@ const primary = definePartsStyle({
...
       bgColor: "dark.ui.bg.default",
-      color: "dark.ui.typography.body",
+      color: getTextColor("body", "dark", foregroundColor, isDarkText),
+      ".chakra-heading": {
+        color: getTextColor("heading", "dark", foregroundColor, isDarkText),
+      },
     },
   },
   heading: {
     color: "dark.ui.typography.heading",
```
EdwinGuzman: Should this also use the `getTextColor` function?

src/theme/components/hero.ts
```diff
@@ -105,8 +105,65 @@ const getSecondaryVariantStyles = (bgColor: string = "") => {
...
+    : isDarkText
+    ? `${prefix}typography.inverse.${type}`
+    : `${prefix}typography.${type}`;
+  const colorLink = foregroundColor
+    ? foregroundColor
+    : isDarkText
+    ? `ui.link.primary`
+    : `dark.ui.link.primary`;
+  const finalColor =
+    type === "link" ? colorLink : mode === "dark" ? colorDark : colorLight;
```
EdwinGuzman: You don't have an example where "link" is being passed to the `getTextColor` function.

Can you write a few tests for these utility functions? Do you think it's worth it to verify the values returned are the expected ones?

src/theme/components/hero.ts
```diff
@@ -105,8 +105,65 @@ const getSecondaryVariantStyles = (bgColor: string = "") => {
...
+  return finalColor;
+};
+const getLinkColor = (state, mode, foregroundColor, isDarkText) => {
+  let finalColor;
+  const isDarkMode = mode === "dark" ? true : false;
```
EdwinGuzman: This can just be `const isDarkMode = mode === "dark";`

src/theme/components/hero.ts
```diff
@@ -105,8 +105,65 @@ const getSecondaryVariantStyles = (bgColor: string = "") => {
...
+  switch (state) {
+    default:
+      finalColor = foregroundColor
+        ? foregroundColor
+        : isDarkText
+        ? `var(--nypl-colors-ui-link-primary) !important`
+        : isDarkMode
+        ? `dark.ui.link.primary`
+        : `dark.ui.link.primary`;
```
EdwinGuzman: Both of these are the same but I don't think they should be. Wouldn't it be checked against `isDarkMode` here so whether it's true or false you'll get the same value.

src/theme/components/hero.ts
```diff
@@ -105,8 +105,65 @@ const getSecondaryVariantStyles = (bgColor: string = "") => {
...
+    : `dark.ui.link.primary`;
+  const finalColor =
+    type === "link" ? colorLink : mode === "dark" ? colorDark : colorLight;
+  return finalColor;
+};
+const getLinkColor = (state, mode, foregroundColor, isDarkText) => {
+  let finalColor;
+  const isDarkMode = mode === "dark" ? true : false;
+  switch (state) {
+    default:
```
EdwinGuzman: Conventionally, the default case goes last in the list of cases.

EdwinGuzman: Hmm it might also help to make it:
```
case "default":
default:
  ....
```

That way it's explicit that the "default" case is also the default for the switch statemetn.

src/theme/components/hero.ts
```diff
@@ -105,8 +105,65 @@ const getSecondaryVariantStyles = (bgColor: string = "") => {
...
+        : `dark.ui.link.secondary`;
+      break;
+    }
+    case "visited": {
+      finalColor = foregroundColor
+        ? foregroundColor
+        : isDarkText
+        ? `var(--nypl-colors-ui-link-tertiary) !important`
+        : isDarkMode
+        ? `dark.ui.link.tertiary`
```
EdwinGuzman: We don't have any rules or patterns for this but I'd like to keep tertiary to two levels or else it gets confusing to read. That said, I'm sure we do this in other files so it's not something you need to change but just to keep in mind for readability.

EdwinGuzman (changes requested):
Looks good but pointing out some minor changes. I think tests for the util functions would be good to have.

src/components/Hero/Hero.stories.tsx
bigfishdesign13: Ugh.  I did remove those lines, but I had to revert something and POOF, they were back in.

src/components/Hero/Hero.stories.tsx
bigfishdesign13: Flipping from `Canvas` to `Story` did the trick.

src/theme/components/hero.ts
bigfishdesign13: Using the class name was the easiest way to target and override the Heading styles.

src/theme/components/hero.ts
bigfishdesign13: Yes.

src/theme/components/hero.ts
bigfishdesign13: Seeing that I also have a `getLinkColor` function, I must have initially added this condition here and then realized I needed something more powerful.

I will see what I can do with the tests.

src/theme/components/hero.ts
bigfishdesign13: Yes, this is a little confusing to see.  However, the default background color (without settign the `backgroundColor` prop) is dark for both the light mode and dark mode variants of the Hero component, so the default color of the links should always be that one value (lighter).  

That being the case, you might ask, why even use a ternary condition her?  I left that in there in case we did eventually need to differentiate the color values for light mode and dark mode.  

But in the end, I know you left another comment about nesting too many ternary conditions.  Based on that, I have removed this last ternary condition.  I also added a comment to somewhat explain the reasoning.

src/theme/components/hero.ts
bigfishdesign13: Right.  Forgot about that.  It's been a while since I used a switch-case statement.

src/theme/components/hero.ts
bigfishdesign13: When I put this together, I felt like I was being bad.  But it worked, so I left it.  That said, I simplified the statement and removed the last ternary condition.

src/components/Hero/Hero.test.tsx
```diff
@@ -672,3 +673,114 @@ describe("Hero", () => {
...
+      "brand.primary"
+    );
+  });
+  it("returns default heading text color", () => {
+    type = "heading";
+    mode = "light";
+    foregroundColor = undefined;
+    isDarkText = false;
+    expect(getTextColor(type, mode, foregroundColor, isDarkText)).toBe(
+      "ui.typography.inverse.heading"
```
EdwinGuzman, line 723: why does the default have `inverse`? I would expect it to just be `ui.typography.heading`.

src/components/Hero/Hero.stories.tsx
```diff
@@ -189,20 +206,42 @@ export const WithControls: Story = {
...
+            text="Hero Primary"
+          />
+        }
+        heroType="primary"
+      />
+      <Source
+        code={`
+<Hero
+  backgroundImageSrc="//loremflickr.com/1600/800/kitten"
+  heading={
```
EdwinGuzman, line 230: Missing `backgroundColor="ui.white"` here. Also, this looks off...
Screen Shot 2024-04-05 at 11 53 48 AM

src/components/Hero/Hero.stories.tsx
```diff
@@ -273,15 +380,45 @@ export const Campaign: Story = {
           imageProps={imageProps}
           subHeaderText={otherSubHeaderText}
         />
+        <Source
+          code={`
+<Hero
+  backgroundImageSrc="//loremflickr.com/1600/800/kitten"
```
EdwinGuzman: the background image is a bit too gruesome to use...

Screen Shot 2024-04-05 at 11 56 31 AM

EdwinGuzman (commented):
The update looks good. It looks like when an image does not expand it's container, the background becomes red:

Screen Shot 2024-04-05 at 11 57 24 AM

src/components/Hero/Hero.test.tsx
bigfishdesign13: Yeah, it starts to get a little convoluted, but in the end it makes sense.  Our standard text color is dark (for a light background) and the inverse text color is light (for a dark background).  The default background colors for the Hero variants are dark, so the colors of the text that sits atop those dark background colors need to be light and that is the "inverse" of our standard text colors.

src/components/Hero/Hero.stories.tsx
bigfishdesign13: The inclusion of `backgroundColor="ui.white"` was a mistake, so I just remove it.

src/components/Hero/Hero.stories.tsx
bigfishdesign13: Yes, I agree! The images are random, so we'll just need to pick a keyword that will be the least likely to delivery oddball images.  For now, I am going to use `"new york public library"`.

https://loremflickr.com/2400/800/new+york+public+library

bigfishdesign13:
>  It looks like when an image does not expand it's container, the background becomes red:

@EdwinGuzman  I'm not sure what to do about that.  I was hoping there was at least a param that could be passed in the URL that would control that filler color, but I was not able to find anything.

EdwinGuzman (approved):

PR #1486, opened 2024-01-05
POC Testing the heading and icon in modal component

EdwinGuzman:
Just a test!


EdwinGuzman:
@bigfishdesign13 check out the last two Modal examples https://nypl-design-system-git-testing-modal-heading-nypl.vercel.app/?path=/docs/components-overlays-switchers-modal--docs

I made this comment in the figma file but adding it here as well:
The problems I see right now are mostly styling. I had to add the right size for the icon but the bigger issue is the alignment. In this design, the text aligns and the icon is further to the left of the text. In my brief testing, the left side of the icon aligns with the rest of the text by default. Even though this works, we may need to control the alignment of the heading ourselves or tell apps how to do it (e.g. put a negative margin so the heading and content text align).

bigfishdesign13:
> @bigfishdesign13 check out the last two Modal examples https://nypl-design-system-git-testing-modal-heading-nypl.vercel.app/?path=/docs/components-overlays-switchers-modal--docs
> 
> I made this comment in the figma file but adding it here as well: The problems I see right now are mostly styling. I had to add the right size for the icon but the bigger issue is the alignment. In this design, the text aligns and the icon is further to the left of the text. In my brief testing, the left side of the icon aligns with the rest of the text by default. Even though this works, we may need to control the alignment of the heading ourselves or tell apps how to do it (e.g. put a negative margin so the heading and content text align).

I am so curious what I was doing so that I could not make this work.  Ragardless, this looks great.  We can overcome the spacing and alignment issues with styles right on the `Heading` component.  It actually works quite well.  I was going to suggest that we add a new prop to handle the icon, but for the other components where icons might be used (`Button` and `Link`), we rely on the `Icon` component being passed as a child, so we are not establishing a new pattern if we do that here as well.

```
headingText: (
  <>
    <Heading
      alignItems="center"
      display="flex"
      gap="xs"
      level="h3"
      size="heading5"
    >
      <>
        <Icon
          name="actionCheckCircleFilled"
          size="large"
          color="ui.success.primary"
        />
        Modal Heading Text
      </>
    </Heading>
  </>
),
```

I [created a ticket](https://jira.nypl.org/browse/DSD-1668) to add an example like this in the modal Storybook docs.

bigfishdesign13:
On a related note, should we utilize the `useDSHeading` hook here as well?

EdwinGuzman:
Yes, we should. Can you add to use that hook in the ticket so we don't forget?

PR #1487, opened 2024-01-08
DSD-1645: Updating Form, Fieldset, Label, and FeedbackBox to Chakra 2.8 theme API

oliviawongnyc:
Fixes JIRA ticket [DSD-1645](https://jira.nypl.org/browse/DSD-1645)

- Updates Fieldset, Label, and FeedbackBox components to use the Chakra 2.8 updated API for their theme object.
- Form doesn't have a theme object, so I just updated its type in order to eliminate the "any".
- One thing I am noticing is that you can import some of the functions (defineStyle, defineStyleConfig, etc) from either `"@chakra-ui/system";` or `"@chakra-ui/react";`. I assume we should make this consistent throughout, but it is not as of now. 

## How has this been tested?

- Locally in Storybook.

- None.

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/FeedbackBox/FeedbackBox.mdx
+++ b/src/components/FeedbackBox/FeedbackBox.mdx
@@ -9,10 +9,10 @@ import Link from "../Link/Link";
 
 # FeedbackBox
 
-| Component Version | DS Version |
-| ----------------- | ---------- |
-| Added             | `1.3.0`    |
-| Latest            | `2.1.1`    |
+| Component Version | DS Version   |
+| ----------------- | ------------ |
+| Added             | `1.3.0`      |
+| Latest            | `Prerelease` |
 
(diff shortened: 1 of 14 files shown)


EdwinGuzman (commented):
There is one minor issue with the `FeedbackBox` close button:

Screen Shot 2024-01-08 at 4 02 24 PM

Otherwise, this looks good.

oliviawongnyc:
@EdwinGuzman, the focus ring was looking vertically oblong because for some reason the width was no longer 40px as it is on prod. I added a new `minWidth` and now it's square again, but I am not sure why I had to do this.

EdwinGuzman (approved):
Not sure why you have to add it but it looks like the minHeight had to be overridden  before.

PR #1489, opened 2024-01-09
DSD-1639: Updating media & icons components to chakra 2.8

EdwinGuzman:
Fixes JIRA ticket [DSD-1639](https://jira.nypl.org/browse/DSD-1639)

- Updates `AudioPlayer`, `Icon`, `Image`, `Logo`, and `VideoPlayer` to use Chakra 2.x style API.
- Adds more Reservoir colors to the `Icon` color array.
- Renames `Custom-` to `Reservoir-`. For example, `CustomImage` to `ReservoirImage` for the Chakra component style theme object name.
- Internally adds `ratio` as a prop to the `ImageWrapper` component. In the theme file, the `aspectRatio` prop did not propagate down to the theme function correctly but the `ratio` prop value did. Not sure why but this is not a prop users or devs need to be concerned about since it's internal in the `Image.tsx` file.

## How has this been tested?

Locally and comparing it with prod

- N/A

### Checklist:

### Front End Review:


src/components/Image/Image.tsx
```diff
@@ -81,6 +82,9 @@ export interface ImageProps
   additionalImageStyles?: { [key: string]: any };
   /** Alternate text description of the image */
   alt?: string;
+  /** Optional value to control the aspect ratio of the card image; default
+   * value is `"original"` */
+  aspectRatio?: ImageRatios;
```
7emansell, line 87: Question just to clarify, devs should use 'ratio' then?

7emansell (approved):
Looks good to me. Also was there a particular reason for Reservoir vs. Custom or just for more clarity?

EdwinGuzman:
> Also was there a particular reason for Reservoir vs. Custom or just for more clarity?

It's for more clarity and maybe _looks_ better to have the project name than "custom".

src/components/Image/Image.tsx
EdwinGuzman: No, devs should keep using `aspectRatio` for the `Image` component. The `ImageWrapper` component should use `ratio` but this component is only used internally, so devs wouldn't be able to access it anyway.

EdwinGuzman:
@jackiequach when you get a chance, can you review this PR specifically checking on the `Image` component update? It includes the size based on height update and I want to make sure I didn't mess the merge.

jackiequach (approved):
LGTM

PR #1491, opened 2024-01-12
DSD-1652: Fixing a DatePicker styling issue for month/year selected values

EdwinGuzman:
Fixes JIRA ticket 
- [DSD-1652](https://jira.nypl.org/browse/DSD-1652)
- [DSD-1661](https://jira.nypl.org/browse/DSD-1661)

- Fixes minor typo in docs.
- Fixes styling issue for range `DatePicker` for hovered/in-range selected values.
Before:
Screen Shot 2024-01-12 at 3 30 53 PM

After:
Screen Shot 2024-01-12 at 3 36 54 PM

## How has this been tested?

Storybook, best to review the range type along with year or month type selected:  https://nypl-design-system-git-dsd-1652-datepicker-color-update-nypl.vercel.app/?path=/story/components-form-elements-datepicker--with-controls&args=dateType:year;isDateRange:!true

-

### Checklist:

### Front End Review:


jackiequach (commented):
For the month DatePicker, the selected month styles apply to the year of the maxDate (in this image: 1/1/2025)

image

The highlighted dates' styles are applied when going to previous years in the From DatePicker. (dates selected in image 6/2024-12/2024)

image

bigfishdesign13:
I am seeing the same issues that Jackie noted.  Those styles are all controlled by classes that are being applied to the various data elements.  Those styles are completed controlled by the `react-datepicker` package, but the change you made to the CSS should not affect the styles like this.  I checked the [react-datepicker site](https://reactdatepicker.com/#example-date-range) and it appears that the example there are also showing problems like this.

EdwinGuzman:
The react-datepicker site is using a newer version which some updates we don't have. I'm still going to try to fix the issue but if it's too difficult, I'll report back.

EdwinGuzman:
@jackiequach @bigfishdesign13 Okay, I better understand the issue and this is an issue in production and coming from the `react-datepicker` package. This is a bug from the package and I tried updating it but this is more complicated than expected. It is based on the classes they add and even if we override the styles, the problem is how the package itself applies the classes to the element. The logic isn't right and I'm not sure if it's worth us trying to fix this in our app.

I'll create a ticket for this to work on later on. This also adds more support for finding a different package.

EdwinGuzman:
@jackiequach @bigfishdesign13 @oliviawongnyc @7emansell please review this PR soon if you can. The last issue found won't be fixed in this PR as that's a plugin bug.

oliviawongnyc (approved):
Looks nice!

jackiequach (approved):
Looks good!

7emansell (approved):

PR #1495, opened 2024-01-19
DSD-1669: responsive font sizes for Heading and Text components (React 18 / Chakra 2)

bigfishdesign13:
Fixes JIRA ticket [DSD-1669](https://jira.nypl.org/browse/DSD-1669)

## PR Review Estimate

`medium`

- NOTE: A large portion of the files included in the PR are snapshots

- Updates the `Heading` component to use native Chakra responsive styles to handle the font sizes of the component's internal `heading`, `overline` and `subtitle` elements. This also resolves the flashing font size bug that is most noticeable on slower internet connections.
- Updates the `Text` component to use native Chakra responsive styles to handle the font sizes of the `"subtitle1"` and `"subtitle2"` variants.

NOTE: On a whim, I cloned the `react18-and-chakra` branch (Chakra v2) and successfully applied the responsive font size styles as you would expect.  This was something I was not able to do with the current `development` branch, which is running Chakra v1 (and yes, I did verify one more time that the Chakra native responsive styles are not working for `fontSize` in the `Heading` component).  That is why this PR is pointed at the `react18-and-chakra`.

## How has this been tested?

- local Storybook

- n/a

### Checklist:

### Front End Review:


EdwinGuzman (approved):
This looks good and I'm glad the update makes this so much easier to work with. It'd be great if we could set this into `development` because we don't know when the React18 update will launch and when DXP can use it. Unless if this is not a blocker for DXP then it's not a problem.

nypl-wluisi:
> This looks good and I'm glad the update makes this so much easier to work with. It'd be great if we could set this into `development` because we don't know when the React18 update will launch and when DXP can use it. Unless if this is not a blocker for DXP then it's not a problem.

Off development would be better for us, as this bug blocks our upgrade to DS V2

bigfishdesign13:
> This looks good and I'm glad the update makes this so much easier to work with. It'd be great if we could set this into `development` because we don't know when the React18 update will launch and when DXP can use it. Unless if this is not a blocker for DXP then it's not a problem.

@EdwinGuzman To make this change in the `development` branch would require a significant refactoring of the whacky logic that I was able to remove as part of this update.  For whatever reason, the version of Chakra that we are using on the `development` branch is not respecting the responsive styles for `fontSize`.  I'll take a look tomorrow or Wednesday to see how complicated it would actually be to do that.

Cc @nypl-wluisi

PR #1497, opened 2024-01-19
DSD-1632: Tabs left/right scrolling 

jackiequach:
Fixes JIRA ticket [DSD-1632](https://jira.nypl.org/browse/DSD-1632).

- Updates `Tabs` component to be scrollable when the width of the tab set is greater than the width of the viewport.

## How has this been tested?

Locally on storybook

-

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/Tabs/Tabs.mdx
+++ b/src/components/Tabs/Tabs.mdx
@@ -2,6 +2,8 @@ import { ArgTypes, Canvas, Description, Meta, Source } from "@storybook/blocks";
 
 import * as TabsStories from "./Tabs.stories";
 import Link from "../Link/Link";
+import ComponentChangelogTable from "../../utils/ComponentChangelogTable";
+import { changelogData } from "./tabsChangelogData";
 
 <Meta of={TabsStories} />
 
@@ -19,9 +21,11 @@ import Link from "../Link/Link";
 - {<Link href="#accessibility" target="_self">Accessibility</Link>}
 - {<Link href="#composing-with-a-data-object" target="_self">Composing with a Data Object</Link>}
 - {<Link href="#responsive-mobile-carousel" target="_self">Responsive Mobile Carousel</Link>}
+- {<Link href="#scrollable-desktop-tab-set" target="_self">Scrollable Desktop Tab Set</Link>}
 - {<Link href="#callback-event-function" target="_self">Callback Event Function</Link>}
 - {<Link href="#url-hash-option" target="_self">URL Hash Option</Link>}
 - {<Link href="#children-components" target="_self">Children Components</Link>}
+- {<Link href="#changelog" target="_self">Changelog</Link>}
 
 ## Overview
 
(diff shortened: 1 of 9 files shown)


src/components/Tabs/Tabs.mdx
```diff
@@ -128,6 +132,15 @@ the <Link href="#children-components" target="_self">Children Components</Link>
 To view and test this in Storybook, go to the "Canvas" tab and change the
 viewport in the Storybook toolbar at the top of the page.
 
+## Scrollable Desktop Tab Set
+
+For desktop viewports (600px and greater), the full tab set scrolls left/right when
+the width of the tab set is greater than the width of the viewport.
+
+To scroll horizontally, use the `shift + scroll wheel` combination.
```
EdwinGuzman, line 140: Nice! Did not know this. I usually scroll horizontally on the mac trackpad but this is so useful.

src/components/Tabs/Tabs.stories.tsx
```diff
@@ -49,6 +49,27 @@ const animalCrossingData = [
   },
 ];
 
+const extraACData = [
+  {
+    label: "Filbert",
```
EdwinGuzman, line 54: 👍

EdwinGuzman (approved):
This looks good to me. Can you tag Clare and ask for an accessibility review on the new section and feature in the Jira ticket with the link to the Vercel Preview?

src/components/Tabs/Tabs.tsx
```diff
@@ -155,6 +156,7 @@ export const Tabs = chakra(
         useHash = false,
         ...rest
       } = props;
+      const [tabIndex, setTabIndex] = useState(0);
```
EdwinGuzman: Use the `defaultIndex` prop value as the initial state value.

src/hooks/useScrollTabIntoView.ts
```diff
@@ -0,0 +1,17 @@
+import React, { useEffect, useRef } from "react";
+
+export const useScrollTabIntoView = (index: number) => {
```
EdwinGuzman, line 9: Can you add a comment block that this is meant to only be used (as of now it seems) in the `Tabs` component. Also that this is an internal hook that is not meant to be used by devs outside of the DS or `Tabs` component. There are other similar non-consuming app hooks like `useDSHeading` and `useStateWithDependencies`.

src/index.ts
```diff
@@ -166,6 +166,7 @@ export { default as useCloseDropDown } from "./hooks/useCloseDropDown";
 // export { default as useMultiSelect } from "./hooks/useMultiSelect";
 export { default as useNYPLBreakpoints } from "./hooks/useNYPLBreakpoints";
 export { default as useNYPLTheme } from "./hooks/useNYPLTheme";
+export { default as useScrollTabIntoView } from "./hooks/useScrollTabIntoView";
```
EdwinGuzman: Ah but here it is being exported. I think it could be generalized if the query text is an argument and `"[role=tab][aria-selected=true]"` is the default value. Otherwise, this might be hard to use outside of the specific DS `Tabs` component.

EdwinGuzman (changes requested):
This works well on desktop but needs some minor updates.

On mobile, the tabs don't really scroll into view and it seems that the left/right button arrows are blocking the tabs from fully coming into view. I think we should go with this and the mobile update be a separate ticket unless if you want to update it here as well.

src/index.ts
jackiequach: Removed this since it is not ideal to create/pass a parameter for the dependency array used in the useEffect hook.

EdwinGuzman (approved):
Thanks for the update!

bigfishdesign13 (approved):
So awesome!  I love it.

PR #1499, opened 2024-01-22
DSD-1644: Fixes from VQA of form elms pt 1

oliviawongnyc:
Fixes JIRA ticket [DSD-1644](https://jira.nypl.org/browse/DSD-1644).

- The `width` property on the label was not being applied because the label was previously using `display: inline`. I am not sure why it was working before without a different display style.
- I also removed the "stack" part of CheckboxGroup because we are not using it to style. I don't think removing it broke anything as far as I could tell.
- I fixed the styling for the DatePicker's calendar triangle in dark mode. This was broken before, but Alkim just pointed it out.
- Finally, I adjusted the alignment of the DatePicker calendar's navigation arrows as they were not properly aligned with the name of the month.

## How has this been tested?

- Locally in Storybook.

- None.

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/DatePicker/_DatePicker.scss
+++ b/src/components/DatePicker/_DatePicker.scss
@@ -26,7 +26,7 @@
   // The left and right arrow buttons.
   .react-datepicker__navigation {
     &-icon {
-      top: 5px;
+      top: 2px;
       width: 0px;
       &::before {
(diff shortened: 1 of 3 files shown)


EdwinGuzman (approved):

bigfishdesign13 (approved):
Looks good.

PR #1500, opened 2024-01-23
DSD-1669: responsive font sizes for Heading and Text components (React 17 / Chakra 1)

bigfishdesign13:
Fixes JIRA ticket [DSD-1669](https://jira.nypl.org/browse/DSD-1669)

- Updates the `Heading` component to use a desktop-first approach to implement the responsive `font-size`

NOTE: These updates are based on the `development` branch which uses React 17 and Charka v1.

## How has this been tested?

- local Storybook
- To validate these changes, Turbine will be updated to import the DS from this brach.

- n/a

### Checklist:

### Front End Review:

### DIFF
--- a/src/theme/components/heading.ts
+++ b/src/theme/components/heading.ts
@@ -52,6 +52,12 @@ export const headings = {
     lineHeight: "1.10",
     width: "auto",
     a: { textUnderlineOffset: "7px" },
+    "@media (min-width: 0px)": {
+      fontSize: "mobile.heading.display1",
+    },
+    "@media (min-width: 600px)": {
+      fontSize: "desktop.heading.display1",
+    },
   },
   heading1: {
     fontWeight: "heading.heading1",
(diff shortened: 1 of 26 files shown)


EdwinGuzman (commented):
This looks good just update the Heading mdx prerelease and changelog.

The best way to test this would be in Turbine through a release candidate because of how Storybook renders components in iframes.

isastettler (commented):
Is it correct that this will just move the bug from visible on desktop to visible on mobile?

bigfishdesign13:
> Is it correct that this will just move the bug from visible on desktop to visible on mobile?

@isastettler In theory, yes.  The lesser of two evils?  We'll have a release candidate prepared soon for testing.

EdwinGuzman (approved):

src/theme/components/heading.ts
```diff
@@ -52,46 +52,88 @@ export const headings = {
     lineHeight: "1.10",
     width: "auto",
     a: { textUnderlineOffset: "7px" },
+    "@media (min-width: 0px)": {
+      fontSize: "mobile.heading.display1",
+    },
+    "@media (min-width: 600px)": {
+      fontSize: "desktop.heading.display1",
+    },
```
EdwinGuzman, line 95: I'm not sure if you need the media query for "0px" because that's the default. So it should be
```
fontSize: "mobile.heading.display1",
"@media (min-width: 600px)": {
  fontSize: "desktop.heading.display1",
},
```

bigfishdesign13: Yes, I thought that as well.  I gave it a try at one point, but the default styles on the Chakra `Heading` seem to override those styles.  However, when I put it into the media query, it worked.

EdwinGuzman (approved):
Reapproving after checking the latest updates.

PR #1502, opened 2024-01-24
DSD-1673: FeedbackBox form fixes

EdwinGuzman:
Fixes JIRA ticket [DSD-1673](https://jira.nypl.org/browse/DSD-1673)

Fixes two accessibility issues:
- Removes the `required` (through the DS `isRequired` prop) so that the native HTML validation popup does not display. Instead, if the form is submitted with an empty "Comment" field, then it will set the invalid state. This requires an update to keeping the invalid state within the component.

Screen Shot 2024-01-24 at 3 24 47 PM

- Forcibly focuses on the open button when the form is closed. This feature worked before but when it was implemented and rendered in various apps, the feature broke. It could be because of how different apps rerender but it's not clear what the issue is. This update will work in Storybook but we need to test the release candidate in Turbine to verify.

## How has this been tested?

Storybook locally.

- Fixes issues Clare found while testing in the DC Facelift app.

### Checklist:

### Front End Review:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,15 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 ## Prerelease
 
+### Updates
+
+- Updates the required comment field so it cannot be submitted when empty for the `FeedbackBox` component.
+
+### Fixes
+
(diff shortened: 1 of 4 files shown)


bigfishdesign13 (approved):
These updates look good.

PR #1505, opened 2024-01-29
DSD-1641: QA fixes on Overlays/Switchers Chakra 2.8

7emansell:
Fixes JIRA ticket [DSD-1641](https://jira.nypl.org/browse/DSD-1641)

- QA fixes on accordion, modal, and tooltip

## How has this been tested?

Locally, Storybook

### Checklist:

### Front End Review:


src/components/Modal/Modal.tsx
```diff
@@ -72,7 +72,7 @@ export const BaseModal: ChakraComponent<
 
           <ModalFooter>
             <ButtonGroup>
-              <Button id="modal-close-btn" onClick={onClose}>
+              <Button id="modal-close-btn" height="40px" onClick={onClose}>
```
EdwinGuzman: Can you update this and the open button to `height="10"`? It's the same 40px value but in the design token name.

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -46,7 +46,9 @@ export const WithControls: Story = {
   },
   render: (args) => (
     <Tooltip {...args}>
-      <Button id="tooltip-button">Hover me</Button>
+      <Button height="40px" id="tooltip-button">
```
EdwinGuzman: This makes me think that the height should be added to the `Button` component instead of here multiple times.

src/theme/components/accordion.ts
```diff
@@ -13,6 +13,11 @@ const containerStyles = {
 };
 const buttonStyles = {
   borderWidth: "1px",
+  " > span ": {
+    display: "flex",
+    alignItems: "center",
+    minHeight: "24px",
```
EdwinGuzman: If you update this to "20px", you can remove the "6.5px 16px" value in `Accordion.tsx`. I think updating it here and once is better.

EdwinGuzman (changes requested):
It seems like the base height for the `Button` is off everywhere. Can you update it in the `button.ts` file and see if that fixes it in all these components? It _might_ throw off other components so check `SearchBar` and `ButtonGroup`.

src/components/Modal/Modal.tsx
EdwinGuzman: Just a reminder that this update is now in the `react18-and-chakra` base branch so pull the latest updates instead of making this update.

src/theme/components/accordion.ts
7emansell: I'm not sure there's a number for the span height that allows design token spacing for both the case where there's only one accordion button and the case where they're stacked– one of them will need custom padding

7emansell:
Looks okay on `Searchbar` and `ButtonGroup` to me as well

EdwinGuzman (approved):
It's like 1px off in the Accordion but that's not a big blocker. Perhaps preface that when passing this back to QA.

PR #1506, opened 2024-01-29
DSD-1331: Menu component

7emansell:
Fixes JIRA ticket [DSD-1331](https://jira.nypl.org/browse/DSD-1331).

- Creates Menu component (and documentation/tests)
- ** See notes on highlight color, alignment, and icon

## How has this been tested?

Locally, Storybook

### Checklist:

### Front End Review:


src/theme/components/menu.ts
```diff
@@ -0,0 +1,113 @@
...
+const Menu = {
+  parts: ["menuItem", "menuButton", "menuGroup", "menuList"],
+  baseStyle: ({
+    highlightColor = "blogs",
+    showBorder,
+    showLabel,
+  }: MenuStyleProps) => {
+    const sectionColor = getSectionColors({
+      type: highlightColor,
+      colorValue: highlightColor === "blogs" ? "secondary" : "primary",
```
7emansell: @bigfishdesign13 There's no design token for section colors with 5% opacity, so currently I don't have the background of the selected item set. I tried a couple CSS workarounds that did not work mainly bc they're not passed as rgb values– is there a plan to add these colors to the DS

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,286 @@
...
+
+          return [...lst, menuItem];
+        }, []);
+
+      return (
+        <ChakraMenu id={id} autoSelect={false} {...rest}>
+          {({ isOpen }) => (
+            <Flex
+              flexDirection={listAlignment === "right" ? "row-reverse" : "row"}
+            >
```
7emansell: @bigfishdesign13 had some doubts about how this alignment works so just flagging this

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,286 @@
...
+        <MenuButton
+          sx={styles.menuButton}
+          backgroundColor={isOpen ? "ui.link.primary-05 !important" : "unset"}
+        >
+          {showLabel && (
+            <>
+              <span style={{ paddingRight: "8px" }}>{labelText}</span>
+              <Icon
+                name="arrow"
+                iconRotation={isOpen ? "rotate180" : "rotate0"}
```
7emansell, line 212: Icon does not exist yet in DS, this is the stand in

src/theme/components/menu.ts
bigfishdesign13: That token, and a whole array of related tokens, is being completed in the work I am doing for `StatusBadge`.

https://github.com/NYPL/nypl-design-system/pull/1507

EdwinGuzman: Small issue, the `StatusBadge` PR is for the React18 branch and this PR is for the `development` branch.

Can we live without the proper design token value for now until the `Menu` component is promoted to the new version?

bigfishdesign13: Or we can can go ahead and add the the necessary token in this branch.  

@7emansell I can show you what I did in the orther branch and then you can apply that here.

Thoughts?

EdwinGuzman: Adding them to this PR sounds good to me. When merging upstream, we should just watch out for any conflicts.

7emansell: @bigfishdesign13 sounds good!

src/components/Menu/Menu.mdx
```diff
@@ -0,0 +1,75 @@
...
+import * as MenuStories from "./Menu.stories";
+import Menu from "../Menu/Menu";
+
+<Meta of={MenuStories} />
+
+# Menu
+
+| Component Version | DS Version   |
+| ----------------- | ------------ |
+| Added             | `0.0.10`     |
```
EdwinGuzman: This is a new component so this should also be "prerelease"!

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,286 @@
...
+
+interface DividerItem {
+  type: "divider";
+  id: string;
+}
+
+/** List items can be one of the three types: */
+export type ListItemsData = ActionItem | GroupItem | DividerItem;
+
+const Menu = chakra(
```
EdwinGuzman: Export this variable as well. The description for all the props are not appearing in the args table. I think it's because Storybook can't pick up the component as this export (the default export should be enough but I forgot why it needs both).

src/components/Menu/Menu.mdx
```diff
@@ -0,0 +1,75 @@
...
+## Table of Contents
+
+- {<Link href="#overview" target="_self">Overview</Link>}
+- {<Link href="#component-props" target="_self">Component Props</Link>}
+- {<Link href="#accessibility" target="_self">Accessibility</Link>}
+- {<Link href="#menu-types" target="_self">Menu Types</Link>}
+- {<Link href="#menu-content" target="_self">Menu Content</Link>}
+- {<Link href="#menu-highlight-colors" target="_self">Menu Highlight Colors</Link>}
+
+## Overview
```
EdwinGuzman, line 27: Can you add a code snippet example here? We explicitly have code examples such as https://nypl-design-system-git-dsd-1331-menu-nypl.vercel.app/?path=%2Fdocs%2Fcomponents-overlays-switchers-accordion--docs

The "show code" preview in storybook has a code snippet as well, but it's not ideal because some of it is garbled storybook code.

src/components/Menu/Menu.tsx
7emansell: Actually the stand in is line 210 but you know what I mean

src/components/Menu/Menu.mdx
```diff
@@ -0,0 +1,75 @@
...
+- [MDN Menu accessibility](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role)
+
+## Menu Types
+
+There are four types: button, text button, icon (with container), and icon (without container). These types are toggled with `showLabel` and `showBorder`.
+Use the icon type when there is limited space such as a table or on mobile.
+
+<Canvas of={MenuStories.MenuTypes} />
+
+## Menu Content
```
EdwinGuzman, line 62: I think it would be helpful to have code snippet examples of each as well as a paragraph that states what `listItemsData` expects. Or perhaps it can be a different section?

What I'm thinking is more information about the structure of each object in this array. For example, the `ActionItem` has specific properties that only work for that type, so `onClick` would only work for this type and if a dev passes that property as a `GroupItem` or `DividerItem`, it won't do anything.

Typescript will (I hope) catch the above issue but it's helpful to know how a dev should structure their data.

src/components/Menu/Menu.stories.tsx
```diff
@@ -0,0 +1,461 @@
...
+  decorators: [withDesign],
+  argTypes: {
+    className: { control: false },
+    id: { control: false },
+    highlightColor: {
+      control: "select",
+      options: sectionTypeArray,
+      defaultValue: "blogs",
+    },
+    listAlignment: {
```
EdwinGuzman, line 21: Add the rest of the props. Some can be as simple as `showLabel: { control: { type: "boolean" } },`.

src/components/Menu/Menu.mdx
EdwinGuzman: Update: Yes, typescript does catch this and it's helpful!
Screen Shot 2024-02-01 at 9 51 12 AM

src/components/Menu/Menu.stories.tsx
```diff
@@ -0,0 +1,461 @@
...
+ * and `parameters` properties in this object.
+ */
+export const WithControls: Story = {
+  args: {
+    className: undefined,
+    id: "Menu-id",
+    labelText: "I'm the menu",
+    showLabel: true,
+    showBorder: true,
+    listAlignment: undefined,
```
EdwinGuzman: The default is "left" so it's okay to set this to "left" instead of undefined.

src/components/Menu/Menu.stories.tsx
```diff
@@ -0,0 +1,461 @@
...
+
+import Menu, { ListItemsData } from "./Menu";
+
+const meta: Meta<typeof Menu> = {
+  title: "Components/Navigation/Menu",
+  component: Menu,
+  decorators: [withDesign],
+  argTypes: {
+    className: { control: false },
+    id: { control: false },
```
EdwinGuzman: Do something similar for the `listitemsData` prop so users can't update it in storybook.

src/components/Menu/Menu.stories.tsx
```diff
@@ -0,0 +1,461 @@
...
+      url: "https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Main?node-id=80460%3A101982",
+    },
+    jest: ["Menu.test.tsx"],
+  },
+};
+
+// The following are additional Menu Stories.
+
+export const MenuTypes: Story = {
+  args: {
```
EdwinGuzman: Our pattern is to only have `args` set for the `WithControls` story. All the other stories are static so users can't update it. So it's okay to remove the args for the rest of the stories and explicitly set the props in the `Menu` component.

EdwinGuzman: This also makes it cleaner because if you set `showLabel` to false, for example, _all_ of the `Menu`s in this story will not display the label even though it should for the first two `Menu`s.

src/components/Menu/Menu.stories.tsx
```diff
@@ -0,0 +1,461 @@
...
+    design: {
+      type: "figma",
+      url: "https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Main?node-id=80460%3A101982",
+    },
+    jest: ["Menu.test.tsx"],
+  },
+};
+
+export const MenuContent: Story = {
+  args: {
```
EdwinGuzman: Same idea here, no need for this and the props can be set explicitly.

src/components/Menu/Menu.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+    expect(await axe(container)).toHaveNoViolations();
+  });
+});
+
+describe("Menu opens and closes", () => {
+  it("renders content when it is opened", async () => {
+    render(<Menu labelText={"Menu"} listItemsData={defaultListItems} />);
+    const openButton = screen.getByText("Menu").closest("button");
+    expect(openButton).toBeInTheDocument();
+    openButton.click();
```
EdwinGuzman: In the next test, there's isn't an `await waitFor` call. If you use fireEvent.click, I think you can remove the await here.

src/components/Menu/Menu.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+  it("closes on Escape key press", () => {
+    render(<Menu labelText={"Menu"} listItemsData={defaultListItems} />);
+    const openButton = screen.getByText("Menu");
+
+    openButton.click();
+    const menuList = screen.getByTestId("menuList");
+
+    fireEvent.keyDown(document, { key: "Escape" });
+    expect(menuList).not.toBeVisible();
```
EdwinGuzman: Add an expectation above the keydown event to verify that `menuList` is visible.

src/components/Menu/Menu.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+  });
+
+  it("closes when clicking outside of the menu", () => {
+    render(<Menu labelText={"Menu"} listItemsData={defaultListItems} />);
+    const openButton = screen.getByText("Menu");
+
+    openButton.click();
+    const menuList = screen.getByTestId("menuList");
+    fireEvent.mouseDown(document.body);
+    expect(menuList).not.toBeVisible();
```
EdwinGuzman: Also make sure that `menuList` is visible before the mousedown event.

src/components/Menu/Menu.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+describe("Menu allows selection", () => {
+  it("registered an item has been selected", () => {
+    render(<Menu labelText={"Menu"} listItemsData={defaultListItems} />);
+    const openButton = screen.getByText("Menu");
+    openButton.click();
+    const button1 = screen.getByText("I'm item 1").closest("button");
+    expect(button1).toBeInTheDocument();
+
+    button1.click();
+    console.log(button1);
```
EdwinGuzman: This can be removed.

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,286 @@
...
+      id,
+      labelText,
+      listAlignment = "left",
+      selectedItem,
+      highlightColor = "blogs",
+      showBorder = true,
+      showLabel = true,
+      listItemsData,
+      ...rest
+    }) => {
```
EdwinGuzman: When using `forwardRef`, pass in `ref?` as the second argument. I'm not sure if you actually need it but it could probably go in the `ChakraMenu` component as `ref={ref}`.

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,286 @@
...
+      const [selected, setSelected] = useState(selectedItem);
+      const handleSelect = (id, customHandler) => {
+        if (customHandler) {
+          customHandler(id);
+        }
+        setSelected(id);
+      };
+
+      /**  Check props. */
+      const validateProps = (listItemsData, selectedItem) => {
```
EdwinGuzman: All this logic check does not need to be wrapped in a function. The function is invoked right away so this block of code will be called every time when the component renders or is updated.

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,286 @@
...
+          listItemsData.length
+        ) {
+          console.warn(
+            "NYPL Reservoir Menu: The `id` values for the list items are not all unique."
+          );
+        } else if (
+          listItemsData.some(
+            (item) =>
+              !item.type ||
+              item.type === "" ||
```
EdwinGuzman: I get this typescript issue locally:
Screen Shot 2024-02-01 at 11 16 52 AM

EdwinGuzman: It seems safe to remove because the other two checks are very similar.

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,286 @@
...
+        if (media) {
+          if (media.type === "icon") {
+            return <Icon name={media.name} size="medium" />;
+          } else {
+            return (
+              <Image
+                src={media.src}
+                alt={media.alt}
+                size="xxxsmall"
+                sx={{ borderRadius: "24px" }}
```
EdwinGuzman, line 190: This is fine but another option is setting `aspectRatio` to "square" and `imageType` to "circle"

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,286 @@
...
+
+      /** Renders the trigger button for the Menu List to open or close. */
+      const getButton = (isOpen) => (
+        <MenuButton
+          sx={styles.menuButton}
+          backgroundColor={isOpen ? "ui.link.primary-05 !important" : "unset"}
+        >
+          {showLabel && (
+            <>
+              <span style={{ paddingRight: "8px" }}>{labelText}</span>
```
EdwinGuzman: "xs" is the same as "8px".

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,286 @@
...
+          if (item.type === "group") {
+            // If item is a group: adds group label then passes children
+            return [
+              ...lst,
+              <MenuGroup
+                key={item.id}
+                sx={styles.groupItem}
+                title={item.label}
+              />,
+              ...getMenuElements(item.children, selectedItem),
```
EdwinGuzman: Like the recursive nature of this implementation 👍

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,286 @@
...
+          )}
+          {!showLabel && (
+            <Icon name="arrow" size={showBorder ? "small" : "medium"} />
+          )}
+        </MenuButton>
+      );
+
+      /** Renders all Menu Items to be passed to the Menu List. Flattened into a reduce() so that
+       Groups of items display together. */
+      const getMenuElements = (data: ListItemsData[] = [], selectedItem) =>
```
EdwinGuzman: Double check the `selectedItem` argument here. It doesn't seem to be used and is only passed recursively.

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,286 @@
...
+              isFocusable={true}
+              data-testid={isSelected ? "selected-item" : ""}
+              onClick={() => handleSelect(item.id, item.onClick)}
+              sx={{
+                ...styles.actionItem,
+                ...(isSelected && styles.selected),
+              }}
+            >
+              <>
+                <div
```
EdwinGuzman: The react fragment is not needed here.
The div can also be a `Box` as
```
<Box display="flex" alignItems="center" gap="8px">
  ...
```

src/index.ts
```diff
@@ -87,6 +87,7 @@ export { default as List } from "./components/List/List";
 export type { DescriptionProps, ListTypes } from "./components/List/List";
 export { default as Logo } from "./components/Logo/Logo";
 export type { LogoNames, LogoSizes } from "./components/Logo/Logo";
+export { default as Menu } from "./components/Menu/Menu";
```
EdwinGuzman, line 90: I _think_ you may need to export the types such as  `ListItemsData` , `ActionItem`, `GroupItem`, and `DividerItem`. When using it in a consuming app, how a dev structures their objects might be an issue if the types are not available.

src/theme/components/menu.ts
```diff
@@ -0,0 +1,113 @@
...
+    });
+    return {
+      groupItem: {
```
EdwinGuzman: Add `groupItem`, `dividerItem`, etc to the `parts` array.

EdwinGuzman (changes requested):
This looks and works well! There are many comments but they're not blockers and mostly nitpicks and clean up in the code. Let me know if you have any questions.

EdwinGuzman:
I did not review the section color update but I can take a look after your next round of updates (but I'll let @bigfishdesign13 take a closer look at that as well since he also worked on that in a different PR).

src/components/Menu/Menu.mdx
7emansell: Code snippet example of the menu or of `listItemsData`?

src/components/Menu/Menu.tsx
7emansell: For some reason the tokens do not translate

src/components/Menu/Menu.tsx
7emansell: `ref` can't go on `ChakraMenu` so it's on a wrapper

src/components/Menu/Menu.mdx
```diff
@@ -0,0 +1,150 @@
...
+        label: "I'm item 2",
+        onClick: () => {
+          console.log("Item Title 2 clicked");
+        },
+      },
+    ],
+  },
+  { type: "divider", id: "divider-1" },
+  {
+        type: "action",
```
EdwinGuzman: Nitpick: minor indentation issue.

src/components/Menu/Menu.mdx
```diff
@@ -0,0 +1,150 @@
...
+];
+
+...
+
+<Menu labelText={"Menu"} listItemsData={exampleListItemsData} />
+`} language="tsx" />
+
+All `ActionItem`s require an `onClick` function, but the `media` prop is optional. Media will display
+to the left of the menu text, and will render either an `Icon` or `Image`. The `media` prop requires a type (icon or image)
+and then either the name of the icon, or the source url and alt text of the image.
```
EdwinGuzman: Mention that it's the name of a *DS icon, just to be clear to use the DS' set of names. Perhaps link to the Icon doc page as well.

src/components/Menu/Menu.stories.tsx
```diff
@@ -0,0 +1,389 @@
...
+      <Menu
+        labelText={"LPA"}
+        highlightColor={"researchLibraryLpa"}
+        selectedItem={"item-title-1"}
+        listItemsData={defaultListItems}
+      />
+      <Menu
+        labelText={"Education"}
+        highlightColor={"education"}
+        selectedItem={"item-title-1"}
```
EdwinGuzman, line 442: No need to change this but just pointing out that `{"string-value"}` is the same as `"string-value"` for react props.

src/components/Menu/Menu.tsx
EdwinGuzman: Oh, make this a `Box` and then add `as="span"` as a prop.

EdwinGuzman (commented):
The latest updates look good but added some minor comments.

The tests look like they should pass and we briefly talked about adding back the `waitFor`. I also mentioned taking a look at the `Tooltip` tests which is somewhat similar to the `Menu`'s behavior.

Something else to look at: https://github.com/chakra-ui/chakra-ui/blob/v1/packages/menu/tests/menu.test.tsx

Chakra does not use `isVisible` in their own tests so we can follow the same pattern and not be tied down to using that assertion function.

EdwinGuzman:
The color update also looks fine. I think most of the color update Marty did was for the UI colors and not for the brand colors, so this update doesn't seem to appear in his PR.

EdwinGuzman:
One more thing, I think the stories could use a min-height so they render in a bigger window for the Docs page. Right now, if you open up a menu and scroll, you lose context of what is rendered:
Screen Shot 2024-02-02 at 1 38 52 PM

7emansell:
> Something else to look at: https://github.com/chakra-ui/chakra-ui/blob/v1/packages/menu/tests/menu.test.tsx

This as a paradigm was super helpful– definitely not the way I would intuitively write these tests but using `aria-expanded` works

7emansell:
Chakra menu button adds a <span> wrapper to the contents of the button, so had to move some stuff around with the svg. Unsure if the padding/margin created on it looks right so let me know

src/theme/foundations/colors.ts
```diff
@@ -56,21 +56,25 @@ const whatsOnSecondary = black;
 const brandObj = {
   light: {
     primary: brandPrimary,
+    "primary-05": hexToRGB(brandPrimary, 0.05),
```
bigfishdesign13, line 81: For the colors where you have added a new option, I'd like to see both `05` and `10` options added.  Please go ahead and do that.  Thanks.

src/theme/components/menu.ts
```diff
@@ -0,0 +1,134 @@
...
+            highlightColor === "whatsOn"
+              ? "dark.ui.bg.hover"
+              : darkSectionColor,
+        },
+      },
+      menuButton: {
+        padding: showLabel ? "8px 16px" : "",
+        borderRadius: "2px",
+        fontSize: "14px",
+        color: showBorder ? "ui.gray.xx-dark" : "ui.link.primary",
```
bigfishdesign13: Replace `ui.gray.xx-dark` with `ui.typography.heading`.  Where did you see that?  If that is being used somewhere in the VDL component, it should be fixed.

src/components/Menu/Menu.mdx
```diff
@@ -111,7 +111,8 @@ const exampleListItemsData = [
 `} language="tsx" />
 
 All `ActionItem`s require an `onClick` function, but the `media` prop is optional. Media will display
-to the left of the menu text, and will render either an `Icon` or `Image`. The `media` prop requires a type (icon or image)
+to the left of the menu text, and will render either an {<Link href="https://nypl-design-system.vercel.app/?path=/docs/components-media-icons-icon--docs"target="">Icon</Link>}
+or {<Link href="https://nypl-design-system.vercel.app/?path=/docs/components-media-icons-image--docs" target="">Image</Link>}. The `media` prop requires a type (icon or image)
```
EdwinGuzman: Use the production link: https://nypl.github.io/nypl-design-system/reservoir/v2/

src/theme/components/menu.ts
```diff
@@ -0,0 +1,134 @@
...
+        fill: "ui.black",
+        border: showBorder ? "1px solid #616161" : "unset",
```
bigfishdesign13: Replace the hex value with the design token `ui.border.default`.

bigfishdesign13: Also, we have traditionally broken the border style into separate `border` and `borderColor` attributes to allow us to use the design tokens properly.  With everything all in one attribute, you'll need to use the CSS var syntax.

src/theme/components/menu.ts
```diff
@@ -0,0 +1,134 @@
...
+        _hover: {
+          bg: "ui.link.primary-05",
+          color: showBorder ? "" : "ui.link.secondary",
```
bigfishdesign13: With the change above, you'll need to set an actual color value rather than leaving it empty.  In this case, you should use `ui.border.hover`.

src/theme/components/menu.ts
```diff
@@ -0,0 +1,134 @@
...
+        },
+        _active: { bg: "ui.link.primary-05" },
+        "> span > svg ": {
+          fill: showBorder
+            ? "ui.gray.xx-dark"
```
bigfishdesign13: Stay away from the primative values as much as possible and use the sematic tokens instead.  Again, I am curious where you saw this.  The VDL design should only be using design tokens for the color values.

src/theme/components/menu.ts
```diff
@@ -0,0 +1,134 @@
...
+            : showLabel
+            ? "ui.link.primary"
+            : "ui.black",
+          paddingTop: showLabel ? "" : "5px",
+          margin: showLabel ? "" : "7px",
+        },
+        _dark: {
+          color: showBorder ? "ui.white" : "dark.ui.link.secondary",
```
bigfishdesign13: These tokens are not correct.  For showBoder, use `dark.ui.typography.heading`. Else, use `dark.ui.link.primary`.

src/theme/components/menu.ts
```diff
@@ -0,0 +1,134 @@
...
+          borderColor: showBorder ? "dark.ui.border.hover" : "unset",
+          _hover: { bg: "dark.ui.link.primary-10" },
```
bigfishdesign13: Also add:
```
borderColor: showBorder ? "dark.ui.border.hover" : "unset",
color: showBorder ? undefined : "dark.ui.link.secondary",
```

src/theme/components/menu.ts
```diff
@@ -0,0 +1,134 @@
...
+            ? "ui.gray.xx-dark"
+            : showLabel
+            ? "ui.link.primary"
+            : "ui.black",
+          paddingTop: showLabel ? "" : "5px",
+          margin: showLabel ? "" : "7px",
+        },
+        _dark: {
+          color: showBorder ? "ui.white" : "dark.ui.link.secondary",
+          borderColor: showBorder ? "dark.ui.border.hover" : "unset",
```
bigfishdesign13: Change to:
```
borderColor: showBorder ? "dark.ui.border.default" : "unset",
```

src/theme/components/menu.ts
```diff
@@ -0,0 +1,134 @@
...
+          _hover: { bg: "dark.ui.link.primary-10" },
+          _active: { bg: "dark.ui.link.primary-10" },
+          "> span > svg": {
+            fill: showBorder ? "ui.white" : "dark.ui.link.secondary",
```
bigfishdesign13: Change this to:
```
fill: "dark.ui.typography.heading",
```

src/theme/components/menu.ts
```diff
@@ -0,0 +1,134 @@
...
+        _dark: {
+          textColor: "ui.white",
+          background:
+            highlightColor === "whatsOn"
+              ? "dark.ui.bg.hover"
+              : darkSectionColor,
+        },
+      },
+      menuButton: {
+        padding: showLabel ? "8px 16px" : "",
```
bigfishdesign13: Use design tokens to replace the `px` values.

src/theme/components/menu.ts
```diff
@@ -0,0 +1,146 @@
...
+          highlightColor === "blogs" ? "ui.bg.default" : lightSectionColor,
+        _dark: {
+          textColor: "ui.white",
+          background:
+            highlightColor === "whatsOn"
+              ? "dark.ui.bg.hover"
+              : darkSectionColor,
+        },
+      },
+      menuButton: {
```
bigfishdesign13: For general formatting, add the following to `menuButton`:
```
minHeight: { base: defaultElementSizes.mobileFieldHeight, md: "40px" },
minWidth: { base: defaultElementSizes.mobileFieldHeight, md: "40px" },
whiteSpace: "nowrap",
```

And for the icon fix, add this:
```
"> span ": {
  alignItems: showLabel ? undefined : "center",
  display: showLabel ? undefined : "flex",
  justifyContent: showLabel ? undefined : "center",
},
```

src/theme/components/menu.ts
```diff
@@ -0,0 +1,146 @@
...
+        },
+        _active: { bg: "ui.link.primary-05" },
+        "> span > svg ": {
+          fill: showBorder
+            ? "ui.gray.xx-dark"
+            : showLabel
+            ? "ui.link.primary"
+            : "ui.black",
+          paddingTop: showLabel ? "" : "5px",
+          margin: showLabel ? "" : "7px",
```
bigfishdesign13: For icon fix: remove these.

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,287 @@
...
+                {labelText}
+              </Box>
+              <Icon
+                name="arrow"
+                iconRotation={isOpen ? "rotate180" : "rotate0"}
+                size="xsmall"
+              />
+            </>
+          )}
+          {!showLabel && <Icon name="navigationMoreVert" size="medium" />}
```
bigfishdesign13: For the icon fix, change the size to `large`.

bigfishdesign13 (changes requested):
I did not make a comment on all of the necessary color changes.  However, I did add some specific comment to address the icon issue.  They are noted as such.

7emansell:
Noting that there is still some use of `ui.gray.*` and `ui.black` on the icon fills and the border of the menu list– this is what's in Figma, but I'm happy to update to different tokens if that would be clearer @bigfishdesign13

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,288 @@
...
+          }
+        }
+        return null;
+      };
+
+      /** Renders the trigger button for the Menu List to open or close. */
+      const getButton = (isOpen) => (
+        <MenuButton
+          sx={styles.menuButton}
+          aria-label={labelText}
```
EdwinGuzman: _I think_ you only want the aria label to render when `showLabel` is false. Otherwise, there's visible label text and then it's added in this aria-label and it might be redundant. Our typical pattern is to add aria-labels when `showLabel` is false, but this is not always true depending on the use-case.

EdwinGuzman (commented):
Regarding the color contrast issue Clare raised, here's a visual of it:
Screen Shot 2024-02-07 at 6 07 55 PM

Not sure what the solution is but does this mean a redesign of the highlighted item?

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,288 @@
...
+      const handleSelect = (id, customHandler) => {
+        if (customHandler) {
+          customHandler(id);
+        }
+        setSelected(id);
+      };
+
+      /**  Check props. */
+      if (!listItemsData) {
+        console.warn(
```
EdwinGuzman, line 116: I missed this earlier but can you add tests to verify that these messages are being logged? Here's an example https://github.com/NYPL/nypl-design-system/blob/development/src/components/Checkbox/Checkbox.test.tsx#L232

For this specific message (and perhaps others), typescript will complain about required props but you can get around that by adding `// @ts-ignore` directly above. For example: https://github.com/NYPL/nypl-design-system/blob/development/src/components/Checkbox/Checkbox.test.tsx#L256

src/theme/components/menu.ts
```diff
@@ -0,0 +1,191 @@
...
+        },
+      },
+      actionItem: {
+        outline: "none !important",
+        textColor: "ui.typography.body",
+        paddingY: "xs",
+        paddingX: "s",
+        fontSize: "desktop.body.body2",
+        fontWeight: "body.body2",
+        lineHeight: "150%",
```
bigfishdesign13: I am not sure you need this.  `lineHeight: 1.5` is the base lineHeight for all of our typography.

src/theme/components/menu.ts
```diff
@@ -0,0 +1,191 @@
...
+        "> div > svg": {
+          fill: "ui.black",
+        },
```
bigfishdesign13: We're going to leave the icons as black.  However, black is the default color for the `Icon` component, so you can remove this.

src/theme/components/menu.ts
```diff
@@ -0,0 +1,191 @@
...
+        textColor: showBorder ? "ui.typography.heading" : "ui.link.primary",
+        border: showBorder ? "1px solid" : "unset",
+        borderColor: "ui.border.default",
+        _hover: {
+          bg: "ui.link.primary-05",
+          textColor: showBorder ? "ui.typography.heading" : "ui.link.secondary",
+          borderColor: "ui.border.hover",
+          "> span > svg ": {
+            fill: showBorder
+              ? "ui.gray.xx-dark"
```
bigfishdesign13: Change this to `ui.typography.heading`.

src/components/Menu/Menu.stories.tsx
```diff
@@ -0,0 +1,414 @@
...
+        highlightColor={"research"}
+        selectedItem={"item-title-1"}
+        listItemsData={defaultListItems}
+      />
+      <Menu
+        labelText={"LPA"}
+        highlightColor={"researchLibraryLpa"}
+        selectedItem={"item-title-1"}
+        listItemsData={defaultListItems}
+      />
```
bigfishdesign13: To enhance the visual contrast of the various highlight colors, change this one to Books & More (booksAndMore).

EdwinGuzman:
The latest update looks good. I think from the dev implementation, this is complete but the last missing piece is the color contrast issue.

src/components/Menu/Menu.mdx
```diff
@@ -0,0 +1,151 @@
...
+
+<Canvas of={MenuStories.MenuTypes} />
+
+## Menu Content
+
+Menu items can be displayed with images, icons, or just as text. They can be grouped with headings, dividers, or both.
+
+The maximum height of the menu list container is `320px`, or roughly eight menu items. Past that height, it will scroll. The
+maximum width is `300px`. Past that width, menu items will wrap onto the next line. It is recommended that, like the menu button,
+menu item labels are kept short.
```
bigfishdesign13: The note about the length of the menu item labels is important.  Pull that our and move it to the bottom of this section and use the following format:
```
**IMPORTANT:**  Avoid long labels for list items.  Keep the item labels short and avoid sentences from wrapping to the next line. 
```

src/theme/components/menu.ts
```diff
@@ -0,0 +1,191 @@
...
+    "menuList",
+  ],
+  baseStyle: ({
+    highlightColor = "blogs",
+    showBorder,
+    showLabel,
+  }: MenuStyleProps) => {
+    const sectionColor = getSectionColors({
+      type: highlightColor,
+      colorValue: highlightColor === "blogs" ? "secondary" : "primary",
```
bigfishdesign13: To address the contrast issue that Clare noted, use `secondary` for both `blogs` and `research`.

src/components/Menu/Menu.mdx
```diff
@@ -0,0 +1,151 @@
...
+The Menu component renders a collection of button and/or link elements within a dropdown menu format.
+The component allows users to perform multiple actions from an element with a small footprint.
+
+## Component Props
+
+<Canvas of={MenuStories.WithControls} />
+
+<ArgTypes of={MenuStories.WithControls} />
+
+## Accessibility
```
bigfishdesign13, line 40: Add the following after the first paragraph in this section: 

```
After a list action item is clicked, consuming apps should be mindful of the action that is performed by the action item's onClick handler and adjust the focus accordingly.
```

bigfishdesign13 (changes requested):
Looking very good.  I just have some minor requests.

EdwinGuzman (approved):
Updates look great to me, thank you!

bigfishdesign13 (approved):
Wonderful!

PR #1507, opened 2024-01-30
DSD-1670: updated UI styles for StatusBadge

bigfishdesign13:
Fixes JIRA ticket [DSD-1670](https://jira.nypl.org/browse/DSD-1670)

## PR Review Estimate

`medium`

- Updates the `StatusBadge` component to add the `type` prop.
- Updates the `StatusBadge` component to change the UI styles.
- Adds `5%` and `10%` opacities for the semantic colors in the color pallette.

## How has this been tested?

- local Storybook

- Color choices are compliant with WCAG 2.1 standards.

### Checklist:

### Front End Review:


src/components/StatusBadge/StatusBadge.mdx
```diff
@@ -34,25 +36,41 @@ import Link from "../Link/Link";
 
 ## Accessibility
 
-The visual aspect of the `StatusBadge` component displays three different status
-levels through the use of three different colors. It is important to assign
-the appropriate text so that the user can understand the priority level of the
-`StatusBadge`, and not rely just on the color.
+The visual aspect of the `StatusBadge` component displays multiple different
```
EdwinGuzman: Nitpick: "multiple different" sounds a bit awkward.

src/components/StatusBadge/StatusBadge.stories.tsx
```diff
@@ -40,25 +49,178 @@ export const WithControls: Story = {
...
 // The following are additional StatusBadge example Stories.
-export const HigherBarrier: Story = {
+
+export const TypeVariants: Story = {
+  render: () => (
+    <Table
+      columnHeaders={["", "Variant", "Purpose", "Examples"]}
+      tableData={[
+        [
+          <>
```
EdwinGuzman, line 64: You shouldn't need to wrap a single element in a react fragment. Did this throw an error?

src/components/StatusBadge/StatusBadge.tsx
```diff
@@ -26,8 +42,16 @@ export const StatusBadge: ChakraComponent<
 > = chakra(
   forwardRef<HTMLDivElement, React.PropsWithChildren<StatusBadgeProps>>(
     (props, ref?) => {
-      const { children, className, id, level = "low", ...rest } = props;
-      const styles = useStyleConfig("StatusBadge", { variant: level });
+      const { children, className, id, level, type, ...rest } = props;
+
+      // let finalVariant;
```
EdwinGuzman: Okay to remove this.

src/components/StatusBadge/StatusBadge.tsx
```diff
@@ -26,8 +42,16 @@ export const StatusBadge: ChakraComponent<
...
+      // if (!level && !type) finalVariant = "low";
+      // else if (level) finalVariant = level;
+      // else if (type) finalVariant = type;
+
+      const finalVariant = level ? level : type ? type : "low";
```
EdwinGuzman, line 47: Should "neutral" be the default value?

src/components/StatusBadge/StatusBadge.tsx
```diff
@@ -1,20 +1,36 @@
...
 export type StatusBadgeTypes = typeof statusBadgeTypeArray[number];
 export interface StatusBadgeProps {
   /** Additional class for the component */
   className?: string;
   /** ID that other components can cross reference for accessibility purposes */
   id?: string;
-  /** Level of the status badge. Defaults to `"low"`. */
-  level?: StatusBadgeTypes;
+  /** Level of the status badge. */
+  level?: StatusBadgeLevels;
```
EdwinGuzman, line 26: This description should note that `level` is deprecated.

src/components/StatusBadge/StatusBadge.tsx
EdwinGuzman, line 47: This should also separately include a check to see if `level` is being used to log a deprecation message.

src/components/StatusBadge/statusBadgeChangelogData.ts
```diff
@@ -14,6 +14,6 @@ export const changelogData: ChangelogData[] = [
     version: "Prerelease",
     type: "Update",
     affects: ["Styles"],
-    notes: ["Chakra 2.8 update."],
+    notes: ["Chakra 2.8 update.", "Added 'type' prop.", "Updated UI styles."],
```
EdwinGuzman: And also include that `level` is deprecated.

EdwinGuzman (changes requested):
Some comments but mostly minor. Will or should the new colors be added to the colors doc page? Not needed for this PR but just curious if they should be added.

src/components/StatusBadge/StatusBadge.mdx
```diff
@@ -34,25 +36,41 @@ import Link from "../Link/Link";
...
+The `level` variants have been deprecated in favor of the semantic `type`
+variants. However, for backward compatability, the `level` prop will override
+the `type` prop.
 
-<Canvas of={StatusBadgeStories.LowerBarrier} />
+<Canvas of={StatusBadgeStories.LevelVariants} />
+
+## Labeling Variations
+
+The capitalization of a badge label will be controlled by the component and
```
charmingduchess: This section is a little confusing. "The capitalization of a badge label will be controlled by the component" implies to me that there is a prop in the component that determines capitalization. Unless I'm misunderstanding, I think all that needs to be here is "Labels will render exactly as they are passed into the component. If a use case calls for all caps, then an all caps string should be passed into the component."

src/components/StatusBadge/StatusBadge.mdx
```diff
@@ -34,25 +36,41 @@ import Link from "../Link/Link";
...
+with the general meaning associated with the semantic colors used in a specific
+variant. In this way, the user will be able to understand the priority of the
+`StatusBadge` and not rely just on the color.
 
-### Higher Barrier
+## Semantic Type Variants
 
-Used in instances where the associated message is the highest barrier. Use
-`"high"` and `"medium"` for these levels.
+The `semantic` variants align with specific uses. Please refer to the purpose of each variant.
```
charmingduchess: should this be "semantic `type` variants"?

src/theme/foundations/colors.ts
```diff
@@ -34,9 +35,30 @@ const graySemiMedium = "#D9D9D9";
 const grayLightCool = "#E9E9E9";
 const grayxLightCool = "#F5F5F5";
 const grayxxLightCool = "#FAFAFA";
+// semantic
+const errorPrimary = "#97272C";
+const errorSecondary = "#6F0106";
 const linkPrimary = "#0069BF";
 const linkSecondary = "#004B98";
 const linkTertiary = "#551A8B";
+const statusPrimary = "#F9E08E";
```
charmingduchess, line 44: I notice that these names don't correspond 1:1 with the options for type. What is the reasoning behind the names here?

src/components/StatusBadge/StatusBadge.tsx
```diff
@@ -1,20 +1,36 @@
...
+export type StatusBadgeLevels = typeof statusBadgeLevelArray[number];
+export const statusBadgeTypeArray = [
+  "informative",
+  "negative",
+  "neutral",
+  "positive",
+  "recommendation",
+  "warningHigh",
+  "warningLow",
+  "low",
```
charmingduchess, line 14: should low medium and high be included in this array? we don't see any examples with type having any of these values.

src/components/StatusBadge/StatusBadge.stories.tsx
bigfishdesign13: Yes, I tried that, but I get the following error message:
```
Missing "key" prop for element in array
```

src/components/StatusBadge/StatusBadge.tsx
bigfishdesign13: For backward compatibility, I set a default from the legacy values.

bigfishdesign13: While we say the prop is deprecated, like in other components, I did not add a console warning to announce that something is deprecated.  We just never know if/when something will get updated in our consuming apps.

src/components/StatusBadge/StatusBadge.mdx
bigfishdesign13: Yes, that could be better.  In fact, the word `not` was missing from one spot, which changed the whole meaning.  I just rewrote this description.

src/theme/foundations/colors.ts
bigfishdesign13: I would consider the `status` colors somewhat legacy, in that the creation of the other semantic colors negated the need for this color.  The `primary` is a yellow, which could be interpreted as a `low` status, and the `secondary` is red, which could be seen as a `high` status.  But the other colors now cover those sentiments much better.

That said, this is our only yellow color style, so it has remained in the palette.  Also, a `status` badge variant didn't make sense.

src/components/StatusBadge/StatusBadge.tsx
bigfishdesign13: Oddly, in the original component, the `level` prop was typed as `StatusBadgeTypes`.  The `type` prop did not exist yet.  In the updated component, I aligned the prop types with the prop names, so, for backward compatability, I left those value in the array.

EdwinGuzman (approved):
Update looks good.

PR #1508, opened 2024-01-30
DSD-1647: QA fix - minor update to button for consistent height

EdwinGuzman:
JIRA ticket [DSD-1647](https://jira.nypl.org/browse/DSD-1647)

Fixes an issue found by Alkim where the button height is not right for the `Select` example but this should fix it globally where other issues were pointed out -- @7emansell this should fix the issue you've encountered in the other PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -231,9 +231,9 @@
       }
     },
     "node_modules/@babel/helper-create-class-features-plugin": {
-      "version": "7.23.7",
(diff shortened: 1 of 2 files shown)


7emansell:
Looks good– should I remove this bit from my PR

EdwinGuzman:
Yes, I'll merge this in now so you can pull it in.

PR #1516, opened 2024-02-07
DSD-1636: React18 for Basic Elements

bigfishdesign13:
Fixes JIRA ticket [DSD-1636](https://jira.nypl.org/browse/DSD-1636)

- Updates `Card`, `FeaturedContent`, and `Hero` components to use the Chakra 2.8 updated API for their theme object.
- No functional changes, only how the theme file is defined.

## How has this been tested?

- local Storybook

- n/a

### Checklist:

### Front End Review:


bigfishdesign13:
@EdwinGuzman The work on the `Card` component is not going so smoothly and I am wondering if my struggles are stemming from the use of child components.  Do you think that would add additional complexities?  I was going to take a look at the `Notification` component, which also uses child components.

EdwinGuzman:
How is it not going smoothly? From a quick glance it looks ok but I may be missing the difficulty in  upgrading the `Card` styles.

src/theme/components/card.ts
```diff
@@ -67,86 +72,91 @@ const getBodyPaddingStyles = ({
...
+          marginBottom: "xs",
+          a: mainActionLink
+            ? {
+                ...baseLinkStyles,
+                textDecoration: "none",
+              }
+            : null,
+        },
+        ...baseBorderStyles,
+        ...layoutStyles,
```
EdwinGuzman: Try moving these to the `base:` object

src/components/Card/Card.tsx
```diff
@@ -166,13 +166,19 @@ const CardWrapper: React.FC<any> = chakra(
           id={id}
           className={className}
           ref={ref}
-          sx={styles}
+          sx={styles.base}
```
EdwinGuzman: Ah, here and on line 179, you don't need to do `styles.base` because `styles.base` is already being passed to the `<CardWrapper` component on line 321. So in this case you should only need to do `sx={styles}`. Same on line 179.

bigfishdesign13: Wow, that seemed to do the trick.  Thanks.

src/components/Card/Card.tsx
```diff
@@ -244,7 +244,7 @@ export const Card: React.FC<React.PropsWithChildren<any>> = chakra(
       backgroundColor && (customColors["backgroundColor"] = backgroundColor);
       foregroundColor && (customColors["color"] = foregroundColor);
 
-      const styles = useMultiStyleConfig("Card", {
+      const styles = useMultiStyleConfig("CustomCard", {
```
EdwinGuzman: Let's use `ReservoirCard`. I know some values still use "Custom" but those can be updated in a "clean up" pr soon.

src/theme/components/card.ts
```diff
@@ -247,4 +255,4 @@ const CardImage = {
   },
 };
 
-export default { Card, CardActions, CardContent, CardImage };
+export default { CustomCard, CardActions, CardContent, CardImage };
```
EdwinGuzman: For consistency, update `CardActions`, `CardContent`, and `CardImage` with a similar update as `CustomCard`. All of these use `useStyleConfig` so you won't need to use `defineMultiStyleConfig`. Here's an example: https://github.com/NYPL/nypl-design-system/blob/react18-and-chakra/src/theme/components/buttonGroup.ts

src/theme/components/featuredContent.ts
```diff
@@ -1,75 +1,84 @@
...
+        bgColor: "ui.bg.default",
+        left: full ? "50%" : "auto",
+        marginLeft: full ? "-50vw" : "auto",
+        marginRight: full ? "-50vw" : "auto",
+        position: "relative",
+        right: full ? "50%" : "auto",
+        width: full ? "100vw" : "100%",
+        _dark: {
+          bgColor: "dark.ui.bg.default",
+        },
```
EdwinGuzman: All of these should be in a `base` property "part" as well, and in `FeaturedContent`, it should be used as `styles.base`.

src/theme/components/hero.ts
```diff
@@ -255,11 +283,11 @@ const fiftyFifty = {
       lg: "50%",
     },
   },
-};
-const Hero = {
-  baseStyle: {
+});
+const Hero = defineMultiStyleConfig({
+  baseStyle: definePartsStyle({
```
EdwinGuzman: Didn't realize this before. Was this always an empty object? You can remove the new lines and the `// ...`.

EdwinGuzman (changes requested):

EdwinGuzman:
Forgot to add this comment before clicking "request changes".

The descriptions in all the args table are missing. The three components still have `React.FC<any>`. This should be updated so that typescript can better pick up the props. The pattern can be found in almost all the other components but it's generally the following but the interface/types values should be updated accordingly for every component.

```
ChakraComponent<
  React.ForwardRefExoticComponent<
    React.PropsWithChildren<NewsletterSignupProps> &
      React.RefAttributes<HTMLDivElement>
  >,
  React.PropsWithChildren<NewsletterSignupProps>
>
```

src/theme/components/hero.ts
bigfishdesign13: Yes, it was always empty.

src/theme/components/hero.ts
```diff
@@ -255,11 +282,9 @@ const fiftyFifty = {
       lg: "50%",
     },
   },
-};
+});
 const Hero = {
```
EdwinGuzman: Even if `bastStyle` is an empty object, make it `Hero = defineMultiStyleConfig({...` since the variants still use the parts.

EdwinGuzman (approved):
Update looks good but just one small comment.

PR #1517, opened 2024-02-08
DSD-1649: Updates focus ring colors to be more accessible

oliviawongnyc:
Fixes JIRA ticket [DSD-1649](https://jira.nypl.org/browse/DSD-1649)

- Updates focus ring color in `Notification` and `Breadcrumbs` to match color of text.

## How has this been tested?

- Locally, in Storybook. 

- None, this is an accessibility improvement.

### Checklist:

### Front End Review:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,6 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 - Updates `Tabs` to be scrollable when the width of the tab set is greater than the viewport.
 - Updates the required comment field so it cannot be submitted when empty for the `FeedbackBox` component.
 - Updates the `Heading` component to use a custom `@media query` method to handle the responsive `font-size` styles.
-- Updates the `useNYPLBreakpoints` hook to include `isSmallerThanMedium` and `isSmallerThanDesktop`, both of which implement a desktop-first design approach.
 - Updates the `Text` component to use native Chakra responsive styles for the font sizes of the `subtitle1` and `subtitle2` variants.
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 - Updates the required comment field so it cannot be submitted when empty for the `FeedbackBox` component.
 - Updates the `Heading` component to use a custom `@media query` method to handle the responsive `font-size` styles.
 - Updates the `useNYPLBreakpoints` hook to include `isSmallerThanMedium` and `isSmallerThanDesktop`, both of which implement a desktop-first design approach.
+- Updates focus ring color in `Notification` and `Breadcrumbs` to match color of text.
 
 ### Fixes
 
--- a/src/theme/components/breadcrumb.ts
+++ b/src/theme/components/breadcrumb.ts
@@ -14,6 +14,11 @@ const blogs = {
         color: "dark.ui.typography.heading",
       },
     },
+    _focus: {
+      outline: "none",
+      ring: "2px",
+      ringColor: "ui.black",
+    },
   },
   "li:last-child": {
     ".chakra-breadcrumb__link": {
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -24,6 +28,7 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 - Updates `Tabs` to be scrollable when the width of the tab set is greater than the viewport.
 - Updates the required comment field so it cannot be submitted when empty for the `FeedbackBox` component.
 - Updates the `Heading` component to use a custom `@media query` method to handle the responsive `font-size` styles.
+- Updates the `useNYPLBreakpoints` hook to include `isSmallerThanMedium` and `isSmallerThanDesktop`, both of which implement a desktop-first design approach.
```
EdwinGuzman: I'm pretty sure this was recently removed by Marty. Try updating this PR with the latest from `development`.

src/theme/components/breadcrumb.ts
```diff
@@ -14,6 +14,11 @@ const blogs = {
         color: "dark.ui.typography.heading",
       },
     },
+    _focus: {
+      outline: "none",
+      ring: "2px",
+      ringColor: "ui.black",
```
EdwinGuzman: This is the first time `ring` and `ringColor` are used in the DS. We have been using `outline` and `outlineColor`. It seems that the difference is that `ringColor` uses the box-shadow for the outline.

We also are not very consistent about this but in many places the box-shadow is set to "none". Did you have a specific reason why ring is used over outline? Otherwise, it might be better to stick with outline for now.

src/theme/components/notification.ts
```diff
@@ -121,6 +121,11 @@ const NotificationContent = {
             color: "dark.ui.link.secondary",
           },
         },
+        _focus: {
```
EdwinGuzman: This general pattern would be great to have a function where the color is passed as an argument. There is a function which sort of does this already that _might_ be usable here: https://github.com/NYPL/nypl-design-system/blob/development/src/theme/components/global.ts#L7

If it's not useable here, that's okay, but make sure that this focus object also includes a `_dark` option.

EdwinGuzman (changes requested):
It looks good but my biggest question is why the use of ring over outline?

src/theme/components/breadcrumb.ts
```diff
@@ -14,6 +16,7 @@ const blogs = {
         color: "dark.ui.typography.heading",
       },
     },
+    _focus: { ...customFocusColor("ui.black", "dark.section.blogs.secondary") },
```
EdwinGuzman: You should be able to do `_focus: customFocusColor("...")`

src/theme/components/breadcrumb.ts
```diff
@@ -95,6 +98,9 @@ const Breadcrumb = {
         color: "ui.gray.light-cool",
         textDecoration: "none",
       },
+      _focus: {
```
EdwinGuzman: Same here, you can use the function directly since that returns an object.

src/theme/components/notification.ts
```diff
@@ -121,6 +122,7 @@ const NotificationContent = {
             color: "dark.ui.link.secondary",
           },
         },
+        _focus: { ...customFocusColor("ui.black", "dark.ui.link.primary") },
```
EdwinGuzman: Use the function directly here, too.

EdwinGuzman (commented):
Looks fine but just update how the function is being used so it's not spread.

src/theme/components/breadcrumb.ts
bigfishdesign13: Use `dark.ui.typography.heading` for dark mode and keep the focus ring color in alignment with the text color.

src/theme/components/notification.ts
bigfishdesign13: Use `dark.ui.typography.body` for dark mode.  The dark mode link color does work, but, for consistency, I would like to use the same color as the text.

bigfishdesign13 (changes requested):
Looking good.  Just one change request comment.  In addition to that, apply the same focus ring color change to the `Notification` component's close X button.  I just realized the need for that as I was reviewing the PR.

bigfishdesign13 (approved):
Looks good to me, but I will allow Edwin to give the final code approval.

EdwinGuzman (approved):

PR #1476, opened 2023-12-11
DSD-1609: MultiSelect v2

gonuguntla:
Fixes JIRA ticket [DSD-1609](https://jira.nypl.org/browse/DSD-1609)

## This PR does the following:

- Adds the `MultiSelect` v2 component.
- Adds the `useMultiSelect` hook.
- Updates the `Accordion` component to include the `ariaLabel` and `buttonInteractionRef` props.

## How has this been tested?
Locally, Storybook

## Accessibility concerns or updates

-

### Checklist:

- [ ] I have updated the Storybook documentation accordingly.
- [ ] I have added relevant accessibility documentation for this pull request.
- [ ] All new and existing tests passed.

### Front End Review:

- [ ] Review the Vercel preview deployment once it is ready.

PR #1521, opened 2024-02-08
REDO - DSD-1609/Multiselect v2

EdwinGuzman:
Edwin: Made a mistake and merged in #1476 too early before Isa got a chance to review. Had to revert `development` to recreate this PR with Deepika's original commits and branch.

Fixes JIRA ticket [DSD-1609](https://jira.nypl.org/browse/DSD-1609)

- Adds the MultiSelect v2 component.
- Adds the useMultiSelect hook.
- Updates the Accordion component to include the ariaLabel and buttonInteractionRef props.

- Locally, Storybook 

### Checklist:

### Front End Review:


src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,480 @@
...
-  /** The id of the MultiSelect. */
+type MultiSelectOnChange = (event: React.ChangeEvent<HTMLInputElement>) => void;
+
+export type MultiSelectProps = {
+  /** The button text rendered within the MultiSelect. */
+  buttonText: string;
+  /** The number of items that will be visible in the list when the component first loads. */
+  defaultItemsVisible?: number;
+  /** The helperText sets the text for the internal HelperErrorText */
+  helperText: string;
```
isastettler: Remove `helperText` since it is not used within the component at this time. Once it is removed. the prop will have to be also removed from all tests (including MultiSelectGroup and FilterBar tests) and from the MultiSelect stories.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,480 @@
...
+export type MultiSelectListOverflowTypes =
+  typeof multiSelectListOverflowArray[number];
 export interface SelectedItems {
   [name: string]: { items: string[] };
 }
-interface MultiSelectCommonProps {
-  /** The id of the MultiSelect. */
+type MultiSelectOnChange = (event: React.ChangeEvent<HTMLInputElement>) => void;
+
+export type MultiSelectProps = {
```
isastettler: Make this a `interface` to be picked up by the `argsTable`.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,480 @@
...
-
+export const multiSelectListOverflowArray = ["scroll", "expand"] as const;
+export type MultiSelectListOverflowTypes =
+  typeof multiSelectListOverflowArray[number];
 export interface SelectedItems {
   [name: string]: { items: string[] };
 }
-interface MultiSelectCommonProps {
-  /** The id of the MultiSelect. */
+type MultiSelectOnChange = (event: React.ChangeEvent<HTMLInputElement>) => void;
```
isastettler: Remove this custom type for the change handler function. It seems that in the DS the pattern is to use the `(event: React.ChangeEvent<HTMLInputElement>) => void` directly on the prop declaration.
This implementation caused an issue in the consuming app as I was unable to match the type. 
The other option would be to export this type and make it available for the consuming app but it seems that that would be a new pattern (see Button,  Checkbox, Select etc.)

src/components/MultiSelect/MultiSelect.stories.tsx
```diff
@@ -0,0 +1,585 @@
...
+  },
+];
+
+const meta: Meta<typeof MultiSelect> = {
+  title: "Components/Form Elements/MultiSelect",
+  component: MultiSelect,
+  decorators: [withDesign],
+  argTypes: {
+    buttonText: {
+      description: "The button text rendered within the MultiSelect.",
```
isastettler: Once the MultiSelectProp is changed to `interface` all descriptions can be removed.

src/components/MultiSelect/MultiSelect.stories.tsx
```diff
@@ -0,0 +1,585 @@
...
+        "The items to be rendered in the Multiselect as checkbox options.",
+    },
+    listOverflow: {
+      description:
+        "Value used to view the all items for the MultiSelect component",
+      control: "radio",
+      options: multiSelectListOverflowArray,
+      table: { defaultValue: { summary: "scroll" } },
+    },
+    onChange: {
```
isastettler: Handlers seem to usually have `control: false`.

src/components/MultiSelect/MultiSelect.stories.tsx
```diff
@@ -0,0 +1,585 @@
...
+      description:
+        "The action to perform on the checkbox's onChange function. <br /> `onChange`: `(event: React.ChangeEvent<HTMLInputElement>) => void` <br /> ",
+    },
+    onMixedStateChange: {
```
isastettler, line 229: Same and onChange. Add `control: false`.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,476 @@
...
-  /** The action to perform for clear/reset button of MultiSelect. */
-  onClear?: () => void;
-  /** The type of MultiSelect that will be rendered. */
-  type: "listbox" | "dialog";
+  /** listOverflow is a property indicating how the list should handle overflow, with options limited to either "scroll" or "expand." */
+  listOverflow?: MultiSelectListOverflowTypes;
   /** The selected items state (items that were checked by user). */
   selectedItems: SelectedItems;
   /** Value used to set the width for the MultiSelect component. */
   width?: MultiSelectWidths;
```
isastettler, line 59: I noticed that this prop does not get picked up by storybook. 
This might be because width is a style prop and used here as a custom prop. Easy solution is to change the prop name to `multiSelectWidth` or something similar.

It can be adjusted in the theme file to match or it can stay in the theme file and simply be passed in the MultiSelect as `width: multiSelectWidth`

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,476 @@
...
+      const accordianButtonRef: React.RefObject<HTMLDivElement> =
+        useRef<HTMLDivElement>();
+
+      // Control the open or closed state of the MultiSelect.
+      const DEFAULT_ITEMS_LIST_HEIGHT = "0px";
+      const MINIMUM_ITEMS_LIST_HEIGHT = "215px";
+      const MAXIMUM_ITEMS_LIST_HEIGHT = "270px";
+      const [itemsList, setItemsList] = useState(items);
+      const [viewAllLabel, setViewAllLabel] = useState("View all");
+      const [listHeight, setListHeight] = useState(MINIMUM_ITEMS_LIST_HEIGHT);
```
isastettler: `useState` hook vs `const` variables
    
For values that will not change during the lifecycle of a component there is no need to use `useState` . I would recommend for the values of `listHeight`to use a ternary operator statement that can set the value at first render since it’s value is only dependent on passed props like this: 
```
    const listHeight =
        listOverflow === "expand"
          ? "unset"
          : isSearchable
          ? MAXIMUM_ITEMS_LIST_HEIGHT
          : MINIMUM_ITEMS_LIST_HEIGHT;
```
You could even just add the px values instead and remove the the hight variables together.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,476 @@
...
       } = props;
 
-      const commonProps = {
-        id,
+      // Create a ref to hold a reference to the accordian button, enabling us to programmatically focus it.
+      const accordianButtonRef: React.RefObject<HTMLDivElement> =
+        useRef<HTMLDivElement>();
+
+      // Control the open or closed state of the MultiSelect.
+      const DEFAULT_ITEMS_LIST_HEIGHT = "0px";
```
isastettler: Remove the `DEFAULT_ITEMS_LIST-HEIGHT`

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,476 @@
...
+          setListHeight(MINIMUM_ITEMS_LIST_HEIGHT);
+          if (isSearchable) {
+            setListHeight(MAXIMUM_ITEMS_LIST_HEIGHT);
+          }
+          setItemsList(items);
+        }
+      }, [listOverflow, items, isSearchable]);
+
+      // Sets the selected items count on the menu button.
+      const getSelectedItemsCount: number =
```
isastettler: Nit: Since this is a variable that hold a value and not a function I would call it `selectedItemsCount`.
I would also add it to the top of the component to group all variable declarations in one place.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,476 @@
...
+        if (item.children.every((childItem) => childItem.isDisabled)) {
+          item.isDisabled = true;
+        } else {
+          item.isDisabled = false;
+        }
+        // Return the updated isDisabled state for the parent item
+        return item.isDisabled;
+      };
+
+      const onChangeSearch = (event) => {
```
isastettler, line 193: I would recommend putting the SearchInput and it’s logic into a separate file to help future engineers when reading through the code. If kept in this file group it all SearchInput code together.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,476 @@
...
       };
 
-      if (type === "listbox") {
-        const listboxOnChange = onChange as ListboxOnChange;
+      /**
+       * Display default items based on the specified limit.
+       * This function creates a list of items to be displayed, limited by the `defaultItemsVisible` parameter.
+       * The resulting list is set using the `setItemsList` function.
+       */
+      const displayDefaultItems = useCallback(() => {
```
isastettler: This can be a variable set at the top of the component when it first renders. In Javascript there is a method for arrays called `.slice()`.  You can use it to create a shallow copy of an array and at the same time trim the array passing the start and end index to `slice()`.
So in this case I would do:
```
const defaultItemsList = isOverflowExpand
        ? items.slice(0, defaultItemsVisible)
        : items;
```
You can set the default to the correct list for both listOverflow views and set the itemsList after to the correct value without using a `useEffect`.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,476 @@
...
+        useRef<HTMLDivElement>();
+
+      // Control the open or closed state of the MultiSelect.
+      const DEFAULT_ITEMS_LIST_HEIGHT = "0px";
+      const MINIMUM_ITEMS_LIST_HEIGHT = "215px";
+      const MAXIMUM_ITEMS_LIST_HEIGHT = "270px";
+      const [itemsList, setItemsList] = useState(items);
+      const [viewAllLabel, setViewAllLabel] = useState("View all");
+      const [listHeight, setListHeight] = useState(MINIMUM_ITEMS_LIST_HEIGHT);
+      const [listItemsCount, setListItemsCount] = useState(defaultItemsVisible);
```
isastettler: Nit: I brought all variables to the top of the component so that they are all in one place (helps when reading through the component)
Here is a full list of variables (I personally would add to the top): 
```
   const MINIMUM_ITEMS_LIST_HEIGHT = "215px";
      const MAXIMUM_ITEMS_LIST_HEIGHT = "270px";
      const listHeight =
        listOverflow === "expand"
          ? "unset"
          : isSearchable
          ? MAXIMUM_ITEMS_LIST_HEIGHT
          : MINIMUM_ITEMS_LIST_HEIGHT;

      const isOverflowExpand =
        items.length >= defaultItemsVisible && listOverflow === "expand";
      const defaultItemsList = isOverflowExpand
        ? items.slice(0, defaultItemsVisible)
        : items;
      const [itemsList, setItemsList] = useState(defaultItemsList);
      const [isExpandable, setIsExpandable] = useState(true);

      const selectedItemsCount: number = selectedItems[id]?.items.length || 0;

      const selectedItemsString = `item${selectedItemsCount === 1 ? "" : "s"}`;
      const ariaLabelValue = `${buttonText}, ${selectedItemsCount} ${selectedItemsString} currently selected`;


      // Get the styles for the component
      const styles = useMultiStyleConfig("MultiSelect", {
        isBlockElement,
        width: multiSelectWidth,
      });
```
Note: I also changed slightly the way the ariaLabelValue is created. This way `selectedItemString` and `selectedItemsCount` can be passed to the `MultiSelectItemsCountButton` and don't have to recalculated.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,476 @@
...
+        // Return the updated isDisabled state for the parent item
+        return item.isDisabled;
+      };
+
+      const onChangeSearch = (event) => {
+        const value = event.target.value.trim().toLowerCase();
+
+        if (value === "") {
+          return listItemsCount === defaultItemsVisible
+            ? displayDefaultItems()
```
isastettler: If using the `isExpandable` variable, you could change this to:
```
       isExpandable ? setItemsList(defaultItemsList) : setItemsList(items);
          return;
```

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,476 @@
...
+          item.isDisabled = false;
+        }
+        // Return the updated isDisabled state for the parent item
+        return item.isDisabled;
+      };
+
+      const onChangeSearch = (event) => {
+        const value = event.target.value.trim().toLowerCase();
+
+        if (value === "") {
```
isastettler: An empty string `""` is falsy in Javascript, so you could just check `if(!value){`

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,476 @@
...
+      };
+
+      // The clearSearchKeyword method resets the search keyword in the MultiSelect component.
+      // If the listOverflow is set to "scroll," it sets the list of items to the full list (items).
+      // If listOverflow is set to a value other than "scroll,"
+      // it checks if the current count of displayed items (listItemsCount) is equal to the default number of visible items (defaultItemsVisible).
+      // If true, it displays the default items; otherwise, it sets the list of items to the full list.
+      // This method is responsible for clearing the search and displaying the appropriate items based on the listOverflow and item count conditions.
+      const clearSearchKeyword = () => {
+        if (listOverflow === "scroll") {
```
isastettler: If using the `isExpandable` variable you can simplify as such: 
```
  const clearSearchKeyword = () => {
        isExpandable ? setItemsList(defaultItemsList) : setItemsList(items);
      };
```

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,476 @@
...
+          list.push(currentItem);
+          if (count >= defaultItemsVisible) {
+            break;
+          }
+        }
+        setItemsList(list);
+      }, [items, defaultItemsVisible, setItemsList]);
 
+      // Separate effect for handling listOverflow "expand"
+      React.useEffect(() => {
```
isastettler, line 231: This useEffect is not necessary.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,476 @@
...
+        if (listOverflow === "expand") {
+          setListHeight("");
```
isastettler: Instead of setting setListHeight to an empty string it would be better to pass a css value ("unset") or `null`.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,476 @@
...
+              {viewAllLabel}
+            </Button>
+          );
+        }
 
+        // Return null if conditions are not met
+        return null;
+      };
+
+      const itemsNotFound = () => {
```
isastettler: Typescript allows to add return types to function. Since the DS is using Typescript, I would recommend using for the return types too. The type is added after the parenthesis. Since this function returns a component I would use a variable using component convention. I would do this on all functions that return a component.
Here is my version:
```
     const NoSearchResults = (): JSX.Element => {
        return (
          <Box id="items-not-found-text-id" marginTop="xs">
            No options found
          </Box>
        );
      };
```

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,476 @@
...
         return (
-          <MultiSelectListBox
-            {...commonProps}
-            ref={ref}
-            onChange={listboxOnChange}
-            {...rest}
-          />
+          <Box>
+            {showSearchInputBox()}
+            {itemsNotFound()}
```
isastettler: In general it is good practice to define functions above the invocations. In this case `itemsNotFound` is declared on line 385 but called here on line 340.
I would also make the function call conditionally. So in this case you could write: 
```
{listItems.length === 0 && itemsNotFound()}
```
Further I would make the itemsNotFound a component and create a ternary operator statement that will either render the view when no results are found or the CheckboxGroup with the (potentially) toggle button. This could looks something like this
```
   <Box>
        {isSearchable && (
          <TextInput
            id="multi-select-text-input-id"
            labelText={`Search ${buttonText}`}
            isClearable={true}
            isClearableCallback={clearSearchKeyword}
            placeholder={`Search for options`}
            onChange={onChangeSearch}
            showLabel={false}
            showRequiredLabel={false}
            type="text"
            __css={styles.searchInputBox}
            marginBottom="s"
          />
        )}
        {itemsList.length === 0 ? (
          <NoSearchResults />
        ) : (
          <>
            <CheckboxGroup
              id="multi-select-checkbox-group"
              layout="column"
              isFullWidth
              isRequired={false}
              labelText="Multi select checkbox group label"
              showLabel={false}
              name="multi-select-checkbox-group"
            >
              {itemsList.map((item: MultiSelectItem) => (
                <MultiSelectItem key={id} item={item} />
              ))}
            </CheckboxGroup>
            {isOverflowExpand && <ExpandToggleButton />}
          </>
        )}
      </Box>
```

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,476 @@
...
+      const DEFAULT_ITEMS_LIST_HEIGHT = "0px";
+      const MINIMUM_ITEMS_LIST_HEIGHT = "215px";
+      const MAXIMUM_ITEMS_LIST_HEIGHT = "270px";
+      const [itemsList, setItemsList] = useState(items);
+      const [viewAllLabel, setViewAllLabel] = useState("View all");
+      const [listHeight, setListHeight] = useState(MINIMUM_ITEMS_LIST_HEIGHT);
+      const [listItemsCount, setListItemsCount] = useState(defaultItemsVisible);
+
+      // Separate effect for handling listOverflow "scroll"
+      React.useEffect(() => {
```
isastettler: This useEffect is not needed. Generally speaking if the dependencies of a useEffect are never changing in the lifecycle of a component, it does not require a useEffect in most cases.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,476 @@
...
+      // Create a ref to hold a reference to the accordian button, enabling us to programmatically focus it.
+      const accordianButtonRef: React.RefObject<HTMLDivElement> =
+        useRef<HTMLDivElement>();
+
+      // Control the open or closed state of the MultiSelect.
+      const DEFAULT_ITEMS_LIST_HEIGHT = "0px";
+      const MINIMUM_ITEMS_LIST_HEIGHT = "215px";
+      const MAXIMUM_ITEMS_LIST_HEIGHT = "270px";
+      const [itemsList, setItemsList] = useState(items);
+      const [viewAllLabel, setViewAllLabel] = useState("View all");
```
isastettler: There are a few values that change upon the toggle of the list view. 
Mainly the itemsList and this label. I would recommend having a toggle boolean set on state that then handles both values. I tested this adding 
```
 const [isExpandable, setIsExpandable] = useState(true);
```
Then you can control the button label like this:
```
       <Button
            buttonType="text"
            fontSize="desktop.button.default"
            id="view-all-text-btn"
            onClick={toggleItemsList}
            __css={styles.viewAllButton}
          >
            {isExpandable ? "View all" : "View less"}
         </Button>
```
and add one `useEffect` to control the `itemsList` like this:
```
   React.useEffect(() => {
        setItemsList(isExpandable ? defaultItemsList : items);
      }, [isExpandable]);
```

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,476 @@
...
+
+      /**
+       * Render a search input box if the MultiSelect is searchable.
+       * This function conditionally renders a TextInput component for searching options
+       * based on the isSearchable prop. It returns null if the MultiSelect is not searchable.
+       *
+       * @returns {JSX.Element | null} The rendered search input box or null if not searchable.
+       */
+      const showSearchInputBox = () => {
+        if (isSearchable) {
```
isastettler: Rather than checking here for `isSearchable` you can check before you make the function call. 
For readability I would recommend creating a file for the SearchInput component and it's logic.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,476 @@
...
+              id="multi-select-checkbox-group"
+              layout="column"
+              isFullWidth
+              isRequired={false}
+              labelText="Multi select checkbox group label"
+              showLabel={false}
+              name="multi-select-checkbox-group"
+            >
+              {itemsList.map((item: MultiSelectItem) => (
+                <>{generateCheckboxArray(item)}</>
```
isastettler: React is login warnings as it can't find the `key` prop on the mapped items. This can be fixed by making this function a React component. 
So this would look something like this here:
```
 {itemsList.map((item: MultiSelectItem) => (
     <MultiSelectItem key={id} item={item} />
 ))}
```
The Design System is also logging warnings as it is not detecting that the children of the `CheckboxGroup` are `Checkbox` components. I am not sure if that can be fixed.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,476 @@
...
+          );
+        }
+        return null; // Return null when itemsList.length !== 0
+      };
+
+      /**
+       * Toggle between viewing all items and a limited number of default items.
+       * Updates the state to control the visibility and height of the items list.
+       */
+      const viewAllItems = () => {
```
isastettler: If the toggle is handled though a boolean value this can be shortened to: 
```
      const toggleItemsList = () => {
        setIsExpandable((prevProp) => !prevProp);
      };
```

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,476 @@
...
+      const ariaLabelValue = `${buttonText}, ${getSelectedItemsCount} item${itemPlural} currently selected`;
+      return (
+        <Box id={id} __css={styles} {...rest}>
+          <Accordion
+            accordionData={[
+              {
+                accordionType: "default",
+                // Pass the ref for interaction with the accordion button.
+                buttonInteractionRef: accordianButtonRef,
+                label: buttonTextLabel(),
```
isastettler: Rather than functions buttonTextLabel and displayAccordionData can be variables. I would name them both referring to what they are so that in the code it becomes more obvious right away. For example:
accordionLabel
accordionPanel

src/theme/components/multiSelect.ts
```diff
@@ -13,6 +7,9 @@ export const multiSelectWidths = {
     minWidth: {
       md: "175px",
     },
+    maxWidth: {
+      md: "200px",
```
isastettler: In Scout we use the MultiSelect for selecting a library branch. The branch and center names can be very long. We would recommend to bump the maxWidth to `"412px"` 

This will allow branch names like "Belmont Library and Enrico Fermi Cultural Center" and 
"Andrew Heiskell Braille and Talking Book Library" to still be on one line.

src/components/MultiSelect/MultiSelect.stories.tsx
```diff
@@ -0,0 +1,565 @@
...
+    name: "Business",
+  },
+];
+
+const meta: Meta<typeof MultiSelect> = {
+  title: "Components/Form Elements/MultiSelect",
+  component: MultiSelect,
+  decorators: [withDesign],
+  argTypes: {
+    buttonText: {
```
isastettler: buttonText can be removed.

src/components/MultiSelect/MultiSelect.stories.tsx
```diff
@@ -0,0 +1,565 @@
...
+    items: {
+      control: false,
+    },
+    listOverflow: {
+      control: "radio",
+      options: multiSelectListOverflowArray,
+      table: { defaultValue: { summary: "scroll" } },
+    },
+    onChange: {
+      table: { defaultValue: { summary: false } },
```
isastettler: Swap 
`table: { defaultValue: { summary: false } },` with ` { control: false },`
for both, `onChange` and `onMixedStateChange`

src/components/MultiSelect/MultiSelect.stories.tsx
```diff
@@ -0,0 +1,565 @@
...
+    },
+    onMixedStateChange: {
+      table: { defaultValue: { summary: false } },
+    },
+    selectedItems: {
+      control: false,
+    },
+    width: {
```
isastettler, line 235: If the prop name gets changed, make sure to change it here too.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,476 @@
...
+            ariaLabel={ariaLabelValue}
+            id="multi-select-accordion-id"
+            isDefaultOpen={isDefaultOpen}
+            isAlwaysRendered
+            panelMaxHeight={listHeight}
+            sx={{
+              ...styles.accordionStyles,
+            }}
+          />
+          <MultiSelectItemsCountButton
```
isastettler: Render this button conditionally `{selectedItemsCount > 0 && <MultiSelectItemsCountButton/>}`

src/components/MultiSelect/MultiSelectItemsCountButton.tsx
```diff
@@ -0,0 +1,87 @@
+import React, { forwardRef } from "react";
+import { useMultiStyleConfig } from "@chakra-ui/react";
+import Button from "../Button/Button";
+import Icon from "../Icons/Icon";
+import { SelectedItems } from "./MultiSelect";
```
isastettler: Remove this type.

src/components/MultiSelect/MultiSelectItemsCountButton.tsx
```diff
@@ -0,0 +1,87 @@
...
+export interface MultiSelectItemsCountButtonProps {
+  id: string;
+  /** The id of the MultiSelect using this button. */
+  multiSelectId: string;
+  /** The label text rendered within the MultiSelect using this button. */
+  multiSelectLabelText: string;
+  /** The open status of the MultiSelect menu. */
+  isOpen: boolean;
+  /** The selected items state (items that were checked by user). */
+  selectedItems: SelectedItems;
```
isastettler: Change this to `selectedItemsCount: string;`
and add `selectedItemsCount: number;` as a prop

src/components/MultiSelect/MultiSelectItemsCountButton.tsx
```diff
@@ -0,0 +1,87 @@
...
+ * A second button is rendered above the main button that displays the current
+ * number of selected items. Clicking on the second button will clear all
+ * the selected items and the main button's close event will not be fired
+ * (as expected).
+ */
+const MultiSelectItemsCountButton = forwardRef<
+  HTMLButtonElement,
+  MultiSelectItemsCountButtonProps
+>((props, _ref) => {
+  const {
```
isastettler, line 37: Update the props to match the new props passed.

src/components/MultiSelect/MultiSelectItemsCountButton.tsx
```diff
@@ -0,0 +1,87 @@
...
+    isOpen,
+    multiSelectId,
+    multiSelectLabelText,
+    accordianButtonRef,
+    onClear,
+    selectedItems,
+  } = props;
+
+  // Sets the selected items count on the menu button.
```
isastettler, line 47: Remove all of this and just create the aria-label with the new props:
```  
let selectedItemsAriaLabel = `remove ${selectedItemsCount} ${selectedItemsString} selected from ${multiSelectLabelText}`;
```

src/components/MultiSelect/MultiSelectItemsCountButton.tsx
```diff
@@ -0,0 +1,87 @@
...
+  let selectedItemsAriaLabel;
+  if (selectedItems[multiSelectId]?.items.length > 0) {
+    getSelectedItemsCount = `${selectedItems[multiSelectId].items.length}`;
+    const itemPlural = getSelectedItemsCount === "1" ? "" : "s";
+    selectedItemsAriaLabel = `remove ${getSelectedItemsCount} item${itemPlural} selected from ${multiSelectLabelText}`;
+  }
+
+  const styles = useMultiStyleConfig("MultiSelectItemsCountButton", {
+    isOpen,
+    hasSelectedItems: getSelectedItemsCount,
```
isastettler: Replace all the `getSelectedItemsCount` with the new passed prop `selectedItemsCount`.

src/components/MultiSelect/MultiSelectItemsCountButton.tsx
```diff
@@ -0,0 +1,87 @@
...
+  });
+
+  return getSelectedItemsCount ? (
```
isastettler: Remove the ternary since this is is handled in the MultiSelect.tsx

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,476 @@
...
+      const [itemsList, setItemsList] = useState(items);
+      const [viewAllLabel, setViewAllLabel] = useState("View all");
+      const [listHeight, setListHeight] = useState(MINIMUM_ITEMS_LIST_HEIGHT);
+      const [listItemsCount, setListItemsCount] = useState(defaultItemsVisible);
+
+      // Separate effect for handling listOverflow "scroll"
+      React.useEffect(() => {
+        if (listOverflow === "scroll") {
+          setListHeight(MINIMUM_ITEMS_LIST_HEIGHT);
+          if (isSearchable) {
```
isastettler: To fix a bug with storybook where the `itemsList` are not being updated correctly when the listOverflow type is changes in the controls I added this useEffect. 
```
    // Only for storybook
      React.useEffect(() => {
        setItemsList(defaultItemsList);
      }, [listOverflow]);
```
This is only an issue in storybook as in a actual use case the MultiSelect would not just randomly change the way it is displaying the options (I don't think?). This is only a temporary fix.

isastettler (changes requested):
This is one big component! Great job Deepika.
Just a few things to note here as there is no particular file to add this: 
1. I noticed that the `isDefaultOpen` prop is not working in storybook. I realized it is also not working for the Accordion component. I have not spent too much time investigating as I wanted to finish this review and possibly you already know this.
2. Due to the nested logic the console gets warnings from the checkbox group. This is not Pr related but It think it would be great to rethink those warnings.
3. I feel like the logic for the isSearchable MultiSelect in connection with the `"expand"` functionality is not intuitive? I am not sure what would be better but maybe someone else has an idea. 

Thank you for giving me time to work through this.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,138 +1,476 @@
...
+            title={buttonText}
+          >
+            {buttonText}
+          </Box>
         );
-      }
+      };
 
-      return null;
+      // const itemsTotal = selectedItems[id].items.length;
```
isastettler: Remove this comment.

src/components/MultiSelect/MultiSelect.tsx
EdwinGuzman: I think we should stick with `width` for now. I would expect this to work because the component has `shouldForwardProp` set to `true`. If this is only in Storybook, then the comment can go directly in the `.stories.tsx` file (as tedious as it is).

src/theme/components/multiSelect.ts
EdwinGuzman: Yes, this is something for @bigfishdesign13 to view as well. Just remember that we're building this for all teams and 412 is an awkward value. If we can bump it to up to 420 or something more rounded, it might fit better in layouts as well.

src/components/MultiSelect/MultiSelect.tsx
EdwinGuzman: No, this prop would not change randomly. If this is a limitation/bug from Storybook, then mention it in the docs and add two different static examples. I don't like the idea of having and shipping code to fix Storybook in the actual component.

src/theme/components/multiSelect.ts
bigfishdesign13: I missed this initially.  These values are off.  I will add clarifying comments on the other PR.

src/theme/components/multiSelect.ts
```diff
@@ -1,3 +1,17 @@
...
+
+// This function creates a set of function that helps us
+// create multipart component styles.
+const { defineMultiStyleConfig, definePartsStyle } =
+  createMultiStyleConfigHelpers([
+    "base",
+    "accordionStyles",
+    "buttonTextLabel",
+    "menuChildren",
+    "menuSearchInputBox",
```
bigfishdesign13, line 11: I know these are the names that were already in there, but for consistency with the docs, change the prefix from `menu` to `list`.

```
"listChildren",
"listSearchInputbox",
```

bigfishdesign13 (changes requested):
Looks good.  One minor request.

PR #1512, opened 2024-02-02
DSD-1679: vertical ellipsis icon

bigfishdesign13:
Fixes JIRA ticket [DSD-1679](https://jira.nypl.org/browse/DSD-1679)

## This PR does the following:

- Adds the `"navigationMoreVert"` option (vertical ellipsis) to the `Icon` component.

## How has this been tested?

- local Storybook

## Accessibility concerns or updates

- n/a

### Checklist:

- [ ] I have updated the Storybook documentation accordingly.
- [ ] I have added relevant accessibility documentation for this pull request.
- [ ] All new and existing tests passed.

### Front End Review:

- [ ] Review the Vercel preview deployment once it is ready.

PR #1533, opened 2024-02-23
DSD-1691: Banner

EdwinGuzman:
Fixes JIRA ticket [DSD-1691](https://jira.nypl.org/browse/DSD-1691)

- This builds the new `Banner` component using semantic grouping of color variants similar to `StatusBadge`.
- Contains the new lightbulb icon.
- This points to the `react18-and-chakra` branch.
- Need for My Account project (LSP)

## How has this been tested?

Locally in Storybook.

- Will go through an accessibility review soon.

### Checklist:

### Front End Review:


EdwinGuzman:
I think the colors could be refactored a bit but leaving that for another PR.

src/theme/components/banner.ts
```diff
@@ -0,0 +1,164 @@
...
+    },
+  },
+});
+const negative = definePartsStyle({
+  base: {
+    ...baseNegative,
+    color: "ui.error.primary",
+    a: {
+      color: "ui.error.primary",
+    },
```
EdwinGuzman: Note: the links in the "negative" variant are red but in the My Account Figma design they are blue.

src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,239 @@
...
+      variant: backgroundColor || highlightColor ? undefined : type,
+    });
+    const generalHeadingProps = {
+      size: "heading6" as HeadingSizes,
+      maxWidth: "800px",
+      noSpace: true,
+      color: type === "negative" ? "ui.error.primary" : null,
+      _dark: {
+        color: type === "negative" ? "dark.ui.error.primary" : null,
+      },
```
EdwinGuzman, line 174: Not happy that we have to override heading color styles here because the chakra style theme object does not override.

src/theme/components/statusBadge.ts
```diff
@@ -39,45 +38,45 @@ const high = defineStyle({
   },
 });
 // Type variants
-const neutral = defineStyle({
+export const neutral = defineStyle({
```
EdwinGuzman, line 41: Exporting to use in `banner.ts`. I think this can be much cleaner but this works for now.

src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,239 @@
...
+      ...rest
+    } = props;
+    const [isOpen, setIsOpen] = useState(true);
+    const handleClose = () => setIsOpen(false);
+    const styles = useMultiStyleConfig("Banner", {
+      backgroundColor,
+      highlightColor,
+      // If either `backgroundColor` or `highlightColor` are set, then it
+      // overrides the Banner types.
+      variant: backgroundColor || highlightColor ? undefined : type,
```
EdwinGuzman: Not a fan of passing undefined but this works to override the variant if either backgroundColor or highlightColor is passed.

icons/svg/action-lightbulb.svg
charmingduchess, line 1: Is this svg supposed to be blank?

src/components/Banner/Banner.mdx
```diff
@@ -0,0 +1,184 @@
...
+- {<Link href="#with-html-content" target="_self">With HTML Content</Link>}
+- {<Link href="#custom-color" target="_self">Custom Color</Link>}
+- {<Link href="#dismissible" target="_self">Dismissible</Link>}
+- {<Link href="#changelog" target="_self">Changelog</Link>}
+
+## Overview
+
+<Description of={BannerStories} />
+
+**IMPORTANT:** The `Banner` component mirrors the `Notification` component but
```
charmingduchess: do `Banner` and `Notification` have any functional or display differences? Also, when you say on top of the screen, does that mean above the DS header and or the local header?

src/theme/components/banner.ts
7emansell: Is this a problem with the design or something that will be need to be created with an override in My Account ?

src/components/Banner/Banner.stories.tsx
```diff
@@ -0,0 +1,380 @@
...
+                  at eros. Cum sociis natoque penatibus et magnis dis parturient
+                  montes, nascetur ridiculus mus. Vivamus sagittis lacus vel
+                  augue laoreet rutrum faucibus dolor auctor.{" "}
+                  <Link href="#">This is a link</Link>.
+                </>
+              }
+              heading="Neutral Banner"
+            />
+          </>,
+          "Indicates a general, non-ciritical status update.",
```
7emansell: nitpick, non-critical

7emansell (approved):
Lgtm plus minor question/typo

icons/svg/action-lightbulb.svg
bigfishdesign13: There is code in the file.  And the icon is working.  The only issue I can see in the code is that it needs a line return after the closing `</svg>` tag.

src/components/Banner/Banner.mdx
bigfishdesign13: @EdwinGuzman After reading this here and seeing how Vera responded, I think `mirrors` is too strong a word.  Let's change this paragraph to the following:
```
**IMPORTANT:** The `Banner` component is similar in form and function to the `Notification` component. However, the `Notification` component should be used for global messaging located at the top of the site, above the NYPL header, while the `Banner` component should be used for all messaging that sits within the flow of the page.
```

src/components/Banner/Banner.mdx
```diff
@@ -0,0 +1,184 @@
...
+const customH3 = <Heading level="h3">Custom H3 Heading</Heading>;
+const customH5 = <Heading level="h5">Custom H5 Heading</Heading>;
+
+<Banner heading={customH3} {...props} />
+<Banner heading={customH5} {...props} />
+`} language="tsx" />
+
+<Canvas of={BannerStories.CustomBannerHeading} />
+
+## Banner Icon
```
bigfishdesign13, line 149: Can you add a code example for this story?

src/components/Banner/Banner.mdx
```diff
@@ -0,0 +1,184 @@
...
+
+The default icon can be overridden by using the `icon` prop to pass a
+custom `Icon` component.
+
+<Canvas of={BannerStories.CustomBannerIcon} />
+
+## With HTML Content
```
bigfishdesign13, line 183: Same here.  I think a code example would be nice to see.

src/components/Banner/Banner.stories.tsx
```diff
@@ -0,0 +1,380 @@
...
+        ],
+        [
+          <>
+            <Banner
+              content={`
+              This is a "informative" Banner with a heading. Cras mattis consectetur
```
bigfishdesign13: ...`an` "informative"...

src/components/Banner/Banner.tsx
bigfishdesign13: This sounds like issues similar to what I was dealing with when trying to normalize the text colors in the `Hero` component.

src/components/Icons/iconChangelogData.ts
```diff
@@ -10,22 +10,29 @@ import { ChangelogData } from "../../utils/ComponentChangelogTable";
...
   },
   {
-    date: "2024-01-04",
-    version: "2.1.4",
+    date: "Prerelease",
+    version: "Prerelease",
     type: "Update",
     affects: ["Styles"],
     notes: ["Chakra 2.8 update."],
   },
```
bigfishdesign13: These can be combined into one entry in the changelog.

src/theme/components/statusBadge.ts
bigfishdesign13: Can these styles be moved into a global space?

src/theme/components/banner.ts
bigfishdesign13: That is an oversight in the My Account designs.  We need to account for all color posibilities, so the link color should align with the text color for all link states (default, hover, visited).

src/theme/components/banner.ts
```diff
@@ -0,0 +1,164 @@
...
+      paddingEnd: "l",
+      paddingStart: {
+        base: "s",
+        md: "l",
+      },
+      position: "relative",
+      width: "100%",
+      a: {
+        color: "ui.typography.heading",
+        textDecoration: "underline",
```
bigfishdesign13: Did you add this to enforce the underline in the `Banner` component?  The `Link component is underlined by default, so I think we can rely on that and remove this attribute here.  Unless there was somethign else.  As it is, this is causing the underline to show as a fat solid line.  If we opt to include this here, we will also need to add the additional attributes that control the dotted style and the line weight.

src/theme/components/banner.ts
```diff
@@ -0,0 +1,164 @@
...
+      justifyContent: "flex-start",
+      py: "s",
+      paddingEnd: "l",
+      paddingStart: {
+        base: "s",
+        md: "l",
+      },
+      position: "relative",
+      width: "100%",
+      a: {
```
bigfishdesign13, line 132: Add a `_hover` object and set the same color value.

src/theme/components/banner.ts
```diff
@@ -0,0 +1,164 @@
...
+        color: "ui.typography.heading",
+        textDecoration: "underline",
+      },
+      svg: {
+        fill: highlightColor ? highlightColor : null,
+      },
+      _dark: {
+        a: {
```
bigfishdesign13, line 142: Same here.  Add a `_hover` object and set the same color value.

src/theme/components/banner.ts
```diff
@@ -0,0 +1,164 @@
...
+      _dark: {
+        fill: "dark.ui.link.tertiary",
+      },
+    },
+  },
+});
+
+const Banner = defineMultiStyleConfig({
+  baseStyle: definePartsStyle(({ backgroundColor, highlightColor }) => ({
+    base: {
```
bigfishdesign13, line 114: Add `fontSize: "desktop.body.body2"`.  Sorry, this was not explicitly defined in the spec doc nor the Figma designs.

src/components/Banner/bannerChangelogData.ts
```diff
@@ -0,0 +1,19 @@
...
+ */
+import { ChangelogData } from "../../utils/ComponentChangelogTable";
+
+export const changelogData: ChangelogData[] = [
+  {
+    date: "Prerelease",
+    version: "Prerelease",
+    type: "Update",
+    affects: ["Styles"],
+    notes: ["Chakra 2.8 update."],
```
bigfishdesign13: Should we also add an entry to state that the component was added?

src/theme/components/banner.ts
```diff
@@ -0,0 +1,164 @@
...
+      fill: "ui.link.tertiary",
+      _dark: {
+        fill: "dark.ui.link.tertiary",
+      },
+    },
+  },
+});
+
+const Banner = defineMultiStyleConfig({
+  baseStyle: definePartsStyle(({ backgroundColor, highlightColor }) => ({
```
bigfishdesign13, line 113: I'm not sure where it should be included, but we need to set a `max-width` of 800px for the `heading` and `content` elements?

src/components/Banner/Banner.tsx
bigfishdesign13: I am seeing some issues if `backgroundColor` or `highlightColor` is set without settign the other.  The original idea was for the colors based on the `type` value to be used unless explicitly overridden with these props.  If only one is set, then the default color for the color that was not overridden should still be used.  However, that is not what is happening currently.

Let's talk about this functionality.  Maybe we should make both required if one is set.

src/components/Banner/Banner.mdx
```diff
@@ -0,0 +1,184 @@
...
+
+The `Banner` component can be rendered with and without a heading. By
+default, when a title is passed to the `heading` prop, the
+`Banner` component will render an `h2` element sized as a `heading6`
+heading. If this causes accessibility issues in your app, you can pass in a
+custom `Heading` element.
+
+In the following examples, the `Banner` component will render custom `h3`
+and `h5` elements. Note that no matter what heading level is set, the size will
+always be set to `heading6` internally by the `Banner` component.
```
bigfishdesign13, line 135: Yay!

bigfishdesign13 (changes requested):
This is looking very good, but there are few things that need to be addressed.

icons/svg/action-lightbulb.svg
EdwinGuzman: Don't think it's a big problem. Previous PRs have also shown a blank SVG preview: https://github.com/NYPL/nypl-design-system/pull/1512/files

src/components/Banner/Banner.mdx
EdwinGuzman: Sounds good. I took most of this, including the word `mirrors` from the spec doc so that may need an update as well.

src/components/Banner/Banner.mdx
EdwinGuzman: Unfortunately, I could not use the `useDSHeading` hook so not an ideal solution.

src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,239 @@
...
+export const bannerTypesArray = [
+  "informative",
+  "negative",
+  "neutral",
+  "positive",
+  "recommendation",
+  "warning",
+] as const;
+export type BannerTypes = typeof bannerTypesArray[number];
+export const bannerBgColorsArray = [
```
EdwinGuzman, line 22: @wluisi @isastettler is this what you had in mind where each component has a set of colors? This does not include the general colors and only section colors but I'm referring to the pattern.

src/components/Banner/Banner.mdx
oliviawongnyc: May also describe an example?

src/components/Banner/Banner.tsx
EdwinGuzman: Hmm so if the variant is "informative" but `highlightColor` is passed as "brand.primary", then it would still have the light blue background and only update the highlightColor?

It's possible but it might me re-doing this so that it uses the `type` prop instead of the `variant` prop, or just redo each variant object.

src/components/Banner/Banner.tsx
EdwinGuzman: It's only for the `Heading`. It has its own styles which cannot be easily overridden.

src/components/Banner/bannerChangelogData.ts
EdwinGuzman: Yes.

src/theme/components/banner.ts
EdwinGuzman: To verify, what is here is correct?

src/theme/components/banner.ts
EdwinGuzman: I did this because the link in the figma file has a solid underline and I thought that was intentional. I can remove this and use the default dotted underline.

src/theme/components/statusBadge.ts
EdwinGuzman: yes, but later?

src/components/Banner/Banner.mdx
```diff
@@ -0,0 +1,184 @@
...
+## Accessibility
+
+The `Banner` component renders with an HTML `aside` element as its
+wrapper. This is an HTML landmark element that is similar to adding an attribute
+of `role="complementary"`. For accessibility purposes, landmark elements should
+not be rendered inside other landmark elements such as the `header` and `footer`
+landmark elements. Adding a `Banner` component inside an HTML `main`
+landmark element is acceptable.
+
+### ARIA Label
```
oliviawongnyc, line 55: If this is required, should it have a red asterisk next to it in the props table?

src/components/Banner/Banner.mdx
```diff
@@ -0,0 +1,184 @@
...
+is set through the `ariaLabel` prop. A unique `aria-label` value, along with the
+`<aside>` HTML landmark element, helps screen readers better navigate a page
+with multiple `Banner`s.
+
+### Icons
+
+Icons rendered in the `Banner` component are decorative by default which
+means that they are hidden to screen readers. Since the "X" close icon inside
+the dismissible button is decorative and because there is no discernible text
+inside the button, an `aria-label` attribute is added to the button.
```
oliviawongnyc: The last sentence is confusing when first read. Maybe could just be: "If `isDismissable` is true, the "X" close icon is wrapped inside a button with an appropriate `aria-label` attribute."

src/components/Banner/Banner.mdx
```diff
@@ -0,0 +1,184 @@
...
+should always include wording that aligns with the type of notitication that is
+being delivered. For example, if the Banner is a warning or error, the
+heading or body copy within the Banner should use the words `"warning"` or
+`"error"` and provide a clear description of the existing or potential isssue.
+
+### Dynamic Banners
+
+If dynamically generated Banners are required, a container into which the
+Banner will be injected should be included in the DOM. The container
+element should have `aria-live="polite"` applied. The container element should
```
oliviawongnyc: Nit: wrap these `Banner`s in backticks

src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,239 @@
...
+  "dark.section.research-library-lpa.primary",
+  "dark.section.research-library-schomburg.primary",
+  "dark.section.research-library-schwarzman.primary",
+  "dark.section.whats-on.primary",
+] as const;
+export type BannerHighlightColors = typeof bannerHighlightColorsArray[number];
+
+interface BannerProps {
+  /** Label used to describe the `Banner`'s aside HTML element. */
+  ariaLabel?: string;
```
oliviawongnyc, line 75: Per my earlier comment, should this be required?

oliviawongnyc (commented):
Looks good!! I agree there should be more clarity between Notification and Banner, maybe an example in Storybook and eventually in Turbine? Also, I think the icon in the header would look better center aligned with heading, but of course defer to the designers.

src/components/Banner/Banner.tsx
nypl-wluisi: Yeah seems like per component there's a set of colors that would be allowed. What you have here is more/less what I'm doing for Newsletter Signup, but that might have a diff set of colors. What is the `-05` after all the section colors do ?

src/theme/components/banner.ts
bigfishdesign13: Yes, for the initial color.  But the hover state color definitely needs to be addressed.  And possibly the visited state.

src/theme/components/statusBadge.ts
bigfishdesign13: Sure, but let's not forget about this.  Perhaps this can be tackled when as part of the big "color" shift.

src/components/Banner/Banner.tsx
bigfishdesign13: I don't want this to be overly difficult or complex in the code.  Let's make it so the custom colors are required as a pair.  If you set one, you much also set the other.  Conditions and warnings will be necessary.

src/components/Banner/Banner.mdx
bigfishdesign13: @EdwinGuzman I updated the spec doc.

@oliviawongnyc I am going to stay away from specifics content examples.  I am hoping that the location distinctions will be enough to differentiate the intended uses.

src/theme/components/banner.ts
bigfishdesign13: Figma can not render all of the CSS styles available for link underlines, so we included a disclaimer in the VDL referencing this discrepancy.

https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Main?type=design&node-id=79380-138172&mode=design&t=GjUSgxQmuNw7yPtS-11

bigfishdesign13:
> Also, I think the icon in the header would look better center aligned with heading, but of course defer to the designers.

@oliviawongnyc The problem is that the heading is not required and the alignment with the body text is different.  But let's see what we can do, becuase I agree...the alignment could be better.

@EdwinGuzman For the icon styles, can you add logic to add `marginTop: "xxxs"` when `heading` is passed?

src/components/Banner/Banner.mdx
EdwinGuzman: Updated the language to better be reflect that it's optional.

src/components/Banner/Banner.tsx
EdwinGuzman: `-05` and `-10` is the percent opacity.

src/components/Banner/Banner.tsx
EdwinGuzman: Updated the text for this prop.

src/helpers/types.ts
```diff
@@ -48,7 +48,7 @@ export const sectionColorsDataMap = {
     secondary: null,
   },
   researchLibrarySchwarzman: {
-    primary: "section.research-library.schwartzman",
+    primary: "section.research-library.schwarzman",
```
EdwinGuzman, line 51: There's no "t" in "Schwarzman" so updated it throughout the repo. I noticed this when trying to apply the design token color. @bigfishdesign13

src/theme/components/banner.ts
EdwinGuzman: Added this to the container in `Banner.tsx`.

src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,252 @@
...
+  "dark.brand.primary-05",
+  "dark.section.blogs.primary-05",
+  "dark.section.books-and-more.primary-05",
+  "dark.section.connect.primary-05",
+  "dark.section.education.primary-05",
+  "dark.section.locations.primary-05",
+  "dark.section.research.secondary-05",
+  "dark.section.research-library.lpa-05",
+  "dark.section.research-library.schomburg-05",
+  "dark.section.research-library.schwarzman-05",
```
EdwinGuzman, line 43: @bigfishdesign13 the spec doc has these light/dark colors as `dark.section.research-library-lpa.primary-05`. But in the colors file, the design token value is `dark.section.research-library.lpa-05`. So there are some inconsistencies. I updated this to reflect the current design tokens in the repo but we can go either way: update the design token values or update the spec doc.

src/helpers/types.ts
bigfishdesign13: Good catch.  The token name is fine in the VDL.

src/components/Banner/Banner.tsx
bigfishdesign13: I will update the spec doc.  That is an error on my part.

src/theme/components/banner.ts
```diff
@@ -0,0 +1,170 @@
...
+      width: "100%",
+      a: {
+        color: "ui.typography.heading",
+        _hover: {
+          color: "ui.typography.heading",
+        },
+      },
+      svg: {
+        fill: highlightColor ? highlightColor : null,
+      },
```
bigfishdesign13, line 140: The icon alignment could be better.  Can you also add a condition to add `marginTop: "xxxs"` when `heading` is passed?

src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,252 @@
...
+    const styles = useMultiStyleConfig("Banner", {
+      backgroundColor,
+      highlightColor,
+      // If `backgroundColor` and `highlightColor` are set, then it
+      // overrides the Banner types.
+      variant: backgroundColor && highlightColor ? undefined : type,
+    });
+    const generalHeadingProps = {
+      size: "heading6" as HeadingSizes,
+      maxWidth: "800px",
```
bigfishdesign13: The max width needs to be applied to all of the text (i.e. heading and content).  Can you add this to the `<Box>` element on line 241?

src/components/Banner/Banner.mdx
```diff
@@ -0,0 +1,236 @@
...
+import ComponentChangelogTable from "../../utils/ComponentChangelogTable";
+import { changelogData } from "./bannerChangelogData";
+
+<Meta of={BannerStories} />
+
+# Banner
+
+| Component Version | DS Version   |
+| ----------------- | ------------ |
+| Added             | `0.23.2`     |
```
bigfishdesign13: This is leftover from `Notification`, right?  This should be set as `Prerelease`.

src/components/Banner/Banner.mdx
```diff
@@ -0,0 +1,236 @@
...
+  language="tsx"
+/>
+
+<Canvas of={BannerStories.HTMLContent} />
+
+## Custom Color
+
+It is recommended to rely on the color styles set with the `type` prop.
+However, in cases where a different color is necessary, a value from a defined
+set of colors can be used.
```
bigfishdesign13: Include something about the colors being used in tandem.  When one is set, the other is required.

src/components/Banner/Banner.mdx
```diff
@@ -0,0 +1,236 @@
...
+**IMPORTANT:** For the "negative" variant only, the heading and text color will
+both be set to the same color, `ui.error.primary` and `dark.ui.error.primary`
+for dark mode.
+
+<Canvas of={BannerStories.TypeVariants} />
+
+## Banner Heading
+
+The `Banner` component can be rendered with and without a heading. By
+default, when a title is passed to the `heading` prop, the
```
bigfishdesign13: This would be cleared if the word `title` was changed to `string`.

src/components/Banner/Banner.mdx
```diff
@@ -0,0 +1,236 @@
...
+`Banner` component will render an `h2` element sized as a `heading6`
+heading. If this causes accessibility issues in your app, you can pass in a
+custom `Heading` element.
```
bigfishdesign13: Same here.  This would be cleared if changed to 
```
If this causes accessibility issues in your app, you can set a custom heading by 
passing in a DS `Heading` component.
```

src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,252 @@
...
+      heading,
+      highlightColor,
+      icon,
+      id,
+      isDismissible = false,
+      type = "neutral",
+      ...rest
+    } = props;
+    const [isOpen, setIsOpen] = useState(true);
+    const handleClose = () => setIsOpen(false);
```
bigfishdesign13, line 157: Should this be opened up to consuming apps so they can keep track of banners that have been closed?

src/helpers/types.ts
bigfishdesign13: Do we need to leave the typo spelling in there along with the correct spelling to avoid a breaking change?

bigfishdesign13 (changes requested):
There are a few more things that need to be addressed.

src/helpers/types.ts
EdwinGuzman: No, this is a change that might break but is easy enough and should be corrected.

src/theme/components/banner.ts
EdwinGuzman: Added it in the component file because that's easier.

src/components/Banner/Banner.tsx
EdwinGuzman: Yes.

EdwinGuzman: But it would mean passing in an additional `onClose` prop. Perhaps for a v2?

EdwinGuzman:
Ready for another review.

CHANGELOG.md
```diff
@@ -6,13 +6,25 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
 
-## Unreleased
+## Prerelease
 
 ### Adds
 
 - Adds the `"actionLightbulb"` option to the `Icon` component.
 - Adds the `Banner` component.
 
+## Prerelease
```
bigfishdesign13: This should be `3.0.1 (March 28, 2024)`.

src/components/Banner/Banner.tsx
bigfishdesign13: Yes.  https://jira.nypl.org/browse/DSD-1756

src/components/Icons/iconChangelogData.ts
```diff
@@ -16,6 +16,13 @@ export const changelogData: ChangelogData[] = [
     affects: ["Documentation", "Functionality"],
     notes: ["Added the actionLightbulb icon."],
   },
+  {
+    date: "Prerelease",
+    version: "Prerelease",
+    type: "Update",
+    affects: ["Documentation", "Functionality"],
+    notes: ["Addsed the editorMode (pencil) icon."],
```
bigfishdesign13: This came in with the merge from `development`, but `Addsed` should be `Added`.  Want to fix it here?

src/theme/components/banner.ts
bigfishdesign13: This works well, but it doesn't look like it is applied when a custom icon is used.

bigfishdesign13 (changes requested):
One last fix needed for the icon spacing.  Also, I just realized there are no examples without a heading.  Can you add additional examples in the `Banner Heading` section?  Depending on how you do that, it may also make sense to change the story name in the sidebar to `Banner Heading` rather than `Custom Banner Heading`.

src/theme/components/banner.ts
EdwinGuzman: I think it should be the responsibility of the consuming app to pass the marginTop rule into the custom Icon. I can add an example but it's not something we can easily enforce in the code.

bigfishdesign13: Ok, understood.  Leave it as is.

bigfishdesign13 (approved):
Looks great!

EdwinGuzman:
Merging this in. The `"schwartzman"` typo should not be an issue -- from a search in the org https://github.com/search?q=org%3ANYPL+schwartzman&type=code&p=1 it does not seem like anyone is using this design token explicitly.

PR #1535, opened 2024-02-26
DSD-1701: Storybook preview sizes

bigfishdesign13:
Fixes JIRA ticket [DSD-1701](https://jira.nypl.org/browse/DSD-1701)

- Updates the `viewport` preview options in Storybook to align with the Reservoir breakpoints.

## How has this been tested?

- local Storybook

- n/a

### Checklist:

### Front End Review:

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -38,14 +38,14 @@ const customViewports = {
   xl: {
     name: "xl - large tablet (less than 1280)",
     styles: {
-      width: "1280px",
+      width: "1200px",
       height: "1024px",
     },
   },
   "2xl": {
     name: "desktop (greater than or equal to 1280)",
     styles: {
-      width: "1440px",
+      width: "1280px",
       height: "1024px",
     },
   },
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -15,38 +15,38 @@ import { StorybookHeading } from "./storybookComponents";
 // Custom viewport options
 const customViewports = {
   sm: {
-    name: "sm (320px)",
+    name: "sm - small mobile (less than 480)",
     styles: {
-      width: "320px",
-      height: "568px",
+      width: "390px",
+      height: "667px",
     },
   },
   md: {
-    name: "md (600px)",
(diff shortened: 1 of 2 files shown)


.storybook/preview.tsx
```diff
@@ -15,38 +15,38 @@ import { StorybookHeading } from "./storybookComponents";
 // Custom viewport options
 const customViewports = {
   sm: {
-    name: "sm (320px)",
+    name: "sm - small mobile (less than 480)",
     styles: {
-      width: "320px",
-      height: "568px",
+      width: "390px",
```
EdwinGuzman, line 20: It's intentional to not have 480 or 479 but any width less than 480?

.storybook/preview.tsx
```diff
@@ -15,38 +15,38 @@ import { StorybookHeading } from "./storybookComponents";
...
       width: "960px",
-      height: "800px",
+      height: "1024px",
     },
   },
   xl: {
-    name: "xl (1280px)",
+    name: "xl - large tablet (less than 1280)",
     styles: {
       width: "1280px",
```
EdwinGuzman: This may need to be 1279, no?

EdwinGuzman (commented):
This may come up later but is the idea that QA will review every component in all viewports?

Just spot checking right now: the `FeaturedContent` loses its image when its set to "sm" under 480.

.storybook/preview.tsx
bigfishdesign13: Yes.  I set the Storybook preview sizes to align with the new Figma art board sizes.  And those sizes were setup as good working sizes within the breakpoint ranges.

bigfishdesign13:
> This may come up later but is the idea that QA will review every component in all viewports?

I don't think we say it outright, but yes.  This should be the case even now.

.storybook/preview.tsx
bigfishdesign13: Yes, good point.  Ultimately, there should be no UI differences between the `large mobile` and `desktop` designs and I got hung up on that.  However, after you pointed this out, I see that this should be changed.

EdwinGuzman (approved):

PR #1537, opened 2024-02-28
DSD-1698: Removing errors and refactoring the `Link` component

EdwinGuzman:
Fixes JIRA ticket [DSD-1698](https://jira.nypl.org/browse/DSD-1698)

- Initially was to just remove the error that is thrown and breaks the app when the DS `Link` component is used in next.js 13. Next.js 13 own `Link` component now renders an anchor element and the nested `<a><a>...</a></a>` elements threw a hydration error.
- The `Link` component has now been refactored to use Chakra's `Link` component which better supports framework's routing components through the `as` prop. We've already been using the `as` prop in the `Box` component internally to render certain elements but this is the first time we are doing so in a DS component. This is a special case as it's for support for external libraries.
- I tested this in the Research Catalog.
  - Currently, the next.js routing system is disabled for links: https://github.com/NYPL/research-catalog/blob/c89a1d436a8122a2b95bbc3a5f2c5705b9d3e0d9/src/components/RCLink/RCLink.tsx#L33
- I made a test npm release and it works well in two cases:
  - The first is using as recommended by Chakra in next.js 13+
```
import Link from "next/link";

<DSLink as={Link} href="...">...</DSLink>
```
  - The second case is the legacy case:
```
import Link from "next/link";

<Link href="..." passHref legacyBehavior>
  <DSLink>...</DSLink>
</Link>
```
- The `href` prop was never required but the component logged an error if it was missing when the child was string. I think we should leave it up to the consuming app to be careful about this. It would be noticed if a link is broken.
- There's a breaking change, which aligns with the v3 release, where we don't support anchor elements passed as a child:
```
<DSLink ...>
  <a ...>...</a>
</DSLink>
```
It's easier, logically in the app, to always render an anchor element ourselves. For consuming apps and devs using the `Link`, it's also easier and makes more sense to know that the DS `Link` will render an anchor element. Otherwise, all this is doing is adding styles and devs still have to add their own anchor element.
- All the other variants work as expected!

## How has this been tested?

Locally in Storybook and also in the Research Catalog.

- N/A

### Checklist:

### Front End Review:


jackiequach (approved):
These changes look great! DRB is currently using the legacyBehavior for Next.js Link, so this will be a nice update.

bigfishdesign13 (approved):
This looks good to me, but more developers should review as well.

7emansell (approved):
Lgtm!

PR #1541, opened 2024-02-28
Exporting helpers from Chakra

EdwinGuzman:

- Found out while testing the new version in the Header app that we need to export a few more helpers. This exports `FocusLock` which is now part of Chakra in 2.8, as well as `useStyleConfig` and `useMultiStyleConfig` for apps that need to create their own theme object (like the Header).
- This also exports the helper for the MatchMedia object.

## How has this been tested?

N/A will be tested in a release candidate.

- N/A

### Checklist:

### Front End Review:


bigfishdesign13 (approved):
💯

PR #1547, opened 2024-03-01
DSD-1711: MultiSelectGroup Chakra update

EdwinGuzman:
Fixes JIRA ticket [DSD-1711](https://jira.nypl.org/browse/DSD-1711)

- Updates `MultiSelectGroup` to use Chakra 2.8 and general update for its v2.

-

### Checklist:

### Front End Review:


EdwinGuzman:
@bigfishdesign13 I think we should discuss some of the props and combination. The non `isBlockElement` and column view does not work well. And there are general zindex issues:
Screen Shot 2024-03-01 at 5 08 29 PM

Try different permutations to see what I mean.

bigfishdesign13:
> I think we should discuss some of the props and combination. The non `isBlockElement` and column view does not work well. And there are general zindex issues:

The recommendations were to definitely not use `isBlockElement` for the column view, so that will resolve some of the potential issues you are seeing.  In the `FilterBar` component, we can enforce some fo the `MultiSelect` props based on the `FilterBar` layout, but other than really **STRONG** recommendations, I am not sure what we can do if and when a `MultiSelect` is used on its own.

src/components/MultiSelectGroup/MultiSelectGroup.stories.tsx
```diff
@@ -0,0 +1,317 @@
...
+    }
+    if (actionName === "onClear") {
+      action(actionName)(selectedItems);
+    }
+  }, [actionName, selectedItems]);
+
+  return (
+    <MultiSelectGroup
+      {...args}
+      renderMultiSelect={({ isBlockElement, multiSelectWidth }) => {
```
EdwinGuzman, line 81: @oliviawongnyc @jackiequach @bigfishdesign13 this is the render prop approach we were chatting about for the two Filter components. Let me know what you think. 

@charmingduchess this approach is also based on the recommendation you gave for the `CheckboxGroup`. The `react.cloneElement` function was removed internally from the `MultiSelectGroup` component and instead this render prop is used to control the the `isBlockElement` and `multiSelectWidth` props. I think something similar can be done for the `CheckboxGroup`.

src/components/MultiSelectGroup/MultiSelectGroup.mdx
```diff
@@ -0,0 +1,121 @@
...
+ease of use is prefered.
+
+The hook returns an object containing all the props and state needed to handle
+the selectedItems. That includes the functions `onChange`, `onClearAll`,
+`onMixedStateChange` for handling any changes to the selection of items and the
+current state of the selection: `selectedItems`. Additionally it returns
+`setSelectedItems` for setting an initial state of `selectedItems` or manipulate
+the `selectedItems` if needed.
+
+Due to it's data structure and functions setup, it is capable to handle a group
```
7emansell: nit its

src/components/MultiSelectGroup/MultiSelectGroup.stories.tsx
```diff
@@ -0,0 +1,317 @@
+import { VStack } from "@chakra-ui/react";
```
7emansell, line 1: Getting an error displaying the docs in Vercel: 
Screenshot 2024-04-02 at 1 23 00 PM

7emansell (commented):
The z-index behavior of one `Multiselect` over another when they are stacked in a row still seems unintuitive but I can see how there's no clear default. Assuming we're leaving that to the developer

EdwinGuzman:
> The z-index behavior of one Multiselect over another when they are stacked in a row still seems unintuitive but I can see how there's no clear default. Assuming we're leaving that to the developer

Yes. I added a note about how it's not a recommended pattern to use.

src/components/MultiSelectGroup/MultiSelectGroup.tsx
```diff
@@ -18,93 +22,77 @@ export interface MultiSelectGroupProps {
...
+    multiSelectWidth?: MultiSelectWidths;
+  }) => React.ReactNode;
   /** Is set to `true` by default and determines if the `labelText` is visible on the site. */
   showLabel?: boolean;
 }
 
 /**
- * `MultiSelectGroup` is a wrapper component specific for `MultiSelect` components. The wrapped `MutliSelect` components can be displayed in a
+ * `MultiSelectGroup` is a helper wrapper component specific for `MultiSelect`
+ * components. The wrapped `MutliSelect` components can be displayed in a
```
7emansell: Final nit multi

7emansell (approved):

src/components/MultiSelectGroup/MultiSelectGroup.mdx
```diff
@@ -0,0 +1,125 @@
...
+
+For accessibility information on the `MultiSelect` component, see the
+[MultiSelect Accessibility](../?path=/docs/components-form-elements-multiselect--docs#accessibility) documentation.
+
+## Layout Patterns
+
+The width of the MultiSelect components can be passed through the
+`multiSelectWidth` prop, with values of `"fitContent"` or `"full"`.
+On screen width under 600px width the `layout` is automatically set to "column"
+and the `multiSelectWidth` to "full". Note: when `layout = "column"` is passed,
```
bigfishdesign13: For consistency with other number comparisons, update this sentence to the following:
```
When the screen width is less than 600px, the `layout` prop is automatically set 
to `"column"` and the `multiSelectWidth` prop is set to `"full"`.
```

src/components/MultiSelectGroup/MultiSelectGroup.mdx
```diff
@@ -0,0 +1,125 @@
...
+the MultiSelect components will have `isBlockElement` set to `true` by default.
+
+**IMPORTANT**: we do not recommended setting `multiSelectWidth` to `"full"` and
+`layout` to `"row"` at the same time. When the `MultiSelect` components are
```
bigfishdesign13: Let's remove the `we`.
```
It is recommended to not set `multiSelectWidth` to `"full"` and `layout` 
to `"row"` at the same time.
```

src/components/MultiSelectGroup/MultiSelectGroup.mdx
```diff
@@ -0,0 +1,125 @@
...
+`useMultiSelect` is an optional but recommended hook for cases where managing
+the state of the component in the consuming app is less of a concern and general
+ease of use is prefered.
+
+The hook returns an object containing all the props and state needed to handle
+the selectedItems. That includes the functions `onChange`, `onClearAll`,
+`onMixedStateChange` for handling any changes to the selection of items and the
+current state of the selection: `selectedItems`. Additionally it returns
+`setSelectedItems` for setting an initial state of `selectedItems` or manipulate
+the `selectedItems` if needed.
```
bigfishdesign13: Rewording this paragraph...
```
The hook returns an object containing all the props and state needed to 
handle the `selectedItems` object. Included are the functions `onChange`, 
`onClearAll`, and `onMixedStateChange` for handling any changes to the 
items that are currently selected. Also included is the `selectItems` object 
to reflect the current state of the items that are currently selected. 
Additionally, the hook returns the `setSelectedItems` function for setting 
an initial state of `selectedItems` or, if needed, to manipulate `selectedItems`.
```

src/components/MultiSelectGroup/MultiSelectGroup.mdx
```diff
@@ -0,0 +1,125 @@
...
+
+Due to its data structure and functions setup, it is capable to handle a group
+of `MultiSelect` components. Find the full documentation under
```
bigfishdesign13:
```
Due to its data structure and the setup of its functions, the `useMultiSelect` 
hook is capable of handling a group of `MultiSelect` components.
```

src/components/MultiSelectGroup/MultiSelectGroup.tsx
```diff
@@ -1,13 +1,17 @@
-import { chakra, Stack, useMultiStyleConfig } from "@chakra-ui/react";
+import {
+  chakra,
+  ChakraComponent,
+  Stack,
+  // useMultiStyleConfig,
```
oliviawongnyc: Can this be removed?

bigfishdesign13 (changes requested):
Overall, this looks good.  I made some suggestions for the language in the docs.  Also, the spacing between the `MultiSelect` elements is missing from the column layout.  It should be set to `xs` (8px).

src/components/MultiSelectGroup/MultiSelectGroup.tsx
oliviawongnyc: Can we simplify this description a little? Maybe: "`MultiSelectGroup` is a wrapper for Reservoir `MultiSelect` components. It specifies display styles (ie, whether `MultiSelect`s are rendered in a column or row) and any associated text (ie, the group's label). Note, the `MultiSelectGroup` will not work with non-Reservoir `MultiSelect`s."

src/components/MultiSelectGroup/MultiSelectGroup.mdx
oliviawongnyc: I am a little confused by the fact that we say not to do something (Row (default) & multiSelectWidth='full' (default)) and then we show it in the examples. If we want it to be there can we mark it  as "Don't use this combination" or something? Also why is that example rendering in a column?

src/components/MultiSelectGroup/MultiSelectGroup.mdx
```diff
@@ -0,0 +1,125 @@
...
+[useMultiSelect](../?path=/docs/hooks-usemultiselect--docs).
+
+## Passing MultiSelect Children
```
oliviawongnyc, line 60: What do you think of putting this section above the `useMultiSelect` hook? I feel it's more important bc using the `renderMultiSelect` prop is required.

src/components/MultiSelectGroup/MultiSelectGroup.mdx
```diff
@@ -0,0 +1,125 @@
...
+
+In order to optimize rendering `MultiSelect` children and control the
+`isBlockElement` and `multiSelectWidth` props, the `MultiSelectGroup` component
+accepts `MultiSelect` children through its render prop `renderMultiSelect`
+function. This callback function returns an object argument with two optional
+properties: `isBlockElement` and `multiSelectWidth`.
+
+Both `isBlockElement` and `multiSelectWidth` are passed to the `MultiSelect`
+children.
```
oliviawongnyc: Do we need this: "In order to optimize rendering `MultiSelect` children and control the
`isBlockElement` and `multiSelectWidth` props,"? I already find this paragraph a bit confusing, and I think this adds to the confusion without explaining how it optimizes rendering or controls the props.

src/components/MultiSelectGroup/MultiSelectGroup.mdx
EdwinGuzman: It's there to show that this combination is possible but not working as intended. One reason it is not recommended is the issue you pointed out, that it's rendering in a column. When it's "row" and "full", the "full" makes it expand the entire width so it _is_ set to column and working, but taking up the entire width makes it float down to separate rows. I'll add a "Not recommended" prefix in the example.

EdwinGuzman: I think the bigger issue is that we are not recommending this but both prop values are the default. @bigfishdesign13 I think these values should be updated to "row" and "fitContent" as default. Yes?

src/components/MultiSelectGroup/MultiSelectGroup.mdx
EdwinGuzman: I think it's needed. It's not explicit why it's optimizing the rendering but it's doing so because the `react.cloneElement` function was removed internally. I do not want to flag that in the docs because I don't think it's useful to explain/write about internal code. I can reword this.

EdwinGuzman:
Updated all comments except for flagging or removing the layout="row" and width="full" default prop values example. We should either have different defaults or state in the story example that it is not recommended.

src/components/MultiSelectGroup/MultiSelectGroup.tsx
```diff
@@ -18,93 +17,77 @@ export interface MultiSelectGroupProps {
...
+        <Stack
+          className={className}
+          columnGap="xs"
+          data-testid={`multiselect-group-${id}`}
+          direction={finalLayout}
+          id={id}
+          ref={ref}
+          rowGap="xs"
+          spacing="xs"
+          wrap={layout === "row" ? "wrap" : null}
```
bigfishdesign13, line 77: This is what is causing the `MultiSelect` components to go full width for the row layout.  Update the condition to set the value as `"wrap"` when `layout === "row"` and `multiSelectWidth === "fitContent"`.

src/components/MultiSelectGroup/MultiSelectGroup.mdx
bigfishdesign13: Good points.  Let's fix the full width problem.  Please see the comment I just left about the `wrap` attribute for the `Stack` component.

src/components/MultiSelectGroup/MultiSelectGroup.mdx
bigfishdesign13: With the addition of my suggested test, the previous sentence can be removed.  They are both saying the same thing.

src/components/MultiSelectGroup/MultiSelectGroup.tsx
```diff
@@ -18,93 +22,77 @@ export interface MultiSelectGroupProps {
...
+        <Stack
+          className={className}
+          columnGap="xs"
+          data-testid={`multiselect-group-${id}`}
+          direction={finalLayout}
+          id={id}
+          ref={ref}
+          rowGap={finalLayout === "row" ? "xs" : "0"}
+          spacing="xs"
+          wrap={layout === "row" ? "wrap" : null}
```
bigfishdesign13: This is what is causing the MultiSelect components to go full width for the row layout. Update the condition to set the value as "wrap" when layout === "row" and multiSelectWidth === "fitContent".

bigfishdesign13 (changes requested):
I am aiming to fix the full width issue.  Refer to the coment about the `wrap` attribute on the `Stack` component.

EdwinGuzman:
@bigfishdesign13 ready for another review.

bigfishdesign13 (approved):
This looks good! :)

PR #1552, opened 2024-03-06
DSD-1710: Use onClearAll  function in multi-select component

gonuguntla:
Fixes JIRA ticket xxx

- Used onClearAll function in multiSelect component instead of onClear.
- Updated onClearAll function in multiSelectGroup and FilterBar.
- Fixed focus issue when click on the ViewAll button.
- Fixed scroll issue when isBlockElement flag is false.

## How has this been tested?
Local and storybook

-

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/MultiSelect/MultiSelect.tsx
+++ b/src/components/MultiSelect/MultiSelect.tsx
@@ -94,6 +94,8 @@ export const MultiSelect: ChakraComponent<
       // Create a ref to hold a reference to the accordian button, enabling us to programmatically focus it.
       const accordianButtonRef: React.RefObject<HTMLDivElement> =
         useRef<HTMLDivElement>();
+      const buttonRef: React.RefObject<HTMLButtonElement> =
+        useRef<HTMLButtonElement>();
 
       const MINIMUM_ITEMS_LIST_HEIGHT = "215px";
       const MAXIMUM_ITEMS_LIST_HEIGHT = "270px";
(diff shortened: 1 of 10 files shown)


src/theme/components/multiSelect.ts
```diff
@@ -110,6 +110,7 @@ const MultiSelect = defineMultiStyleConfig({
       },
       ".chakra-collapse": {
         bgColor: "ui.bg.page",
+        overflow: "unset !important",
```
gonuguntla, line 113: This code is related to [DSD-1713](https://jira.nypl.org/browse/DSD-1713)

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -222,6 +224,11 @@ export const MultiSelect: ChakraComponent<
       /** Toggle for listOverflow = "expand" */
       const toggleItemsList = () => {
         setIsExpandable((prevProp) => !prevProp);
+        setTimeout(() => {
```
gonuguntla, line 227: This code is related to DSD-1712. I tried without setTimeout, but focus of the button is not working after onClick.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -94,6 +94,8 @@ export const MultiSelect: ChakraComponent<
       // Create a ref to hold a reference to the accordian button, enabling us to programmatically focus it.
       const accordianButtonRef: React.RefObject<HTMLDivElement> =
         useRef<HTMLDivElement>();
+      const buttonRef: React.RefObject<HTMLButtonElement> =
```
EdwinGuzman: Rename this `expandToggleButtonRef` to be a bit more clear.

EdwinGuzman (approved):
The overflow issue seems like it'll work as well as the button focus behavior (just rename the variable).

The `onClearAll` seems like it will work. I thought you would need both `onClear` and `onClearAll` but that does not seem like that's the case. This _might_ change for `MultiSelectGroup` and/or `FilterBar` but we can update this later. I can't say I know why the `onClear` was  there to begin with if the `onChange` is doing the addition and removal in the `selectedItems` array.

EdwinGuzman (approved):

bigfishdesign13 (approved):

PR #1556, opened 2024-03-12
MultiSelect onClear function

EdwinGuzman:

- It turns out that `onClear` is very different than `onClearAll` and both are needed for separate situations. We've been using `onClearAll` incorrectly and this update adds back the `onClear` and gives an example of how it should be used.
- This makes it so that when multiple `MultiSelect` components are used in a single `useMultiSelect` hook, that clicking on the "clear" button won't clear the state for _all_ `MultiSelect` but instead just each individual one.

## How has this been tested?

Locally in Storybook

- N/A

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/FilterBar/FilterBar.test.tsx
+++ b/src/components/FilterBar/FilterBar.test.tsx
@@ -72,6 +72,7 @@ const FilterBarTestComponent = ({
     onChange,
     onMixedStateChange,
     selectedItems,
+    onClear,
     onClearAll,
     isModalOpen,
     onToggle,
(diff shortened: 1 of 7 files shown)


bigfishdesign13:
As it is now, everythign works as expected, but I think the `MultiSelect in a Group` example should include a `Clear all filters` button in order to show how the `onClearAll` function can be used.  That function is really intended for use in the `FilterBar` component, but the function is mentioned in the description above the example and then not used in the example.

Also, does the `onClearAll` prop still need to be part of `MultiSelect` and `MultiSelectItemsCountButton`?  In theory, that type of action would never be called from the component itself.

EdwinGuzman:
Good catch. I added an extra button for the `onClearAll` function and removed it from the `MultiSelectItemsCountButton` component.

bigfishdesign13 (approved):
Looks good!

PR #1557, opened 2024-03-13
Archive version of Reservoir v2.1.6

bigfishdesign13:


src/docs/Welcome.mdx
```diff
@@ -7,7 +7,75 @@ import Icon from "../components/Icons/Icon";
...
+        href="https://nypl-design-system-git-reservoir-v216-nypl.vercel.app"
+        fontSize="desktop.body.body2"
+        type="external"
+      >
+        v2.1.6 in Storybook
+      </Link>
+    </Text>,
+    <Text size="body2" whiteSpace="nowrap">
+      <Link
+        href="https://github.com/NYPL/nypl-design-system/tree/reservoir-v2.1.6"
```
EdwinGuzman, line 44: NBD but I wonder if pointing to the tag might be better than the branch. This is fine but maybe we can offer or point to both?

src/docs/Welcome.mdx
```diff
@@ -7,7 +7,75 @@ import Icon from "../components/Icons/Icon";
...
+  "Documentation",
+  "Code",
+];
+export const versionsData = [
+  [
+    <Text size="body2" whiteSpace="nowrap">
+      Reservoir 2.1.6
+    </Text>,
+    <Text size="body2" whiteSpace="nowrap">
+      September 2023
```
EdwinGuzman: 2.1.6 went out three weeks ago on Feb 22 (20 days ago from 3/13 based on npm).

EdwinGuzman (commented):
This looks good but a minor date comment. This actually shouldn't be merged, right? Because this is very similar to the update in the other PR, right? I think once approved this could be closed and the branch protected.

src/docs/Welcome.mdx
bigfishdesign13: Hmm, yes, that is a good question.  The tag would get a user to the downloadables, but is that more important than the branch in the repo itself?

src/docs/Welcome.mdx
bigfishdesign13: When I put this table together for the v3 branch, I used the dates that the major versions were first released (i.e. `1.0.0` and `2.0.0`).  So that begs the quetions, what is more important...the date of the initial major version release or the date of the final release within a major version?

src/docs/Welcome.mdx
bigfishdesign13: This would work...
image

EdwinGuzman: Hmm either way, if someone wants the 2.1.6 version, they would install it through npm. But I like the idea of pointing to the branch/tag in case someone wants to see the code.

src/docs/Welcome.mdx
EdwinGuzman: I would make it the date of that specific version. Should we poll it?

EdwinGuzman:
Something I forgot to point out, this seems wrong, right? That the visited link color is set for the svg icon but not the link.

Screen Shot 2024-03-13 at 8 35 48 PM

src/docs/Welcome.mdx
bigfishdesign13: No, I will use the exact dates for those versions.

EdwinGuzman (approved):

PR #1562, opened 2024-03-21
DSD-1623: label spacing for DatePicker, Select, Slider, and TextInput

bigfishdesign13:
Fixes JIRA ticket [DSD-1623](https://jira.nypl.org/browse/DSD-1623)

- Updates the `DatePicker`, `Select`, `Slider`, and `TextInput` components to reduce the spacing between the field label and the field itself from `"8px"` to `"2px"`.
- NOTE: The `FilterBar` and `MultiSelectGroup` components were not addressed because they have not yet been reintroduced back into the DS.

## How has this been tested?

- local Storybook

- n/a

### Checklist:

### Front End Review:


src/components/DatePicker/DatePicker.tsx
```diff
@@ -225,6 +226,12 @@ const DatePickerWrapper: React.FC<
           isRequired={isRequired}
           legendText={labelText}
           showRequiredLabel={showRequiredLabel}
+          __css={{
+            legend: {
+              // marginBottom: "xxxs",
+              ...labelLegendTextSpecialSpacing,
+            },
```
EdwinGuzman: This can be `legend: labelLegendTextSpecialSpacing,`

src/theme/components/global.ts
```diff
@@ -125,6 +125,10 @@ const labelLegendText = {
     color: "dark.ui.typography.heading",
   },
 };
+// Used in form inputs that require nuanced spacing.
+const labelLegendTextSpecialSpacing = {
+  marginBottom: "xxxs",
```
EdwinGuzman, line 130: Why not add this rule to `fieldset.ts` and `label.ts`? This seems like it'll be easier to miss if the update is in the component file than the style file.

bigfishdesign13: I was following the pattern of `labelLegendText` because I wanted a reusable object.

EdwinGuzman: Yes, but `labelLegendText` is used in `fieldset.ts` and `label.ts` and not in the `__css` prop for the components. So what's here is fine but if we can keep the update in the theme files, that'd be ideal.

EdwinGuzman (approved):
Approving but would like to see this in the theme file at some point if possible.

src/theme/components/datePicker.ts
```diff
@@ -1,15 +1,19 @@
...
  */
 
 const { defineMultiStyleConfig, definePartsStyle } =
-  createMultiStyleConfigHelpers(["subLabels"]);
+  createMultiStyleConfigHelpers(["fieldset", "subLabels"]);
 
 const DatePicker = defineMultiStyleConfig({
   baseStyle: definePartsStyle({
+    fieldset: {
+      legend: { ...labelLegendTextSpecialSpacing },
```
EdwinGuzman: This can be `legend: labelLegendTextSpecialSpacing,`

EdwinGuzman (commented):
Update looks good but just needs one minor update.

EdwinGuzman (approved):

PR #1578, opened 2024-04-12
DSD-1760: implementing NYPL placeholder  images

bigfishdesign13:
Fixes JIRA ticket [DSD-1760](https://jira.nypl.org/browse/DSD-1760)

- Updates placeholder images to come from `loremflickr.com`, rather than `placekitten.com`.

## How has this been tested?

- local Storybook

- n/a

### Checklist:

### Front End Review:


EdwinGuzman (changes requested):
The update looks good but the dimensions should be updated. In some cases, the image is very blurry and looks bad (check `Hero`, `Card`, and `Image`).

One thing I'm not a fan of is NYPL images that have people in the foreground. I don't think it's a risk but if we can get images of the Library itself without people, that will be better (in one case the image is a picture of a young child and the Library in the back).

src/utils/utils.ts
```diff
@@ -150,3 +150,25 @@ export const sanitizeStringForAttribute = (str: string) => {
   const sanitizedStr = str.replace(/[^a-z0-9]/gi, "-").toLowerCase();
   return sanitizedStr;
 };
+
+/** Provides a random image to use as a placeholder. */
+// export interface GetPlaceholderImageProps {
+//   size?: "small" | "medium" | "large";
+// }
+export const getPlaceholderImage = (size: string = "medium") => {
```
bigfishdesign13: @EdwinGuzman I added this function and it works well.  However, I first tried to set it up using an interface and ran into issues where typescript was saying that at least one arg was required.  To see this fucntion in action, take a look at the `Hero` component.

EdwinGuzman: I'm not sure how `GetPlaceholderImageProps` was assigned, but because this is a single argument value and not an object, you can't do something like `getPlaceholderImage = (size: GetPlaceholderImageProps = "medium") => {`.

What you can do is make this a type instead:
```
export type GetPlaceholderImageProps = "small" | "medium" | "large";
export const getPlaceholderImage = (
  size: GetPlaceholderImageProps = "medium"
) => {
```

src/utils/utils.ts
```diff
@@ -150,3 +150,25 @@ export const sanitizeStringForAttribute = (str: string) => {
...
+  const imageArray = [
+    "https://images.nypl.org/index.php?id=61705",
+    "https://images.nypl.org/index.php?id=61707",
+    "https://images.nypl.org/index.php?id=swope_243048",
+    "https://images.nypl.org/index.php?id=swope_243025",
+    "https://images.nypl.org/index.php?id=swope_244712",
+    "https://images.nypl.org/index.php?id=1252839",
+    "https://images.nypl.org/index.php?id=1252803",
+  ];
+  const t = size === "small" ? "r" : "w";
```
EdwinGuzman: The "large" value does change anything in this function.

src/utils/utils.ts
```diff
@@ -150,3 +150,25 @@ export const sanitizeStringForAttribute = (str: string) => {
...
+export const getPlaceholderImage = (size: string = "medium") => {
+  // const { size = "small" } = props;
+  const imageArray = [
+    "https://images.nypl.org/index.php?id=61705",
+    "https://images.nypl.org/index.php?id=61707",
+    "https://images.nypl.org/index.php?id=swope_243048",
+    "https://images.nypl.org/index.php?id=swope_243025",
+    "https://images.nypl.org/index.php?id=swope_244712",
+    "https://images.nypl.org/index.php?id=1252839",
+    "https://images.nypl.org/index.php?id=1252803",
```
EdwinGuzman: Can we include these as well?...
- https://images.nypl.org/index.php?id=836241&t=w
- https://images.nypl.org/index.php?id=836235&t=w
- https://images.nypl.org/index.php?id=1587878&t=w
- https://images.nypl.org/index.php?id=1588120&t=w
- https://images.nypl.org/index.php?id=58691098&t=w
- https://images.nypl.org/index.php?id=1203012&t=w

EdwinGuzman (commented):
Looks good. I added a suggestion for updating the utils function.

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -170,7 +170,11 @@ export const UsingIconContent: Story = {
 
 export const UsingImageContent: Story = {
   render: () => (
-    <Tooltip content={<Image src="//placekitten.com/300/300" alt="" />}>
+    <Tooltip
+      content={
+        <Image src="//loremflickr.com/300/300/new+york+public+library" alt="" />
```
oliviawongnyc: Nit, you switched these to alphabetical order elsewhere, if you want to do here too.

oliviawongnyc (commented):
Looks good to me! Agree with Edwin, the "large" argument should do something in the utility function, and I also noticed that the images are fuzzy, but I am not sure how you would change that if you're just pulling random images. One other small nit, we are inconsistent about whether we are passing alt text or not in these stories. Not really important, but thought I would mention jic.

bigfishdesign13:
@oliviawongnyc Yes, it looks like we pass a lot of empty strings for the alt text just to pass the component requirements.  That said, I did not look to address anything related to the alt text in this ticket.

EdwinGuzman (approved):

EdwinGuzman:
>  One other small nit, we are inconsistent about whether we are passing alt text or not in these stories. Not really important, but thought I would mention jic.

In some cases, Clare flags redundant alt tests and suggests to have an empty string value instead. So it does depend on the context but maybe it's acceptable to always have one in the DS as an example and give clearer information on alt text in the Storybook docs -- but a change for another ticket!

oliviawongnyc (approved):

PR #1579, opened 2024-04-12
DSD-1556: better layout for Icon examples

bigfishdesign13:
Fixes JIRA ticket [DSD-1556](https://jira.nypl.org/browse/DSD-1556)

- Updates the `Icon` component story page to improve the layout of the examples.

## How has this been tested?

- local Storybook

- n/a

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/Icons/Icon.stories.tsx
+++ b/src/components/Icons/Icon.stories.tsx
@@ -89,9 +89,10 @@ export const WithControls: Story = {
   },
 };
 
-const iconRow = (iconName, opts: any = {}) => {
-  // We'll use this setup function to render all the icons in a list item.
-  // Some icons display better with a dark background.
+const iconElement = (iconName, opts: any = {}, type: string = "grid") => {
+  // Use this setup function to render each icon packaged as a list item or
+  // wrapped in a basic div based on its intended usage. Note: some icons
+  // display better with a dark background.
   const styles: any = { display: "block" };
   const {
     size = "large",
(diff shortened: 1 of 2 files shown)


src/components/Icons/Icon.stories.tsx
```diff
@@ -129,9 +128,56 @@ const iconRow = (iconName, opts: any = {}) => {
           color={color}
         />
       </span>
+      <Text size="caption">{displayValue}</Text>
     </li>
   );
 };
+const iconBlock = (iconName, opts: any = {}) => {
+  // We'll use this setup function to render all the icons in a list item.
```
EdwinGuzman: This should say "in a grid" rather than in a list item.

src/components/Icons/Icon.stories.tsx
```diff
@@ -129,9 +128,56 @@ const iconRow = (iconName, opts: any = {}) => {
...
+        <Icon
+          name={iconName}
+          size={size}
+          iconRotation={iconRotation}
+          color={color}
+        />
+      </span>
+      <Text size="caption">{displayValue}</Text>
+    </Box>
+  );
```
EdwinGuzman: The only difference between this function and `iconRow` is the DOM that gets returned. Can you simplify this so that the variables and `key` value logic from both functions are in one place?

EdwinGuzman (changes requested):
- Add this to the repo changelog and component changelog.
- Reuse the logic from both functions (maybe create one logic function and the existing two functions call this new function and return the DOM).

EdwinGuzman (approved):

PR #1581, opened 2024-04-17
DSD-1767: Adding fallback and onerror feature to Image and Hero components

EdwinGuzman:
Fixes JIRA ticket [DSD-1767](https://jira.nypl.org/browse/DSD-1767)

- Adds `fallbackSrc` and `onError` props to the `Image` component.
- Updates the `imageProps` prop for the `Hero` component to include these new props.
- Example of an approach on how to use it.

## Pages to review
- `Image` "Fallback Image" story. The broken image icon will appear and then the fallback image should render: https://nypl-design-system-git-dsd-1767-hero-img-component-nypl.vercel.app/?path=%2Fstory%2Fcomponents-media-icons-image--fallback-image
- `Hero` "Fallback Campaign Image story. The broken image icon does _not_ appear since the visible images are background images. The background will be dark and then the fallback image should render https://nypl-design-system-git-dsd-1767-hero-img-component-nypl.vercel.app/?path=%2Fstory%2Fcomponents-basic-elements-hero--fallback-campaign-image
- Example code under "Fallback Campaign Image" section on https://nypl-design-system-git-dsd-1767-hero-img-component-nypl.vercel.app/?path=%2Fdocs%2Fcomponents-basic-elements-hero--docs

## How has this been tested?

Storybook

-

### Checklist:

### Front End Review:


src/components/Hero/Hero.mdx
```diff
@@ -198,6 +198,48 @@ subHeaderText={otherSubHeaderText}
...
+<Source
+  code={`
+const FallBackExample = () => {
+  // Or get image src from external service.
+  const [imageSrc, setImageSrc] = useState("foobar.jpg");
+  const fallbackImageSrc = "https://loremflickr.com/cache/resized/65535_51121136167_40f20af042_z_300_300_nofilter.jpg";
+  // For some reason IIIF image as fallback does not work...
+  // Perhaps the better approach is to store the image locally in the app.
+  // This cannot be easily tested in Storybook.
+  // const fallbackImageSrc = "https://iiif.nypl.org/iiif/2/105224/full/!900,900/0/default.jpg";
```
EdwinGuzman: Not sure why this fails but if you can try locally to verify, that'd be great. I don't think it has to do with the DS or how images are loading but maybe instead iiif? loremflicker and other images that are used as fallback images load.

7emansell: Seeing the same issue locally, and images not from IIIF do load

7emansell (approved):
Lgtm as in i understand the approach. Tested with a random image stored locally and that worked too, so definitely seems like a IIIF problem

src/components/Hero/Hero.tsx
```diff
@@ -258,7 +262,7 @@ export const Hero: ChakraComponent<
           ...(foregroundColor && { color: foregroundColor }),
           ...(backgroundColor
             ? { backgroundColor }
-            : { defaultBackgroundColor }),
+            : { bgColor: defaultBackgroundColor }),
```
EdwinGuzman, line 265: @bigfishdesign13 this will fix the issue we noticed during our chat.

EdwinGuzman:
This is now ready for a review.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 ### Adds
 
 - Adds docs for `"Visited State Patterns"` to the `Link` component story page.
+- Adds `fallbackSrc` and `onError` props to the `Image` component.
```
bigfishdesign13, line 15: Also include an entry for the `Hero` component updates.

src/components/Hero/Hero.mdx
```diff
@@ -198,6 +201,55 @@ subHeaderText={otherSubHeaderText}
 
 <Canvas of={HeroStories.CampaignBackgroundColors} />
 
+## Fallback Campaign Image
```
bigfishdesign13, line 204: Can this same approach be used for the `"primary"` variant?  What about the `FeaturedContent` component?

src/components/Hero/Hero.stories.tsx
```diff
@@ -371,6 +372,48 @@ export const CampaignBackgroundColors: Story = {
   ),
 };
 
+const FallBackExample = () => {
+  const [imageSrc, setImageSrc] = useState("foobar.jpg");
+  const fallbackImageSrc =
+    "https://loremflickr.com/cache/resized/65535_51121136167_40f20af042_z_300_300_nofilter.jpg";
```
bigfishdesign13: This image path may expire.  I had an image from loremflickr.com open in a browser tab for a day or two and eventually I was unable to reload the image, leading me to believe the path had expired.  It worries me that `cache` is part of the path.

src/components/Image/Image.tsx
```diff
@@ -168,6 +178,14 @@ export const Image: ChakraComponent<
       size,
       sizeBasedOn,
     });
+    // Function that gets called when an image fails to load.
+    const onImageError = (event: React.SyntheticEvent<HTMLImageElement>) => {
+      console.warn(
+        "NYPL Reservoir: `Image` failed to load initial image, using fallback image."
```
bigfishdesign13: Nit: use complete sentences in the warning.

src/components/Hero/Hero.stories.tsx
```diff
@@ -371,6 +372,48 @@ export const CampaignBackgroundColors: Story = {
...
+  // For some reason IIIF image as fallback does not work...
+  // const fallbackImageSrc = "https://iiif.nypl.org/iiif/2/105224/full/!900,900/0/default.jpg";
+
+  return (
+    <Stack spacing="l">
+      <Heading
+        id="campaign-hero-default"
+        size="heading6"
+        text="Campaign Hero with Custom Image component"
```
bigfishdesign13: It's not really using a `Custom Image component`.  Rather, it's a more complex method for handling the `imageSrc` prop.  The title should reflect that.

bigfishdesign13 (changes requested):
The functionality seems to work well, but docs need some changes.

src/components/Hero/Hero.mdx
EdwinGuzman: Yes, I was focusing on the needed variant first before doing the rest but I can work on updating the "primary" variant and have an example. How about `FeaturedContent` as it's own example? Right now the raised issue is with the `Hero` since that uses dynamic images but `FeaturedContent` does not (in DC Facelift).

src/components/Hero/Hero.stories.tsx
EdwinGuzman: Yes, right. My first approach was "custom image" but then the fallback solution worked better. Will update.

src/components/Hero/Hero.mdx
EdwinGuzman: Actually, I can't! No image is loaded for the `primary` variant so it's not possible to check for an image loading error. "primary" only uses the background-image CSS property. We'll need a different solution for this...

I'd say not for this PR.

src/components/Hero/Hero.mdx
```diff
@@ -198,6 +201,55 @@ subHeaderText={otherSubHeaderText}
...
+the `"campaign"` variant. The internal `Image` component will use the `onerror`
+event to load the `fallbackSrc` image.
+
+The `Hero` component does not use the `Image` component, however, for the
+background image. The background image is rendered through the CSS
+`background-image` property. This means that more work is needed to set a
+fallback for the background image.
+
+The following example demonstrates how to set a fallback image for both the main
+image and the background image. Note: this example assumes that both the main
```
avertrees, line 216: re: "background-image" not having a fallback image prop - would this only be a issue when the main and background image sources are different?

EdwinGuzman: Yes, that's right. Luckily (?), for DCF the same image is used for the main and background image.

This is a problem I will revisit later for this "campaign" variant and the "primary" variant.

src/components/Image/Image.mdx
```diff
@@ -109,6 +111,40 @@ the `onError` function callback.
...
+Set a fallback image source through the `fallbackSrc` prop. Internally, the
+`Image` component will call the `img` element's `onerror` attribute with a
+callback function to load the `fallbackSrc` image. Additionally, a warning will
+be logged to the console when the error occurs and the fallback image is loaded.
+
+If an _additional_ action needs to be performed when the fallback image is
+loaded, pass a callback function to the `onError` prop.
+
+**IMPORTANT**: If a custom image is passed through the `component` prop, then
+`fallbackSrc` and `onError` won't work. It is expected that the custom image
```
avertrees: this is a little confusing to me. `onError` won't work but it's expected?

EdwinGuzman: Hmm yes. I can add another example. The following is what I mean but does it make sense?

```
**IMPORTANT**: If a custom image, such as the Next.js `Image`, is passed through the `component`
prop, then the `fallbackSrc` and `onError` props won't work. It is expected that the custom image
that is passed in the implements the `onError` attribute.

<Source
  code={`
import { Image } from "@nypl/design-system-react-components";
import Image as NextImage from "next/image";

// ...
    <Image
      component={{
         <NextImage
            src="https://placekitten.com/540/420"
            onError={(event) => {
              // implement your own fallback event
              event.target.src =  "https://loremflickr.com/540/420/cat"
            }}
         />
      }}
    />
`}
  language="jsx"
/>
```

avertrees: This makes more sense and the example also helps. 

I'd update the last sentence to say: "It is expected that the custom image that is passed in implements the `onError` attribute."

src/components/Hero/Hero.mdx
bigfishdesign13: The solution that is suggested here for DC facelift is only possible because they use the same image in the foreground and background.  Technically, the solution that is eventually used for `primary` should also be used here.  But yes, I agree, that should be handled in a separate ticket.

bigfishdesign13 (approved):
Right on!

avertrees (approved):
Looks good to me!

7emansell (approved):
Lgtm

PR #1583, opened 2024-04-24
[WIP] Featured content container query testing

jackiequach:
Fixes JIRA ticket xxx

- Testing swapping media queries for container queries in FeaturedContent component.

-

### Checklist:

### Front End Review:

### DIFF
--- a/src/theme/components/featuredContent.ts
+++ b/src/theme/components/featuredContent.ts
@@ -2,6 +2,7 @@ import { createMultiStyleConfigHelpers } from "@chakra-ui/styled-system";
 import { StyleFunctionProps } from "@chakra-ui/system";
 import { wrapperStyles } from "./global";
 import { screenreaderOnly } from "./globalMixins";
+import breakpoints from "../foundations/breakpoints";
 
 // This function creates a set of function that helps us
 // create multipart component styles.
@@ -48,7 +49,7 @@ const FeaturedContent = defineMultiStyleConfig({
           img: screenreaderOnly(),
           containerType: "inline-size",
           "@container (min-width: 0px)": {
-            " > div": {
+            "[data-wrapper]": {
               flexDirection: imageAtEnd ? "column-reverse" : "column",
             },
             "[data-image]": {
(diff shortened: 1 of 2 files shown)


bigfishdesign13:
@oliviawongnyc and @jackiequach 
It's interesting that you had to include two container queries with unique width values in order to trigger the changes at our mobile breakpoint.  Why would text element and the image element be reading different width values?

EdwinGuzman:
Adding @7emansell as a draft reviewer since she knows `FeaturedContent` very well.

Emma, this is what I'm thinking about when reviewing this draft: at some point, the DS and maybe the entire department will move to container queries. Before that happens, does the update for this component make sense? Is it easy to read and know what's going on? How does it compare to media queries in terms of syntax and how we create components.

Just some questions to start, but do feel free to point out other things or ask other questions.

oliviawongnyc:
> @oliviawongnyc and @jackiequach It's interesting that you had to include two container queries with unique width values in order to trigger the changes at our mobile breakpoint. Why would text element and the image element be reading different width values?

@bigfishdesign13, Jackie and I are discussing this now. We are not sure why the container breakpoints are different if the media query breakpoint was the same for both, but we were measuring the container while changing the screen size and we chose the breakpoints based on when we saw the styles being applied. Jackie did notice that the difference between 568 and 600 px is 32, which is the value of our large spacing variable, but we tried to remove the `padding: "l"` from the text element and that didn't seem to change anything.

7emansell (commented):
Lgtm as someone new to container queries! i do find it less readable than media queries rn but i think that'll change w more practice visualizing/actually implementing it. Also confused by the 568px/600px breakpoint difference..

src/theme/components/featuredContent.ts
```diff
@@ -46,37 +46,57 @@ const FeaturedContent = defineMultiStyleConfig({
...
+          "@container (min-width: 0px)": {
+            " > div": {
+              flexDirection: imageAtEnd ? "column-reverse" : "column",
+            },
+            "[data-image]": {
+              height: "320px",
+              width: "100%",
+            },
+          },
+          "@container (min-width: 568px)": {
```
EdwinGuzman: Can you explain a bit why this one is needed and then the 600px one?

Also, I think you and Olivia already mentioned this, but you cannot interpolate a variable into the strings, right? It'd be nice to use the values from the breakpoint object and set this as 
```
[`@container (min-width: ${breakpoints.md})`]
```

src/theme/components/featuredContent.ts
```diff
@@ -46,37 +46,57 @@ const FeaturedContent = defineMultiStyleConfig({
             bgColor: "dark.ui.bg.default",
           },
           img: screenreaderOnly(),
+          containerType: "inline-size",
+          "@container (min-width: 0px)": {
+            " > div": {
```
EdwinGuzman: If possible, I'd like this to be similar to the `data-image` attributes and not rely so much on the DOM structure. Try to make the rules more general so that if the DOM gets updated, little needs to be updated on the CSS side.

EdwinGuzman:
This looks great! I think you and @oliviawongnyc should create a working feature branch where all work in progress and reviewed code can go into. This bigger feature branch would be in "under review" stage until the next step where  a TAD or other documentation is worked on.

src/theme/components/featuredContent.ts
jackiequach: There is 1rem padding around the components on Storybook pages, so when the image is full width the featured content container is 600px since there is no padding. I think we can use 600px as a standard.

It is possible to interpolate the variable, I will update this.

PR #1586, opened 2024-05-01
DSD-1764: Extending Heading prop interface

EdwinGuzman:
Fixes JIRA ticket [DSD-1764](https://jira.nypl.org/browse/DSD-1764)

- Extends the prop interface for the `Heading` component.
- This issue was found in the Research Catalog where devs needed to pass the `tabIndex` prop but it threw a ts error:
  - https://github.com/NYPL/research-catalog/blob/6920382982806ccae7c208745b739e94995e7d7b/pages/search/index.tsx#L168
  - https://github.com/NYPL/research-catalog/blob/6920382982806ccae7c208745b739e94995e7d7b/pages/search/index.tsx#L221

@charmingduchess  just FYI.

## How has this been tested?

Through unit tests and checking the linter.

- Will help in some scenarios where the Heading needs to be programmatically focused.

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/Heading/Heading.tsx
+++ b/src/components/Heading/Heading.tsx
@@ -4,6 +4,7 @@ import {
   ChakraComponent,
   Heading as ChakraHeading,
   useMultiStyleConfig,
+  HeadingProps as ChakraHeadingProps,
 } from "@chakra-ui/react";
 import React, { forwardRef } from "react";
 
@@ -41,7 +42,7 @@ export const headingLevelsArray = [
 export type HeadingSizes = typeof headingSizesArray[number];
 export type HeadingLevels = typeof headingLevelsArray[number];
 
-export interface HeadingProps {
+export interface HeadingProps extends ChakraHeadingProps {
   /** Optional className that appears in addition to `heading` */
   className?: string;
   /** Optional ID that other components can cross reference for accessibility
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 ## Prerelease
 
+### Updates
+
+- Updates the `Heading` component to allow more HTML attributes as props.
+
 ## 3.1.1 (April 25, 2024)
 
 ### Adds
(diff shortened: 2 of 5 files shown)


src/components/Heading/Heading.tsx
```diff
@@ -41,7 +42,7 @@ export const headingLevelsArray = [
 export type HeadingSizes = typeof headingSizesArray[number];
 export type HeadingLevels = typeof headingLevelsArray[number];
 
-export interface HeadingProps {
+export interface HeadingProps extends ChakraHeadingProps {
```
EdwinGuzman, line 45: We use this same pattern for the `Link` component and this should be the pattern for other components if they need it. I'm not sure if we should just go ahead and do this for all components now or wait until it's needed but it's not a big blocker in consuming apps.

src/components/Heading/Heading.tsx
```diff
@@ -4,6 +4,7 @@ import {
   ChakraComponent,
   Heading as ChakraHeading,
   useMultiStyleConfig,
+  HeadingProps as ChakraHeadingProps,
```
bigfishdesign13, line 7: Are all default Chakra component props available in this way?

src/components/Heading/Heading.tsx
bigfishdesign13: If we apply this method to other components, will this open up potential problems because it will expose props that we may not want consuming apps to use?  And if that is the case, should we get over ourselves and not worry too much about it?

bigfishdesign13 (approved):
Such a simple change. :)

src/components/Heading/Heading.tsx
EdwinGuzman: I _think_ so. I can't see why not.

src/components/Heading/Heading.tsx
EdwinGuzman: Ha, yes sort of. But it's already restrictive because it's for the Chakra `Heading` component. So, for example, `alt` is not applicable here so there's some safety in there.

PR #1587, opened 2024-05-01
DSD-1762: Modal variant

7emansell:
Fixes JIRA ticket [DSD-1762](https://jira.nypl.org/browse/DSD-1762).

- Adds the "confirmation" `Modal` variant, with 2 buttons in the footer
- Adds tests and documentation for variant

## How has this been tested?

Locally

### Checklist:

### Front End Review:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@nypl/design-system-react-components",
-  "version": "3.1.1-modal-rc",
+  "version": "3.1.1-modal-rc-2",
   "description": "NYPL Reservoir Design System React Components",
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@nypl/design-system-react-components",
-  "version": "3.1.0",
+  "version": "3.1.1-modal-rc",
   "description": "NYPL Reservoir Design System React Components",
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@nypl/design-system-react-components",
-  "version": "3.1.1-modal-rc-2",
+  "version": "3.1.1-modal-rc-3",
   "description": "NYPL Reservoir Design System React Components",
   "repository": {
     "type": "git",
--- a/src/components/Modal/Modal.mdx
+++ b/src/components/Modal/Modal.mdx
@@ -25,6 +25,7 @@ import Link from "../Link/Link";
 - {<Link href="#modaltrigger-component-props" target="_self">ModalTrigger Component Props</Link>}
 - {<Link href="#usemodal" target="_self">useModal</Link>}
 - {<Link href="#usemodal-component-props" target="_self">useModal Component Props</Link>}
+- {<Link href="#confirm-cancel-variant" target="_self">Confirm/Cancel Variant</Link>}
 - {<Link href="#default-heading-text" target="_self">Default Heading Text</Link>}
 - {<Link href="#custom-heading-text" target="_self">Custom Heading Text</Link>}
 - {<Link href="#content-window-scrolling" target="_self">Content Window Scrolling</Link>}
@@ -222,6 +223,57 @@ export const ModalStory = (args) => {
 <Canvas of={ModalStories.useModalStory} />
 
 <ArgTypes of={ModalStories.useModalStory} />
+
+## Confirm Cancel Variant
+
+The "confirmCancel" `Modal` variant is used when the modal can confirm or revoke the modal's content. Accordingly, there are two buttons
+in the footer instead of the one in the default modal. Each button has its own custom function (`onCancel`, `onConfirm`) that will close the modal.
+The X button in the upper right will call the `onCancel` function.
+
+Notes:
+
+- The `onConfirm`, `onCancel`, `confirmButtonLabel`, and `type` props are required for this variant
+- The `onClose` prop should not be passed when using the `ModalTrigger`– `onConfirm` and `onCancel` will call the `onClose` from the `useModal` hook internally
+- If using `useModal` hook, not the `ModalTrigger`, pass the `onClose` function from the hook to `onCancel` and `onConfirm`
+
+<Source
+  code={`
+// Example for the confirmCancel variant.
+modalProps={{
+  type: "confirmCancel",
+  bodyContent: "The action is happening",
+  cancelButtonLabel: "Cancel action",
+  confirmButtonLabel: "Confirm action",
+  onCancel: () => {
+    console.log("Custom cancel");
+  },
+  onConfirm: () => {
+    console.log("Custom confirm");
+  },
+  headingText: (
+    <Heading
+      level="h3"
+      display="flex"
+      alignItems="center"
+      gap="xs"
+      size="heading5"
+    >
+      <>
+        <Icon
+          color="ui.success.primary"
+          name="actionCheckCircleFilled"
+          size="large"
+        />
+        This is an action
+      </>
+    </Heading>
+  ),
+}}
+`}
+  language="jsx"
+/>
+
+<Canvas of={ModalStories.ConfirmCancelVariant} />
 
 ## Default Heading Text
 
--- a/src/components/Modal/Modal.mdx
+++ b/src/components/Modal/Modal.mdx
@@ -125,6 +125,7 @@ import { ModalTrigger } from "@nypl/design-system-react-components";
   buttonText="Button Text"
   id="modal-trigger"
   modalProps={{
+    type: "default",
     bodyContent: "body text",
     closeButtonLabel: "Close Button",
     headingText: (
@@ -178,6 +179,7 @@ import { useModal } from "@nypl/design-system-react-components";
 export const ModalStory = (args) => {
   const { onClose, onOpen, Modal } = useModal();
   const modalProps = {
+    type: "default",
     bodyContent: (
       <>
         <Button id="custom-close" onClick={onClose}>
@@ -303,6 +305,7 @@ additional icon.
   code={`
 // Example for the ModalTrigger component.
 modalProps={{
+  type: "default",
   bodyContent: "body text",
   closeButtonLabel: "Close Button",
   headingText: (
--- a/src/components/Modal/Modal.mdx
+++ b/src/components/Modal/Modal.mdx
@@ -233,8 +233,9 @@ The X button in the upper right will call the `onCancel` function.
 Notes:
 
 - The `onConfirm`, `onCancel`, `confirmButtonLabel`, and `type` props are required for this variant
-- The `onClose` prop should not be passed when using the `ModalTrigger`– `onConfirm` and `onCancel` will call the `onClose` from the `useModal` hook internally
-- If using `useModal` hook, not the `ModalTrigger`, pass the `onClose` function from the hook to `onCancel` and `onConfirm`
+- The `onClose` prop should never be passed with `onConfirm` and `onCancel`
+- When using the `ModalTrigger`, `onConfirm` and `onCancel` will call the `onClose` from the `useModal` hook internally
+- When using the `useModal` hook, as with the `onClose` prop, pass the `onClose` function from the hook to `onCancel` and `onConfirm`
 
 <Source
   code={`
--- a/src/components/Modal/Modal.mdx
+++ b/src/components/Modal/Modal.mdx
@@ -224,7 +224,7 @@ export const ModalStory = (args) => {
 
 <ArgTypes of={ModalStories.useModalStory} />
 
-## Confirmation Modal Variant
+## Confirmation Variant
 
 The "confirmation" `Modal` variant is used when the modal can confirm or revoke the modal's content. Accordingly, there are two buttons
 in the footer instead of the one in the default modal. Each button has its own custom function (`onCancel`, `onConfirm`) that will close the modal.
--- a/src/components/Modal/Modal.mdx
+++ b/src/components/Modal/Modal.mdx
@@ -25,7 +25,7 @@ import Link from "../Link/Link";
 - {<Link href="#modaltrigger-component-props" target="_self">ModalTrigger Component Props</Link>}
 - {<Link href="#usemodal" target="_self">useModal</Link>}
 - {<Link href="#usemodal-component-props" target="_self">useModal Component Props</Link>}
-- {<Link href="#confirm-cancel-variant" target="_self">Confirm/Cancel Variant</Link>}
+- {<Link href="#confirmation-variant" target="_self">Confirmation Modal Variant</Link>}
 - {<Link href="#default-heading-text" target="_self">Default Heading Text</Link>}
 - {<Link href="#custom-heading-text" target="_self">Custom Heading Text</Link>}
 - {<Link href="#content-window-scrolling" target="_self">Content Window Scrolling</Link>}
@@ -224,9 +224,9 @@ export const ModalStory = (args) => {
 
 <ArgTypes of={ModalStories.useModalStory} />
 
-## Confirm Cancel Variant
+## Confirmation Modal Variant
 
-The "confirmCancel" `Modal` variant is used when the modal can confirm or revoke the modal's content. Accordingly, there are two buttons
+The "confirmation" `Modal` variant is used when the modal can confirm or revoke the modal's content. Accordingly, there are two buttons
 in the footer instead of the one in the default modal. Each button has its own custom function (`onCancel`, `onConfirm`) that will close the modal.
 The X button in the upper right will call the `onCancel` function.
 
@@ -238,9 +238,9 @@ Notes:
 
 <Source
   code={`
-// Example for the confirmCancel variant.
+// Example for the Confirmation variant.
 modalProps={{
-  type: "confirmCancel",
+  type: "confirmation",
   bodyContent: "The action is happening",
   cancelButtonLabel: "Cancel action",
   confirmButtonLabel: "Confirm action",
@@ -273,7 +273,7 @@ modalProps={{
   language="jsx"
 />
 
-<Canvas of={ModalStories.ConfirmCancelVariant} />
+<Canvas of={ModalStories.ConfirmationVariant} />
 
 ## Default Heading Text
 
--- a/src/components/Modal/Modal.mdx
+++ b/src/components/Modal/Modal.mdx
@@ -223,6 +223,7 @@ export const ModalStory = (args) => {
 ## useModal Component Props
 
 <Canvas of={ModalStories.useModalStory} />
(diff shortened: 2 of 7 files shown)


7emansell:
@EdwinGuzman @bigfishdesign13 Including the "type" prop is the biggest departure from what we discussed in the ticket, but I think it really clarifies the code/ makes it more readable. It's also consistent with the pattern set by `Tagset`. Let me know what you think– my biggest hesitation is that it escalates how we handle passing the wrong combination of props to a breaking error, rather than a warning and an adjustment to just take the one correct onClose prop.

src/components/Modal/Modal.mdx
```diff
@@ -223,6 +224,58 @@ export const ModalStory = (args) => {
 
 <ArgTypes of={ModalStories.useModalStory} />
 
+## Confirmation Variant
+
+The "confirmation" `Modal` variant is used when the modal can confirm or revoke the modal's content. Accordingly, there are two buttons
+in the footer instead of the one in the default modal. Each button has its own custom function (`onCancel`, `onConfirm`) that will close the modal.
+The X button in the upper right will call the `onCancel` function.
```
EdwinGuzman: Say, "The 'X' Close button ..."

src/components/Modal/Modal.mdx
```diff
@@ -223,6 +224,58 @@ export const ModalStory = (args) => {
...
+
+- The `onConfirm`, `onCancel`, `confirmButtonLabel`, and `type` props are required for this variant
+- The `onClose` prop should never be passed with `onConfirm` and `onCancel`
+- When using the `ModalTrigger`, `onConfirm` and `onCancel` will call the `onClose` from the `useModal` hook internally
+- When using the `useModal` hook, as with the `onClose` prop, pass the `onClose` function from the hook to `onCancel` and `onConfirm`
+
+<Source
+  code={`
+// Example for the Confirmation variant.
+modalProps={{
```
EdwinGuzman: Include the rest of the component in the example

```
<ModalTrigger
      buttonText="Confirm"
      id="modal-confirmation"
      ....
```

src/components/Modal/Modal.mdx
```diff
@@ -223,6 +224,58 @@ export const ModalStory = (args) => {
...
+The "confirmation" `Modal` variant is used when the modal can confirm or revoke the modal's content. Accordingly, there are two buttons
+in the footer instead of the one in the default modal. Each button has its own custom function (`onCancel`, `onConfirm`) that will close the modal.
+The X button in the upper right will call the `onCancel` function.
+
+Notes:
+
+- The `onConfirm`, `onCancel`, `confirmButtonLabel`, and `type` props are required for this variant
+- The `onClose` prop should never be passed with `onConfirm` and `onCancel`
+- When using the `ModalTrigger`, `onConfirm` and `onCancel` will call the `onClose` from the `useModal` hook internally
+- When using the `useModal` hook, as with the `onClose` prop, pass the `onClose` function from the hook to `onCancel` and `onConfirm`
```
EdwinGuzman: Can you add a code snippet example of this?.

src/components/Modal/Modal.mdx
```diff
@@ -223,6 +224,58 @@ export const ModalStory = (args) => {
...
 
+## Confirmation Variant
+
+The "confirmation" `Modal` variant is used when the modal can confirm or revoke the modal's content. Accordingly, there are two buttons
+in the footer instead of the one in the default modal. Each button has its own custom function (`onCancel`, `onConfirm`) that will close the modal.
+The X button in the upper right will call the `onCancel` function.
+
+Notes:
+
+- The `onConfirm`, `onCancel`, `confirmButtonLabel`, and `type` props are required for this variant
```
EdwinGuzman: Nitpick: add period at the end of each note.

src/components/Modal/Modal.tsx
```diff
@@ -70,32 +113,75 @@ export const BaseModal: ChakraComponent<
...
-              <Button id="modal-close-btn" onClick={onClose}>
-                {closeButtonLabel}
-              </Button>
-            </ButtonGroup>
-          </ModalFooter>
-        </ModalContent>
-      </ChakraModal>
+    return (
+      <>
+        {isDefaultType(type) && (
```
EdwinGuzman: It might make this easier to read if the component blocks were variables. But I think because of the type guard that it's not really an option.

EdwinGuzman (commented):
I'll do another round of review for the `Modal` component file soon but left some feedback and the biggest issue which we talked about was the `onClose` function not working for the default variant.

I think having the `type` prop is fine because it does clarify the two variants. 
> my biggest hesitation is that it escalates how we handle passing the wrong combination of props to a breaking error
Where is this happening?

src/components/Modal/Modal.tsx
```diff
@@ -70,32 +113,75 @@ export const BaseModal: ChakraComponent<
...
-        scrollBehavior="inside"
-        size={size}
-        {...rest}
-      >
-        <ModalOverlay />
-        <ModalContent>
-          <ModalHeader>{finalTitle}</ModalHeader>
-          <ModalCloseButton />
-          <ModalBody>{bodyContent}</ModalBody>
+    if (
```
7emansell: @EdwinGuzman i'm not really describing it well but I mean you'll never hit this warning, you'll just have a type error and it'll display as a "default" modal, which is not exactly how we said errors should be handled. So maybe i need to change how this part works

7emansell:
@bigfishdesign13 after some discussion- Edwin and I are not sure there's a way around making the `type` prop required, even for the default modal, which is a breaking change.

src/components/Modal/modalChangelogData.ts
```diff
@@ -9,6 +9,13 @@
 import { ChangelogData } from "../../utils/ComponentChangelogTable";
 
 export const changelogData: ChangelogData[] = [
+  {
+    date: "Prerelease",
+    version: "Prerelease",
+    type: "Update",
+    affects: ["Functionality", "Styles"],
+    notes: ["Adds `Confirmation` variant to the `Modal`."],
```
bigfishdesign13: Remove `to the Modal` from the description.

src/components/Modal/Modal.mdx
```diff
@@ -223,6 +226,63 @@ export const ModalStory = (args) => {
 
 <ArgTypes of={ModalStories.useModalStory} />
 
+## Confirmation Variant
+
+The "confirmation" `Modal` variant is used when the modal can confirm or revoke the modal's content. Accordingly, there are two buttons
+in the footer instead of the one in the default modal. Each button has its own custom function (`onCancel`, `onConfirm`) that will close the modal.
+The 'X' Close button in the upper right will call the `onCancel` function.
```
bigfishdesign13: As well as the ESC button and clicking anywhere outside the modal.  Maybe change this to the following:
```
Additionally, the `onCancel` function will be called when using 
any of the other available methods for closing the modal, 
including clicking the "x" close button in the upper right corner 
of the modal, clicking anywhere outside the modal UI, or using 
the keyboard ESC button.
```

src/components/Modal/Modal.tsx
```diff
@@ -11,54 +11,97 @@ import {
...
+// Type guard for the `Modal` variant.
+export function isDefaultType(type: BaseModalProps["type"]): type is "default" {
+  return type === "default";
+}
+
 export const BaseModal: ChakraComponent<
   React.ForwardRefExoticComponent<
-    React.PropsWithChildren<BaseModalProps> &
-      React.RefAttributes<HTMLButtonElement>
+    BaseModalProps & React.RefAttributes<HTMLButtonElement>
```
EdwinGuzman: This is not necessarily related to your changes, but this component is using the `forwardRef` function but the `ref` prop is not defined. So in this case, the second part of `BaseModalProps & React.RefAttributes...` is not doing anything. You could add a ref on line 91, `(props: BaseModalProps, ref: ...)`, but I'm not sure if you can pass it to the `ChakraModal` or if that is even the correct element to pass this ref to. Perhaps this is the PR where the `forwardRef` can just be removed.

src/components/Modal/Modal.mdx
```diff
@@ -223,6 +226,63 @@ export const ModalStory = (args) => {
...
+                color="ui.success.primary"
+                name="actionCheckCircleFilled"
+                size="large"
+              />
+              This is an action
+            </>
+          </Heading>
+        ),
+      }}
+    />
```
bigfishdesign13: This code block is not aligned proply.  Most of it needs to be slid to the left.

src/components/Modal/Modal.mdx
```diff
@@ -223,6 +226,63 @@ export const ModalStory = (args) => {
...
+`}
+  language="jsx"
+/>
+
+<Canvas of={ModalStories.ConfirmationVariant} />
```
bigfishdesign13, line 291: I see there are console messages being set for the button actions, but can the story be updated to include real-time updates onscreen?  If not, let's at least update the docs above to mention that users should take a look at the console as they are clicking the buttons.

src/components/Modal/Modal.tsx
EdwinGuzman: I see, yes, typescript will catch the issue in your editor or when the app is built. This is helpful, however, in applications that don't use typescript. One example is MLN (MyLibraryNyc) where these warnings do get logged.

src/components/Modal/Modal.tsx
EdwinGuzman: I think it's worth trying to make these into variables.

src/components/Modal/Modal.mdx
```diff
@@ -25,6 +25,7 @@ import Link from "../Link/Link";
 - {<Link href="#modaltrigger-component-props" target="_self">ModalTrigger Component Props</Link>}
 - {<Link href="#usemodal" target="_self">useModal</Link>}
 - {<Link href="#usemodal-component-props" target="_self">useModal Component Props</Link>}
+- {<Link href="#confirmation-variant" target="_self">Confirmation Modal Variant</Link>}
```
bigfishdesign13, line 28: I feel this functionality is significant and should be moved further up in the TOC.  What do you think?  What was your reasoning for putting it in this slot?

src/components/Modal/Modal.tsx
```diff
@@ -70,32 +113,75 @@ export const BaseModal: ChakraComponent<
...
+                <ButtonGroup>
+                  <Button id="modal-close-btn" onClick={onClose}>
+                    {closeButtonLabel}
+                  </Button>
+                </ButtonGroup>
+              </ModalFooter>
+            </ModalContent>
+          </ChakraModal>
+        )}
+        {!isDefaultType(type) && (
```
EdwinGuzman: I know this is using the `TagSet` pattern but I wonder why we didn't use a tertiary there. Do you think this could be easier to read if it was used here?

bigfishdesign13 (commented):
Overall, this looks good.  I just want to know more about the `type` prop as a required prop.  I will do a final review once we talk about that more.

That being said, I also think the `type` prop is fine to use.  That is definitely a pattern we use in other components.

src/components/Modal/Modal.tsx
```diff
@@ -149,19 +258,42 @@ export function useModal(): any {
...
-        />
+        <>
+          {isDefaultType(type) ? (
+            <BaseModal
+              bodyContent={bodyContent}
+              headingText={headingText}
+              id={id}
+              type={type}
+              isOpen={isOpen}
+              onClose={onClose}
```
EdwinGuzman, line 273: The more I look at this, the more it's confusing and not liking some of the decisions I made before. This is using the local `onClose` from the component, but there is also an `onClose` in the outer scope and this can be confusing.

In the storybook documentation, there's a `useModal` story that is using the "default" type. Can you add another storybook story example for `useModal` that uses the "confirmation" type?

EdwinGuzman (commented):
Minor changes but mostly looking for more updates in the storybook documentation.

src/components/Modal/Modal.tsx
7emansell: I would dearly love to change the `onClose` in the outer scope– if we're already full steam ahead w `type` do you think i could? The only reason i didn't was that i didn't want to create breaking changes. 
useModal example added but let me know if that's where it should be placed in the doc.

src/components/Modal/Modal.tsx
7emansell: yes

src/components/Modal/Modal.mdx
7emansell: Agreed but i wanted someone reading through to understand the difference between `ModalTrigger`/`useModal` before getting to the variant . Not super attached to this slot though

bigfishdesign13: Ok, understood.  Leave it as is.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@nypl/design-system-react-components",
-  "version": "3.1.0",
+  "version": "3.1.1-modal-rc",
```
EdwinGuzman: Did you make a new `rc-2` release to test the latest changes in RC?

src/components/Modal/Modal.tsx
EdwinGuzman: As we talked in our chat with Marty, do not rename the `onClose` function that consuming apps use. But if you want, you can update the internal `onClose` on line 250 as `onClose: [someOtherName]`. You'd just need to also export it on line 296 as `onClose: [someOtherName]`.

EdwinGuzman (approved):
The latest update looks good! I know you tested the initial release candidate in RC but you can also make another `rc` to test the latest update.

package.json
7emansell: I haven't yet but i will!

EdwinGuzman:
There are some flickering issues when the Modal opens but it's not from these changes. The button group does flicker but that's a known issue (mobile vs desktop rendered layout issue @bigfishdesign13 ) that we'll fix separately.

src/components/Modal/Modal.mdx
```diff
@@ -220,9 +223,73 @@ export const ModalStory = (args) => {
...
+Additionally, the `onCancel` function will be called when using
+any of the other available methods for closing the modal,
+including clicking the "X" close button in the upper right corner
+of the modal, clicking anywhere outside the modal UI, or using
+the `ESC` button.
+
+Notes:
+
+- The `onConfirm`, `onCancel`, `confirmButtonLabel`, and `type` props are required for this variant.
+- The `onClose` prop should never be passed with `onConfirm` and `onCancel`.
```
charmingduchess, line 243: Is this because `onCancel` overrides the `onClose`?

src/components/Modal/Modal.mdx
```diff
@@ -220,9 +223,73 @@ export const ModalStory = (args) => {
...
+- When using the `ModalTrigger`, `onConfirm` and `onCancel` will call the `onClose` from the `useModal` hook internally.
```
charmingduchess, line 244: Do you mean here that the developer will have to invoke the `onClose` method within the callbacks? This phrasing makes it sound like it happens automatically.

src/components/Modal/Modal.tsx
```diff
@@ -11,92 +11,173 @@ import {
...
-              </Button>
-            </ButtonGroup>
-          </ModalFooter>
-        </ModalContent>
-      </ChakraModal>
+  if (
+    (!isDefaultType(type) && onClose) ||
+    (isDefaultType(type) && (onCancel || onConfirm))
+  ) {
+    console.warn(
```
charmingduchess, line 117: This message is not totally accurate to what the conditions are. The conditionals are checking for type and presence of the callbacks, not an invalid combination of callbacks.

src/components/Modal/Modal.mdx
charmingduchess, line 243: There's a console.warn in Modal.tsx that explains that onClose will overrride the other callbacks if they are all provided. Not entirely sure, but maybe this should be called out here too?

src/components/Modal/Modal.tsx
charmingduchess: Any reason not to make the same change on line 195? I find that naming convention confusing.

charmingduchess (commented):
Looks good! I think the example in story book might be improved with some feedback on the screen indicating that the confirm or the cancel callback was invoked. I'm imagining like a string that would display after each, which would be updated by either the cancel or confirm callbacks to say "action cancelled" or "action confirmed".

src/components/Modal/Modal.mdx
7emansell: It does happen automatically with `the ModalTrigger`, but not with the `useModal`

src/components/Modal/Modal.mdx
7emansell: i think we need to change that console.warn, so doesn't need to be represented here

src/components/Modal/Modal.tsx
7emansell: agreed: @bigfishdesign13 is it okay to update the language of this warning given the new expected behavior with the type prop?

src/components/Modal/Modal.tsx
EdwinGuzman: For that case, the dev using the DS won't see the renamed function since it's internal. So it could be updated but won't be any less confusing.

src/components/Modal/Modal.tsx
bigfishdesign13: @7emansell Yes, absolutely.  Change the message as needed.

src/components/Modal/Modal.tsx
```diff
@@ -115,7 +115,7 @@ export const BaseModal: ChakraComponent<
     (isDefaultType(type) && (onCancel || onConfirm))
   ) {
     console.warn(
-      "NYPL Reservoir Modal: The `onClose`, `onConfirm`, and `onCancel` props have all been set, but they can not all be used together. Only `onClose` will be used, while `onConfirm` and `onCancel` will be ignored."
+      "NYPL Reservoir Modal: A combination of `onClose`, `onConfirm`, `onCancel`, and `confirmButtonLabel` props have all been set, but they can not all be used together. Either pass the `onClose` (with the default type) or `onCancel`, `onConfirm`, and `confirmButtonLabel` (with the confirmation type)."
```
bigfishdesign13: I think the warning message is still not quite right.  The condition above is ultimately an `or` statement, so to say that the `props have all been set` is not necessarily true.  I would remove any implication of that sentiment.  Maybe like this:
```
A combination of `onClose`, `onConfirm`, `onCancel`, and 
`confirmButtonLabel` props have been passed, but they can 
not be used as they are currently configured. Either pass the 
`onClose` prop (with the default type) or pass the `onCancel`, 
`onConfirm`, and `confirmButtonLabel` props (with the 
confirmation type).
```

bigfishdesign13 (approved):
👍

PR #1588, opened 2024-05-01
DSD-1772: Fixing Table issue where headings rendered twice

EdwinGuzman:
Fixes JIRA ticket [DSD-1772](https://jira.nypl.org/browse/DSD-1772)

- This issue was caught in the Research Catalog. Check the tables for each search result: http://qa-www.nypl.org/research/research-catalog/search?q=shakespeare
- This component used the `useWindowSize` hook which was used to conditionally render headings in the body for the mobile view. This cause a flash while the correct style was applied. The problem with this is that it depends on javascript to load. So, the mobile view rendered the mobile headings on desktop, once the component loaded it called the `useWindowSize` hook, then it conditionally removed the mobile headings on desktop. 
- This removes the `useWindowSize` hook and applies the styles through Chakra's responsive inline styles.

## How has this been tested?

In Storybook.

- N/A

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -1,4 +1,5 @@
 import {
+  Box,
   chakra,
   Table as ChakraTable,
   TableCaption as ChakraTableCaption,
@@ -11,7 +12,6 @@ import {
   ChakraComponent,
 } from "@chakra-ui/react";
 import React, { forwardRef } from "react";
-import useWindowSize from "../../hooks/useWindowSize";
 
 interface CustomColors {
   backgroundColor?: string;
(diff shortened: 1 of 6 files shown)


src/components/Table/Table.tsx
```diff
@@ -79,10 +79,6 @@ export const Table: ChakraComponent<
         useRowHeaders,
       });
 
-      // Based on --nypl-breakpoint-medium
-      const breakpointMedium = 600;
-      const windowDimensions = useWindowSize();
```
bigfishdesign13, line 84: I wonder why I did it this way.  I guess I was just not aware of the breakpoint hooks.

bigfishdesign13 (approved):
Looks good.  Everything flipped as expected.

PR #1589, opened 2024-05-02
DSD-1719/adds container queries to searchbar

oliviawongnyc:
Fixes JIRA ticket DSD-1719

- Another test PR, this time to see how container queries could be used in `SearchBar`, not meant to be merged

### Checklist:

### Front End Review:


src/theme/components/searchBar.ts
```diff
@@ -6,23 +6,24 @@ const { defineMultiStyleConfig, definePartsStyle } =
 const SearchBar = defineMultiStyleConfig({
   baseStyle: definePartsStyle({
     display: "flex",
-    marginBottom: {
-      base: "xs",
-      md: "auto",
+    "@container (min-width: 0px)": {
+      "&": { marginBottom: "xs", flexFlow: "column nowrap" },
+      " > select": { marginBottom: "-1px" },
```
EdwinGuzman: Similar to the other PR, if this can target the element directly without the need to rely on the DOM, that'd be great.

src/components/SearchBar/SearchBar.tsx
```diff
@@ -217,6 +217,7 @@ export const SearchBar: React.FC<any> = chakra(
         invalidText={invalidText}
         isInvalid={isInvalid}
         ref={ref}
+        sx={{ containerType: "inline-size" }}
```
EdwinGuzman: It seems to me that container queries are so new that Chakra does not support them as style props. Is that true?

This doesn't have to be part of this ticket but we should update all components so that devs can do `<Component wrapper containerType="inline-size" ...>`. It might mean having a global base interface with the container query attributes and _all_ component interfaces inherit from it. Maybe? This is a nice-to-have update.

PR #1593, opened 2024-05-09
Release 3.1.2

EdwinGuzman:
### Adds

- Adds the `id` prop to the `Hero`, `HorizontalRule`, `SkeletonLoader`, `TemplateAppContainer`, and `Text` components.
- Adds the `id` property to the `imageProps` prop object in the `Card` component.

### Updates

- Updates the `Heading` component to allow more HTML attributes as props.

### Fixes

- Fixes an issue with the `Table` component where it rendered the headings twice until the component loaded and applied the desktop styles. This was done by adding inline responsive styles and removing the use of the `useWindowSize` hook.


bigfishdesign13 (approved):
💯

PR #1597, opened 2024-05-13
DSD-1748: Adding interaction tests for RadioGroup

EdwinGuzman:
Fixes JIRA ticket [DSD-1748](https://jira.nypl.org/browse/DSD-1748)

- Adds interaction tests for the `RadioGroup` component. It checks for both clicking on radio inputs and using up/down and left/right keys presses to select new values.
- As with other interaction tests, this is viewed in the "Controls" tab in the addon toolbar.

## How has this been tested?

Storybook

- N/A

### Checklist:

### Front End Review:


src/components/RadioGroup/RadioGroup.stories.tsx
```diff
@@ -77,6 +78,19 @@ export const Controls: Story = {
...
+
+    expect(screen.getByRole("radiogroup")).toBeInTheDocument();
+
+    await userEvent.click(screen.getByLabelText("Radio 3"));
+    await userEvent.click(screen.getByLabelText("Radio 2"));
+    await userEvent.click(screen.getByLabelText("Radio 5"));
+    await userEvent.keyboard("{arrowdown}");
+    await userEvent.keyboard("{arrowleft}");
+    await userEvent.keyboard("{arrowup}");
+    await userEvent.keyboard("{arrowright}");
```
bigfishdesign13: This code executes events, but shouldn't there also be tests that verify the outcome of the events?

bigfishdesign13 (commented):
Questions.

src/components/RadioGroup/RadioGroup.stories.tsx
EdwinGuzman: I thought the same. I was a bit confused but the Storybook doc examples don't always have expectations https://storybook.js.org/docs/writing-tests/interaction-testing

The actions and interactions _are_ the tests. When I did expectations, such as `  expect(screen.getByRole("radiogroup")).toBeInTheDocument();` above, they don't show up in the addon panel. So we can add more expectations but what is shown to users are the interactions.

bigfishdesign13: What happens if you add an expectation and it fails?

EdwinGuzman: The error will be shown. It's not very intuitive but it's an implicit success, so you're not told anything until something throws an error.

bigfishdesign13: Do you think this will ever replace the unit tests?

EdwinGuzman: No, it should not. I think there's a way to "share" tests between unit tests and interaction tests but I haven't looked too deep into it.

bigfishdesign13: I ask because I feel an expectation should be added whenever an userEvent is executed.  We do that with the unit tests.  Should we also do that here as well.  We can add all the click events that we want, but if nothing changes after the event and we don't explicitly test to verify that it has changed, does the fact that we executed a click event mean anything?

EdwinGuzman: I think adding interactions and expectations is required for unit tests since it's testing the component in a fundamental and technical manner. In these tests, it's enough that the interactions work since that's what users do on sites and it's also how Storybook write the example tests as well.

There's also this idea of [implicit assertions](https://www.epicweb.dev/implicit-assertions) (it's on the site from the creator of React Testing Library). He posted this recently https://x.com/kentcdodds/status/1765501275003580800 and it's very interesting because it's not explicitly asserting anything, just that the page loads and that's enough.

That's something I'd like to also be considered here where if the keyboard and mouse interactions work, then we don't really need explicit expectations.

Regardless though, I'd like to keep how we think about unit tests and these interaction tests as separate types of tests.

CC @oliviawongnyc @7emansell @jackiequach

EdwinGuzman:
Ready for another review @7emansell @jackiequach @bigfishdesign13

src/components/RadioGroup/RadioGroup.stories.tsx
bigfishdesign13: @EdwinGuzman Looking back at the article about implicit assertions, none of the examples talk about testing user events.  They are all related to values and formats, but not to an event that may cause something to happen or change.  With the interective tests, we are triggering events, but can there always be an implicit assertion that something on screen or in the code has changed simply because an event was triggered?

EdwinGuzman: Right, I'd say yes and no and generally depends on the component we are testing and what we are testing for. If we want to test tabbing and moving arrows around, we visually see that new values get selected so it's working. It doesn't hurt to add expectations though.

I can add some expectations but because of how the addons display the tests, it won't appear in the list of interactions.

bigfishdesign13: > I can add some expectations but because of how the addons display the tests, it won't appear in the list of interactions.

I think that is ok.  They won't verify success, but they will throw errors if there is a problem.

src/components/RadioGroup/RadioGroup.stories.tsx
```diff
@@ -75,6 +76,25 @@ export const Controls: Story = {
...
+    await userEvent.click(screen.getByLabelText("Radio 2"));
+    await userEvent.click(screen.getByLabelText("Radio 5"));
+    expect(screen.getByLabelText("Radio 5")).toBeChecked();
+
+    await userEvent.keyboard("{arrowdown}");
+    await userEvent.keyboard("{arrowleft}");
+    await userEvent.keyboard("{arrowup}");
+    await userEvent.keyboard("{arrowright}");
+    await userEvent.keyboard("{arrowdown}");
+    expect(screen.getByLabelText("Radio 2")).toBeChecked();
```
EdwinGuzman, line 96: Odd, I _now_ see this in the UI. 
Screenshot 2024-06-05 at 3 04 28 PM

bigfishdesign13 (approved):
Wonderful.

PR #1598, opened 2024-05-13
DSD-1729: Adding showRowDividers prop to hide List description list  dividers

EdwinGuzman:
Fixes JIRA ticket [DSD-1729](https://jira.nypl.org/browse/DSD-1729)

- Adds the `showRowDividers` prop to be able to hide horizontal row dividers for the `"dl"` `List` variant.
- The underlying style functions that were updated are also used in `StructuredContent` but this update does not apply to the `StructuredContent` component. So there is no change and does not need to be included in the changelog.
- Try it out in the "Controls" tab.

Question: @bigfishdesign13 I removed the `Heading`'s top border as well. Let me know if it should not be removed as it's technically not a row divider, but it's still a border that is part of the `List` component.

## How has this been tested?

Storybook

- N/A

### Checklist:

### Front End Review:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 ## Prerelease
 
+### Adds
+
+- Adds `showRowDividers` prop to show/hide row dividers for the `List` `"dl"` variant.
+
 ## 3.1.2 (May 9, 2024)
 
 ### Adds
--- a/src/components/List/List.mdx
+++ b/src/components/List/List.mdx
@@ -9,10 +9,10 @@ import { changelogData } from "./listChangelogData";
 
 # List
 
-| Component Version | DS Version |
-| ----------------- | ---------- |
-| Added             | `0.7.0`    |
-| Latest            | `3.0.1`    |
+| Component Version | DS Version   |
+| ----------------- | ------------ |
+| Added             | `0.7.0`      |
+| Latest            | `Prerelease` |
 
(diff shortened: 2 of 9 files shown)


bigfishdesign13 (changes requested):
The functionality works great, but the console warning still needs to be added.  And to answer your question directly, yes, I think it is appropriate to also remove the border above the heading with this functionality.

EdwinGuzman:
Ah yes, I forgot to bring that up. Since the default is already set to `true`, it's hard to know when it is explicitly passed by the consumer. So we can do more logic checks to log the message but I'm not sure if it's worth it. I would know that devs read the documentation and figure out that it's only for the `dl` variant. I'll give it another shot but if it's too much, I'll say we cut it and make it a new ticket for a better pattern (since this might also be used in other components).

bigfishdesign13:
Yes, I figured it might take some odd logic to properly generate the warning.  Like you said, if it requires a lot of code, then skip it.

I've often thought that we chose the wrong pattern for toggling the visibility of elements.   More often than not, our visibility props default to `true` and the props need to be set explicitly to `false` to hide something.  Rather than a `showElement` pattern, I often feel we should have used a `hideElement` pattern with the prop being `false` by default.

EdwinGuzman:
The logic is not straightforward and I prefer to make that a new ticket if we really want it. In my opinion, the dev should use the doc to know that the prop is only applicable to the `dl` variant.

bigfishdesign13 (approved):

PR #1603, opened 2024-05-15
DSD-1741: Fixing checkbox's positioning so it is visible when focused

EdwinGuzman:
Fixes JIRA ticket [DSD-1741](https://jira.nypl.org/browse/DSD-1741)

- When tabbing through checkboxes in a scrollable container (so with a fixed height), a focused checkbox does not go into view. The suggested approach was to use the `useScrollTabIntoView` hook but when attempted, it did not make sense to introduce state to keep track of the focused element. Also, this meant we'd have to check track of user key presses for the tab press and check if a checkbox was focused. This seemed like overengineering a solution.
- After searching online, I saw this https://codepen.io/devadmin/pen/MZMjYx and natively without javascript, the expected behavior works. 
- This is not a `MultiSelect` or `Accordion` issue but rather an issue with how the `Checkbox` is positioned. Adding `position: "relative"` fixed the root of the issue.

## How has this been tested?

Review `MultiSelect`, `Accordion` (last few stories), and `CheckboxGroup` to verify that visually and functionally, everything is working as expected.

- This is an accessibility fix.

### Checklist:

### Front End Review:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 ## Prerelease
 
+### Fixes
+
+- Fixes the `Checkbox` positioning so it is visible when it is focused in a scrollable container.
+
 ## 3.1.3 (May 15, 2024)
 
 ### Adds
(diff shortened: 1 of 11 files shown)


bigfishdesign13 (approved):
Works well.  In many of the `MultiSelect` examples, the last checkbox sits halfway off the bottom of the dropdown element.  It does not shift into full view when it gains focus, but I think that is acceptable.

PR #1605, opened 2024-05-21
DSD-1692: Heading testing link style rules

EdwinGuzman:
Fixes JIRA ticket [DSD-1692](https://jira.nypl.org/browse/DSD-1692)

- Making links in the heading not have the underline unless if they are children links.
- Defining the recommended pattern for having heading text with only some words as links.
- Check how the `url` prop in the style function is not needed for this update. Instead, the `:only-child` pseudo class is used.
- Update: I'm now seeing the the following initial investigation work is not what the final PR renders. I'll check back on this.
- Original comment: The only DOM that doesn't work is this example
```
 <Heading id="heading-with-link-child" level="h2">
  <>
    Heading with a Word <a href="#hello">Link</a>
  </>
</Heading>
```

This renders as 
```
<h2>
  "Heading with a Word"
  <a ....
</h2>
```

You cannot select text nodes in CSS or JS so this specific case won't work. What we can do is recommend apps to use spans for lonely text as the example below it in the story file.

```
<Heading id="heading-with-DS-link-child" level="h2">
 <>
    <span>Some text </span>
    <Link href="#hello">Link</Link>
    <span> more text</span>
  </>
</Heading>
```

(Aside: I'm finding the error we throw that allows only one child to be annoying)

## How has this been tested?

Storybook

- N/A

### Checklist:

### Front End Review:


bigfishdesign13:
@EdwinGuzman Based on our discussion about recommended use cases for the `Heading` component, I think this approach is acceptable.  You'll just need to add clear docs to show how to get the `one word link` styles.

src/components/Heading/Heading.mdx
```diff
@@ -248,8 +248,17 @@ composed heading lockup.
...
+will create a child `<a>` element within the heading element. Alternatively, you
+can pass a DS `Link` component, a custom `Link` component, or an HTML `anchor`
+element as the **only child** of the `Heading` component.
+
+**Important**: Linked headings should have the link color but not be underlined.
+
+If only a specific portion of the heading should be linked, then the `url` prop
+cannot be used. Instead, use a `Link` or HTML `anchor` element to wrap the text
+that should be linked. Note that the non-linked text should be passed as a text
+node and not wrapped in span or other elements.
```
bigfishdesign13, line 261: Add to this...
```
This treatment allows for individual words within the full heading text to be set as 
a link, however, this is not a recommended pattern.  While this treatment is possible 
with the `Heading` component, this pattern SHOULD NOT be used.
```

This is pulled directly from the docs in the VDL.

src/components/Heading/Heading.stories.tsx
```diff
@@ -413,38 +413,38 @@ export const Links: Story = {
...
-      <Heading id="heading-with-DS-link" level="h2">
-        <Link href="#hello">Link</Link>
+      <Heading id="heading-with-DS-link" level="h3">
+        <Link href="#hello">Link using DS `Link` component</Link>
       </Heading>
-      <Heading id="heading-with-link-child" level="h2">
+      <Heading id="heading-with-link-child" level="h3">
         <>
-          Heading with a Word <a href="#hello">Link</a>
+          Loose text heading with a word <a href="#hello">Link</a>
```
bigfishdesign13: I think we need to tweak the messaging here.  This pattern is not recommedned, but I think it needs a stronger statement.  It will not pass a11y, so it should never be used.

```
Not acceptable. Do not use.
```

src/components/Heading/Heading.stories.tsx
```diff
@@ -413,38 +413,38 @@ export const Links: Story = {
...
         </>
       </Heading>
-      <Heading id="heading-with-DS-link-child" level="h2">
+      <Heading id="heading-with-DS-link-child" level="h3">
         <>
-          <span>Some text </span>
+          <span>Text wrapped in a span </span>
           <Link href="#hello">Link</Link>
-          <span> more text</span>
+          <span> Not a recommended pattern</span>
```
bigfishdesign13: Based on the previous comment, change this to...
```
Acceptible, but not recommended.
```

src/components/Heading/Heading.stories.tsx
```diff
@@ -413,38 +413,38 @@ export const Links: Story = {
...
         noSpace
         url="http://apple.com"
         urlClass="special-link"
       >
-        Heading with URL Props
+        Heading with `url` prop value
       </Heading>
-      <Heading id="heading-with-anchor" level="h2">
-        <a href="#hello">Link</a>
+      <Heading id="heading-with-anchor" level="h3">
```
bigfishdesign13: Add `noSpace`.

src/components/Heading/Heading.stories.tsx
```diff
@@ -413,38 +413,38 @@ export const Links: Story = {
...
+        <a href="#hello">Link from HTML anchor element</a>
       </Heading>
-      <Heading id="heading-with-DS-link" level="h2">
-        <Link href="#hello">Link</Link>
+      <Heading id="heading-with-DS-link" level="h3">
```
bigfishdesign13: Add `noSpace`.

src/components/Heading/Heading.stories.tsx
```diff
@@ -413,38 +413,38 @@ export const Links: Story = {
...
+        <Link href="#hello">Link using DS `Link` component</Link>
       </Heading>
-      <Heading id="heading-with-link-child" level="h2">
+      <Heading id="heading-with-link-child" level="h3">
```
bigfishdesign13: Add `noSpace`.

bigfishdesign13 (changes requested):
I want to be a little more clear about the acceptable patterns, so let's adjust the docs a little. Please refer to my comments.

bigfishdesign13 (approved):
Looks good!

PR #1606, opened 2024-05-22
Bug fix: zindex click issue on MultiSelect count button

EdwinGuzman:
This was found in QA reviewing ticket [DSD-1741](https://jira.nypl.org/browse/DSD-1741).

Updates the z-index when the `MultiSelect` count button is focused so it can be clicked/selected.


bigfishdesign13 (approved):
Looks good and works well.

PR #1607, opened 2024-05-22
DSD-1740/close dropdown multiselect

oliviawongnyc:
Fixes JIRA ticket [DSD-1740](https://jira.nypl.org/browse/DSD-1740)

- Adds the ability to close the Accordion if an element within the panel is focused (for example, if a checkbox within the MultiSelect is focused, this PR will close the parent panel when the user presses `esc`).

## How has this been tested?

- Locally, in Storybook.
- Also added another unit test.

- None.

### Checklist:

### Front End Review:


src/components/Accordion/accordionChangelogData.ts
```diff
@@ -14,7 +14,18 @@ export const changelogData: ChangelogData[] = [
     version: "Prerelease",
     type: "Update",
     affects: ["Accessibility", "Functionality"],
-    notes: ["Adds logic to close accordion when 'esc' key is pressed"],
+    notes: [
+      "Adds logic to close accordion when element within panel is focused and 'esc' key is pressed",
```
EdwinGuzman: This is duplicated below.

src/components/Accordion/Accordion.tsx
```diff
@@ -225,15 +225,25 @@ export const Accordion: ChakraComponent<
...
+          focusedPanelIndex = Number(e.target.dataset.index);
+        } else {
+          // If the user is focused on an element within the panel...
+          focusedPanelIndex = Number(
+            e.target.closest("[role='region']").id.split("-").pop()
+          );
+
+          setExpandedPanels(
+            expandedPanels.filter((i) => i !== focusedPanelIndex)
+          );
```
EdwinGuzman: This is within the `else` block and it doesn't seem intentional because the esc key should still be functional when the user is focused on the accordion button.

src/components/Accordion/Accordion.tsx
```diff
@@ -225,15 +225,25 @@ export const Accordion: ChakraComponent<
...
+      // panel is already collapsed.)
+      if (e.code === "Escape") {
+        let focusedPanelIndex;
+        if (e.target.dataset.index) {
+          // If the user is focused on an accordion button...
+          focusedPanelIndex = Number(e.target.dataset.index);
+        } else {
+          // If the user is focused on an element within the panel...
+          focusedPanelIndex = Number(
+            e.target.closest("[role='region']").id.split("-").pop()
```
EdwinGuzman, line 247: Nice, but also brittle if the DOM and/or id ever changes.

EdwinGuzman (changes requested):
I think your tests may break based on my comment so will review that once this is updated.

src/components/Accordion/accordionChangelogData.ts
oliviawongnyc: They're slightly different, but I can combine them if you prefer.

src/components/Accordion/Accordion.tsx
oliviawongnyc: Yes! Good catch. Thank you.

src/components/Accordion/Accordion.tsx
oliviawongnyc: I know :/, I didn't want to do it this way, but it seemed like the best option...the id is really the only way the panel is connected to the button that needs to close.

src/components/Accordion/accordionChangelogData.ts
EdwinGuzman: You can keep both but what I mean is you only need one object with "Prerelease" but now there are two. Unless if this was done because you don't think this will get deployed by 5/23. If that's the case then it's okay to keep both.

src/components/Accordion/Accordion.test.tsx
```diff
@@ -160,19 +160,17 @@ describe("Accordion", () => {
   });
 
   it("closes the accordion when the 'esc' key is pressed", async () => {
-    render(<Accordion accordionData={[accordionData[0]]} isDefaultOpen />);
+    render(<Accordion accordionData={[accordionData[0]]} />);
```
EdwinGuzman, line 168: Can you add another test where the entire `accordionData` array is passed, so now there are three accordions. For one of the accordions, add a button or checkbox, focus on it, and then press escape. This is to test the new feature in this PR but also that it works when there are more than just one accordion.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -4,7 +4,7 @@ import {
   ChakraComponent,
   useMultiStyleConfig,
 } from "@chakra-ui/react";
-import React, { useState, forwardRef, useRef } from "react";
+import React, { forwardRef, useRef, useState } from "react";
```
bigfishdesign13, line 7: I was not sure where to put this comment, but the focus needs to be addressed after the ESC key is pressed when a checkbox has focus.  In that scenario, the focus should return to the main button after the ESC key is pressed.

bigfishdesign13 (changes requested):
Please see my comment about how focus should be handled.

src/components/Accordion/Accordion.test.tsx
EdwinGuzman: I know there's a test that does this for `MultiSelect` but the feature is in the Accordion so it should also be tested there.

src/components/MultiSelect/MultiSelect.tsx
EdwinGuzman: That's a good point. If it involves having state or ref we should talk about this because it's getting overly complex and from what I understood, accordions don't generally behave this way. So if we don't have to implement this then perhaps we should not.

bigfishdesign13: We don't need it for `Accordion`, but we do need it for `MultiSelect`.

EdwinGuzman: That's true but the `Accordion` is a big part of the `MultiSelect` and, to me, `MultiSelect` is inheriting all the features of `Accordion`. There is currently a ref that is being passed from the `MultiSelect` to the `Accordion` to keep track of focus for when the items count button is clicked. This could be used as well when the esc key is pressed. But it we have to keep track of pressing the ESC key in the MultiSelect to close the Accordion, then it feels like the work is being duplicated because the `Accordion` already does check for when the ESC key is pressed.

Perhaps we should reconsider and move all the logic _just_ to the `MultiSelect` from the `Accordion`. @oliviawongnyc  As you try to implement the focus management in the `Accordion`, let me know if my comment makes sense.

src/components/Accordion/accordionChangelogData.ts
oliviawongnyc: I see! Okay. Will combine.

src/components/Accordion/Accordion.test.tsx
oliviawongnyc: Done!

EdwinGuzman (commented):
Update looks good but tests are failing

src/components/MultiSelect/MultiSelect.tsx
oliviawongnyc: Okay @EdwinGuzman and @bigfishdesign13, focus should now return to the button when the `Accordion` (or `MultiSelect`) is closed with the `esc` key.

bigfishdesign13 (commented):
The `esc` focus handling works great in `MultiSelect`, but it is not working in `Accordion`.

src/components/Accordion/Accordion.tsx
```diff
@@ -224,16 +224,40 @@ export const Accordion: ChakraComponent<
...
-      if (e.keyCode === 27) {
-        const focusedPanelIndex = Number(e.target.dataset.index);
-        // collapse the currently focused panel.
-        // (Nothing will happen if the currently
-        // focused panel is already collapsed.)
+      // If the 'esc' key is pressed, find the panel the
+      // user is focused on or within, and remove it as
+      // an expanded panel. (Nothing will happen if the
+      // panel is already collapsed.)
+      if (e.code === "Escape") {
```
EdwinGuzman, line 239: Just curious, why the update from `e.keyCode` to `e.code`?

EdwinGuzman (approved):
The update looks good to me and it works in both the `Accordion` and `MultiSelect`.

src/components/Accordion/Accordion.tsx
oliviawongnyc: Two reasons. 1) I thought it was clearer than. using a keyCode, and 2) for some reason in the tests, the keyCode for Escape was a different number than it is in the code? So I used the code 'Escape' instead for the tests and wanted to be consistent.

EdwinGuzman (approved):
Really nice refactor in the last commit.

bigfishdesign13 (approved):
Awesome!

PR #1599, opened 2024-05-14
DSD-1777: Fixing the desktop styling flicker in ButtonGroup

EdwinGuzman:
Fixes JIRA ticket [DSD-1777](https://jira.nypl.org/browse/DSD-1777)

## This PR does the following:

- Fixes the styling flicker for the `ButtonGroup` component. The mobile styles flickers right before the desktop styles were applied.

## How has this been tested?

Storybook and comparing against prod.

## Accessibility concerns or updates

- N/A

### Checklist:

- [ ] I have updated the Storybook documentation accordingly.
- [ ] I have added relevant accessibility documentation for this pull request.
- [ ] All new and existing tests passed.

### Front End Review:

- [ ] Review the Vercel preview deployment once it is ready.

PR #1608, opened 2024-05-23
DSD-1770: FilterBarInline Component

jackiequach:
Fixes JIRA ticket [DSD-1770](https://jira.nypl.org/browse/DSD-1770).

- Adds the FilterBarInline Component

## How has this been tested?

Locally on Storybook

-

### Checklist:

### Front End Review:

[DSD-1770]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-1770?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/FilterBarInline/FilterBarInline.tsx
```diff
@@ -0,0 +1,137 @@
...
+        layout = "row",
+        onClear,
+        onSubmit,
+        selectedItems,
+        renderChildren,
+        ...rest
+      } = props;
+
+      const { isLargerThanMobile } = useNYPLBreapoints();
+      const finalLayout = isLargerThanMobile ? layout : "column";
```
EdwinGuzman: > Noticed when the page loads, the mobile view is displayed before the desktop view due to the Chakra's useMediaQuery hook.

Regarding this comment, this approached worked for me https://github.com/NYPL/nypl-design-system/pull/1599/files#diff-1885fe882eadb679d28e96b476db7c366b4edabf310cafdffc31160cad0d67c4

I think you can get away without having `useNYPLBreakpoints` and doing more responsive styles if you have a theme object file. That's why youre' seeing the mobile view first before the desktop styles kick in.

src/components/FilterBarInline/FilterBarInline.stories.tsx
```diff
@@ -0,0 +1,1201 @@
...
+        onClear={onClearAll}
+        renderChildren={({ layout, width }) => {
+          return (
+            <MultiSelectGroup
+              showLabel={false}
+              id="row-clear-all"
+              labelText="Row with Clear All button"
+              layout={layout}
+              multiSelectWidth={width}
+              renderMultiSelect={({ isBlockElement, multiSelectWidth }) => {
```
EdwinGuzman, line 357: Double nested render prop pattern. Does this feel or seem difficult to keep track of?

EdwinGuzman (commented):
Even as a draft, this is looking great! Seems there's more documentation than component code so it feels complex as a UI wrapper, but glad to see multiple examples. 

One small issue I ran into was the focus seemed to jump in the example under the "Passing Children" section. I clicked on the "Accessibility" MultiSelect, then checkbox 1, and then the text input. The focus did not persist on the text input even though I clicked on it repeatedly. This went away when I typed but the focus ring was not visible. I can't replicate this anymore but wanted to point it out.

I also ran into the mobile rendering issue and removing the `useNYPLBreakpoints` hook should fix that, though you will need to move the logic as responsive styles for the component. I do see what you mean about needing the value for the children in the `renderChildren` function.

The other option I'm thinking is that this is meant to be the desktop component and the mobile equivalent is the "popup" component. Perhaps you can flip the logic around so that the initial view is desktop, then use the value from `useNYPLBreakpoints` to conditionally render the children. But do want @bigfishdesign13 to verify this.

The component as is does not need track of data, which is good, but I think it needs the companion `useFilterBar` hook in examples to show how consuming devs can use both the hook and the component together to manage state. Sorry, not sure if that was clear or not but it might change a bit of how this is approached, but if it was decided to not use it for this round then ignore the comment.

bigfishdesign13:
> The other option I'm thinking is that this is meant to be the desktop component and the mobile equivalent is the "popup" component.

Yes, what you stated is correct.  While the `FilterBarInline` component is responsive and does offer mobile styles, NYPL will only use it for the desktop viewport.  The `FilterBarPopup` component will be used for mobile viewports.  If you need to flip the style logic to ensure that the desktop styles render firwst, that would be fine.

src/components/FilterBarInline/FilterBarInline.stories.tsx
jackiequach: It was a bit hard to read, so I added a note to breakout the render prop functions in the docs.

src/components/FilterBarInline/FilterBarInline.tsx
jackiequach: The initial load still displays the mobile styles for the `MultiSelectGroup` component, as that component will need to be updated to not use `useNYPLBreakpoints`.

src/components/FilterBarInline/FilterBarInline.mdx
```diff
@@ -0,0 +1,349 @@
...
+import Link from "../Link/Link";
+
+<Meta of={FilterBarInlineStories} />
+
+# FilterBarInline
+
+| Component Version | DS Version    |
+| ----------------- | ------------- |
+| Added             | `Pre-release` |
+| Latest            | `Pre-release` |
```
EdwinGuzman: nitpick and super minor: we write this as `Prerelease`

src/components/FilterBarInline/FilterBarInline.mdx
```diff
@@ -0,0 +1,349 @@
...
+- [TextInput](../?path=/docs/components-form-elements-textinput--docs#accessibility)
+- [Toggle](../?path=/docs/components-form-elements-toggle--docs#accessibility)
+
+## Layout Patterns
+
+The display of the `FilterBarInline` child components can be set to a `"row"` or a
+`"column"` layout. On screen width under 600px width the `layout` is
+automatically set to `"column"`.
+
+**Note:** when `layout = "column"` is passed, `MultiSelect` components will have
```
EdwinGuzman, line 101: This is true but it will only happen when the value is passed down correctly in the render function, right? I think it's worth saying that this will happen when the `FilterBarInline` component's values are passed down appropriately but you can say something like `see the "Passing Children" section for more information...".

This is also true for the next "Note".

EdwinGuzman: Actually, sorry, the `isBlockElement` is specific to the `MultiSelectGroup` but I think the above is still applicable to width and layout for other components (like `CheckboxGroup` and `RadioGroup`) right?

src/components/FilterBarInline/FilterBarInline.mdx
```diff
@@ -0,0 +1,349 @@
...
+    // ...
+  />
+</Box>
+; `}
+  language="jsx"
+/>
+
+### Column Layout
+
+<Canvas of={FilterBarInlineStories.ColumnLayout} />
```
EdwinGuzman: How about reordering this so it first shows the source code and then the relevant UI?

src/components/FilterBarInline/FilterBarInline.mdx
```diff
@@ -0,0 +1,349 @@
...
+/>
+
+## Passing Children
+
+<Canvas of={FilterBarInlineStories.MultipleChildren} />
+
+In order to control the styling and props passed to children form elements,
+the `FilterBarInline` component accepts `CheckboxGroup`, `MultiSelectGroup`,
+`RadioGroup`, `Checkbox`, `DatePicker`, `MultiSelect`, `Radio`, `Select`,
+`Slider`, `TextInput`, and `Toggle` children through its `renderChildren` render
```
EdwinGuzman, line 228: You give really good full examples below, but I think a short snippet to explain the shape of this will be helpful. Something like
```
<FilterBarInline
  id="..."
  heading="..."
  renderChildren={({ isBlockElement, layout, width }) => {
    return (...)
  }}
/>
```

src/components/FilterBarInline/FilterBarInline.mdx
```diff
@@ -0,0 +1,349 @@
...
+    onClear={onClearFilterBar}
+    onSubmit={() => console.log(selectedFilterItems)}
+    selectedItems={selectedFilterItems}
+    renderChildren={renderFilterComponents}
+  />
+); `}
+  language="jsx"
+/>
+
+## useMultiSelect hook
```
EdwinGuzman: I think you can remove this section entirely. You can still mention this and link to the appropriate doc page in the section above near the "To control each of the child form elements...." sentence. At that point, you can say "`useMultiSelect` is an optional but recommended hook ...." and then link to the docs.

It follows the same pattern as "Find documentation on how to do this on each individual component page".

src/components/FilterBarInline/FilterBarInline.mdx
```diff
@@ -0,0 +1,349 @@
...
+`isBlockElement`, `layout`, and `width`.
+
+- The `isBlockElement` argument value should be passed to the `isBlockElement`
+  prop of `MultiSelect` children.
+- The `layout` argument value should be passed to the `layout` prop of
+  `CheckboxGroup`, `MultiSelectGroup`, and `RadioGroup` children.
+- The `width` argument value should be passed to the `multiSelectWidth` prop of
+  `MultiSelectGroup` children and the `width` prop of `MultiSelect` children.
+
+## Controlling State Using selectedItems, onClear, and onSubmit Props
```
EdwinGuzman, line 256: I really like the flow of this section, showing code and explaining it as it gets to the final component render.

src/components/FilterBarInline/FilterBarInline.stories.tsx
```diff
@@ -0,0 +1,1204 @@
...
+      selectedItems={selectedFilterItems}
+      renderChildren={({ layout, width }) => {
+        return (
+          <MultiSelectGroup
+            id="multiselect-group"
+            labelText="MultiSelect Group"
+            layout={layout}
+            multiSelectWidth={width}
+            renderMultiSelect={({ isBlockElement, multiSelectWidth }) => {
+              return (
```
EdwinGuzman: Not a problem but just funny you don't follow the recommendation you gave about splitting this up into smaller functions!

src/components/FilterBarInline/FilterBarInline.tsx
```diff
@@ -0,0 +1,142 @@
+import { chakra, ChakraComponent, Stack } from "@chakra-ui/react";
+import React, { forwardRef } from "react";
+
+import Button from "../Button/Button";
+import ButtonGroup from "../ButtonGroup/ButtonGroup";
+import { LayoutTypes } from "../../helpers/types";
+import { MultiSelectWidths, SelectedItems } from "../MultiSelect/MultiSelect";
+import useNYPLBreapoints from "../../hooks/useNYPLBreakpoints";
```
EdwinGuzman: typo: `useNYPLBreapoints`

src/components/FilterBarInline/FilterBarInline.tsx
```diff
@@ -0,0 +1,142 @@
+import { chakra, ChakraComponent, Stack } from "@chakra-ui/react";
+import React, { forwardRef } from "react";
+
+import Button from "../Button/Button";
+import ButtonGroup from "../ButtonGroup/ButtonGroup";
+import { LayoutTypes } from "../../helpers/types";
+import { MultiSelectWidths, SelectedItems } from "../MultiSelect/MultiSelect";
```
EdwinGuzman: I think `SelectedItems` is a bit confusing here and in the args table. What do you think about renaming this `SelectedItems as MultiSelectItems`? Then it's clear that the `FilterBarItemsType` accepts data from the `MultiSelect` component. I think that might still be confusing and don't have a good answer for this.

src/components/FilterBarInline/FilterBarInline.tsx
```diff
@@ -0,0 +1,142 @@
...
+        renderChildren,
+        ...rest
+      } = props;
+
+      const { isLargerThanMobile } = useNYPLBreapoints();
+      const finalWidth = isLargerThanMobile
+        ? layout === "row"
+          ? "fitContent"
+          : "full"
+        : "full";
```
EdwinGuzman: There a barely visible flicker when the component loads and I think it's because of this. It doesn't seem like there's an easy way around this though, right?

EdwinGuzman: Perhaps load the desktop styles by default so that, if there is a flicker, it happens on mobile where we have less traffic. This still doesn't really solve the issue though...

EdwinGuzman (changes requested):
Looks great! I have mostly doc-related comments.

I am concerned about the flicker issue however. Because those are not values set through responsive styles in Chakra and are prop values instead, I can see that being an issue...

I'm tempted to ask for an example that includes _all_ DS filter components since you listed them down in the component description: `CheckboxGroup`, `MultiSelectGroup`, `RadioGroup`, `Checkbox`, `DatePicker`, `MultiSelect`, `Radio`, `Select`, `Slider`, `TextInput`, and `Toggle`. I'm not sure if we really need this right now. What we can do is create that working example in Turbine with all filter components.

src/components/FilterBarInline/FilterBarInline.mdx
```diff
@@ -0,0 +1,349 @@
...
+
+<Source
+  code={`
+const onClearFilterBar = () => {
+  onClearAll();
+  setSelectedCheckbox([]);
+  setTextValue("");
+};`}
+  language="jsx"
+/>
```
bigfishdesign13, line 305: What do you think about including in this example the code for the `FilterBarInline` component with the `onClear` prop set to `onClearFilterBar`?

bigfishdesign13: Ok, I made it further down the page and I see that there is a culminating code block showing everyting tied together in the `return` statement.  This makes more sense now.  For further clarification, can you add appropriate subheadings for each subsection in this block?

src/components/FilterBarInline/FilterBarInline.mdx
bigfishdesign13: I agree with Edwin.  In a way, we've demoted the significance of this hook, so linking to the `useMultiSelect` docs should be enough.

src/components/FilterBarInline/FilterBarInline.mdx
```diff
@@ -0,0 +1,349 @@
...
+The display of the `FilterBarInline` child components can be set to a `"row"` or a
+`"column"` layout. On screen width under 600px width the `layout` is
+automatically set to `"column"`.
+
+**Note:** when `layout = "column"` is passed, `MultiSelect` components will have
+`isBlockElement` set to `true` by default and the child components `width`
+will be set to `"full"`.
+
+**Note:** `CheckboxGroup`, `MultiSelectGroup`, and `RadioGroup` `layout` prop will
+match the value of the `layout` set for the `FilterBarInline` component.
```
bigfishdesign13, line 106: Feel free to use the `Banner` component to add notes like this.

src/components/FilterBarInline/FilterBarInline.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+      return (
+        <ComponentWrapper
+          className={className}
+          id={`filter-bar-inline-${id}`}
+          headingText={heading}
```
bigfishdesign13: Similar to what is done in `Banner`, the `Heading` element needs to be set to default size.  In this case, it should be set to `heading5`.  To go along with that, add a new `FilterBarInline Heading` section right after the `Accessibility` section.  The new section should more or less mirror what we have in the `Banner` docs.

https://nypl.github.io/nypl-design-system/reservoir/v3/?path=/docs/components-feedback-banner--docs#banner-heading

bigfishdesign13 (changes requested):
This is wonderful!  There are just a few things that need to be addressed.

src/components/FilterBarInline/FilterBarInline.mdx
jackiequach: `isBlockElement` and `width` are specific to `MultiSelectGroup` and `layout` applies to the other components.

src/components/FilterBarInline/FilterBarInline.tsx
jackiequach: I updated the component to default on desktop styles, but there is still a flicker that is caused by the MultiSelectGroup component.

src/components/FilterBarInline/FilterBarInline.mdx
jackiequach: The Banner component would require special styling when using code snippets.

EdwinGuzman (approved):
Works well and the updates look great. We do need others to review before this gets merged in but you can still pass the Vercel preview URL to Clare for an accessibility review.

Not a concern but are we recommending devs/users to use this component with the "Apply filters" and "Clear all filters" buttons? I think it makes sense to recommend this pattern since that's a big reason to use this component. What do you think @bigfishdesign13 @jackiequach ?

jackiequach:
> Not a concern but are we recommending devs/users to use this component with the "Apply filters" and "Clear all filters" buttons? I think it makes sense to recommend this pattern since that's a big reason to use this component. What do you think @bigfishdesign13 @jackiequach ?

@EdwinGuzman I agree, without the buttons it just seems like a container for form elements. I'll add the recommendation to the docs if Marty agrees.

bigfishdesign13:
> Not a concern but are we recommending devs/users to use this component with the "Apply filters" and "Clear all filters" buttons? I think it makes sense to recommend this pattern since that's a big reason to use this component. What do you think @bigfishdesign13 @jackiequach ?

@jackiequach and @EdwinGuzman – The official UX recommendation is to use neither the Apply button nor the Clear button.  But rather to use real-time filtering (no need for Apply button) in conjunction with the TagSet component (includes a Clear button).  That said, we do want to offer the Apply and Clear buttons for apps that are not able to support the full UX recommendations right off the bat.

src/components/FilterBarInline/FilterBarInline.tsx
```diff
@@ -0,0 +1,159 @@
...
+}
+
+/**
+ * `FilterBarInline` is a wrapper component for filter components and individual form input components.
+ * The DS considers the following components as filter components: `CheckboxGroup`, `MultiSelectGroup`,
+ * `RadioGroup`, `Checkbox`, `DatePicker`, `MultiSelect`, `Radio`, `Select`, `Slider`, `TextInput`, and `Toggle`.
+ *
+ * The wrapped components/component groups can be displayed in a column or
+ * in a row layout. `FilterBarInline` can render additional `Clear All` and a `Apply Filters` buttons. The two
+ * _optional_ buttons are controlled by the `onClear` or `onSubmit` props respectively.
```
7emansell: I think it could be clearer here that using these buttons is not the officially recommended pattern (if that's the final call)

7emansell (approved):
Looks fantastic to me!! just wanted a little more explicit clarification about dev usage

jackiequach:
I added an example for using the `TagSet` component. It may be a bit complicated to implement when consuming apps using more filter components since the tag set data with the selected filter items. I think it might be better to add this to Turbine, but let me know your thoughts.

@bigfishdesign13 @EdwinGuzman @7emansell

src/components/FilterBarInline/FilterBarInline.mdx
```diff
@@ -380,6 +402,52 @@ return (
   language="jsx"
 />
 
+## TagSet Example
+
+<Canvas of={FilterBarInlineStories.TagSetExample} />
+
```
EdwinGuzman, line 408: Can you add a short explanation to say that the following code is just one example of how to use it but it could be done differently depending on the app's needs?

src/components/FilterBarInline/FilterBarInline.mdx
```diff
@@ -380,6 +402,52 @@ return (
   language="jsx"
 />
 
+## TagSet Example
+
+<Canvas of={FilterBarInlineStories.TagSetExample} />
```
EdwinGuzman, line 407: Really good example. It feels a big laggy to me but do you notice it as well?

EdwinGuzman:
> I added an example for using the TagSet component. It may be a bit complicated to implement when consuming apps using more filter components since the tag set data with the selected filter items. I think it might be better to add this to Turbine, but let me know your thoughts.

I think the text with the recommended approach and using the Tagset component is clear. I agree that a more complicated example should be shown in Turbine. One specific place to add this example with the TagSet component is in Turbine's Article and Databases page: https://nypl-ds-test-app.vercel.app/fullPages/articles-databases#above-header-notification

Right now, the MultiSelect components are commented out : https://github.com/NYPL/nypl-ds-test-app/blob/main/pages/fullPages/articles-databases.tsx#L142

src/components/FilterBarInline/FilterBarInline.mdx
jackiequach: I do notice some lag when updating the text input. The solution implemented for updating the tag set data is probably not the most efficient.

EdwinGuzman (approved):
Latest updates looks good to me!

src/components/FilterBarInline/FilterBarInline.mdx
bigfishdesign13: Yes, it is a little laggy, but acceptable in this example.  I'm curious to see what happens as the number of input children increases.  Let's keep an eye on it.

bigfishdesign13 (approved):
Looks great!  Let's just keep an eye on the potential lagginess.

PR #1610, opened 2024-05-23
Release v3.1.4

bigfishdesign13:
### Adds

- Adds `showRowDividers` prop to show/hide row dividers for the `List` `"dl"` variant.

### Updates

- Exports `useMediaQuery` and `chakra` from Chakra UI.
- Updates `Accordion` to close focused panel when "esc" key is pressed.
- Updates the `TextInput` and `SearchBar` components to better associate the input element to the entire component's helper text.

### Fixes

- Fixes the z-index issue in the `MultiSelect` component's selected item button.
- Fixes the `Checkbox` positioning so it is visible when it is focused in a scrollable container.
- Fixes a `Breadcrumbs` snapshot test that was duplicated and now has the appropriate `"locations"` variant.


EdwinGuzman (approved):
Looks like `release` to `development` sync done last time did not fix the "out-of-date" issue we keep seeing with the git branches.

PR #1614, opened 2024-06-04
DSD-1750: Select interaction tests

7emansell:
Fixes JIRA ticket [DSD-1750](https://newyorkpubliclibrary.atlassian.net/browse/DSD-1750).

- Adds a test that selects an option (with mouse)
- **See https://github.com/testing-library/user-event/issues/786– looks like there's a standing issue in Testing Library with keyboard inputs to dropdowns, none of the keyboard events I tried seemed to change the state of the select

## How has this been tested?

Locally, in Storybook

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/Select/Select.stories.tsx
+++ b/src/components/Select/Select.stories.tsx
@@ -6,7 +6,7 @@ import Button from "../Button/Button";
 import Form, { FormField } from "../Form/Form";
 import Select, { labelPositionsArray, selectTypesArray } from "./Select";
 import { argsBooleanType } from "../../helpers/storybookUtils";
-import { userEvent, within } from "@storybook/test";
+import { expect, userEvent, within } from "@storybook/test";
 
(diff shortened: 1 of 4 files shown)


src/components/Select/Select.stories.tsx
```diff
@@ -82,6 +83,11 @@ export const WithControls: Story = {
     },
     jest: ["Select.test.tsx"],
   },
+  play: async ({ canvasElement }) => {
+    const screen = within(canvasElement);
+    const select = screen.getByRole("combobox");
+    await userEvent.selectOptions(select, "green");
```
EdwinGuzman, line 90: Add a few more events to select other values. Marty and I still go back and forth on this, but add one expectation that the select's value is what you expect it to be.

EdwinGuzman (approved):
Just update the changelog!

PR #1619, opened 2024-06-10
DSD-1685: Github Action tag/release fix

EdwinGuzman:
Fixes JIRA ticket [DSD-1685](https://newyorkpubliclibrary.atlassian.net/browse/DSD-1685)

- I think this should be enough to fix the issue we keep encountering. The `GITHUB_TOKEN` value is not read because it's not an accepted value in under the "with" property.

## How has this been tested?

Not yet, not until the next release.

- N/A

### Checklist:

### Front End Review:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,8 +23,9 @@ jobs:
       - name: Tag
         id: autotagger
         uses: butlerlogic/action-autotag@stable
-        with:
+        env:
           GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+        with:
           strategy: package
           tag_prefix: v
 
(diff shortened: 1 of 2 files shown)


bigfishdesign13 (approved):
👍

.github/workflows/release.yml
```diff
@@ -23,8 +23,9 @@ jobs:
       - name: Tag
         id: autotagger
         uses: butlerlogic/action-autotag@stable
```
jackiequach: A version number needs to be specified here instead of stable. [Others](https://github.com/ButlerLogic/action-autotag/issues/45#issuecomment-1825726927) have said 1.1.2 worked for them.

jackiequach (changes requested):

jackiequach (approved):

PR #1620, opened 2024-06-12
DSD-1740: Adds ability to close MultiSelect on blur

oliviawongnyc:
Fixes JIRA ticket [DSD-1740](https://newyorkpubliclibrary.atlassian.net/browse/DSD-1787)

- Adds a prop to `MultiSelect` called `closeOnBlur` which, when true, closes the component's accordion when the component loses focus.

## How has this been tested?

- In Storybook

- None that I know of.

### Checklist:

### Front End Review:


src/components/MultiSelect/multiSelectChangelogData.ts
```diff
@@ -9,6 +9,15 @@
 import { ChangelogData } from "../../utils/ComponentChangelogTable";
 
 export const changelogData: ChangelogData[] = [
+  {
```
oliviawongnyc, line 12: I added this second object rather than combining this and the one before because they have different "types", but can adjust if wanted.

src/components/MultiSelect/MultiSelect.mdx
```diff
@@ -27,6 +27,7 @@ import { changelogData } from "./multiSelectChangelogData";
 - {<Link href="#block-element-or-float" target="_self">Block Element or Float</Link>}
 - {<Link href="#width" target="_self">Width</Link>}
 - {<Link href="#default-open-state" target="_self">Default Open State</Link>}
+- {<Link href="#close-on-blur-state" target="_self">Close on Blur State</Link>}
```
EdwinGuzman, line 30: Nit: the link doesn't work. It should be `#closeOnBlur` to match the section's title.

src/components/MultiSelect/MultiSelect.mdx
```diff
@@ -577,6 +578,50 @@ state by default.
...
+When `closeOnBlur` is set to `true`, the `MultiSelect` accordion will close when
+focus leaves the component (the user clicks outside).
+
+<Canvas of={MultiSelectStories.closeOnBlurState} />
+
+<Source
+  code={`
+<MultiSelect
+  buttonText="MultiSelect"
+  closeOnBlur=true
```
EdwinGuzman: This should be either `closeOnBlur={true}` or just the prop `closeOnBlur`.

src/components/MultiSelect/MultiSelect.stories.tsx
```diff
@@ -250,6 +253,7 @@ export const withControls: Story = {
   args: {
     buttonText: "MultiSelect",
     id: "multi-select-id",
+    closeOnBlur: true,
```
EdwinGuzman: Set this to `false` since it's "false" by default.

src/components/MultiSelect/multiSelectChangelogData.ts
EdwinGuzman: You can merge them together and add both types in the `affects` array.

src/components/MultiSelect/MultiSelect.mdx
EdwinGuzman: On line 581, rename `## closeOnBlur` to `## Close on Blur State` to match the section name in `MultiSelectGroup.mdx`.

src/components/MultiSelectGroup/MultiSelectGroup.mdx
```diff
@@ -102,6 +103,14 @@ argument value should be passed to the `width` prop of the `MultiSelect`.
   language="tsx"
 />
 
+## Close on Blur State
+
+When `closeOnBlur` is set to `true`, the `MultiSelect` accordion will close when
+focus leaves the component (the user clicks outside). In the case of a `MultiSelectGroup`,
+this means only one accordion will be open at a time.
+
+<Canvas of={MultiSelectGroupStories.closeOnBlurState} />
```
EdwinGuzman, line 112: I like this behavior and would vote to make it the default (so `closeOnBlur` set to true as the default). I (personally) get a bit annoyed when I select an option in a `MultiSelect`, open another one, but the existing one is still open. I would expect it to close but maybe that's just me.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -93,13 +97,39 @@ export const MultiSelect: ChakraComponent<
...
       // Create a ref to hold a reference to the accordian button, enabling us
       // to programmatically focus it.
       const accordionButtonRef: React.RefObject<HTMLDivElement> =
         useRef<HTMLDivElement>();
+      const containerRef: React.RefObject<HTMLDivElement> =
+        useRef<HTMLDivElement>();
       const expandToggleButtonRef: React.RefObject<HTMLButtonElement> =
         useRef<HTMLButtonElement>();
 
+      const handleClick = (event) => {
```
EdwinGuzman: Chakra has a `useOutsideClick` function that sort of replicates what you have here. Did you get a chance to try that? Check out `useCloseDropDown.ts` for an example.

I think you can reuse the `useCloseDropDown` function here.

EdwinGuzman (changes requested):
Mostly want to see if you can reuse the `useCloseDropDown` function from Chakra.

If this function was the default in the `Accordion`, then you would probably use the `useCloseDropDown` hook for the esc key press and the click outside functionality. But I know that the `Accordion` shouldn't close for clicks outside of it. Just an observation because I was trying to find a way to do this without having to create the `userClickedOutside` state variable but I don't see a way around that.

src/components/MultiSelectGroup/MultiSelectGroup.mdx
bigfishdesign13: @EdwinGuzman I agree with your sentiment about a multiselect not closing when you move away from it, but from the NYPL UX prespective, that pattern is only applicable when the `MultiSelect` component is used in a horizonal layout filter bar.  

Additionally, I've been trying to move away from setting a positive sounding prop to `true` by default and then having to explicitly set it to `false` to negate it.  To better illustrate what I mean here, think about the labels on our form inputs.  Would we have been better off naming the label prop `hideLabel` rather than `showLabel`?  If it was `hideLabel`, then the resulting code would be simpler with `hideLabel` instead of `showLabel={false}`.

bigfishdesign13 (commented):
The "close on blur" works well when clicking away from the component, but it should also be triggered for any event that removes focus from the component.  For example, when using the keyboard to tab away from the component.  This type of functionality will be very useful when the `MultiSelect` component is used in a horizontal filter bar.

oliviawongnyc:
> The "close on blur" works well when clicking away from the component, but it should also be triggered for any event that removes focus from the component. For example, when using the keyboard to tab away from the component. This type of functionality will be very useful when the `MultiSelect` component is used in a horizontal filter bar.

@EdwinGuzman I played around with Chakra's `useOutsideClick` hook, but ultimately abandoned it since it doesn't work for keyboard events as described in Marty's comment above, so I was going to have to address them anyway and now I am addressing both keyboard events and mouse clicks with the same function. LMK your thoughts.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -93,13 +97,48 @@ export const MultiSelect: ChakraComponent<
...
         useRef<HTMLDivElement>();
+      const containerRef: React.RefObject<HTMLDivElement> =
+        useRef<HTMLDivElement>();
       const expandToggleButtonRef: React.RefObject<HTMLButtonElement> =
         useRef<HTMLButtonElement>();
 
+      const handleEvent = (e) => {
+        if (
+          e.type === "mousedown" ||
+          (e.type === "keydown" && e.code === "Enter")
```
EdwinGuzman: I was a bit confused by this. The latest request was to close it on blur when moving focus away from the multiselect such as by tabbing. When you tab, it doesn't close the MultiSelect. Also, I think you are using "Enter" because you can open/close a MultiSelect with Enter, but you can also open/close with the "Space" key.

I can see this being difficult to implement with "tab" because you want to check for the "tab" key _only_ when moving away from the component (or "shift + tab" when moving backwards). 

If it's true that it's too complicated to implement tab/space keys, let's make it a new ticket. I'd like to merge this in with just the click outside action and add on top of it later on.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -93,13 +97,48 @@ export const MultiSelect: ChakraComponent<
...
+          if (
+            containerRef.current &&
+            !containerRef.current.contains(e.target)
+          ) {
+            setUserClickedOutside(true);
+          } else {
+            setUserClickedOutside(false);
+          }
+        } else {
+          setUserClickedOutside(false);
```
EdwinGuzman: Do you need to fire these events for any and all key presses? I think you can get away with removing line 126 but you do need it on line 123.

src/components/MultiSelect/MultiSelect.tsx
EdwinGuzman: I was able to manage to have two open by tabbing and opening with "space". So this should be fixed to align with the text in the doc about only having one open at a time.

Screenshot 2024-06-24 at 5 54 27 PM

EdwinGuzman (changes requested):
The update looks good and yes, it makes sense to have both type of events in the same function. Only thing now is to update this to include "space" and "tab"/"shift tab" keyboard presses to close on those blur events.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -93,13 +97,51 @@ export const MultiSelect: ChakraComponent<
...
+
+        // Handles when user tabs outside of the container
+        if (e.type === "blur") {
+          if (!e.currentTarget.contains(e.relatedTarget)) {
+            setUserClickedOutside(true);
+          } else {
+            setUserClickedOutside(false);
+          }
+        }
+      };
```
EdwinGuzman, line 135: Can you try breaking this out into a separate function so that the mousedown event triggers `handleEvent` and the onblur event on the div element triggers another function with this code block? This is mostly to separate the two concepts.

EdwinGuzman:
@oliviawongnyc the latest update works great! Just one minor request to try before I approve.

src/components/MultiSelect/MultiSelect.tsx
oliviawongnyc: Yup, shouldn't force it if it doesn't make sense

EdwinGuzman (approved):
Thanks for the updates!

src/components/MultiSelect/MultiSelect.mdx
```diff
@@ -577,6 +578,50 @@ state by default.
   language="jsx"
 />
 
+## Close on Blur State
+
+When `closeOnBlur` is set to `true`, the `MultiSelect` accordion will close when
+focus leaves the component (the user clicks outside).
```
bigfishdesign13: Minor update.  Update the text in the parentheses.
```
(the user clicks outside or uses the keyboard to tab away)
```

bigfishdesign13 (approved):
Works great! Just make the one minor text update that I noted and update the `Latest` version to `Prerelease` before merging.

PR #1628, opened 2024-06-24
DSD-1746: FeedbackBox interaction tests

7emansell:
Fixes JIRA ticket [DSD-1746](https://newyorkpubliclibrary.atlassian.net/browse/DSD-1746)

- Adds interaction tests in Storybook for the `FeedbackBox` component

### Checklist:

### Front End Review:

[DSD-1746]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-1746?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/FeedbackBox/FeedbackBox.stories.tsx
+++ b/src/components/FeedbackBox/FeedbackBox.stories.tsx
@@ -103,14 +103,14 @@ export const WithControls: Story = {
     hiddenFields: undefined,
     id: "feedbackBox-id",
     isInvalidComment: false,
-    isInvalidEmail: false,
+    isInvalidEmail: true,
     isOpen: undefined,
     notificationText: undefined,
     onClose: undefined,
     onOpen: undefined,
     onSubmit: undefined,
-    showCategoryField: false,
-    showEmailField: false,
+    showCategoryField: true,
+    showEmailField: true,
     title: "Help and Feedback",
     view: "form",
   },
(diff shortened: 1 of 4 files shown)


src/components/FeedbackBox/FeedbackBox.stories.tsx
```diff
@@ -1,6 +1,7 @@
 import { Box } from "@chakra-ui/react";
 import type { Meta, StoryObj } from "@storybook/react";
 import { useState } from "react";
+import { userEvent, expect, screen } from "@storybook/test";
```
EdwinGuzman, line 4: Did not know you can get the mounted DOM directly from the package. I've been doing `const screen = within(canvasElement);`

EdwinGuzman (approved):
It'd be nice to get tests for the other fields but I'm not sure if you can rerender the component and set `showCategoryField` and `showEmailField` to true. Maybe those should be set to true for storybook, but that can be separate and a later step.

src/components/FeedbackBox/FeedbackBox.stories.tsx
7emansell: Right, something to remember for `Modal` for sure bc `screen` here is not the same as `within(canvasElement)`

7emansell:
> It'd be nice to get tests for the other fields but I'm not sure if you can rerender the component and set `showCategoryField` and `showEmailField` to true. Maybe those should be set to true for storybook, but that can be separate and a later step.

@bigfishdesign13 I think we've done this before with the `Searchbar` interaction tests (added props to the default component just so they can be tested) so I think it would make sense to do here as well, but lmk

bigfishdesign13:
> > It'd be nice to get tests for the other fields but I'm not sure if you can rerender the component and set `showCategoryField` and `showEmailField` to true. Maybe those should be set to true for storybook, but that can be separate and a later step.
> 
> @bigfishdesign13 I think we've done this before with the `Searchbar` interaction tests (added props to the default component just so they can be tested) so I think it would make sense to do here as well, but lmk

@7emansell Yes, I think that would be fine to set `showCategoryField` and `showEmailField` to `true` for the `With Controls` story. 

@EdwinGuzman The next release is next Thursday, so there would be time to tackle this change as part of this ticket.

EdwinGuzman (approved):

PR #1630, opened 2024-06-27
Container query feature branch

oliviawongnyc:
Part of work for [DSD-1719](https://newyorkpubliclibrary.atlassian.net/browse/DSD-1719)

- Release candidate for container query work
- Adds container queries to `Card`, `FeaturedContent`, and `Searchbar`

-

### Checklist:

### Front End Review:

[DSD-1719]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-1719?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jackiequach:
@bigfishdesign13 @EdwinGuzman This PR should be reviewed before merging in

src/components/Card/Card.tsx
```diff
@@ -320,46 +321,49 @@ export const Card: ChakraComponent<
...
+              />
+            )}
+            <Box className="card-body" __css={styles.body} data-body>
+              {cardContents}
             </Box>
-          ) : null}
-        </CardWrapper>
+            {cardRightContents.length ? (
+              <Box
+                className="card-right"
```
EdwinGuzman: This doesn't seemed to be used, can it be removed?

src/theme/components/card.ts
```diff
@@ -119,8 +129,68 @@ const ReservoirCard = defineMultiStyleConfig({
           bodyMargin = "0";
         }
       }
+      // These sizes are only for the "row" layout.
+      // const imageSize = size ? imageSizes[size] : {};
```
EdwinGuzman: Remove just the second line of the comment.

src/theme/components/card.ts
```diff
@@ -119,8 +129,68 @@ const ReservoirCard = defineMultiStyleConfig({
...
       return {
         base: {
+          containerType: "inline-size",
+          "@container (min-width: 0px)": {
```
EdwinGuzman: This is intentional, correct? Using media queries, you wouldn't need this clause, but it seems setting this allows it to allow shrink past the default size.

src/theme/components/card.ts
```diff
@@ -119,8 +129,68 @@ const ReservoirCard = defineMultiStyleConfig({
...
+                ...imageSizesMinWidth0[imageSize],
+              },
+            }),
+          },
+          [`@container (min-width: ${breakpoints.sm})`]: {
+            "[data-body]": {
+              flexBasis: isRow ? "100%" : null,
+            },
+          },
+          [`@container (min-width: ${breakpoints.md})`]: {
```
EdwinGuzman: I wonder if this needs a +1 here. The following is the container query on the left and production on the right, setting the width to "md (600px)". I think we won't have parity but it's ok, I would rather use the breakpoint value than `{breakpoint} + 1`. Thoughts?

Screenshot 2025-01-29 at 9 27 40 PM

src/theme/components/featuredContent.ts
```diff
@@ -46,37 +47,53 @@ const FeaturedContent = defineMultiStyleConfig({
...
+          "@container (min-width: 0px)": {
+            "[data-wrapper]": {
+              flexDirection: imageAtEnd ? "column-reverse" : "column",
+            },
+            "[data-image]": {
+              height: "320px",
+              width: "100%",
+            },
+          },
+          [`@container (min-width: ${breakpoints.md})`]: {
```
EdwinGuzman: Similar observation as the Card at 600px, but I think the container query update looks better. We should just flag that we expect this to look different.

Screenshot 2025-01-29 at 9 30 48 PM

src/theme/components/searchBar.ts
```diff
@@ -1,22 +1,38 @@
...
-      md: "auto",
+    "@container (min-width: 0px)": {
+      "&": { marginBottom: "xs", flexFlow: "column nowrap" },
+      " > select": { marginBottom: "-1px" },
+      "[data-button]": {
+        borderRightRadius: "none",
+        maxWidth: "unset",
+      },
+    },
+    [`@container (min-width: ${breakpoints.md})`]: {
```
EdwinGuzman: This view is also off but it just needs a minor update for the select's right border:
Screenshot 2025-01-29 at 9 20 11 PM

EdwinGuzman (commented):
Looks good but just some questions.

src/theme/components/card.ts
jackiequach: Yes, this is to set the base styles of the elements within the container. This was done using array syntax for the media query implementation.

src/theme/components/featuredContent.ts
jackiequach: Yes there will be visual breaking changes since its against the container's width instead of the screen's width. The storybook padding causes the styles to update earlier at a larger screen width.

EdwinGuzman:
Yes, okay, I'm still getting used to this and it makes a lot more sense. Thanks!!

I removed the storybook padding and the display is the same as prod.

EdwinGuzman (approved):
Just fix the tests. Thanks!

src/theme/components/card.ts
bigfishdesign13: @EdwinGuzman I created a POC combining the container query and v4 feature branches. With the new breakpoint updates, I also reconfigured the values for the viewport sizing tool and using that tool make more sense now.

https://nypl-design-system-git-container-query-breakpoints-nypl.vercel.app/?path=/story/components-basic-elements-featuredcontent--with-controls

src/theme/components/card.ts
bigfishdesign13: Would it be possible to set up a reusable method for formatting the code block for the container styles? Something like this (please excuse any syntax errors)?
```
const setContainerStyles = (props) => {
  const { breakpoint, styles } = props;
  const minWidth = {
    base: "0px",
    mobile: breakpoints.mobile,
  };
  return `"@container (min-width: ${minWidth[breakpoint]})": { ${styles} }`;
};
```

src/theme/components/featuredContent.ts
```diff
@@ -46,37 +47,53 @@ const FeaturedContent = defineMultiStyleConfig({
             bgColor: "dark.ui.bg.default",
           },
           img: screenreaderOnly(),
+          containerType: "inline-size",
+          "@container (min-width: 0px)": {
+            "[data-wrapper]": {
+              flexDirection: imageAtEnd ? "column-reverse" : "column",
+            },
+            "[data-image]": {
```
bigfishdesign13: This aligns with the `imageWrapper` styles below, right? Can we make it a rule that the same name be used across the container query styles and the standard styles?
```
data-imageWrapper
```

src/theme/components/featuredContent.ts
```diff
@@ -46,37 +47,53 @@ const FeaturedContent = defineMultiStyleConfig({
...
-           * positioned at the start.
-           * */
-          paddingStart: full && imageAtEnd ? { base: null, md: 0 } : null,
         },
         imgWrapper: {
           backgroundPosition: "center",
           backgroundSize: "cover",
-          height: { base: "320px", md: "auto" },
-          width: { base: "100%", md: wrapperWidth },
         },
```
bigfishdesign13, line 103: How would you feel about merging the remaining `base` styles into the block for the base styles under the container query styles? That would simplify the code. With that change, my comment above would be irrelevant.

src/theme/components/card.ts
bigfishdesign13: Also, with the updated breakpoints, the breakpoints object includes semantic breakpoint names as well. `md` would be replaced with `mobile` or `largeMobile`.

bigfishdesign13 (commented):
I have some questions.

src/theme/components/featuredContent.ts
jackiequach: I feel like this wouldn't be a best practice since container queries are designed to target styling based on the parent's size and not the overall viewport size, and applying them too broadly may lead to unnecessary complexity and potential performance issues

src/theme/components/card.ts
jackiequach: added this and updated styles accordingly @bigfishdesign13

src/utils/setContainerStyles.ts
```diff
@@ -0,0 +1,19 @@
+import breakpoints from "../theme/foundations/breakpoints";
+
+/**
+ * Generates container query styles based on the provided breakpoint and styles.
+ *
+ * @param {string} props.breakpoint - The breakpoint at which the container query should apply. It can be "base" or a key from the breakpoints object.
+ * @param {object} props.styles - The styles to apply within the container query.
+ * @returns {object} An object containing the container query styles.
+ */
+export const setContainerStyles = (props: {
```
EdwinGuzman, line 10: This is great. At first, I was a bit hesitant about this. It seemed like a bit more added complexity for an already complicated CSS way of adding styles. The initial approach of adding the `@containers` rules in the style files made more to me as I was reading and understanding container queries. This adds an easier way to write them, but hope it's not more complicated when learning container queries.

EdwinGuzman (approved):

src/utils/setContainerStyles.ts
jackiequach: I think there is a ticket for writing up a Storybook page for a guide to adding container queries to components. We can clarify any complicated approaches there.

EdwinGuzman: Sounds great, thanks!

src/theme/components/featuredContent.ts
jackiequach: Just rechecked this, data-wrapper refers to the wrapper of the image wrapper and text. data-image aligns to the imgWrapper styles

src/utils/setContainerStyles.ts
```diff
@@ -0,0 +1,19 @@
+import breakpoints from "../theme/foundations/breakpoints";
+
+/**
+ * Generates container query styles based on the provided breakpoint and styles.
+ *
+ * @param {string} props.breakpoint - The breakpoint at which the container query should apply. It can be "base" or a key from the breakpoints object.
```
bigfishdesign13, line 6: When talking about container queries, we have always thought about the styles in terms of breakpoints, but that does not need to always be the case. The style adjustments could be based on any width value. That being said, should this argument also allow standard style measurements like `300px`, `80rem`, and `25%`?

src/theme/components/card.ts
```diff
@@ -133,60 +133,69 @@ const ReservoirCard = defineMultiStyleConfig({
...
-              marginTop: "0",
-              maxWidth: "180px",
-            },
-            "[data-body]": {
-              display: "block",
-              flexFlow: "row nowrap",
-              width: "auto",
+          }),
+          ...setContainerStyles({
+            breakpoint: "sm",
```
bigfishdesign13, line 162: @jackiequach and @EdwinGuzman – During dev, I would be more comfortable using the semantic breakpoint names here (`smallMobile`, `tablet`, etc.). Do you have a preference one way or the other?

bigfishdesign13 (commented):
More questions.

src/theme/components/featuredContent.ts
bigfishdesign13: Good point.

src/utils/setContainerStyles.ts
jackiequach: I agree that the min-width for container queries do not necessarily need to be at our current breakpoints, but I think it would still be beneficial to have set values to be used in this function to minimize needing to handle many different container sizes. Having a focus on key breakpoints would likely allow for better performance and maintanence.

src/theme/components/card.ts
jackiequach: I think once the breakpoints PR is integregated the semantic names would be used. For now it can use the current breakpoint keys.

bigfishdesign13 (approved):
Looks great!

PR #1631, opened 2024-06-28
Storybook version bump

EdwinGuzman:

This was done to fix a few dependabot security issues.

- Updates Storybook and similar packages to version 8.1.11.
- Removes a package I thought was previously removed.

## How has this been tested?

Locally. Should be tested in Turbine as an `rc`.

- N/A does not affect components, just Storybook.

### Checklist:

### Front End Review:


src/components/Button/Button.stories.tsx
```diff
@@ -38,8 +38,8 @@ const meta: Meta<typeof Button> = {
       description: "**Only used for Storybook – ** Select an icon. ",
     },
     id: { control: false },
-    isDisabled: { table: { defaultValue: { summary: false } } },
```
bigfishdesign13, line 41: This seems odd.  Why the need for quotes on a boolean value?

src/components/MultiSelect/MultiSelect.stories.tsx
```diff
@@ -206,22 +206,22 @@ const meta: Meta<typeof MultiSelect> = {
   component: MultiSelect,
   argTypes: {
     defaultItemsVisible: {
-      table: { defaultValue: { summary: 5 } },
+      table: { defaultValue: { summary: "5" } },
```
bigfishdesign13, line 209: And this.

src/helpers/storybookUtils.ts
```diff
@@ -1,7 +1,7 @@
 /**
  * For usage in Storybook stories to set the default value of a boolean prop.
  */
-export const argsBooleanType = (defaultValue = false) => ({
+export const argsBooleanType: any = (defaultValue = false) => ({
   control: { type: "boolean" },
-  table: { defaultValue: { summary: defaultValue } },
+  table: { defaultValue: { summary: `${defaultValue}` } },
```
bigfishdesign13: This may be the weirdest one of all.

bigfishdesign13 (approved):
It works and everything seems to look fine, but I am very curious about the need for the quotes.

src/helpers/storybookUtils.ts
EdwinGuzman: Ah yes, it's because the `summary` must now be a string so I had to cast it. I will actually make an update to update this to be more correct.

EdwinGuzman:
The need for quotes is coming from Storybook and their update. It now requires a string as the correct type which makes sense but it's an annoyance for us.

EdwinGuzman:
@bigfishdesign13 made an update. Lmk what you think.

bigfishdesign13:
@EdwinGuzman Looks good.

bigfishdesign13 (approved):
Approving!

PR #1635, opened 2024-07-03
Release v3.1.7

bigfishdesign13:
### Adds

- Adds Storybook interaction tests for the `CheckboxGroup`, `DatePicker`, `FeedbackBox`, and `Slider` components.
- Adds `digitalCollections` color variant to the `Breadcrumbs` component.
- Adds `closeOnBlur` prop to the `MultiSelect` component which, when set to true, closes the component if it loses focus.

### Updates

- Updates base font stack to only `system-ui` and `sans-serif`.
- Updates the `Accordion` component to close panel when element within panel is focused and "esc" key is pressed.


src/components/Accordion/accordionChangelogData.ts
```diff
@@ -14,7 +14,10 @@ export const changelogData: ChangelogData[] = [
     version: "3.1.4",
```
jackiequach, line 23: Should this version number be updated?

bigfishdesign13: Ah, good catch.  Those should be split into two separate entries.  I made the adjustment and updated the PR.

jackiequach (approved):
Looks good!

PR #1643, opened 2024-07-29
DSD-1797: Breadcrumbs custom link prop

7emansell:
Fixes JIRA ticket [DSD-1797](https://newyorkpubliclibrary.atlassian.net/browse/DSD-1797).

- Adds `customLinkComponent` prop to `Breadcrumbs`
- Adds `linkProps` prop to `breadcrumbsData`

## How has this been tested?

Locally, and in [this Turbine branch](https://nypl-ds-test-b59lbu4si-nypl.vercel.app/components/breadcrumbs)

### Accessibility Checklist

## Open Questions

- `customLink` is resistant to stricter typing but I would like for it not to be `any`?

### Checklist:

### Front End Review:

[DSD-1797]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-1797?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Breadcrumbs/Breadcrumbs.mdx
+++ b/src/components/Breadcrumbs/Breadcrumbs.mdx
@@ -20,6 +20,7 @@ import Link from "../Link/Link";
 - {<Link href="#component-props" target="_self">Component Props</Link>}
 - {<Link href="#accessibility" target="_self">Accessibility</Link>}
 - {<Link href="#long-titles" target="_self">Long Titles</Link>}
+- {<Link href="#custom-link-and-link-props" target="_self">Custom Link and Link Props</Link>}
 - {<Link href="#color-variations" target="_self">Color Variations</Link>}
 - {<Link href="#changelog" target="_self">Changelog</Link>}
 
--- a/src/components/Breadcrumbs/Breadcrumbs.mdx
+++ b/src/components/Breadcrumbs/Breadcrumbs.mdx
@@ -9,10 +9,10 @@ import Link from "../Link/Link";
 
 # Breadcrumbs
 
-| Component Version | DS Version |
-| ----------------- | ---------- |
-| Added             | `0.0.3`    |
-| Latest            | `3.1.7`    |
+| Component Version | DS Version   |
+| ----------------- | ------------ |
+| Added             | `0.0.3`      |
+| Latest            | `Prerelease` |
 
 ## Table of Contents
 
--- a/src/components/Breadcrumbs/Breadcrumbs.mdx
+++ b/src/components/Breadcrumbs/Breadcrumbs.mdx
@@ -70,9 +70,9 @@ the truncated page title.
 
 <Canvas of={BreadcrumbsStories.LongText} />
 
-## Custom Link and Link Props
+## Custom Link Component and Link Props
 
-The `customLink` prop allows apps with internal routers (Next.js, for example) to pass their custom link component
+The `customLinkComponent` prop allows apps with internal routers (Next.js, for example) to pass their custom link component
 into the `Breadcrumbs` component. This allows the `Breadcrumbs` links to use client-side routing, so users can navigate between pages without reloading the full page.
 
 Additionally, the `linkProps` prop on `breadcrumbsData` allows any of the framework's [custom props](https://nextjs.org/docs/app/api-reference/components/link) to be applied on a link-by-link basis.
@@ -90,7 +90,10 @@ const breadcrumbsDataWithLinkProps = [
 { url: "#", text: "Grandchild" },
 ];
 
-<Breadcrumbs customLink={Link} breadcrumbsData={breadcrumbsDataWithLinkProps} />
+<Breadcrumbs
+  customLinkComponent={Link}
+  breadcrumbsData={breadcrumbsDataWithLinkProps}
+/>
 `} language="tsx" />
 
 Below story uses the DS `Link` component.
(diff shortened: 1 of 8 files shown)


src/components/Breadcrumbs/Breadcrumbs.mdx
```diff
@@ -69,6 +70,31 @@ the truncated page title.
 
 <Canvas of={BreadcrumbsStories.LongText} />
 
+## Custom Link and Link Props
+
+The `customLink` prop allows apps with internal routers (Next.js, for example) to pass their custom link component into Breadcrumbs so the links can use client-side routing.
```
EdwinGuzman: nit: "into the `Breadcrumbs` component so ...."

src/components/Breadcrumbs/Breadcrumbs.mdx
```diff
@@ -69,6 +70,31 @@ the truncated page title.
...
+To maximize these benefits, the `linkProps` prop on `breadcrumbsData` allows accompanying custom props on a link-by-link basis.
```
EdwinGuzman: > To maximize these benefits,

I think I know what you're trying to say but you mentioned client-side routing and I'm assuming that's a benefit. I think you can add one or two more benefits or go into why client-side routing is beneficial in an app before this sentence.

src/components/Breadcrumbs/Breadcrumbs.mdx
```diff
@@ -69,6 +70,31 @@ the truncated page title.
...
+
+Note: assigning a value for `href` in `linkProps` will override the `url` prop, so avoid this.
```
EdwinGuzman: We try to include "recommend" in these types of notes so how about, ", ... we strongly recommend against this." or something along those lines.

src/components/Breadcrumbs/Breadcrumbs.mdx
```diff
@@ -69,6 +70,31 @@ the truncated page title.
...
+<Source
+  code={`
+  // Example with Next.js Link and associated props.
+const breadcrumbsData = [
+  { url: "#", text: "Parent", linkProps: { replace: true, scroll: false }}
+  { url: "#", text: "Child", , linkProps: { replace: true }}
+  { url: "#", text: "Grandchild" },
+];
+// ...
+<Breadcrumbs customLink={Link} breadcrumbsData={breadcrumbsData} />
```
EdwinGuzman: At the top of this code example snippet, add the import for the `Link` component from next/link, to explicitly show where it's imported from.

src/components/Breadcrumbs/Breadcrumbs.mdx
EdwinGuzman: > accompanying custom props on a link-by-link basis

Can you also describe some examples and link to the nextjs link prop docs? You show it below in the code snippet but it'll be useful to say that it's really any prop that the framework's link component accepts.

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,10 +1,10 @@
 import {
   Breadcrumb as ChakraBreadcrumb,
   BreadcrumbItem,
-  BreadcrumbLink,
   chakra,
   useStyleConfig,
   ChakraComponent,
+  BreadcrumbLink,
```
EdwinGuzman: Why the move?

src/components/Breadcrumbs/breadcrumbsChangelogData.ts
```diff
@@ -9,6 +9,13 @@
 import { ChangelogData } from "../../utils/ComponentChangelogTable";
 
 export const changelogData: ChangelogData[] = [
+  {
+    date: "Prerelease",
+    version: "Prerelease",
+    type: "Update",
+    affects: ["Functionality"],
+    notes: ["Adds `customLink` and `linkProps` props."],
```
EdwinGuzman: Add why they're added or what it supports.

EdwinGuzman (changes requested):
Looks good but some documentation update. I haven't tested it beyond that time we tested it together and it was working but I'll try to review it in Turbine or DCF if you have a working branch/vercel preview.

7emansell:
> Looks good but some documentation update. I haven't tested it beyond that time we tested it together and it was working but I'll try to review it in Turbine or DCF if you have a working branch/vercel preview.

https://nypl-ds-test-b59lbu4si-nypl.vercel.app/components/breadcrumbs!

EdwinGuzman (approved):
Looks and works great. I'm debating whether `as` is a better prop name than `customLink`. `as` follows Chakra's convention and that's consistent.

There are various components that use the DS Link internally but I think only `Heading` and `Pagination` would benefit from using the nextjs link (from a quick glance). So if you were to add this feature to those components, would `as` or `customLink` work better?

I'm approving the functionality but updating the name is a small update.

7emansell:
I feel like `customLink` is clearer to me as a dev than `as` (which even as a Chakra convention is slightly confusing) but I agree either could work, so happy to change it!

jackiequach (approved):
I think that the `customLink` prop makes sense here since `as` would imply the `Breadcrumbs` component is a `Link`, which is not true since it would be the children of the component would be links.

bigfishdesign13:
@7emansell,
I don't think `as`, in this case, is appropriate.  As Jackie noted, it doesn't feel right as a component level prop name.  Something like `linkAs` would be better.  That said, I am leaning toward the other naming convention.  But I think even that could be better.  `customLinkType` or `customLinkComponent` feels better to me.

Cc @jackiequach @EdwinGuzman

PR #1646, opened 2024-08-12
DSD-1811: Searchbar responsive styling

7emansell:
Fixes JIRA ticket [DSD-1811](https://newyorkpubliclibrary.atlassian.net/browse/DSD-1811).

- Updates the responsive styles on `Searchbar` (no stacking on mobile)
- Adds `<span>` to "Search" text so DCF can use new breakpoints

### Accessibility Checklist

### Checklist:

### Front End Review:

[DSD-1811]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-1811?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/SearchBar/searchBarChangelogData.ts
+++ b/src/components/SearchBar/searchBarChangelogData.ts
@@ -10,6 +10,13 @@ import { ChangelogData } from "../../utils/ComponentChangelogTable";
 
 export const changelogData: ChangelogData[] = [
   {
+    date: "Prerelease",
+    version: "Prerelease",
+    type: "Update",
+    affects: ["Styles"],
+    notes: ["Updates Searchbar style to remove stacking on mobile breakpoints."],
+  },
+  {
     date: "2024-06-20",
     version: "3.1.6",
     type: "Update",
--- a/src/components/SearchBar/searchBarChangelogData.ts
+++ b/src/components/SearchBar/searchBarChangelogData.ts
@@ -15,7 +15,7 @@ export const changelogData: ChangelogData[] = [
     type: "Update",
     affects: ["Styles"],
     notes: [
-      "Updates Searchbar style to remove stacking on mobile breakpoints.",
+      "Updates Searchbar style to remove stacking on mobile breakpoints, adds span element to text for CSS targeting.",
     ],
   },
   {
--- a/src/components/SearchBar/searchBarChangelogData.ts
+++ b/src/components/SearchBar/searchBarChangelogData.ts
@@ -15,7 +15,7 @@ export const changelogData: ChangelogData[] = [
     type: "Update",
     affects: ["Styles"],
     notes: [
-      "Updates Searchbar style to remove stacking on mobile breakpoints, adds span element to text for CSS targeting.",
+      "Updates styles on mobile breakpoints, adds span element to text for CSS targeting.",
     ],
   },
   {
--- a/src/components/SearchBar/searchBarChangelogData.ts
+++ b/src/components/SearchBar/searchBarChangelogData.ts
@@ -14,7 +14,9 @@ export const changelogData: ChangelogData[] = [
     version: "Prerelease",
     type: "Update",
     affects: ["Styles"],
-    notes: ["Updates Searchbar style to remove stacking on mobile breakpoints."],
+    notes: [
+      "Updates Searchbar style to remove stacking on mobile breakpoints.",
+    ],
   },
   {
     date: "2024-06-20",
--- a/src/theme/components/searchBar.ts
+++ b/src/theme/components/searchBar.ts
@@ -10,7 +10,7 @@ const SearchBar = defineMultiStyleConfig({
       base: "xs",
       md: "auto",
     },
-    ".textInput" :{
+    ".textInput": {
       flexGrow: 1,
     },
     flexFlow: { base: "column nowrap", md: "row" },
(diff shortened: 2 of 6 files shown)


src/components/SearchBar/searchBarChangelogData.ts
```diff
@@ -9,6 +9,15 @@
 import { ChangelogData } from "../../utils/ComponentChangelogTable";
 
 export const changelogData: ChangelogData[] = [
+  {
+    date: "Prerelease",
+    version: "Prerelease",
+    type: "Update",
+    affects: ["Styles"],
+    notes: [
+      "Updates Searchbar style to remove stacking on mobile breakpoints.",
```
EdwinGuzman: Add the span update as well.

EdwinGuzman (changes requested):
This looks good but small stylistic issue: remove the rounded right border radius of the input element so it flows into the search button:

production:
Screenshot 2024-08-13 at 12 03 50 PM

this PR:
Screenshot 2024-08-13 at 12 03 59 PM

EdwinGuzman (approved):

src/components/SearchBar/searchBarChangelogData.ts
```diff
@@ -9,6 +9,15 @@
 import { ChangelogData } from "../../utils/ComponentChangelogTable";
 
 export const changelogData: ChangelogData[] = [
+  {
+    date: "Prerelease",
+    version: "Prerelease",
+    type: "Update",
+    affects: ["Styles"],
+    notes: [
+      "Updates Searchbar style to remove stacking on mobile breakpoints, adds span element to text for CSS targeting.",
```
bigfishdesign13: This is a little misleading because the component will still "stack" when the select element is enabled.  Just update this statement to mirror what is in the CHANGELOG.

src/theme/components/searchBar.ts
```diff
@@ -10,14 +10,29 @@ const SearchBar = defineMultiStyleConfig({
...
-      base: "column nowrap",
-      md: "row nowrap",
-    },
     ".textInput": {
-      flexGrow: "1",
+      flexGrow: 1,
+      " div > input": {
+        borderRightRadius: 0,
+        borderLeftRadius: "sm",
+      },
```
bigfishdesign13: This is overriding the logic that sets the left radius to `0` when the select element is enabled and the left radius is now rounded when the select element is enabled.  Make sure the left radius is `0` when the select element is enabled.

src/theme/components/searchBar.ts
```diff
@@ -10,14 +10,29 @@ const SearchBar = defineMultiStyleConfig({
...
+    },
+    flexFlow: { base: "column nowrap", md: "row" },
+    button: {
+      borderLeftRadius: "none",
+      lineHeight: "1.70",
+      marginBottom: "auto",
+      padding: "xs",
```
bigfishdesign13: This should only be set for mobile.

src/theme/components/searchBar.ts
```diff
@@ -10,14 +10,29 @@ const SearchBar = defineMultiStyleConfig({
...
     ".textInput": {
-      flexGrow: "1",
+      flexGrow: 1,
+      " div > input": {
+        borderRightRadius: 0,
+        borderLeftRadius: "sm",
+      },
+    },
+    flexFlow: { base: "column nowrap", md: "row" },
+    button: {
```
bigfishdesign13: Add `minWidth: "44px"` to ensure that the button is square on moble.

bigfishdesign13 (changes requested):
I have a couple of small style requests.

Also, I didn't think we were going to do the full mobile updates in the `SearchBar` component yet becuase the DS breakpoints do not align with DC's breakpoints.

7emansell:
@bigfishdesign13 Sorry I misunderstood our conversation– if we are not doing mobile updates for the `Searchbar` is any of this needed? None of this will be helpful to DC right now (besides adding the span element) but it does move the DS a step closer to what's in the designs for DC, and once all breakpoints align we can definitely use this

src/theme/components/searchBar.ts
```diff
@@ -10,14 +10,34 @@ const SearchBar = defineMultiStyleConfig({
...
-    },
     ".textInput": {
-      flexGrow: "1",
+      flexGrow: 1,
+      "div > input": {
+        borderRightRadius: 0,
+      },
+    },
+    flexFlow: { base: "column nowrap", md: "row" },
+    ".searchButton": {
```
EdwinGuzman, line 20: Why did you switch from the previous `button` keyword to the class name? CSS precedence rules?

7emansell: Yes styles were getting applied to the clear button

EdwinGuzman: That's annoying. I know there are other instances of class name as the key in the them object but I'd like to move away from that so we don't use class names. Leave this as is but keep that in mind.

src/theme/components/searchBar.ts
```diff
@@ -10,14 +10,34 @@ const SearchBar = defineMultiStyleConfig({
...
+      minWidth: "44px",
+      maxWidth: "80px",
```
bigfishdesign13: This should be removed.  This is what is causing the button spacing to get out of whack.

bigfishdesign13 (changes requested):
Looks good.  Just needs one final adjustment.

src/theme/components/searchBar.ts
7emansell: On >md breakpoints it seems like this is needed though?

bigfishdesign13 (approved):
💯

PR #1652, opened 2024-08-26
DSD-1827: Adding Tooltip positioning props

7emansell:
Fixes JIRA ticket [DSD-1827](https://newyorkpubliclibrary.atlassian.net/browse/DSD-1827).

- Adds `placement` and `offset` props to the `Tooltip`

## How has this been tested?

Locally

### Accessibility Checklist

### Checklist:

### Front End Review:

[DSD-1827]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-1827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Card/Card.mdx
```diff
@@ -208,6 +209,14 @@ This example can be found in the [Turbine homepage](https://nypl-ds-test-app.ver
 
 <Canvas of={CardStories.CardFullClickTurbineExample} />
 
+## Card With Full-Click Functionality and Tooltip
+
+The `Tooltip` relies on pointer events that are overridden by the card's `LinkOverlay`.
```
EdwinGuzman: `LinkOverlay` is not mentioned in this doc besides this. Can you say that it's an internal component that is added through the  "full-click" functionality? Just to be clear how/when it's added and why it affects the tooltip.

src/components/Card/Card.mdx
```diff
@@ -208,6 +209,14 @@ This example can be found in the [Turbine homepage](https://nypl-ds-test-app.ver
...
+
+If you need a tooltip within a clickable `Card`, wrap the entire `Card` component inside the `Tooltip` and use the `placement` and `offset` props to move the `Tooltip` to display next to the intended reference. Keep in mind that `offset` is static pixel values, so check where the `Tooltip` displays on all breakpoints.
```
EdwinGuzman: I'm a bit inclined to say add an example of the responsive `offset` values since you mention it, but I think it may be enough to say for now to keep track of it through state and maybe the `useNYPLBreakpoint` hook, if the app needs to update the value.

src/components/Card/Card.tsx
```diff
@@ -8,7 +8,6 @@ import {
   useStyleConfig,
 } from "@chakra-ui/react";
 import React, { forwardRef } from "react";
-
```
EdwinGuzman, line 11: Nit: I like separating packages with internal files.

src/components/Tooltip/Tooltip.mdx
```diff
@@ -96,6 +96,19 @@ on a disabled `Button`, pass the `shouldWrapChildren` prop.
 
 <Canvas of={TooltipStories.DisablingTooltipButton} />
 
+### Placement
+
+The `placement` prop adjusts where the tooltip is displayed relative to its children.
```
EdwinGuzman: List down some of the (commonly used) values you expect users to use.

EdwinGuzman: Or point to the "show code" button so they can see the values, but it's probably better to list them down explicitly.

src/components/Tooltip/Tooltip.mdx
```diff
@@ -96,6 +96,19 @@ on a disabled `Button`, pass the `shouldWrapChildren` prop.
...
+
+<Canvas of={TooltipStories.TooltipPlacement} />
+
+### Offset
+
+The `offset` prop sets the distance along the main and cross axis between the tooltip and its reference element.
+`offset` is just static pixel values, so be careful to check where your tooltip displays on all breakpoints.
```
EdwinGuzman: Give a visual example of the prop value in " pixel values, such as `[100, 100]`, so be careful..."  or somewhere else.

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -13,6 +11,8 @@ import Toggle from "../Toggle/Toggle";
 import Tooltip from "./Tooltip";
 import { argsBooleanType } from "../../helpers/storybookUtils";
 import { getPlaceholderImage } from "../../utils/utils";
+import Card, { CardContent, CardHeading } from "../Card/Card";
+import { VStack } from "../..";
```
EdwinGuzman: Ah, are you pointing to the internal import/export?

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -42,6 +42,8 @@ export const WithControls: Story = {
     id: "tooltip-id",
     isDisabled: false,
     shouldWrapChildren: false,
+    placement: "top",
+    offset: [0, 8],
```
EdwinGuzman: Why 8?

src/components/Tooltip/tooltipChangelogData.ts
```diff
@@ -9,6 +9,13 @@
 import { ChangelogData } from "../../utils/ComponentChangelogTable";
 
 export const changelogData: ChangelogData[] = [
+  {
+    date: "Prerelease",
+    version: "Prerelease",
+    type: "Update",
+    affects: ["Styles"],
+    notes: ["Adds placement and offset props."],
```
EdwinGuzman: nit: backticks around prop names

src/components/Tooltip/tooltipChangelogData.ts
```diff
@@ -9,6 +9,13 @@
 import { ChangelogData } from "../../utils/ComponentChangelogTable";
 
 export const changelogData: ChangelogData[] = [
+  {
+    date: "Prerelease",
+    version: "Prerelease",
+    type: "Update",
+    affects: ["Styles"],
```
EdwinGuzman: Add "functionality" as well.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -36,6 +40,8 @@ export const Tooltip: ChakraComponent<
       className,
       content,
       id,
+      placement = "top",
+      offset = [0, 8],
```
EdwinGuzman, line 44: This answers my question above, 8 is the default value. I see it's too close to the element when it's `0,0`.

EdwinGuzman (commented):
Works well but very minor updates.

EdwinGuzman (approved):

src/components/Card/Card.mdx
```diff
@@ -208,6 +209,15 @@ This example can be found in the [Turbine homepage](https://nypl-ds-test-app.ver
 
 <Canvas of={CardStories.CardFullClickTurbineExample} />
 
+## Card With Full-Click Functionality and Tooltip
+
+When the `Card` is fully clickable, an internal component wraps everything in a link, which in turn overrides the pointer events that trigger a `Tooltip`.
+
+If you need a tooltip within a clickable `Card`, wrap the entire `Card` component inside the `Tooltip` and use the `placement` and `offset` props to move the `Tooltip` to display next to the intended reference.
+Keep in mind that `offset` is static pixel values, so check where the `Tooltip` displays on all breakpoints. You may need to keep track of the offset values through state, and update them conditionally with the `useNYPLBreakpoints` hook.
```
bigfishdesign13, line 217: The checks for the position values need to be more granular than just the breakpoints.  In fact, the position values should be generated dynamically looking at the width of the viewport.  And it should be recalculated any time the viewport (i.e. browser window) changes size.  This is readily apparent in the new example that was added to the `Card` story docs.

https://nypl-design-system-git-dsd-1827-tooltip-position-props-nypl.vercel.app/?path=/story/components-basic-elements-card--full-click-with-tooltip

Resposition the `addon` sidebar to the right side of the screen and then adjust the width of the `addon` sizebar.  Check the tooltip each time you change the width.

Many of the NYPL sites, and especially the DC site, allows cards (and the internal images) to resize based on the available width.  As the image resizes, the Y position of the card heading changes.  If a tooltip is intended for the heading, the position values will always need to be recalculated based on the available width.

Can you update this example to use dynamic position values that are updated as the viewport width changes?

src/components/Tooltip/Tooltip.mdx
```diff
@@ -96,6 +96,19 @@ on a disabled `Button`, pass the `shouldWrapChildren` prop.
...
+### Placement
+
+The `placement` prop adjusts where the tooltip is displayed relative to its children: "top", "bottom", "left", or "right".
+
+<Canvas of={TooltipStories.TooltipPlacement} />
+
+### Offset
+
+The `offset` prop sets the distance along the main and cross axis between the tooltip and its reference element.
+`offset` is just static pixel values, such as [100, 100], so be careful to check where your tooltip displays on all breakpoints.
```
bigfishdesign13, line 110: This statement will also need to be updated based on my comment on the `Card` component.

Also add the following:
```
While it is possible to set the offset of the tooltip, it is 
recommended to not set custom offset values unless there 
are extenuating circumstances.
```

src/components/Tooltip/Tooltip.mdx
```diff
@@ -96,6 +96,19 @@ on a disabled `Button`, pass the `shouldWrapChildren` prop.
 
 <Canvas of={TooltipStories.DisablingTooltipButton} />
 
+### Placement
+
+The `placement` prop adjusts where the tooltip is displayed relative to its children: "top", "bottom", "left", or "right".
```
bigfishdesign13, line 101: Add the following:
```
While it is possible to set the placement of the tooltip, it is 
recommended to use the default `top` placement.
```

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -42,6 +42,8 @@ export const WithControls: Story = {
     id: "tooltip-id",
     isDisabled: false,
     shouldWrapChildren: false,
+    placement: "top",
```
bigfishdesign13: `"top"` is the default, so this is not needed here.

bigfishdesign13 (changes requested):
There are somethings we need to think about.  Please see my comments.

Also, make sure to update the `argTypes` in `Tooltip.stories.tsx` to properly alphabetize `offset` and `placement` and to show the default values for each new prop.

src/components/Card/Card.mdx
EdwinGuzman: Good catch. I was not resizing and checking every time because I don't think the use-case of DC needs it for for widths smaller than desktop and the image is always the same size.

I'm concern about dynamically keeping track of this value because, if it's used on a site like DC with various tooltips, then the page might get laggy from all the calculations. The height of the image can probably be used to dynamically generate the offset. Otherwise, I'm thinking we don't advertise this as a feature because not getting the exact location of the tooltip on a responsive Card can be tricky and expensive. 

@7emansell try it out and let us know how it goes.

EdwinGuzman (approved):
The responsive update is a good baseline for developers so they can implement a more fluid version in their own app. @bigfishdesign13 , I don't think it needs to be perfect since no team is really asking for this feature and it's more of an implementation detail that came up for the DC Facelift use case (using full-click `Card` with the `Tooltip`).

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -22,7 +22,13 @@ const meta: Meta<typeof Tooltip> = {
     className: { control: false },
     id: { control: false },
     isDisabled: argsBooleanType(),
+    offset: { control: false },
     shouldWrapChildren: argsBooleanType(),
+    placement: {
+      control: "select",
+      options: ["top", "bottom", "left", "right", "auto"],
+      table: { defaultValue: { summary: "top" } },
+    },
```
bigfishdesign13, line 30: Nit: move this above `shouldWrapChildren`.

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -41,9 +47,9 @@ export const WithControls: Story = {
       "This is a tooltip with information describing the hovered object.",
     id: "tooltip-id",
     isDisabled: false,
+    offset: undefined,
     shouldWrapChildren: false,
-    placement: "top",
-    offset: [0, 8],
+    placement: undefined,
```
bigfishdesign13, line 51: Nit: move this above `shouldWrapChildren`.

bigfishdesign13 (approved):
Very nice.  Just cleanup the alphabetization before merging.

PR #1658, opened 2024-09-03
DSD-1814: table enhancements

bigfishdesign13:
Fixes JIRA ticket [DSD-1814](https://newyorkpubliclibrary.atlassian.net/browse/DSD-1814)

- Adds the `isScrollable` prop to enable horizontal scrolling.
- Add the `columnStyles` prop to allow consuming apps to set custom column styles (i.e. width, text alignment, etc.).
- Adds the `tableTextSize ` prop to set the size of the text within the table.
- Adds the `titleText` and `showTitleText` props to control the `<caption>` element and `arial-label` attribute.

The documentation related to horizontal scrolling has been added as well as infomation about the difference in the mobile view for scrolling vs. non-scrolling tables.

## How has this been tested?

- locally in Storybook

- When horizontal scrolling is enabled, `role` and `tabindex` attributes are added to the table wrapper in order to meet a11y requirements related to focus and keyboard control.
- Accessibility concerns listed under the Storybook `Accessibility` tab are inconsequential because they are related to the background gradient used to imply depth when scrolling.  The background gradient does not impact text readability.

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[DSD-1814]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-1814?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Image/Image.tsx
```diff
@@ -31,9 +31,9 @@ export const imageSizesArray = [
   "large",
 ] as const;
 export const imageTypesArray = ["default", "circle"] as const;
-export type ImageRatios = (typeof imageRatiosArray)[number];
-export type ImageSizes = (typeof imageSizesArray)[number];
-export type ImageTypes = (typeof imageTypesArray)[number];
```
EdwinGuzman, line 36: Automatic eslint update?

src/components/Table/Table.tsx
```diff
@@ -143,11 +171,30 @@ export const Table: ChakraComponent<
...
                   key === 0 && useRowHeaders ? (
-                    <ChakraTh scope="row" key={key}>
+                    // row header cell
+                    <ChakraTh
+                      scope="row"
+                      key={key}
+                      sx={
+                        isScrollable
+                          ? columnStyles[key]
+                          : { base: undefined, md: columnStyles[key] }
```
EdwinGuzman: I think this also works if you leave out the `base` property, right?

src/components/Table/Table.tsx
```diff
@@ -170,20 +217,50 @@ export const Table: ChakraComponent<
...
+            overflow: "auto",
+            role: "region",
+            tabIndex: 0,
+            whiteSpace: "wrap",
+          }
+        : {
+            overflow: "hidden",
+            whiteSpace: "wrap",
+          };
+
```
EdwinGuzman: Can these rules be moved to the theme object and then assigned as `<TableContainer sx={styles}>`?

src/components/Table/Table.tsx
```diff
@@ -27,11 +32,21 @@ export interface TableProps {
   columnHeadersBackgroundColor?: string;
   /** Hex value to set the text color of the column headers. */
   columnHeadersTextColor?: string;
+  /** Array of style objects used to set custom width values for the table
+   * columns. Will accept "width" and "maxWidth" attributes. */
```
EdwinGuzman: are `width` and `maxWidth` examples or the only expected values that are passed?

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 ### Updates
 
 - Updates `Tooltip` component with `placement` and `offset` props for custom positioning.
+- Updates the `Table` component to add the `isScrollable` prop to enable horizontal scrolling.
+- Updates the `Table` component to add the `customStyles` prop to allow for custom column styles (i.e. width, text alignment, etc.).
```
EdwinGuzman: This should be `columnStyles`.

src/components/Table/Table.mdx
```diff
@@ -95,6 +129,74 @@ color contrast.
...
+  { minWidth: "200px", width: "15%", textAlign: "right" },
+];`}
+  language="jsx"
+/>
+
+<Canvas of={TableStories.CustomColumnStyles} />
+
+## Horizontal Scrolling
+
+When using `customStyles`, if the `min-width` values cause a table to exceed the
```
EdwinGuzman: `columnStyles`

src/components/Table/Table.mdx
```diff
@@ -95,6 +129,74 @@ color contrast.
...
+<Canvas of={TableStories.HorizontalScrolling} />
+
+## Responsive Mobile Layout
+
+If the `isScrollable` prop is true, the layout of the `Table` component is the
+same for all viewports and horizontal scrolling will always be enabled. The only
+difference between desktop and mobile views is that the row headers column will
+not be locked in place for mobile viewports.
+
+If the `isScrollable` prop is false (or omitred), the `Table` component layout
```
EdwinGuzman: typo "omitred"

EdwinGuzman (commented):
Looks good but minor updates. @dgcohen This does have the scrollable mobile table so it can/should be used for the Bib details page when this goes out. Please review soon.

I'm a bit concerned about the `columnStyles` prop. It has to perfectly match the length of the columns in order to work. I was wondering if it could be part of the data object (similar to other components like `Menu`) but the data is split up into `tableData` and `columnHeaders` so that won't work. 

Can you add an example where maybe one or two columns does not need custom styles so the array looks like `[{ width: ..., minWidth: ... }, {}, {}, {width:..., minWidth: ..}]` to show that an empty array should be passed.

src/components/Table/tableChangelogData.ts
```diff
@@ -9,6 +9,18 @@
...
 
 export const changelogData: ChangelogData[] = [
+  {
+    date: "Prerelease",
+    version: "Prerelease",
+    type: "Update",
+    affects: ["Accessibility", "Documentation", "Functionality", "Styles"],
+    notes: [
+      "Added the `isScrollable` prop to enable horizontal scrolling.",
+      "Added the `customStyles` prop to allow for custom column styles (i.e. width, text alignment, etc.).",
```
dgcohen: This should also be columnStyles

src/theme/components/customTable.ts
```diff
@@ -1,38 +1,82 @@
...
+});
 
 export const baseTRStyles = (
   columnHeadersBackgroundColor = "",
+  isScrollable = false,
   showRowDividers = false,
   useRowHeaders = false
 ) => ({
-  borderBottom: { base: "2px solid", md: 0 },
+  borderBottom: isScrollable ? "0" : { base: "2px solid", md: "0" },
```
dgcohen: Doesn't make a difference in terms of functionality, but is there a reason these 0 's are being passed as strings?

src/theme/components/customTable.ts
```diff
@@ -42,40 +86,40 @@ export const baseTRStyles = (
 export const baseCellStyles = (
   columnHeadersBackgroundColor = "",
   columnHeadersTextColor = "",
-  showRowDividers = false,
-  useRowHeaders = false
+  tableTextSize = "",
```
dgcohen: Looks like tableTextSize is optional, can this be omitted?

src/components/Image/Image.tsx
bigfishdesign13: I am not sure where the parentheses initially came from.  I feel like they appeared when I merged with `development` at one point.

src/components/Table/Table.tsx
bigfishdesign13: Tthis code actually does not work as expected, regardless of using `base` or not.  I had to rewrite it using `isLargerThanMobile`, but I am concerned about possible style flickering.

src/components/Table/Table.tsx
bigfishdesign13: I think I wrote that at the beginning of this update with the expectation those would be the only styles that would be accepted.  Any style can be passed in, but `width` and `maxWidth` are the styles I see as the most applicable. I am also using `textAlign`, which is another one that makes sense when thinking about displaying a column of numbers.  I will update that description.

src/theme/components/customTable.ts
bigfishdesign13: Nope.  I just removed the quotes.  All good.

src/theme/components/customTable.ts
bigfishdesign13: `tableTextSizeI` is being used on line 97, so it is needed here.  It is optional on the component level, so that is why I gave it a default value.

src/components/Table/Table.tsx
bigfishdesign13: These styles and attributes are for a conditional parent of the actual table element.  The parent is not always applied to the component.  Is there a way to set styles for a parent element from within the `Table` theme object?

dgcohen (approved):

src/components/Image/Image.tsx
EdwinGuzman: I think they appeared from Diego's update. We briefly chatted about that.

src/components/Table/Table.tsx
EdwinGuzman: What is the conditional parent? Is it `TableContainer` and why is it not always applied to the component?

EdwinGuzman: You can/should think of the base theme object styles, the outermost object, as the parent. If it always renders then it would get `sx={styles}` as the prop. `ChakraTable` would need something like `sx={styles.innerTable}` (random name).

bigfishdesign13: I was able to rework the styles as a multi-part configuration.  :)

EdwinGuzman (approved):

PR #1659, opened 2024-09-05
DSD-1585: HelperErrorText styling bug

oliviawongnyc:
Fixes JIRA ticket [DSD-1585](https://newyorkpubliclibrary.atlassian.net/browse/DSD-1585)

- Fixes issue with HelperErrorText where, if text value was not a string, necessary styles weren't applied.
- Important note!! I realize my change broke some functionality, but I wonder if we are ok with it? If so, I will update the documentation. 

Basically, the following test failed because the code now sees strings as strings and JSX elements as JSX elements, rather than interpreting HTML strings like "<b>This text is bold</b>". (Marty this is why we were using dangerouslySetInnerHtml, I believe), but I feel like this maybe should be the case anyway. Let me know what you think.

  it("renders the text passed as an HTML string", () => {
    render(<HelperErrorText text="<b>This text is bold</b>" />);
    expect(screen.getByText("This text is bold")).toBeInTheDocument();
  });

## How has this been tested?

- Locally, in Storybook.

- None.

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[DSD-1585]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-1585?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bigfishdesign13:
With this update, when the `text` prop is passed as a string, HTML tags that are included in the string will no longer be rendered.  That is definitely a breaking change and it will either need to be reverted or annother solution will need to be implemented to allow the HTML tags to be rendered propperly.  Form field helper text is everywhere and we can't take this risk.

I see that we are using `dangerouslySetInnerHTML` in a number of other components, so I think we can stick with that.

Thoughts?

EdwinGuzman:
This very closely follows the requirement in the ticket to remove the conditional logic but I still think it's needed. We just recently added this to the `Banner` component because of the use-case where HTML is stored as a string from either a database, API call, or environment variable.

I think the conditional should still be there so it's backwards compatible, but wrap the `text` in `<Box __css={styles.innerChild}>{text}</Box>`. The issue might also come from how the parent Box wrapper is styled.

If the user passes in HTML that has top margin or padding, the text will be off and we cannot help that. So maybe add a warning against doing so or being careful about what gets passed -- the simpler the better.

bigfishdesign13 (approved):
The spacing now looks good across the board!

EdwinGuzman (approved):

PR #1680, opened 2024-10-04
DSD-1798: Add labelPlacement prop to ProgressIndicator

jackiequach:
Fixes JIRA ticket [DSD-1798](https://newyorkpubliclibrary.atlassian.net/browse/DSD-1798)

- Adds `labelPlacement` prop to the `ProgressIndicator` component to allow users to set the placement of labels relative to the indicator. It is only applicable to the circular variant. 

## How has this been tested?

- Locally in Storybook

-

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[DSD-1798]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-1798?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/ProgressIndicator/ProgressIndicator.tsx
```diff
@@ -95,7 +99,7 @@ export const ProgressIndicator: ChakraComponent<
       if (indicatorType === "circular") {
         // For the small size, since the label won't be visible, we need to add
```
EdwinGuzman: I don't think this comment is true anymore since the circular small size variant does display the label on the right side.

src/components/ProgressIndicator/ProgressIndicator.tsx
```diff
@@ -107,8 +111,14 @@ export const ProgressIndicator: ChakraComponent<
                 </ChakraCircularProgressLabel>
               )}
             </ChakraCircularProgress>
-            {showLabel && size !== "small" && (
-              <Label id={`${id}-label`} htmlFor={id}>
+            {showLabel && (size !== "small" || labelPlacement === "right") && (
```
EdwinGuzman: Can you add a few more examples in the storybook doc section? The current example shows size="small" and labelPlacement="right", but it'd be good to see others such as the regular size and labelPlacement="right".

EdwinGuzman (commented):
Looks good but minor comments regarding documentation. I think the ticket needs to be updated to reflect these changes for the AC.

src/components/ProgressIndicator/ProgressIndicator.tsx
```diff
@@ -93,9 +97,10 @@ export const ProgressIndicator: ChakraComponent<
       // Only display the percentage text for the default size, not in the
       // indeterminate state, and when `showLabel` is true.
       if (indicatorType === "circular") {
-        // For the small size, since the label won't be visible, we need to add
-        // it to the parent component's `aria-label` attribute.
-        if (size === "small") {
+        // For the small size and the label placement is on the bottom,
```
EdwinGuzman: nit: "For the small size and the bottom label placement, ..."

EdwinGuzman (approved):
Looks good, thanks!

src/components/ProgressIndicator/ProgressIndicator.mdx
```diff
@@ -103,6 +103,10 @@ Accessibility Note: when `showLabel` is false, the `aria-label` prop is set in
 the progress element to provide a description of the progress for screen
 readers.
 
+The `labelPlacement` prop can be used to set the placement of a label relative to
```
oliviawongnyc: Nit: I would move this fact "This prop only applies to circular indicator variants" up bc it's an important detail. Then you can delete the last sentence. Additionally, I would make it clear that the options are only "bottom" or "right". Example:

`For circulator indicator variants, the "labelPlacement" prop can be used to move the label to the right of the indicator. By default, the label will be displayed below the indicator. `

src/components/ProgressIndicator/ProgressIndicator.tsx
```diff
@@ -93,9 +97,10 @@ export const ProgressIndicator: ChakraComponent<
       // Only display the percentage text for the default size, not in the
       // indeterminate state, and when `showLabel` is true.
       if (indicatorType === "circular") {
-        // For the small size, since the label won't be visible, we need to add
-        // it to the parent component's `aria-label` attribute.
-        if (size === "small") {
+        // For the small size and the bottom label placement, since the
```
oliviawongnyc: Slightly clearer phrasing: 

```
// If the indicator size is set to small and the labelPlacement is set to bottom,
// the label won't be visible, so we need to add it to the parent
// component's `aria-label` attribute.

Also, should we include this fact in the Storybook docs? That size small and labelPlacement bottom won't show the label?

File: src/components/ProgressIndicator/ProgressIndicator.tsx [hunk 6]
```diff
@@ -93,9 +97,10 @@ export const ProgressIndicator: ChakraComponent<
...
+        // label won't be visible, we need to add it to the parent
+        // component's `aria-label` attribute.
+        if (size === "small" && labelPlacement !== "right") {
```
oliviawongnyc: Nit: I would personally make this `&& labelPlacement === "bottom"` for simpler readability.

src/theme/components/progressIndicator.ts
```diff
@@ -53,9 +53,13 @@ const ProgressIndicator = defineMultiStyleConfig({
         circularContainer: {
           alignItems: "center",
           display: "flex",
-          flexDirection: "column",
+          flexDirection: labelPlacement === "bottom" ? "column" : "row",
           width: "fit-content",
         },
+        circularLabel: {
```
oliviawongnyc, line 87: Maybe this should be `circularLabelRight`? Or something that indicates it is only styles for when the label is on the right hand side.

src/theme/components/progressIndicator.ts
```diff
@@ -53,9 +53,13 @@ const ProgressIndicator = defineMultiStyleConfig({
...
+          marginBottom: 0,
+          marginLeft: "xs",
```
oliviawongnyc: Since there is a marginLeft here, when the label is on the right, should there be a marginTop for then the label is below the indicator?

oliviawongnyc (commented):
Couple nits! Also, what do you think of adding a warning is someone passes a `labelPlacement` if the variant is not circular?

jackiequach:
Made some updates based on additional requirements. For the warning, since there is no way to check if a prop was passed in or set by the default prop value, a warning is sent if the variant is linear and labelPlacement is not the default value (bottom).

src/components/ProgressIndicator/ProgressIndicator.tsx
jackiequach: This logic was removed.

src/theme/components/progressIndicator.ts
```diff
@@ -18,9 +22,34 @@ const { defineMultiStyleConfig, definePartsStyle } =
     "linearPercentage",
   ]);
 
+const getCircularContainerFlexDir = (labelPlacement) => {
+  console.log(labelPlacement);
```
bigfishdesign13: This can be removed.

src/components/ProgressIndicator/ProgressIndicator.tsx
```diff
@@ -79,6 +91,12 @@ export const ProgressIndicator: ChakraComponent<
       );
       finalValue = 0;
     }
+    if (indicatorType === "linear" && labelPlacement !== "bottom") {
+      console.warn(
+        "NYPL Reservoir Progress Indicator: The `labelPlacement` prop has been passed, " +
+          "but the `'linear'` `indicatorType` variant will not use it."
+      );
+    }
```
bigfishdesign13: I think we're going to need to do more for this error warning. Right now, if a consuming app passes both `linear` and `bottom`, no error message will show, leaving the possibility for confusion. What if you removed the default value for `labelPlacement` and created a new `finalLabelPlacement` prop where the value of `finalLabelPlacement` was set to something like this:

```
const finalLabelPlacement = !labelPlacement ? "bottom" : labelPlacement;
```

`finalLabelPlacement` could be passed to `useMultiStyleConfig` and `labelPlacement` could be used for the condition that controls the warning.

```
if (indicatorType === "linear" && labelPlacement) { ... }
```

src/components/ProgressIndicator/ProgressIndicator.stories.tsx
```diff
@@ -147,6 +155,76 @@ export const Labels: Story = {
   ),
 };
 
+export const LabelPlacements: Story = {
+  render: () => (
+    <SimpleGrid columns={1} gap="grid.m">
```
bigfishdesign13: To help with visual clarity, bump the gap value to `"grid.xl"`.

src/components/ProgressIndicator/ProgressIndicator.stories.tsx
```diff
@@ -147,6 +155,76 @@ export const Labels: Story = {
...
+      <HStack spacing="s">
```
bigfishdesign13: Similarly, to help with visual clarity, bump the spacing value to `"l"` for the `<HStack>` components.

bigfishdesign13 (changes requested):
Some minor spacing changes, but also a suggestion for the new warning message.

EdwinGuzman (commented):
Something seems off in the latest round. When checking storybook, all the labels display on the bottom and not in the other places where it should display.

The other thing that should work is the "With Controls" tab since that's also what QA will use to test the label placement.

src/components/ProgressIndicator/ProgressIndicator.tsx
oliviawongnyc: Should we just call the prop `circularIndicatorLabelPlacement`? It could also be a boolean instead that is `circularIndicatorLabelToRight` or something like that, since there is only one option other than the default.

jackiequach: The requirements for the prop were updated, so now there are 4 options for placement bottom, top, right, and left.

oliviawongnyc: Oh I see, thank you @jackiequach! 

One more suggestion: rather than the console warning, what if we instead used a [discriminated union](https://basarat.gitbook.io/typescript/type-system/discriminated-unions) to make the type such that you could only pass the `labelPlacement` prop if the variant was `circular`?

bigfishdesign13: > Should we just call the prop `circularIndicatorLabelPlacement`? It could also be a boolean instead that is `circularIndicatorLabelToRight` or something like that, since there is only one option other than the default.

There is a chance that the label placement functionality is eventually added to the linear variant as well, so let's keep the prop name agnostic for now.

src/components/ProgressIndicator/ProgressIndicator.tsx
```diff
@@ -65,17 +65,18 @@ export const ProgressIndicator: ChakraComponent<
...
       isIndeterminate = false,
-      labelPlacement = "bottom",
+      labelPlacement,
       labelText,
       showLabel = true,
       size = "default",
       value = 0,
       ...rest
     } = props;
+    const finalLabelPlacement = labelPlacement ?? "bottom";
```
bigfishdesign13, line 88: Nice. A `nullish coalescing operator`. I love the name. I've never used this type of operator before, so I had to look it up!

src/components/ProgressIndicator/ProgressIndicator.tsx
```diff
@@ -79,6 +92,12 @@ export const ProgressIndicator: ChakraComponent<
       );
       finalValue = 0;
     }
+    if (indicatorType === "linear" && labelPlacement) {
+      console.warn(
+        "NYPL Reservoir Progress Indicator: The `labelPlacement` prop has been passed, " +
```
bigfishdesign13: Sorry, I just noticed this. We always use the camel case component names in the warnings, so this should be 'NYPL Reservoir ProgressIndicator`.

src/components/ProgressIndicator/ProgressIndicator.tsx
jackiequach: Oh sorry, didn't see this. I will revert my change.

jackiequach: I think the discriminated union is a great idea! It would flag devs at typecheck instead of in the browser.

oliviawongnyc (approved):
Looks great to me!

EdwinGuzman (approved):
Very nice update and good use of the discriminate union type.

bigfishdesign13 (approved):
Very nice!

src/components/ProgressIndicator/ProgressIndicator.tsx
bigfishdesign13: This is no longer an issue. ;)

PR #1683, opened 2024-10-11
DSD-1840: Allowing custom image component in Hero

7emansell:
Fixes JIRA ticket [DSD-1840](https://newyorkpubliclibrary.atlassian.net/browse/DSD-1840)

- Reconfigures the `imageProps` so that a custom image component can be passed into the `Hero`

## How has this been tested?

Locally, release candidates in DC (passing in a Next Image)

### Accessibility Checklist

## Open Questions

- More a note: this one really confused me for a bit because I thought it was intentional for `component` NOT to be one of the props picked from the `ComponentImageProps`. But if it is, we can use the already existing functionality in the DS `Image` that replaces the component. So there ends up being very few code changes, and I'm worried I'm totally missing something..
- Since the Image component is not being used for the background in the "campaign"/"primary" variants, I don't think it makes sense to build a whole new approach to the component that allows you to pass a custom image component for the background, especially because it involves layered styling that would have to be more restrictive if we didn't know what component was being passed in? What do you guys think?

### Checklist:

### Front End Review:

[DSD-1840]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-1840?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Hero/Hero.tsx
```diff
@@ -59,14 +55,15 @@ export interface HeroProps {
...
+  /** Object used to create and render the `Image` component. You can pass `component`
+   * (with its own internal props, which will override) or `src`, `alt`, `id`, `fallBackSrc`, and `onError`.
+   * If `imageProps.alt` is left blank, a warning will be logged to the console and
    * will cause accessibility issues. For `imageProps.src`, it will only work for
-   * the "secondary", "fiftyFifty" and "campaign" `Hero` types; Note: `imageProps.src`
-   * can only be used in conjunction with `backgroundImageSrc` for the "campaign"
-   * `Hero` type. Note: not all `Hero` variations utilize this prop. */
-  imageProps?: HeroImageProps;
+   * the "secondary", "fiftyFifty" and "campaign" `Hero` types. */
+  imageProps?: Pick<
```
EdwinGuzman: This is fine but any reason why it's directly here rather than in the interface?

EdwinGuzman: This is how it renders in storybook which is not better or worse, just different.

Screenshot 2024-10-15 at 6 16 54 PM

src/components/Hero/Hero.tsx
```diff
@@ -120,6 +117,23 @@ export const Hero: ChakraComponent<
...
+          onError={imageProps.onError}
+          src={imageProps.src}
+        />
+      );
+
+      // Using the custom Image component's props to check requirements.
+      if (imageProps.component) {
+        const { src, alt } = imageProps.component.props;
+        Object.assign(imageProps, { src, alt });
+      }
```
EdwinGuzman: Regarding your first note in the PR description, it is confusing and you can use the default behavior for the `component` prop in the `Image` component. It should be so that what you pass as `component` is used so you don't need the other props in `imageProps`.

In the story example
```
const componentImageProps = {
  component: (
    <Image alt={"Custom Image component example"} src={getPlaceholderImage()} />
  ),
};
```

This should propagate to the `Image` component so assigning the prop values to `imageProps` should not be necessary.

EdwinGuzman: The test you wrote passed when I commented this out.

EdwinGuzman (commented):
> Since the Image component is not being used for the background in the "campaign" variant, I don't think it makes sense to build a whole new approach to the component that allows you to pass a custom image component for the background, especially because it involves layered styling that would have to be more restrictive if we didn't know what component was being passed in? What do you guys think?

I agree, it's not necessary to use a custom component for the background image since that's set through CSS.

src/components/Hero/Hero.tsx
7emansell: It works technically but then the DS warnings don't show up if you're missing props

EdwinGuzman: I think that's fine. We should log the warnings if the props we support are not passed in. But if the user passes in their own image component, they're responsible for setting it correctly.

EdwinGuzman (approved):
Let's get more input re: the warning message and background-image css concern.

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 ## Prerelease
 
+### Updates
+
+- Updates image props in the `Hero` component to allow custom components.
```
bigfishdesign13: Nit: to be clearer, update to `...allow custom image components.`

src/components/Hero/Hero.mdx
```diff
@@ -484,6 +485,47 @@ should not be used at the same time.
   language="jsx"
 />
 
+## Custom Image Component
+
+Instead of passing a path for the `imageProps.src` prop, you can pass a custom image component in using the `imageProps.component` prop.
```
bigfishdesign13: Nit: remove `in`

src/components/Hero/Hero.mdx
```diff
@@ -484,6 +485,47 @@ should not be used at the same time.
...
+Note that this only impacts the "campaign" variant, and the custom component will only replace the foreground image, since the background image is set using `background-image`.
+
+<Source
+  code={`
+import Image from "next/image";
```
bigfishdesign13: To hammer the point of using a custom image, what do you think about updating the import statement to something like this?
```
import Image as NextImage from "next/image";
```
And then, of course, updating the reference in the code below.

bigfishdesign13 (approved):
I also agree that we should not worry too much about using a custom image component for the background image unless consuming apps feel they need to programmatically resize or recompress images.

bigfishdesign13:
I did not pull this up in the browser when I reviewed and I am just now seeing that the custom image that is being passed in is not showing in the browser. The `<image>` rendered by the custom `Image` component is in the DOM, but it is hidden and the image URL is not being passed to the `background-image` attribute that is being used to display the image. This pattern mirriors what we discussed about the `FeaturedContent` component.

With that realization, would we gain anything by scraping the image path from the custom component and passing it as the `background-image`? What was the main need for using the NextJS image?

PR #1690, opened 2024-10-24
Release v3.4.1

bigfishdesign13:
### Adds

- Adds the `"decorativeBookBroken"` option (for error pages) to the `Icon` component.
- Adds the `"xxxxlarge"` and `"xxxxxlarge"` size options for the `Icon` component.
- Adds the `"2xlarge"`, `"3xlarge"`, `"4xlarge"`, and `"5xlarge"` sustainable size options for the `Icon` component.

### Updates

- Updates `Select` in the `SearchBar` component to allow long option titles before truncation.
- Updates the `ProgressIndicator` component to add the `labelPlacement` prop to allow for custom placement of a label for circular indicators.
- Updates image props in the `Hero` component to allow custom image components.
- Updates Vite from `5.2.8` to `5.2.14` to fix a security vulnerability.

### Removes

- Removes unused dependencies, two of which were causing npm vulnerabilities.


EdwinGuzman (approved):

src/theme/components/select.ts
```diff
@@ -24,7 +24,7 @@ const select = (labelPosition: string) => ({
   fontSize: "desktop.body.body2",
   minHeight: { base: defaultElementSizes.mobileFieldHeight, md: "auto" },
   paddingTop: "inset.narrow",
-  paddingEnd: "inset.extrawide",
+  paddingEnd: "2.5rem",
```
EdwinGuzman, line 27: @bigfishdesign13 we don't have a design token for this value?

bigfishdesign13: No. And we needed something in-between  what we do offer. `inset.wide` equates to `l` which is `32px` and `inset.extrawide` equates to `xl` which is `48px`. This needed `40px`.

oliviawongnyc: And do we need to update the changelog and Storybook doc version for this or do we feel like this is too small a change?

bigfishdesign13: No, I don't think this is too small a change. This update was included in the `CHANGELOG` and the `Latest Version` number was bumped for the `SearchBar` component.

PR #1691, opened 2024-10-24
[POC] DSD-1851: Updates `Template` to be more responsive using grid

oliviawongnyc:
Fixes JIRA ticket [DSD-1851](https://newyorkpubliclibrary.atlassian.net/browse/DSD-1851)

- Uses one grid rather than two to make a fully responsive template, including responsive gaps, padding, and sidebar
- Uses the `outlineOffset` property to make the focus ring visible even with `overflow: hidden` turned on
- This is merely a POC to help support the work of the new `Template` TAD and is not expected to be merged

## How has this been tested?

- In Storybook
- Plans to create an RC to test in Turbine and other apps

- None that I can think of.

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[DSD-1851]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-1851?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Template/Template.tsx
+++ b/src/components/Template/Template.tsx
@@ -89,8 +89,7 @@ const Template: ChakraComponent<
 const TemplateAboveHeader: React.FC<any> = (
   props: React.PropsWithChildren<TemplateProps>
 ) => {
-  const styles = useStyleConfig("TemplateBreakout", {});
-  return <Box __css={styles}>{props.children}</Box>;
+  return <Box>{props.children}</Box>;
 };
 
 /**
@@ -104,8 +103,7 @@ const TemplateHeader: React.FC<any> = ({
   children,
   renderHeaderElement = true,
 }: React.PropsWithChildren<TemplateHeaderProps>) => {
-  const styles = useStyleConfig("TemplateBreakout", {});
-  let headerElement = <Box __css={styles}>{children}</Box>;
+  let headerElement = <Box>{children}</Box>;
 
   // The user wants to render the `header` HTML element.
   if (renderHeaderElement) {
@@ -120,11 +118,7 @@ const TemplateHeader: React.FC<any> = ({
         );
       }
     });
-    headerElement = (
-      <Box as="header" __css={styles}>
-        {children}
-      </Box>
-    );
+    headerElement = <Box as="header">{children}</Box>;
   }
   return headerElement;
 };
@@ -138,8 +132,7 @@ const TemplateHeader: React.FC<any> = ({
 const TemplateBreakout: React.FC<any> = (
   props: React.PropsWithChildren<TemplateProps>
 ) => {
-  const styles = useStyleConfig("TemplateBreakout", {});
-  return <Box __css={styles}>{props.children}</Box>;
+  return <Box>{props.children}</Box>;
 };
 
 /**
@@ -242,11 +235,7 @@ const TemplateContentPrimary: React.FC<any> = (
 const TemplateContentSidebar: React.FC<any> = (
   props: React.PropsWithChildren<TemplateContentProps>
 ) => {
-  const { sidebar } = props;
-  const styles = useStyleConfig("TemplateContentSidebar", {
-    variant: sidebar,
-  });
-  return <Box __css={styles}>{props.children}</Box>;
+  return <Box>{props.children}</Box>;
 };
 
 /**
@@ -259,8 +248,8 @@ const TemplateFooter: React.FC<any> = ({
   children,
   renderFooterElement = true,
 }: React.PropsWithChildren<TemplateFooterProps>) => {
-  const styles = useStyleConfig("TemplateBreakout", {});
-  let footerElement = <Box __css={styles}>{children}</Box>;
+  // const styles = useStyleConfig("TemplateBreakout", {});
+  let footerElement = <Box>{children}</Box>;
 
   // The user wants to render the `footer` HTML element.
   if (renderFooterElement) {
@@ -274,11 +263,7 @@ const TemplateFooter: React.FC<any> = ({
         );
       }
     });
-    footerElement = (
-      <Box as="footer" __css={styles}>
-        {children}
-      </Box>
-    );
+    footerElement = <Box as="footer">{children}</Box>;
   }
   return footerElement;
 };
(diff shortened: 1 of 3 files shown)


src/components/Template/Template.tsx
```diff
@@ -242,11 +235,7 @@ const TemplateContentPrimary: React.FC<any> = (
 const TemplateContentSidebar: React.FC<any> = (
   props: React.PropsWithChildren<TemplateContentProps>
 ) => {
-  const { sidebar } = props;
```
EdwinGuzman, line 245: No need to act on this note for this PR, but in the TAD you should mention that the template children components are basically now all divs. When you do update this PR, it should be one generic "TemplateChild" (but better named) component instead of `TemplateBreakout`, `TemplateContentSidebar`, etc.

src/theme/components/template.ts
```diff
@@ -1,103 +1,75 @@
 import { defineStyleConfig } from "@chakra-ui/react";
 import { defineStyle } from "@chakra-ui/system";
-/**
- * Grid layout based on https://www.joshwcomeau.com/css/full-bleed/
```
EdwinGuzman, line 4: Add this in the TAD to explain the context and the current implementation in prod.

src/theme/components/template.ts
```diff
@@ -1,103 +1,75 @@
...
-    },
-    right: {
-      gridColumn: { base: "1", md: "2" },
+    overflow: "hidden",
+    /**  We need `overflow: hidden` to ensure larger elements passed to the template
+     * appear correctly, but it also cuts off the default focus ring. The following
+     * offset property ensures the focus ring is fully visible. */
+    "*": {
+      _focus: {
+        outlineOffset: "-2px !important",
```
EdwinGuzman, line 64: Two things with this:
- It works but the focus ring is much closer to every focusable element now. This is a change outside the scope of the template and grid update. We'd have to all be okay with this update independently of the grid/template update.
- This might also work better as a global rule for all elements instead of declaring it here (e.g. set it in the global.ts file)

EdwinGuzman (commented):
Just some notes. It's not a perfect pixel-to-pixel update but as a POC it's coming along nicely.

oliviawongnyc:
> Just some notes. It's not a perfect pixel-to-pixel update but as a POC it's coming along nicely.

@EdwinGuzman, pixel perfect compared to what? Marty's PR or what we have now? It shouldn't be pixel perfect compared to what we have now given it includes a responsive sidebar, padding, and gaps.

EdwinGuzman:
> pixel perfect compared to what?

To what we have now. There are some changes but I think this difference is crucial -- the alignment of the content:

Screenshot 2024-10-28 at 12 08 38 PM

In prod, the "Content Top" and "Content Primary" align with the Hero. I think it's necessary for alignment but not something to fix right now since the main concern is the underlying foundation.

oliviawongnyc:
> > pixel perfect compared to what?
> 
> To what we have now. There are some changes but I think this difference is crucial -- the alignment of the content:
> 
> Screenshot 2024-10-28 at 12 08 38 PM
> In prod, the "Content Top" and "Content Primary" align with the Hero. I think it's necessary for alignment but not something to fix right now since the main concern is the underlying foundation.

@EdwinGuzman ahhh, got it. But this may actually concern the underlying foundation. This problem occurs because Hero is now outside of the grid so I'll have to think about that.

src/theme/components/template.ts
```diff
@@ -1,103 +1,75 @@
...
-    gridTemplateColumns: "1fr",
-    paddingY: 0,
-    paddingX: "s",
-    gap: "grid.l",
+    display: "grid",
+    gridTemplateColumns: {
+      base: "repeat(1, minmax(100px, 1fr))",
+    },
+    gap: responsiveGap,
+    padding: responsivePadding,
```
bigfishdesign13, line 26: Also add the following:
```
maxWidth: "1280px"
```

This will also need something to center this element on the page. Maybe that means adding the following to `Template`?
```
display: "flex",
justifyContent: "center",
```

Or whatever works best for the overall structure of the page.

src/theme/components/template.ts
```diff
@@ -1,103 +1,75 @@
...
+        xl: "1fr 3fr",
+      },
     },
     right: {
-      gridTemplateColumns: { md: "1fr 255px" },
+      gridTemplateColumns: {
+        md: "repeat(2, minmax(100px, 1fr))",
+        lg: "2fr 1fr",
+        xl: "3fr 1fr",
+      },
```
bigfishdesign13, line 44: This is so much more straight forward than what I put together in my POC using all 12 columns. Easier on the eyes and brain. However, I am still curious about the widths of the columns and gaps and how they align with expectations. Once the `responsiveGap` and `responsiveMargin` variables are updated and implemented, I will compare the widths in this POC with the widths in my POC. If they are in alignment, then I don't know what I was seeing that caused me to implement a full 12-column grid.

src/theme/components/template.ts
bigfishdesign13: Right, any changes to the focus styles should go through UX exploration first with an aim toward updating the style for all elements. That said, I don't feel a change to the focus style is the way to go with this. There must be a better way to handle the overflow issues, whether that means updating the styles on `Template` or on `Table`.

src/components/Template/Template.tsx
bigfishdesign13: As we look to enable better composibility, do we need three separate divs at the top of the Template? I like the idea of steering consuming apps in the right direction with three child components that we offer, but that strecture could also be easily recreated with recommended patterns.

oliviawongnyc: Yes, I agree it feels unnecessary to have all of these components at this point.

src/theme/components/template.ts
oliviawongnyc: I removed this. It seems (?) to be working now without it, but would like someone else to check and make sure it looks that way to them too.

bigfishdesign13: @oliviawongnyc Can you push those upates?

PR #1697, opened 2024-11-05
DSD-1852: SubNav Menu Component

gonuguntla:
Fixes JIRA ticket (https://newyorkpubliclibrary.atlassian.net/browse/DSD-1852)

- Implemented subnav 

## How has this been tested?

- local Storybook

-

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:


src/components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,400 @@
...
+   */
+  primaryActions: ({
+    highlightColor,
+    actionBackgroundColor,
+    selectedItem,
+  }: {
+    highlightColor?: string;
+    actionBackgroundColor?: string;
+    selectedItem?: string;
+  }) => React.ReactNode;
```
bigfishdesign13: Why are these props passed into `primaryActions` and `secondaryActions`?

src/components/SubNav/SubNav.stories.tsx
```diff
@@ -0,0 +1,493 @@
...
+    actionBackgroundColor: undefined,
+    highlightColor: undefined,
+    id: "sub-nav-id",
+    selectedItem: undefined,
+  },
+
+  render: (args: any) => (
+    <SubNav
+      {...args}
+      primaryActions={({ highlightColor, actionBackgroundColor }) => (
```
bigfishdesign13: I need to know more about why `highlightColor` and `actionBackgroundColor` are being passed here. I would like to simplify this so the prop value can be written like this:
```
primaryActions={
  <>
    ...
  </>
}
```

src/components/SubNav/SubNav.mdx
```diff
@@ -0,0 +1,44 @@
...
+
+<Meta of={SubNavStories} />
+
+# SubNav
+
+# TODO update the version
+| Component Version | DS Version |
+| ----------------- | ---------- |
+| Added             | `3.4.0`   |
+| Latest            | `3.4.0`   |
```
bigfishdesign13: These can both be set to `Prerelease`.

src/components/SubNav/SubNav.stories.tsx
bigfishdesign13: @EdwinGuzman This is where I had concerns. Can the `primaryActions` and `secondaryActions` props be simplified for end-users?

src/components/SubNav/SubNav.tsx
EdwinGuzman: It should be `(SubNavButton | SubNavLink)[]` but we want to move away from mapping and cloning children to reduce creating new components on the client. This render prop pattern allows the `SubNavButton` or `SubNavLink` children to be styled appropriately based on `SubNav`

src/components/SubNav/SubNav.stories.tsx
EdwinGuzman: Because both props are render prop functions, a function has to be passed with the prop values that are passed from `SubNav`. It can't be simpler than this unfortunately.

EdwinGuzman: One other option can be 
```
<SubNav
  primaryActions={(args) => (
    <>
      <SubNavButton
          id="primary-sub-nav-button-1"
          buttonType="text"
          isOutlined
          {...args}
        >
          Label Text
        </SubNavButton>
    </>
  )}
/>
```

I like being explicit but this should  work as well.

src/theme/components/subnav.ts
```diff
@@ -0,0 +1,169 @@
...
+  textDecoration: "none !important",
+  transition: "background-color 0.2s, color 0.2s !important",
+});
+
+const SubNav = defineMultiStyleConfig({
+  baseStyle: definePartsStyle(
+    ({ backgroundColor, highlightColor, isOutlined }: SubNavStyleProps) => {
+      const defaultLabelColor = "ui.typography.body";
+      const highlightOrLinkColor = highlightColor
+        ? `${highlightColor} !important`
```
EdwinGuzman: I haven't looked too closely to figure out why, but why are there so many `!important` flags? Try not to use them unless if there's a specific reason why. I would understand one or two but there are too many in this file.

gonuguntla: I tried to avoid using !important, but in this case, it's necessary due to multiple components(Buttons, Links, Icons, primary actions and secondary actions) where the colors need to be applied.

src/components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,379 @@
...
+      if (React.isValidElement(child) && isIconComponent(child.type)) {
+        return child; // Only render the icon on smaller screens
+      }
+      return null;
+    });
+  }
+  return children; // If no icon, render text
+};
+
+export const SubNavButton: React.FC<React.PropsWithChildren<any>> = chakra(
```
EdwinGuzman: Since we want to ignore custom styles that might be passed to `SubNavButton` and `SubNavLink`, the `chakra()` function should not be used here. If it's removed, style props and the `sx` and `__css` props won't be available which is what we want.

src/components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,379 @@
...
+    if (React.isValidElement(child)) {
+      const childType = child.type as React.ComponentType<any>;
+      return isIconComponent(childType);
+    }
+    return false;
+  });
+};
+
+// Function to render content based on the screen size
+export const renderContent = (
```
EdwinGuzman: I understand why this was done but I don't think we need `renderContent`, `hasIcon`, or `isIconComponent`. From the spec doc, the best practice and recommendation is that  the _user_ passes in the correct link or button with icon for the appropriate device width. They need to consider this and it's their responsibility and not a responsibility of `SubNav`. So these functions and checks can be removed.

src/components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,379 @@
...
+        const { scrollWidth, clientWidth } = refCurrent;
+        setShowRightFade(scrollWidth > clientWidth); // Show fade if content is wider than the visible area
+      }
+
+      return () => {
+        if (refCurrent) {
+          refCurrent.removeEventListener("scroll", handleScroll);
+        }
+      };
+    }, []);
```
EdwinGuzman: There's a linting error here. `handleScroll` should be added here in the dependency array.

src/components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,379 @@
...
+
+    // Function to ensure only valid CSS properties are passed
+    const fadeEffectStyles = (styleObject) => {
+      const { accentColor, ...validStyles } = styleObject; // Exclude non-CSS properties
```
EdwinGuzman: I'm not sure I understand what `accentColor` is. It's not defined in `styles.fadeEffect` so this function doesn't change the input.

src/components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,318 @@
...
+  highlightColor?: highlightColors;
+  /**
+   * Optional unique ID for accessibility, allowing other components
+   * to reference this element.
+   */
+  id?: string;
+  /** Optional string used to identify and highlight an item. The value should
+   * match the id associated with one of the items.
+   */
+  selectedItem?: string;
```
oliviawongnyc: I think we should keep this here, but we aren't currently using it at all in the SubNav component, only in the children. I think we need to have selectedItem passed to the parent and for the parent to somehow pass it to the child, rather than passing this prop to the child directly, because there should only be one selectedItem and right now, the consuming app could pass selectedItem prop to every child.

src/components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,318 @@
...
+          >
+            <List
+              type="ul"
+              m="0"
+              sx={{ ...styles.scrollableButtons, ...styles.primaryActions }}
+              ref={scrollableRef}
+              inline
+              noStyling
+            >
+              <li id="primary-actions">{primaryActions}</li>
```
oliviawongnyc: I feel like each button/link should be an li from an accessibility standpoint rather than putting the whole group of buttons/links in one li.

src/components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,318 @@
...
+              {showRightFade && (
+                // Explicitly cast styles.fadeEffect to CSSProperties
+                <div style={styles.fadeEffect as CSSProperties} />
```
oliviawongnyc: I'm getting a warning because all children of List should be an li. Is there another way to do this to ensure that all direct children are li's?

src/components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,318 @@
...
+
+export interface SubNavProps {
+  /** Additional class name for the `SubNav` component. */
+  className?: string;
+  /**
+   * Primary actions displayed on the left side of the SubNav.
+   * Use SubNavButton and SubNavLink components, which mirror
+   * the DS Button and Link.
+   */
+  primaryActions: React.ReactNode;
```
oliviawongnyc: Maybe we should go back to the idea of this being an array of buttons or links. Then we can pass the selectedItem into the SubNav (parent) and then pass an isSelected prop into the button/link with a matching id. @EdwinGuzman, was your issue with this idea that you didn't want to loop through them?

src/components/SubNav/SubNav.tsx
oliviawongnyc: Also, what do you think about making this prop selectedId rather than selectedItem?

oliviawongnyc (commented):
Looking much cleaner and easier to understand (at least for me 😄), Deepika! A few inita comments to start, but I think there is a reason we didn't want to simply loop through an array of sub nav children.

src/components/SubNav/SubNav.tsx
bigfishdesign13: I just spoke with Deepika. We are going to remove the `selectedItem` prop altogether and instead add a new `isSelected` prop that can be passed to `SubNavButton` and `SubNavLink`, similar to `isOutlined`. This approach would allow more than one child to be marked as selected, so the onus would be on consuming apps to configure the child components accordingly and only mark one child as selected. This will be noted in the docs.

src/components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,312 @@
...
+      <Button
+        id={id}
+        buttonType="text"
+        className={isSelected ? "selectedItem" : ""}
+        sx={{ ...childrenStyles.outLine }}
+        screenreaderOnlyText={screenreaderOnlyText}
+      >
+        {children}
+      </Button>
+    </li>
```
bigfishdesign13: @oliviawongnyc and @EdwinGuzman – The addition of `<li>` elements here makes the semantic DOM structure better and improves a11y. The browser DOM inspector recognizes the `<li>` elements and the list element is rendered correctly in the browser. However, the DS `List` component is still throwing the error that Olivia noted in another comment:

```
NYPL Reservoir List: Direct children of `List` (ul) must be `<li>`s.
```

Any thoughts on why this is happening? Does this return value need a different type definition?

src/components/SubNav/SubNav.mdx
```diff
@@ -0,0 +1,302 @@
...
+## Accessibility
+
+Only one `SubNav` component should be rendered on a page. This is because only
+one HTML `<nav>` element with an `aria-label` attribute value of
+`"Sub-navigation menu"` should be rendered on a page. The DS `SubNav` component
+renders the `<nav>` HTML landmark, so only one component must be rendered on a
+page.
+
+Internally, buttons and links are organized in an unordered list (`<ul>`)
+element. If applicable, the current page link is denoted by an `aria-current`
```
EdwinGuzman, line 53: Is adding `aria-current` a responsibility of the DS? If so, this should be implemented in the `SubNavLink` component through the `isSelected` prop.

src/components/SubNav/SubNav.test.tsx
```diff
@@ -0,0 +1,284 @@
...
+            </SubNavButton>
+            <SubNavButton id="primary-sub-nav-button-2" buttonType="text">
+              Label Text
+            </SubNavButton>
+            <SubNavButton id="primary-sub-nav-button-3" buttonType="text">
+              Label Text
+            </SubNavButton>
+          </>
+        }
+        secondaryActions={
```
EdwinGuzman: Did you mean to include `secondaryActions`? It's not in the test description.

src/components/SubNav/SubNav.test.tsx
```diff
@@ -0,0 +1,284 @@
...
+              />
+            </SubNavLink>
+          </>
+        }
+      />
+    );
+    expect(await axe(container)).toHaveNoViolations();
+  });
+
+  it("passes axe accessibility test with secondary actions", async () => {
```
EdwinGuzman: Did you mean to include `primaryActions`? It's not in the test description.

src/components/SubNav/SubNav.test.tsx
```diff
@@ -0,0 +1,284 @@
...
+            <SubNavButton id="primary-button-2">Primary Button 2</SubNavButton>
+          </>
+        }
+      />
+    );
+
+    const selectedButton = await screen.findByRole("button", {
+      name: /Primary Button 1/i,
+    });
+    expect(selectedButton).toHaveTextContent("Primary Button 1");
```
EdwinGuzman: This test verifies that the component is rendered but not that it's highlighted. You added an extra class name when `isSelected` is passed so check for that instead.

src/components/SubNav/SubNav.test.tsx
```diff
@@ -0,0 +1,284 @@
...
+  });
+
+  it("passes axe accessibility test with primary actions", async () => {
```
EdwinGuzman: The following seem like duplicate tests so you can remove. They're similar as the initial accessibility tests at the top.

src/components/SubNav/SubNav.test.tsx
```diff
@@ -0,0 +1,284 @@
...
+  });
+
+  it("applies selectedItem class when isSelected is true", () => {
+    render(
+      <SubNavButton id="sub-nav-button" isSelected>
+        Selected Button
+      </SubNavButton>
+    );
+    const button = screen.getByText("Selected Button");
+    expect(button).toHaveClass("selectedItem"); // Should have "selectedItem" class
```
EdwinGuzman, line 143: This is great. This is what you should have for the test on line 177

src/components/SubNav/SubNav.tsx
EdwinGuzman: Storybook is loading all component pages so this _might_ be coming from a different story example. Does this happen if you view the page in isolation?

src/components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,312 @@
...
+  href,
+  screenreaderOnlyText = "",
+}) => {
+  const childrenStyles = useMultiStyleConfig("SubNavChildren", {
+    isOutlined: isOutlined,
+  });
+
+  return (
+    <li>
+      <Link
```
EdwinGuzman, line 152: I think you need an additional attribute for `aria-current="page"` if `isSelected` is passed. I'm assuming we always want this. Is this true @bigfishdesign13 ?

src/components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,312 @@
...
+    isOutlined: isOutlined,
+  });
+
+  return (
+    <li>
+      <Button
+        id={id}
+        buttonType="text"
+        className={isSelected ? "selectedItem" : ""}
+        sx={{ ...childrenStyles.outLine }}
```
EdwinGuzman: You should be able to just have `sx={childrenStyles.outLine}`. It's the same as `{ ...childrenStyles.outLine }`

src/components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,312 @@
...
+    <li>
+      <Link
+        key={id}
+        id={id}
+        type="action"
+        href={href}
+        isUnderlined={false}
+        screenreaderOnlyText={screenreaderOnlyText}
+        className={isSelected ? "selectedItem" : ""}
+        sx={{ ...childrenStyles.outLine }}
```
EdwinGuzman: You should be able to just have sx={childrenStyles.outLine}. It's the same as { ...childrenStyles.outLine }

src/components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,312 @@
...
+
+      const backgroundColor =
+        highlightColor !== undefined ? actionBackgroundColor : undefined;
+
+      const styles = useMultiStyleConfig("SubNav", {
+        backgroundColor: backgroundColor,
+        highlightColor: highlightColor,
+      });
+
+      const handleScroll = useCallback(() => {
```
EdwinGuzman: What's the use of `useCallback` here?

src/components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,312 @@
...
+            // Show right fade only if not at the end
+            const atRightEdge = scrollLeft + clientWidth >= scrollWidth;
+            setShowRightFade(!atRightEdge);
+          } else {
+            setShowRightFade(false); // Hide when scrolling right
+          }
+        }
+      }, [lastScrollLeft, setLastScrollLeft, setShowRightFade]);
+
+      useEffect(() => {
```
EdwinGuzman: This is great. This should be a reusable hook instead of declaring it inside this component. Doing so helps separate the component UI with the functionality/styles. This fading style is also used in the `Table` so it might overlap there. For now, you can create a hook file and only apply it to this component.

src/components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,312 @@
...
+                  ...styles.primaryActions,
+                }}
+                ref={scrollableRef}
+                inline
+                noStyling
+              >
+                {primaryActions}
+                {showRightFade && (
+                  // Explicitly cast styles.fadeEffect to CSSProperties
+                  <div style={styles.fadeEffect as CSSProperties} />
```
EdwinGuzman: Ah, @bigfishdesign13 I think this is where the list warning was coming from. This is a div and should be wrapped in an li.

EdwinGuzman: @gonuguntla Wrap the div in an `li`. Also, you can convert this to `<Box sx={styles.fadeEffect} />` to remove the type casting for `CSSProperties`.

src/components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,312 @@
...
+              </List>
+              <List
+                noStyling
+                inline
+                type="ul"
+                sx={{ ...styles.secondaryActions, ...styles.listCleanup }}
+                m="0"
+                width="fit-content"
+              >
+                {secondaryActions ? secondaryActions : null}
```
EdwinGuzman: I don't think empty `ul`s should render. So you might need to wrap the entire `<List ...` with the logic check for `secondaryActions ? ....`.

What do you think @bigfishdesign13 @oliviawongnyc ?

src/components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,312 @@
...
+            (child: React.ReactElement) => {
+              if (child.type !== SubNavButton && child.type !== SubNavLink) {
+                console.warn(
+                  `NYPL Reservoir SubNav: An element that is not a SubNavButton or SubNavLink component has been passed in the \`${propName}\` prop. That element will be ignored.`
+                );
+                return null;
+              }
+            }
+          );
+        }
```
EdwinGuzman: In theory this is good but this only logs the warning. This does not actually ignore the element. In the screenshot I passed a DS `Button` and an HTML anchor link:

Screenshot 2024-11-13 at 7 21 17 PM


If we actually want to ignore the element, we'll have to do more work which I think we can not do for this round. What do you think about this being the responsibility of the consuming application? @bigfishdesign13  @oliviawongnyc ?

EdwinGuzman (changes requested):
@gonuguntla thank you! The latest update looks much cleaner in terms of concise code. I left some comments but the biggest one is removing the logic in the component that handles the scroll check and fading styles.

I know you have some dependencies you need but it's possible to extract out all the logic and return the data you need. Something along the lines of 

```
const { scrollableRef, showRightFade } = useScrollFadeStyles();
```

Let me know if you need help with this.

src/components/SubNav/SubNav.mdx
bigfishdesign13: The DS does handle this for the `AlphabetFilter`, `Breadcrumbs`, and `Pagination` components, so yes, I think we should also do it here in conjunction with `isSelected`.

src/components/SubNav/SubNav.tsx
bigfishdesign13: Yes, each individual story throws the error twice, by default. When the viewport is small enough to cause scrolling, each story throws the error four times.

src/components/SubNav/SubNav.tsx
bigfishdesign13: Yes, that is true. Something like this...
```
aria-current={isSelected ? "page" : null}
```

src/components/SubNav/SubNav.tsx
bigfishdesign13: We do a similar type check of child components in the `ButtonGroup` and `List` components, so yes, I think we can continue that pattern here. But let's only throw a warning instead of ignoring the elements.

@deepikagonuguntla Update the warning to the following:
```
"NYPL Reservoir SubNav: An element that is not a 
`SubNavButton` or `SubNavLink` component has 
been passed in the \`${propName}\` prop. That 
element may not work properly."
```

src/components/SubNav/SubNav.tsx
bigfishdesign13: @deepikagonuguntla For better semantics and a11y, it would be best to move this element outside the `List` component. You should be able to do that by adding another `div` or `Box` around this `List` and then attaching the fade to that new wrapper element.

src/components/SubNav/SubNav.tsx
bigfishdesign13: Yes, this condition should be moved up one level to wrap the `List` element and the `List` component should be rendered only when `secondaryActions` is `true`.

src/hooks/useScrollFadeStyles.mdx
```diff
@@ -0,0 +1,16 @@
...
+
+# useScrollFadeStyles
+
+| Hook Version | DS Version |
+| ------------ | ---------- |
+| Added        | `Prerelease`   |
+| Latest       | `Prerelease`   |
+
+
+ The `useScrollFadeStyles` hook manages the visibility of a right-side fade effect
```
bigfishdesign13: Can you show code examples that include the `div` element that is necessary?

src/components/SubNav/SubNav.tsx
```diff
@@ -287,20 +235,23 @@ export const SubNav: ChakraComponent<
               >
                 {primaryActions}
                 {showRightFade && (
-                  // Explicitly cast styles.fadeEffect to CSSProperties
-                  <div style={styles.fadeEffect as CSSProperties} />
+                  <li>
+                    <Box sx={styles.fadeEffect} />
+                  </li>
```
bigfishdesign13: This is not working well. The fade element is moving with the scroll. Also, the action items are flickering when scrolling. Can you wrap the `List` in a `div` and then move the `Box` up one level so it is next to the `List`? That would be more in alignment with how we are dong this for the `Table` component.

I think you will also need to move the styles that power the scrolling.

bigfishdesign13 (changes requested):
Most of the updates look good. The fade element still needs some work.

src/hooks/useScrollFadeStyles.mdx
gonuguntla: I will add it Marty

src/components/SubNav/SubNav.stories.tsx
```diff
@@ -0,0 +1,569 @@
...
+    id: "subnav-with-controls",
+  },
+
+  render: (args: any) => (
+    <SubNav
+      {...args}
+      primaryActions={
+        <>
+          <SubNavButton
+            buttonType="text"
```
EdwinGuzman: You should not pass `buttonType="text"` since this is already set internally in the `SubNavButton` component. So you can remove this prop from all examples.

src/components/SubNav/SubNav.stories.tsx
```diff
@@ -0,0 +1,569 @@
...
+            onClick={() => subNavButtonOnClick("from primary-subnav-button-2")}
+          >
+            Action
+          </SubNavButton>
+          <SubNavLink
+            id="subnav-with-controls-primary-action-3"
+            screenreaderOnlyText="for the NYPL Research Catalog"
+            href="#link3"
+          >
+            Action
```
EdwinGuzman: Can you keep "Action" as the label text for buttons, but update the text to "Link" for links? Just to differentiate a bit between buttons and links.

src/components/SubNav/SubNav.stories.tsx
```diff
@@ -0,0 +1,569 @@
...
+        <>
+          <SubNavButton
+            buttonType="text"
+            id="subnav-selected-babies-toddlers"
+            onClick={() =>
+              subNavButtonOnClick("from subnav-selected-babies-toddlers")
+            }
+            screenreaderOnlyText="Education for babies and toddlers"
+          >
+            Babies &amp; Toddlers
```
EdwinGuzman: Nitpick: "Toddlers" should be lowercase

src/components/SubNav/SubNav.stories.tsx
EdwinGuzman: This is also not a prop in the component so it's not actually setting the type inside of `SubNavButton`.

src/components/SubNav/SubNav.stories.tsx
```diff
@@ -0,0 +1,569 @@
...
+  ),
+};
+
+export const SelectedItem: Story = {
+  args: {
+    className: undefined,
+    actionBackgroundColor: undefined,
+    highlightColor: undefined,
+    id: "subnav-selected",
+  },
```
EdwinGuzman: We only use `args` in the `WithControls` story. So you can remove the args object from other stories and explicitly set the prop values. For example, from this
```
args: {
    className: undefined,
    actionBackgroundColor: undefined,
    highlightColor: undefined,
    id: "subnav-selected",
  },
  render: (args: any) => (
    <SubNav
      {...args}
      primaryActions={
```

to

```
  render: (args: any) => (
    <SubNav
      id="subnav-selected"
      primaryActions={
```

src/components/SubNav/SubNav.stories.tsx
```diff
@@ -0,0 +1,569 @@
...
+            My account
+          </SubNavLink>
+        </>
+      }
+    />
+  ),
+};
+
+export const OutlinedItem: Story = {
+  args: {
```
EdwinGuzman: Remove the args as explained above.

src/components/SubNav/SubNav.stories.tsx
```diff
@@ -0,0 +1,569 @@
...
+            My account
+          </SubNavLink>
+        </>
+      }
+    />
+  ),
+};
+
+export const UsingIcons: Story = {
+  args: {
```
EdwinGuzman: Remove the args as explained above.

src/components/SubNav/SubNav.stories.tsx
```diff
@@ -0,0 +1,569 @@
...
+            My account
+          </SubNavLink>
+        </>
+      }
+    />
+  ),
+};
+
+export const UsingIconsMobile: Story = {
+  args: {
```
EdwinGuzman: Remove the args as explained above.

src/components/SubNav/SubNav.stories.tsx
```diff
@@ -0,0 +1,569 @@
...
+            <Icon name="actionIdentityFilled" size="medium" />
+          </SubNavLink>
+        </>
+      }
+    />
+  ),
+};
+
+export const CustomColors: Story = {
+  args: {
```
EdwinGuzman: Remove the args as explained above.

src/components/SubNav/SubNav.stories.tsx
```diff
@@ -0,0 +1,569 @@
...
+            </SubNavLink>
+          </>
+        }
+      />
+    </>
+  ),
+};
+
+export const HorizontalScrolling: Story = {
+  args: {
```
EdwinGuzman: Remove the args as explained above.

src/components/SubNav/SubNav.stories.tsx
```diff
@@ -0,0 +1,569 @@
...
+            </SubNavLink>
+          </>
+        }
+      />
+    </>
+  ),
+};
+
+export const Placement: Story = {
+  args: {
```
EdwinGuzman: Remove the args as explained above.

src/components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,263 @@
...
+  primaryActions: React.ReactNode;
+  /**
+   * Secondary actions displayed on the right side of the SubNav.
+   * Use SubNavButton and SubNavLink components, which mirror
+   * the DS Button and Link.
+   */
+  secondaryActions?: React.ReactNode;
+}
+
+export const SubNavButton: React.FC<React.PropsWithChildren<any>> = ({
```
EdwinGuzman: You need an interface for `SubNavButton`. Once you create it, replace `any` with the interface.

src/components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,263 @@
...
+        aria-current={isSelected ? "page" : null}
+        screenreaderOnlyText={screenreaderOnlyText}
+      >
+        {children}
+      </Button>
+    </li>
+  );
+};
+
+export const SubNavLink: React.FC<React.PropsWithChildren<any>> = ({
```
EdwinGuzman: You need an interface here to replace `any`. This one also shares a lot of properties with the new interface for `SubNavButton` except for href. So you can use typescripts `extends` to add the `href` property.

src/components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,263 @@
...
+    (props, _ref?) => {
+      const {
+        className,
+        actionBackgroundColor,
+        highlightColor,
+        primaryActions,
+        secondaryActions,
+      } = props;
+
+      const { scrollableRef, showRightFade } = useScrollFadeStyles();
```
EdwinGuzman, line 191: 👍

src/components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,263 @@
...
+        backgroundColor: backgroundColor,
+        highlightColor: highlightColor,
+      });
+
+      return (
+        <Box as="nav" aria-label="Sub-navigation menu" __css={styles.base}>
+          <Box __css={styles.container}>
+            <Flex
+              alignItems="center"
+              className={className}
```
EdwinGuzman: Move the classname to the parent `Box` wrapper that renders the `nav`

src/components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,263 @@
...
+              gap="1rem"
```
EdwinGuzman: The equivalent design token value is "s". So this can be `gap="s"`.

EdwinGuzman: You can find more of the values in the `spacing.ts` file.

src/components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,263 @@
...
+              justify="space-between"
+            >
+            <div style={styles.primaryList}>
```
EdwinGuzman: I think this is throwing a linting or typescript error. Convert the div into a `Box`, and you can also replace `style` with `sx`.

src/theme/components/subnav.ts
EdwinGuzman: This was true before but now the component is simpler. Can you try removing more `!important` flags to make sure they are not needed?

src/theme/components/subnav.ts
```diff
@@ -0,0 +1,207 @@
...
+        container: {
+          maxWidth: "1280px",
+          // px: "s",
+          px: { base: "0", md: "xs" },
+          width: "100%",
+        },
+        listCleanup: {
+          // p: "xs",
+          p: { base: "s", md: "xs" },
+          gap: "xs",
```
EdwinGuzman: `gap` is not a rule for the `ul` element so you can remove this

src/theme/components/subnav.ts
```diff
@@ -0,0 +1,207 @@
...
+          width: "50px",
+          background:
+            "linear-gradient(to left, rgba(255, 255, 255, 1) 0%, rgba(255, 255, 255, 0) 100%)",
+          pointerEvents: "none",
+          zIndex: 1,
+        },
+        ul: {
+          margin: "0 !important",
+        },
+        primaryList: {
```
EdwinGuzman, line 183: I don't think you're using this.

src/theme/components/subnav.ts
```diff
@@ -0,0 +1,207 @@
...
+          display: "flex",
+          justifyContent: "center",
+        },
+        container: {
+          maxWidth: "1280px",
+          // px: "s",
+          px: { base: "0", md: "xs" },
+          width: "100%",
+        },
+        listCleanup: {
```
EdwinGuzman: You can refactor and update this. There are two ways I'm thinking but you can update this if you think of another way.

1. You are using `...styles.listCleanup` in both `<List ...>` components. Instead of having to pass the styles in this manner, you can rename `listCleanup` as `ul`. This will target all lists in the component, but there are only two so this is ok.

2. You can set these rules in an object and add them to `primaryActions` and `secondaryActions`.

```
const ulStyles = {
  p: { base: "s", md: "xs" },
  li: {
    marginEnd: "unset",
  },
  margin: "0 !important",
};

// ...

primaryActions: {
  ...ulStyles,
  width: "100%",
  button: {
// etc

secondaryActions: {
  ...ulStyles,
  width: "fit-content",
  whiteSpace: "nowrap",
// etc

Comment (EdwinGuzman) [reply src/theme/components/subnav.ts]: Note, I moved the rule from line 181 inside `ulStyles` as an example.

**Review by EdwinGuzman — CHANGES_REQUESTED** [review]
The update looks good and thank you for breaking out the logic into its own hook. This perfectly separates the logic with the UI for the `SubNav` component. Also, your `useScrollFadeStyles` can be used by other developers. I left comments for minor code updates.

File: src/theme/components/subnav.ts (hunk 47 above)
Comment (EdwinGuzman) [reply src/theme/components/subnav.ts:183]: Never mind, you are using it but I commented before you made your latest updates.

File: src/theme/components/subnav.ts (hunk 46 above)
Comment (gonuguntla) [reply src/theme/components/subnav.ts]: @EdwinGuzman 
I need a gap. When isOutlined is true or when I select any button, if there is no gap, the items are getting mixed up.

Comment (EdwinGuzman) [reply src/theme/components/subnav.ts]: Hmm, ok, keep it for now

Comment (bigfishdesign13) [reply src/theme/components/subnav.ts]: @EdwinGuzman The `gap` attribute is not a native attribute for `list`, but `display: flex` has been applied to the `list` element. Do you know if there are semantic implications when mixing these attributes?

File: src/components/SubNav/SubNav.tsx [hunk 49]
```diff
@@ -0,0 +1,272 @@
...
+  screenreaderOnlyText?: string;
+}
+
+interface SubNavLinkProps extends SubNavItemProps {
+  href: string;
+}
+
+interface SubNavButtonProps extends SubNavItemProps {
+  onClick?: (event: React.MouseEvent | React.KeyboardEvent) => void;
+}
```
EdwinGuzman, line 113: Very nice.

src/components/SubNav/SubNav.stories.tsx
EdwinGuzman: Bring this up again but this is minor.

src/hooks/useScrollFadeStyles.mdx
```diff
@@ -0,0 +1,49 @@
...
+<Box>
+  <div>
+    <List
+      type="ul"
+      m="0"
+      ref={scrollableRef}
+      inline
+      noStyling
+    >
+      {primaryActions}
```
EdwinGuzman: This is fine but it's not clear what `primaryActions` is. It's okay to have something like
```
<li>red</li>
<li>blue</li>
... etc
```

because it's an example. You can essentially put down any example you like.

src/hooks/useScrollFadeStyles.mdx
```diff
@@ -0,0 +1,49 @@
...
+    </List>
+    {showRightFade && <Box sx={styles.fadeEffect} />}
```
EdwinGuzman: This is interesting. I didn't make the connection that this hook and the styles are not returned together. Maybe `fadeEffect` should be returned from the function? What do you think? 

Don't do this now for this PR. We can work on it later, just wanted to get your thought first.

src/theme/components/subnav.ts
EdwinGuzman: I don't know, but it's is how it works for multi-column, flex, and grid containers. Did not know before https://developer.mozilla.org/en-US/docs/Web/CSS/gap

EdwinGuzman (approved):
This looks good to me, thanks for working on this and making all the updates.

src/components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,272 @@
...
+  ChakraComponent,
+  useMultiStyleConfig,
+  Flex,
+} from "@chakra-ui/react";
+import Button from "../Button/Button";
+import Link from "../Link/Link";
+import List from "../List/List";
+import useScrollFadeStyles from "../../hooks/useScrollFadeStyles";
+
+export const actionBackgroundColorsArray = [
```
7emansell, line 14: Both arrays of colors are already declared in `Banner`– could they live in one place now that two components are using them?

7emansell (approved):
Quick question about the colors but otherwise looks and works great! Is there already a place where we're planning to use this?

src/components/SubNav/SubNav.tsx
bigfishdesign13: @7emansell Yes, that is the eventual goal, but for now, we are continuing to use unique data objects in the separate components.

src/hooks/useScrollFadeStyles.mdx
bigfishdesign13: I had similar feelings about this. How can it all be tied together into the hook?

src/hooks/useScrollFadeStyles.mdx
```diff
@@ -0,0 +1,60 @@
...
+<Box>
+  <div>
+    <List
+      type="ul"
+      m="0"
+      ref={scrollableRef}
+      inline
+      noStyling
+    >
+      <>
```
bigfishdesign13: This can be removed.

src/hooks/useScrollFadeStyles.mdx
```diff
@@ -0,0 +1,60 @@
...
+          <SubNavButton id="subnav-action">
+            Action
+          </SubNavButton>
+        </li>
+        <li>
+          <SubNavLink id="subnav-link">
+            Link
+          </SubNavLink>
+        </li>
+      </>
```
bigfishdesign13: This can be removed.

src/hooks/useScrollFadeStyles.mdx
```diff
@@ -0,0 +1,60 @@
...
+    </List>
+    {showRightFade && <Box sx={styles.fadeEffect} />}
```
bigfishdesign13: @deepikagonuguntla Can you also include a code example showing the styles?

bigfishdesign13 (changes requested):
Just one minor addition to the story for the hook, but otherwise, it looks good.

src/hooks/useScrollFadeStyles.mdx
gonuguntla: Ok

src/hooks/useScrollFadeStyles.mdx
EdwinGuzman: Hmm what if the hook exports  a component like `RightFade` and it gets added here like `<RightFade />`? Then the style object and DOM can come from the hook. I was originally thinking just the style object which is my preference.

oliviawongnyc:
https://github.com/user-attachments/assets/270a142a-55f9-4997-8a86-9bbd7698ff9b

Just wondering - is the height of the component meant to grow as the screen shrinks? @deepikagonuguntla

bigfishdesign13:
> Screen.Recording.2024-11-22.at.1.58.46.PM.mov 
> Just wondering - is the height of the component meant to grow as the screen shrinks? @deepikagonuguntla

@oliviawongnyc Yes. There are two things happening when it hits the mobile viewport. First, the overall padding within the component jumps from `8px` to `16px`. Second, the buttons increase in height from `32px` to `44px` for mobile a11y compliance.

oliviawongnyc:
> > Screen.Recording.2024-11-22.at.1.58.46.PM.mov
> > Just wondering - is the height of the component meant to grow as the screen shrinks? @deepikagonuguntla
> 
> @oliviawongnyc Yes. There are two things happening when it hits the mobile viewport. First, the overall padding within the component jumps from `8px` to `16px`. Second, the buttons increase in height from `32px` to `44px` for mobile a11y compliance.

I see, thanks, Marty!

bigfishdesign13 (approved):
Looks good!

PR #1705, opened 2024-11-18
Release v3.4.3

EdwinGuzman:
### Adds

- Adds the NYPL Header as a DS component as a temporary solution for a specific application. It's only meant to be used internally and teams are still encouraged to use the `nypl-header-app`.


src/theme/foundations/breakpoints.ts
```diff
@@ -16,7 +21,11 @@
 export default {
   sm: "20em",
   md: "37.5em",
+  // Temporary and just for the header
+  mh: headerBreakpoints.mh,
   lg: "60em",
+  // Temporary and just for the header
+  lh: headerBreakpoints.lh,
```
bigfishdesign13, line 28: I just noticed these.

bigfishdesign13 (approved):
👍

src/theme/foundations/breakpoints.ts
EdwinGuzman: Yeah, not ideal but for now it's ok. No other components will use these breakpoints.

PR #1710, opened 2024-12-05
DSD-1121: SVG testing update

7emansell:
Fixes JIRA ticket [DSD-1121](https://newyorkpubliclibrary.atlassian.net/browse/DSD-1121)

- Adds `jest-transformer-svg` package
- Updates component tests to use the new `data-file-name` prop generated in the SVG mock (closest we can get to testing that the correct icon is being displayed)
   - `Icon`, `Logo`, `SocialMediaLinks`, `Banner`

## How has this been tested?

- Tested a release candidate in Turbine, no reason to expect this would change anything there, but all tests passed and SVGs displayed as expected locally

### Accessibility Checklist

### Checklist:

### Front End Review:

[DSD-1121]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-1121?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


EdwinGuzman (approved):
Great, thanks! Was this package always installed so you didn't have to add it to `package.json`?

I don't think _all_ components need to test the SVGs they render, but I think more components can use the `data-file-name` confirmation test. For example, `Accordion` could use it but that's not needed for `MultiSelect` or `FilterBarInline` because those use `Accordion`s internally. It's not necessary for this PR since it wasn't requested to update all tests, but we should aim to update other component tests later on.

7emansell:
> Great, thanks! Was this package always installed so you didn't have to add it to package.json?

I think so, but since we were including `svg` in the list of elements for `fileMock.ts`, it wasn't being used.

> I don't think all components need to test the SVGs they render, but I think more components can use the data-file-name confirmation test. For example, Accordion could use it but that's not needed for MultiSelect or FilterBarInline because those use Accordions internally. It's not necessary for this PR since it wasn't requested to update all tests, but we should aim to update other component tests later on.

Added `Accordion` test. I think it's definitely within the scope of this PR to update the tests but I guess by the same token, full coverage of the `Icon` component makes any tests we would do on `Accordion`, `SocialMediaLinks`, etc., redundant since they're using `Icon` internally. @EdwinGuzman

EdwinGuzman:
Yes, I'm also thinking about whether we should just write tests or if they're redundant since the `Icon` is already tested. I'm ok with what is already written.

PR #1712, opened 2024-12-09
DSD-1877: Filtering Style Guide

bigfishdesign13:
Fixes JIRA ticket [DSD-1877](https://newyorkpubliclibrary.atlassian.net/browse/DSD-1877)

- Adds the `Filtering` page to the `Style Guide`.

## How has this been tested?

- local Storybook

- n/a

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[DSD-1877]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-1877?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/StyleGuide/FilteringAccessibility.mdx
EdwinGuzman, line 1: Can you create a `Filtering` subfolder in `StyleGuide` for all these files? Then you don't have to repeat the prefix for every file.

EdwinGuzman (commented):
These are great. Thank you for writing this and the imagery makes the docs easy to read. Do you think we need some gifs with examples on doc pages such as the accessibility for focus management? Another example, on the pagination page, it could use information about where focus should go after a button is clicked (the total results heading). Focus management is something that comes up a lot so I want to make it clear what our recommendation for that is as well. 

Before approving, I want all FE devs to read through this, make sure they understand the approach, and approve it since it's a guideline for their apps: @7emansell @dgcohen @avertrees @charmingduchess @jackiequach @isastettler @Toxiapo

src/components/StyleGuide/FilteringApplyingClearing.mdx
```diff
@@ -0,0 +1,106 @@
...
+- update the `active filters` list based on the filter options that have been
+  selected
+- update the `results` list based on the filter options that have been selected
+- update the `total results` heading to reflect the number of items in the
+  updated `results` list
+
+## Clearing filters
+
+Filters can be cleared using various methods within the page structure,
+including global methods as well as methods for clearing individual filters.
```
Toxiapo: I'm curious: If a user searches for a term and applies some filters, but then continues typing in the search box without clearing the filters, should the filters be reset? Not sure if this is a accessibility question or a product question.

src/components/StyleGuide/FilteringAccessibility.mdx
```diff
@@ -0,0 +1,78 @@
...
+
+## Using aria-live
+
+The parent element housing the `total results` heading should include the
+`aria-live="polite"` attribute. This will ensure that real-time updates made to
+the heading text are announced to screen reader users.
+
+## Focus management
+
+### Applying search terms
```
Toxiapo: We’ll eventually need to build a custom filtering modal (we’re currently using an Ionic component) for the Reader, and this will be a great reference for accessibility!

dgcohen: Agree, great to see these AC consolidated here.

src/components/StyleGuide/FilteringCompositionFilterBar.mdx
```diff
@@ -0,0 +1,232 @@
...
+
+- reset all form input elements within the `filter bar`
+- update the `active filters` list to remove all filter options
+- update the `results` list to an unfiltered state
+- update the `total results` heading to reflect the number of items in the
+  updated `results` list
+
+### Mobile
+
+For mobile viewports, the filter bar will display a "Show result" button and can
```
dgcohen: Should this say "Show results"?

src/components/StyleGuide/FilteringApplyingClearing.mdx
bigfishdesign13: Good point. I will make sure to add something for this use case.

src/components/StyleGuide/FilteringAccessibility.mdx
bigfishdesign13: In Reservoir, we recently added two components to house the filter elements, one of which (`FilterBarPopup`) is a modal overlay.

src/components/StyleGuide/FilteringApplyingClearing.mdx
bigfishdesign13: Product and Design has settled on the following: 
```
All filters that have been applied should be removed any time 
a keyword search is executed
```

bigfishdesign13:
@EdwinGuzman 

> Do you think we need some gifs with examples on doc pages such as the accessibility for focus management? 

Yes, let me see what I can do.

> Another example, on the pagination page, it could use information about where focus should go after a button is clicked (the total results heading).

I am not 100% sure what you mean here. Can you provide an example?

src/components/StyleGuide/Filtering/Accessibility.mdx
```diff
@@ -0,0 +1,78 @@
...
+### Applying search terms
+
+When a search term is applied using the `keyword search` field, focus should be moved
+to the `total results` heading.
+
+### Applying filters
+
+#### Live filtering
+
+For live filtering, as a user makes changes to filtering options, focus should
```
charmingduchess: Maybe we want to focus on total results after some period of inactivity? I'm wondering if there is another way to indicate to a screen reader that results are being/have been loaded for live filtering.

src/components/StyleGuide/Filtering/PageCompositionElements/FilterBar.mdx
```diff
@@ -0,0 +1,232 @@
...
+with each filter option. For example, `English (5)` or `Offsite (11)`. The
+number of associated items included in the label text should be updated as
+filter options are applied. This treatment is applicable for the labels of the
+`Checkbox`, `Radio`, and `Toggle` components, and the labels of the checkbox
+options within a `MultiSelect` component, as well as the options within a
+`Select` component.
+
+## Filter option status
+
+If the number of items associated with a filter option is zero (0), that filter
```
charmingduchess, line 165: This would only apply to live filtering, correct? Or else it would only apply to filter options that are known beforehand. 
Most RC filters are based on aggregations that are returned with search responses, and can't be known before.

7emansell (approved):
this is really clear and also consistent with previous filtering guidance I've received, great to have it all in one place

src/components/StyleGuide/Filtering/Accessibility.mdx
7emansell: Also just interested in if there could/should be loading guidance in general (this also has to do with `Pagination` probably)

src/components/StyleGuide/Filtering/PageCompositionElements/ActiveFilters.mdx
```diff
@@ -0,0 +1,65 @@
...
+has been applied.
+
+## Labeling
+
+The `active filters` list should include an "Active filters" label.
+
+## Adding filters
+
+A new tag should be added to the `TagSet` as each new filter option is applied.
+The tags in the `TagSet` should be listed in the order they were selected.
```
charmingduchess: Is there something built into the TagSet or the filterbar that would maintain that order?

charmingduchess (commented):
Looks good! I have a couple questions. Overall, I'm curious what kind of collaboration and buy in was there from UX and product in these guidelines?

src/components/StyleGuide/Filtering/PageCompositionElements/ActiveFilters.mdx
bigfishdesign13: The tags within the `TagSet` component are rendered using the order of the elements in the data object that is passed in. It will be the responsibility of consuming apps to handle the order of the elements in that data object.

src/components/StyleGuide/Filtering/PageCompositionElements/FilterBar.mdx
bigfishdesign13: If the associated items count comes back with the search response, once the response comes in, consuming apps should be able to update the data used to populate the component with the new numbers and states.

src/components/StyleGuide/Filtering/Accessibility.mdx
bigfishdesign13: @charmingduchess Yes, screen reader users will be notified when the `total results` heading is updated. Please see the `Using aria-live` section.

bigfishdesign13: @7emansell I am not 100% sure what you mean. Let's make sure to talk about this.

bigfishdesign13:
> Looks good! I have a couple questions. Overall, I'm curious what kind of collaboration and buy in was there from UX and product in these guidelines?

@charmingduchess Product and Design have been fully involved in the definition and approval of these recommendations. It's been an ongoing process for almost two years now.

src/components/StyleGuide/Filtering/Accessibility.mdx
```diff
@@ -0,0 +1,167 @@
...
+  <Box>
+    When a search term is applied using the `keyword search` field, focus should
+    be moved to the `total results` heading.
+  </Box>
+  <Image
+    alt="Example of full-page filtering app"
+    border="1px solid var(--nypl-colors-ui-border-default)"
+    src={imageFocusKeywordSearch}
+    {...exampleWrapperStyles}
+  />
```
bigfishdesign13, line 57: @EdwinGuzman I added animated examples for most of the focus management recs.

src/components/StyleGuide/FilteringApplyingClearing.mdx
Toxiapo: Thank you!

src/components/StyleGuide/Filtering/Accessibility.mdx
EdwinGuzman: This looks very good!

src/components/StyleGuide/Filtering/PageCompositionElements/Pagination.mdx
```diff
@@ -22,14 +22,19 @@ multiple pages.
...
 
 ## Rendering
 
 The `pagination` menu should be rendered using the DS
 [Pagination](./?path=/docs/components-navigation-pagination--docs) component.
 
-## Display
+## Position
+
+The `pagination` menu should be positioned directly below the `results` list.
```
EdwinGuzman: This is something we've been discussing in DC: when the patron goes to a bad search URL, specifically because they entered a page number that is "out of bounds", then it makes sense to include the page number in the error message:
Screenshot 2025-01-13 at 10 09 04 AM

https://qa-digitalcollections.nypl.org/collections?collection_keywords=collection&page=9000

And this led me to wonder why don't we display the page number at the top of a search page? The patron would have to scroll all the way down to see what page they're on: https://qa-digitalcollections.nypl.org/collections?collection_keywords=collection&page=5#collections

The examples are in DCF but I think this is a general pattern we have.

bigfishdesign13: That's an interesting way to approach the error message. My gut feeling would be to approach this with a different message. Was any research done for this wording?

Regarding the wording in the total results heading, do you mean, why not say something like this?
```
Displaying page 2 of 31
```

7emansell: There was not really any research done for this wording ("No results for x found on page y") since we thought we were solving for a pretty specific edge case– my understanding was that if the page is out of bounds, it should actually be a 404, we just weren't able to do that here

src/components/StyleGuide/Filtering/Accessibility.mdx
7emansell: This ticket came out of this, since this is not something to be addressed in this PR: https://newyorkpubliclibrary.atlassian.net/browse/DSD-1988

src/components/StyleGuide/Filtering/PageCompositionElements/Pagination.mdx
bigfishdesign13: To get to this error, someone would have had to manipulate the URL. Maybe an error saying something about an invalid URL would be more applicable. Going that route would also eliminate the introduction of a data point (i.e. page number) that we don't refer to anywhere else in terms of searching.

EdwinGuzman: The problem is that we won't know that the URL is invalid until the request is made and the API doesn't return any data. For `collection_keywords=collection&page=9000`, "collections" and "9000" are valid values but 9000 is a number out of bounds of what data is available in the API. So I'm more inclined to give an error that's says there are no results rather than saying it's an invalid URL or redirecting to the 404 page. 

But to go back to the core concern, yes something like "Displaying page 2 of 31" could be helpful for patrons so they don't have to scroll all the way down on a page.

src/components/StyleGuide/Filtering/ApplyingClearing.mdx
```diff
@@ -0,0 +1,112 @@
...
+    users have to engage in a trial-and-error process to get useful results.
+
+### Batch filtering
+
+- The user selects multiple filtering options and the data is updated after the
+  user takes the additional action of clicking the "Apply filters" button
+  included in the `filter bar`.
+- This method is recommended for:
+  - very heavy data sets
+  - low performing apps
```
EdwinGuzman, line 51: I think we should define what "low performing" means. If we can't right now, "very heavy data sets" is enough and we should remove "low performing apps".

EdwinGuzman: I'm thinking, if we can get a baseline of web core vital metrics, for example, we can say something along the lines of, "does not meet Digital performance baseline metrics", or something to that effect.

src/components/StyleGuide/Filtering/JavaScriptDisabled.mdx
```diff
@@ -0,0 +1,29 @@
+import { Meta } from "@storybook/blocks";
+
+<Meta title="Style Guide/Filtering/JavaScript disabled" />
+
+# Filtering when JavaScript is disabled
```
EdwinGuzman, line 5: Once I have better progressive enhancement updates and recommendation, I want to update this to say that a fallback message banner should render in cases where features are JS-heavy.

src/components/StyleGuide/Filtering/Overview.mdx
```diff
@@ -0,0 +1,44 @@
+import { Meta } from "@storybook/blocks";
+
+import Link from "../../Link/Link";
+
+export const exampleWrapperStyles = {
```
EdwinGuzman: I think this could go in the `/src/utils` directory to keep it aligned with other storybook-related functions.

src/components/StyleGuide/Filtering/PageCompositionElements/FilterBar.mdx
```diff
@@ -0,0 +1,232 @@
...
+import imageExampleRow from "/static/filteringStyleGuide/exampleRow.png";
+import imageFilterBar from "/static/filteringStyleGuide/filterBar.png";
+import { exampleWrapperStyles } from "../Overview.mdx";
+
+<Meta title="Style Guide/Filtering/Page elements/Filter bar" />
+
+# Filter bar
+
+The `filter bar` is a wrapper for individual form input elements, including the
+following DS components: `CheckboxGroup`, `MultiSelectGroup`, `RadioGroup`,
```
EdwinGuzman, line 19: I think this gives the impression that we are supporting managing state for all components but right now we only manage state for `MultiSelect`.

src/components/StyleGuide/Filtering/PageCompositionElements/FilterBar.mdx
```diff
@@ -0,0 +1,232 @@
...
+The `filter bar` element should be rendered using the DS
+[FilterBarInline](./?path=/docs/components-form-elements-filterbarinline--docs)
+and
+[FilterBarPopup](./?path=/docs/components-form-elements-filterbarpopup--docs)
+components.
+
+<Banner
+  content={
+    <>
+      <strong>IMPORTANT:</strong> The filter bar should be configured based on
```
EdwinGuzman, line 47: Yes, but note, RC only uses `FilterBarInline` for both mobile and desktop.

src/components/StyleGuide/Filtering/PageCompositionElements/Pagination.mdx
bigfishdesign13: > The problem is that we won't know that the URL is invalid until the request is made and the API doesn't return any data. For collection_keywords=collection&page=9000, "collections" and "9000" are valid values but 9000 is a number out of bounds of what data is available in the API. So I'm more inclined to give an error that's says there are no results rather than saying it's an invalid URL or redirecting to the 404 page.

Ok, got it. Let's stick with what is there. I am warming to the idea of the `page`.

> But to go back to the core concern, yes something like "Displaying page 2 of 31" could be helpful for patrons so they don't have to scroll all the way down on a page.

I'll start a discussion about this. Maybe something like this would work...
image

src/components/StyleGuide/Filtering/ApplyingClearing.mdx
bigfishdesign13: @EdwinGuzman That wording is the original language from the UX exploration, so it is very generic. I like your suggestion. Go for it? But then that puts pressure on us/someone to define those baseline metrics.

src/components/StyleGuide/Filtering/JavaScriptDisabled.mdx
bigfishdesign13: @EdwinGuzman Great. Pass it along when you are ready.

src/components/StyleGuide/Filtering/PageCompositionElements/FilterBar.mdx
bigfishdesign13: What is giving that impression? Calling it a `wrapper`?

src/components/StyleGuide/Filtering/PageCompositionElements/FilterBar.mdx
bigfishdesign13: Yes, true. But these are recommendations for a more ideal state. This is what apps should be working toward.

PR #1720, opened 2025-01-03
DSD-1888: Pagination style updates

7emansell:
Fixes JIRA ticket [DSD-1888](https://newyorkpubliclibrary.atlassian.net/browse/DSD-1888)

- Updates the style of the `Pagination` component, particularly the Previous and Next links

### Accessibility Checklist

### Checklist:

### Front End Review:

[DSD-1888]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-1888?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -6,9 +6,11 @@ import {
 } from "@chakra-ui/react";
 import React, { forwardRef, useState, useRef } from "react";
 
+import Text from "../Text/Text";
 import Link from "../Link/Link";
 import List from "../List/List";
 import { range } from "../../utils/utils";
+import Icon from "../Icons/Icon";
 
 export interface PaginationProps {
   /** Additional className. */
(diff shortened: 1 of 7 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -123,76 +125,156 @@ export const Pagination: ChakraComponent<
         handlePageClick(e, nextPageNumber);
       }
     };
+
+    // Returns the "Previous" or "Next" link element, disabled according to current page number.
+    const previousNextElement = (
```
7emansell: A lot of styling added here since this element is a weird halfway point between a `Button` and a `Link`. Should this be moved to `theme/pagination.ts` or is it okay here since the item elements also have their styling in this file?

src/components/Pagination/Pagination.tsx
```diff
@@ -123,76 +125,160 @@ export const Pagination: ChakraComponent<
...
+      pageNumber: number,
+      text: string,
```
EdwinGuzman: Set this to `"Previous" | "Next"`

src/components/Pagination/Pagination.tsx
```diff
@@ -123,76 +125,160 @@ export const Pagination: ChakraComponent<
         handlePageClick(e, nextPageNumber);
       }
     };
+
+    // Returns the "Previous" or "Next" link element, disabled according to current page number.
+    const previousNextElement = (
```
EdwinGuzman: Instead of argument order, create an object structure to pass the arguments.

src/components/Pagination/Pagination.tsx
```diff
@@ -123,76 +125,160 @@ export const Pagination: ChakraComponent<
...
+            changeUrls ? undefined : isPrevious ? previousPage : nextPage
+          }
+          {...{
+            "aria-label": `${isPrevious ? "Previous" : "Next"} page`,
+            "aria-disabled": isDisabled,
+          }}
+        >
+          {!isPrevious && (
+            <Text
+              size="body2"
```
EdwinGuzman: Add an `as="span"` prop. There's technically nothing wrong with a div inside of an anchor, but it's a general convention for inline elements to be inside anchors rather than block elements.

src/components/Pagination/Pagination.tsx
```diff
@@ -123,76 +125,160 @@ export const Pagination: ChakraComponent<
...
+          id={`${id}-${text}`}
+          sx={combinedStyles}
+          type="action"
+          onClick={
+            changeUrls ? undefined : isPrevious ? previousPage : nextPage
+          }
+          {...{
+            "aria-label": `${isPrevious ? "Previous" : "Next"} page`,
+            "aria-disabled": isDisabled,
+          }}
```
EdwinGuzman: Why are these added as spread props? It looks like theyre always added so you can add them directly to the component.

src/components/Pagination/Pagination.tsx
```diff
@@ -123,76 +125,160 @@ export const Pagination: ChakraComponent<
...
+            <Text
+              size="body2"
+              sx={{
+                display: { base: "none", md: "inline" },
+              }}
+            >
+              {text}
+            </Text>
+          )}
+          <Icon
```
EdwinGuzman, line 170: It's technically correct to render both svgs since they're both hidden from screenreader and, depending on which one is rendered, the other one is visually hidden. It just feels wrong looking at the code since you have to pay more attention to the CSS hiding the component than a logic statement. Nothing to do, just pointing out something nitpicky.

src/components/Pagination/Pagination.tsx
EdwinGuzman: I'm not sure where my comment went... but yes, do try moving this into the `pagination.ts` file. I think it'll work best if it's a new theme object within the same file so you can use `useMultiStyleConfig` again but with a different value.

This element also seems like one you can remove from this main component and a create a new component in this same file (or it can be a different file).

If it is moved to `pagination.ts`, will the arguments need to be passed in the `useMultiStyleConfig` function? That seems like the trickiest part to implement without taking a very close look.

7emansell: > This element also seems like one you can remove from this main component and a create a new component in this same file (or it can be a different file).

I would but I think it also makes sense for it to parallel `getItemElement()` ? Now that the styling is in the theme file too it's not as long

src/components/Pagination/Pagination.tsx
7emansell: Just sloppy– i was initially using `linkAttrs` from the item element and then got rid of it

src/components/Pagination/Pagination.tsx
7emansell: Also, let me know if this is what you were thinking for the theme file. It works but I always feel like I'm missing something when I use `defineMultiStyleConfig`

EdwinGuzman: Yep, the theme file looks good. The usage of `defineMultiStyleConfig` is also fine and nothing is missing.

EdwinGuzman (approved):

src/theme/components/pagination.ts
```diff
@@ -1,22 +1,73 @@
...
+        },
+      },
+    },
+    disabledElement: {
+      color: "ui.disabled.primary",
+      pointerEvents: "none",
+      svg: { fill: "ui.disabled.primary" },
+      _dark: {
+        color: "ui.disabled.secondary",
+        svg: { fill: "ui.disabled.secondary" },
```
bigfishdesign13: These color values should be `dark.ui.disabled.secondary`.

bigfishdesign13 (changes requested):
The updated functionality works great. This just needs one color update for dark mode.

bigfishdesign13 (approved):
Looks great!

PR #1728, opened 2025-01-17
DSD-1905: suggestions for format number PR

bigfishdesign13:
Fixes JIRA ticket [DSD-1905](https://newyorkpubliclibrary.atlassian.net/browse/DSD-1905)

- This includes some suggestion for the `useFormatNumber` hook that are intended for ease of use.
- While the code does work, I thought the implementation would be different from what I used.

## How has this been tested?

- local Storybook

- n/a

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[DSD-1905]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-1905?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Text/Text.stories.tsx
+++ b/src/components/Text/Text.stories.tsx
@@ -1,6 +1,7 @@
 import type { Meta, StoryObj } from "@storybook/react";
 import Text, { textSizesArray } from "./Text";
 import { argsBooleanType } from "../../helpers/storybookUtils";
+import formatNumber from "../../hooks/useFormatNumber";
 
 const meta: Meta<typeof Text> = {
   title: "Components/Typography & Styles/Text",
@@ -24,6 +25,12 @@ const meta: Meta<typeof Text> = {
     },
   },
 };
+
+const copiesSold = formatNumber(81450000);
+const suggestedPrice = formatNumber(100, 20000);
+// const getFormattedNumbers = () => {
+//   formatNumber(81450000);
+// };
 
 export default meta;
 type Story = StoryObj<typeof Text>;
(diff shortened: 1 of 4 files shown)


src/components/Text/Text.stories.tsx
```diff
@@ -1,6 +1,7 @@
 import type { Meta, StoryObj } from "@storybook/react";
 import Text, { textSizesArray } from "./Text";
 import { argsBooleanType } from "../../helpers/storybookUtils";
+import formatNumber from "../../hooks/useFormatNumber";
```
bigfishdesign13, line 4: @EdwinGuzman I was expecting for this to be wrapped in brackets when it is imported, but it works fine like this. In fact, when I put it in brackets, it throws an error.

EdwinGuzman: In the hook file, the function is the default export so this is correct for that export. If we want to use braces, then it additionally needs to export the function as a named function `export useFormatNumber;`.

Either is ok and ultimately it will matter the most when it gets exported in the main `index.ts` file... which actually is needed here in order for consuming apps to use. So @deepikagonuguntla make sure to export this hook in the main index.ts file.

src/hooks/useFormatNumber.mdx
```diff
@@ -4,43 +4,40 @@ import { Meta, Source } from "@storybook/blocks";
...
-| Hook Version | DS Version |
-| ------------ | ---------- |
-| Added        | `Prerelease`    |
-| Latest       | `Prerelease`    |
+| Hook Version | DS Version   |
+| ------------ | ------------ |
+| Added        | `Prerelease` |
+| Latest       | `Prerelease` |
 
 `useFormatNumber` provides a number with commas and handles a range of numbers.
```
EdwinGuzman, line 12: Can you add that this is based on digital standards?

EdwinGuzman (approved):

gonuguntla (approved):

PR #1729, opened 2025-01-17
DSD-1924: Heading component mobile line height

bigfishdesign13:
Fixes JIRA ticket [DSD-1924](https://newyorkpubliclibrary.atlassian.net/browse/DSD-1924)

- Updates the `Heading` component to add `line-height` styles for mobile.

## How has this been tested?

- local Storybook

- n/a

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[DSD-1924]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-1924?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/theme/components/heading.ts
```diff
@@ -172,7 +172,7 @@ export const headings = {
       },
       fontWeight: "heading.heading8",
       letterSpacing: "0",
-      lineHeight: "1.50",
+      lineHeight: { base: "1.50", md: "1.50" }, // This value is intensionally the same for mobile and desktop.
```
EdwinGuzman: You should be able to just set this to `lineHeight: "1.50"` and it should be set for both mobile and the medium breakpoint.

As an aside, is `1.50` not the same as `1.5`?

bigfishdesign13: Right. I will simplify and then update my comment to explain why this one is not setup to be responsive like the others.

bigfishdesign13: Also, can you share documentation about the difference of `1.50` and `1.5` for `line-height`?

EdwinGuzman: I don't know if it's true that those two values `1.50` and `1.5` are not the same so I was asking, but I can look for resources.

bigfishdesign13: Sorry, I read that as a statement and not as a question. Yes, `1.50` and `1.5` are numerically the same. However, the designer in me opted to use the same format as those that do require two decimal places.  Symmetry. Consistency. All that.

EdwinGuzman (approved):

PR #1743, opened 2025-02-04
DSD-1931: font weight for "required" label

bigfishdesign13:
Fixes JIRA ticket [DSD-1931](https://newyorkpubliclibrary.atlassian.net/browse/DSD-1931)

- Updates the `Label` component to set the font weight for the `(required)` label to `"light"`.
- Updates the `Fieldset` component to set the font weight for the `(required)` label to `"light"`.
- Updates the `FeedbackBox` component to set the font weight for the `(required)` label on the `Textarea` element to `"light"`.

**NOTE:** I made a change to the `TextInput` component to enable the update in the `FeedbackBox` component. I'd like to get some feedback about that choice. There are 15 total components that implement a `labelText` prop and the `labelText` props under the `Checkbox` and `Radio` components are already configured with the change that I just made under the `TextInput` component. Do you think this is a valid change to the `TextInput` component? And if so, should this update be implemented across the other components that have the `labelText` component?

## How has this been tested?

- local Storybook

- n/a

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[DSD-1931]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-1931?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/theme/components/fieldset.ts
```diff
@@ -18,6 +18,9 @@ const Fieldset = defineStyleConfig({
         color: "ui.typography.heading",
         ...labelLegendText,
         ...screenreaderStyles,
+        span: {
+          fontWeight: "light",
```
EdwinGuzman, line 22: The PR description says "medium" but this is "light". Is this correct?

src/theme/components/label.ts
```diff
@@ -11,6 +11,9 @@ const Label = defineStyleConfig({
     ...labelLegendText,
     flex: props.isInlined ? "1" : null,
     whiteSpace: props.isInlined ? "nowrap" : null,
+    span: {
+      fontWeight: "light",
```
EdwinGuzman, line 15: The PR description says "medium" but this is "light". Is this correct?

EdwinGuzman (approved):
I'd like to revisit the `string` vs `string | JSX.Element` prop value but at a later time. This is consistent with what other components already set their type value.

EdwinGuzman:
Didn't mean to approve -- have that questions about "medium" vs "light"

src/theme/components/fieldset.ts
bigfishdesign13: Sorry. That was a typo in the PR description. It should say `light` to match the description in the Jira ticket.

src/theme/components/label.ts
bigfishdesign13: Same here.

src/theme/components/fieldset.ts
bigfishdesign13: I updated the PR description.

7emansell (changes requested):
Font weight looks good but I think that there needs to be a condition added for the `aria-label` on `<TextInput>` when the passed `labelText` is an `JSX.Element`, so that it doesn't set `aria-label` with that (I'm not sure what else it could use? if the element passed has an ID, it could use `aria-labelledBy`?)

This is with `showLabel={false}`:
Screenshot 2025-02-06 at 11 37 04 AM

Also, since it has `aria-describedBy` maybe this isn't a serious issue

EdwinGuzman:
> (I'm not sure what else it could use? if the element passed has an ID, it could use aria-labelledBy?)

Yes, `aria-labelledBy` is a good alternative if the element with that id is already on the page.

bigfishdesign13:
@7emansell and @EdwinGuzman – When `showLabel` is `false`, the label element is removed altogether from the DOM. This is the pattern used on a number of our components. With that, I see two options:

1. Hide the `<label>` rather than removing it from the DOM and add `aria-labelledBy` (as you mentioned)
2. Remove the `<label>` element from the DOM, as we are doing now, but also add a condition to require that `labelText` is a string when `showLabel` is `false`

The solution used should then be the pattern as we enable the `string | JSX.Element` type for the label in other components.

EdwinGuzman:
Tweak 1) so that you hide it off screen, the "screenreaderOnly" style rule, and then you shouldn't have to add an `aria-labelledby` since it is already in the DOM.

src/utils/utils.ts
```diff
@@ -189,3 +189,20 @@ export const exampleWrapperStyles = {
   overflow: "hidden",
   width: "100%",
 };
+
+/** Return just the text from within a JSX.Element */
+export const getTextFromElement = (
```
bigfishdesign13, line 194: @7emansell and @EdwinGuzman – Rather than reworking the logic around the `aria-label`, I added a utility function to extract plain text from an `Element`. This ensures that the value passed to the aria-label is a string.

EdwinGuzman: Looks good so far. Can you add a few tests, specifically with multiple level of children since this is recursive.

bigfishdesign13: Good point. Done.

7emansell (approved):

EdwinGuzman (approved):
The test for `getTextFromElement` is implictly tested in the `TextInput` component's own test. This is fine but it should have explicit tests in `utils.test.ts`. I'd say not a blocker for now but let's revisit tests in general for v4.

bigfishdesign13:
> The test for `getTextFromElement` is implictly tested in the `TextInput` component's own test. This is fine but it should have explicit tests in `utils.test.ts`. I'd say not a blocker for now but let's revisit tests in general for v4.

@EdwinGuzman 
It was easy enough to add a test for this. However, like the structure in the `hooks` directory, the test file sits alongside `utils.ts`.

PR #1759, opened 2025-03-03
DSD-2012: `TagSetFilter` button wrapper

7emansell:
Fixes JIRA ticket [DSD-2012](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2012)

- Removes the `button` wrapping each tag in the `TagSetFilter` if `isDismissible` is false and no `onClick` has been passed
- Adds/updates `TagSet` tests

- This should correct the existing issue of tags that are essentially labels being identified as buttons, and being tabbable/clickable as if they are buttons

### Accessibility Checklist

### Checklist:

### Front End Review:

[DSD-2012]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-2012?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 - Updates the spacing between heading and content in `Banner` component.
 - Updates the `Card`, `FeaturedContent`, and `SearchBar` components to use container queries.
-- Updated `TagSetFilter` to remove tag button wrapper when `isDismissible` and `onClick` are false.
+- Updated the `"filter"` variant of `TagSet` to remove tag button wrapper when `isDismissible` and `onClick` are false.
 
 ## 3.5.4 (February 13, 2025)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 - Updates the spacing between heading and content in `Banner` component.
 - Updates the `Card`, `FeaturedContent`, and `SearchBar` components to use container queries.
-- Updated `TagSetFilter` to remove tag button wrapper when `isDismissible` is false.
+- Updated `TagSetFilter` to remove tag button wrapper when `isDismissible` and `onClick` are false.
 
 ## 3.5.4 (February 13, 2025)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 - Updates the spacing between heading and content in `Banner` component.
 - Updates the `Card`, `FeaturedContent`, and `SearchBar` components to use container queries.
+- Updated `TagSetFilter` to remove tag button wrapper when `isDismissible` is false.
 
 ## 3.5.4 (February 13, 2025)
 
--- a/src/components/TagSet/TagSet.mdx
+++ b/src/components/TagSet/TagSet.mdx
@@ -112,8 +112,7 @@ Notes:
   function will return the following object as an argument:
   `{ label: "Clear Filters", id: "clear-filters" }`
 - When `isDismissible` is false, the tags will not render as buttons, since they
-  have no `onClick`. In this case, they should be used similarly to `StatusBadge`,
-  as a label.
+  have no `onClick`. In this case, they should be used as a static list of keywords.
 
 <Canvas of={TagSetStories.FilterVariant} />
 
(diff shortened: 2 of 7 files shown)


src/components/TagSet/TagSetFilter.tsx
```diff
@@ -103,8 +99,31 @@ export const TagSetFilter: React.FC<TagSetFilterProps> = chakra(
...
+                      align="left"
+                      color={iconColor}
+                      data-testid="ts-icon"
+                      name={tagSet.iconName}
+                      size="small"
+                    />
+                  ) : null}
+                  <span
+                    style={{
+                      fontWeight: "var(--nypl-fontWeights-button-default)",
```
EdwinGuzman: Does moving this rule to the `Box` also work? Just to reduce having another span.

src/components/TagSet/TagSetFilter.tsx
```diff
@@ -73,28 +73,24 @@ export const TagSetFilter: React.FC<TagSetFilterProps> = chakra(
...
-                    align="left"
-                    color={iconColor}
-                    data-testid="ts-icon"
-                    name={tagSet.iconName}
-                    size="small"
-                  />
-                ) : null}
-                <span>{tagSet.label}</span>
-                {isDismissible ? (
+              {isDismissible || onClick ? (
```
EdwinGuzman, line 76: For this, if it's dismissible but there's no onClick function, how will it be removed?

EdwinGuzman (changes requested):
I think it needs an update to remove a tagset if it's dismissible but there's no onClick function. For styling, hover over the example in storybook: 
Screenshot 2025-03-03 at 4 38 46 PM

The nondismissible example is fine. The second example is also ok. The third, however, loses the cursor pointer style when you hover.

7emansell:
I think `onClick` is required if `isDismissible` is true, so this only needs to handle the case where `onClick` is true but `isDismissible` is false. In that case, the third one in that screenshot, there is no cursor change(on prod) so I don't think it should be added. @EdwinGuzman

src/components/TagSet/TagSetFilter.tsx
bigfishdesign13: > I think onClick is required if isDismissible is true, so this only needs to handle the case where onClick is true but isDismissible is false. In that case, the third one in that screenshot, there is no cursor change(on prod) so I don't think it should be added.

@7emansell Can you verify if the condition you mentioned is set up that way? It appears that the third example here has `isDismissible` set to true, yet `onClick` is not set.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 - Updates the spacing between heading and content in `Banner` component.
 - Updates the `Card`, `FeaturedContent`, and `SearchBar` components to use container queries.
+- Updated `TagSetFilter` to remove tag button wrapper when `isDismissible` and `onClick` are false.
```
bigfishdesign13: Update to say the `"filter"` variant of the `TagSet` component.

src/components/TagSet/TagSet.mdx
```diff
@@ -111,6 +111,9 @@ Notes:
   Clicking this button will also trigger the `onClick` prop. The `onClick`
   function will return the following object as an argument:
   `{ label: "Clear Filters", id: "clear-filters" }`
+- When `isDismissible` is false, the tags will not render as buttons, since they
+  have no `onClick`. In this case, they should be used similarly to `StatusBadge`,
+  as a label.
```
bigfishdesign13: Nit: ...they should be used as a static list of keywords.

This is more in line with the description of the `"filter"` variant.

src/components/TagSet/tagSetChangelogData.ts
```diff
@@ -9,6 +9,15 @@
 import { ChangelogData } from "../../utils/ComponentChangelogTable";
 
 export const changelogData: ChangelogData[] = [
+  {
+    date: "Prerelease",
+    version: "Prerelease",
+    type: "Update",
+    affects: ["Documentation", "Accessibility"],
+    notes: [
+      "Updated `TagSetFilter` to remove button wrapper on each tag when `isDismissible` is false and no `onClick` is passed.",
```
bigfishdesign13: Please make a similar change to the language here to match what is in the CHANGELOG.

bigfishdesign13 (changes requested):
Some minor wording changes.

bigfishdesign13 (approved):
👍

EdwinGuzman (approved):

PR #1760, opened 2025-03-10
DSD-2011: primitive and highlighter colors

bigfishdesign13:
Fixes JIRA ticket [DSD-2011](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2011)

- Adds the `primitives` color object.
- Adds the `highlighter` colors to the `colors` theme object.

## How has this been tested?

- local Storybook

- n/a

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[DSD-2011]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-2011?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/AccessibilityGuide/HighlightingText.mdx
+++ b/src/components/AccessibilityGuide/HighlightingText.mdx
@@ -24,10 +24,6 @@ implement highlighted text for semantic coding purposes, but it should not be
 used to fulfill WCAG 2.1 criteria as is not announced by most screen reading
 technology in its default configuration.
 
-Resources:
-
-- [The Mark Text element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/mark)
-
 ## NYPL Recommendations
 
 The NYPL Reservoir Design System (DS) provides semantic design tokens for a
@@ -124,3 +120,4 @@ Add a 1px border using a contrast compliant color.
 
 - [WCAG Success Criterion 1.4.1 Use of Color](https://www.w3.org/WAI/WCAG21/Understanding/use-of-color.html)
 - [NYPL Reservoir Design System Color Palette: Highlighter Colors](http://localhost:6006/?path=/docs/style-guide-colors--docs#highlighter-colors)
+- [The Mark Text element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/mark)
--- a/src/components/AccessibilityGuide/HighlightingText.mdx
+++ b/src/components/AccessibilityGuide/HighlightingText.mdx
@@ -18,6 +18,16 @@ Highlighting text in the UI requires more than a color indicator, as WCAG 2.1 AA
 standards stipulate that color should not be used as the only visual means of
 distinguishing a visual element.
 
+The `<mark>` HTML element is intended to represent text which is marked or
+highlighted for reference or notation purposes. The element may be used to
+implement highlighted text for semantic coding purposes, but it should not be
+used to fulfill WCAG 2.1 criteria as is not announced by most screen reading
+technology in its default configuration.
+
+Resources:
+
+- [The Mark Text element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/mark)
+
 ## NYPL Recommendations
 
 The NYPL Reservoir Design System (DS) provides semantic design tokens for a
--- a/src/components/AccessibilityGuide/HighlightingText.mdx
+++ b/src/components/AccessibilityGuide/HighlightingText.mdx
@@ -20,10 +20,13 @@ distinguishing a visual element.
 
 ## NYPL Recommendations
 
-The NYPL Reservoir Design System (DS) provides semantic design tokens for colors
-that can be used for highlighting text, but the full implementation of the
-`highlight` requires the addition of at least one of the following styles to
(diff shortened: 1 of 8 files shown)


src/components/AccessibilityGuide/HighlightingText.mdx
```diff
@@ -0,0 +1,116 @@
...
+    word
+  </Box> is highlighted with yellow.
+</p>
+}
+{
+
+<p>
+  This{" "}
+  <Box
+    as="mark"
```
EdwinGuzman, line 107: Can you include info and a link (such as https://developer.mozilla.org/en-US/docs/Web/HTML/Element/mark) for the `mark` element? I haven't used this before and it's interesting for this purpose!

src/components/StyleGuide/ColorCard.tsx
```diff
@@ -590,6 +591,55 @@ export const cssVars = {
       textColor: colorContrastData.ui.gray.xLightWarm.textColor,
     },
   ],
+  highlighter: [
+    {
+      name: "ui.highlighter.red",
+      showColorDataTable: false,
+      textColor: "ui.black",
```
EdwinGuzman, line 557: Can this value be the default so it doesn't have to be added for every object?

src/components/StyleGuide/ColorCard.tsx
```diff
@@ -871,23 +959,21 @@ export const makeColorCard = function (data) {
...
+export const getColorCards = (category: string) => {
   const cards = [];
-  const catArr = cssVars[category];
-  if (category === "researchLibraries") {
-    console.log(catArr);
-  }
-  for (let i = 0; i < catArr.length; i++) {
-    const colorData = catArr[i];
+  const catArr = colorCardsData[category];
+  catArr.map((color: any) => {
```
EdwinGuzman: `.map` returns an array so a better option is `catArr.forEach`.

But, because it returns an array, you can set `cards` to the returned value

```
const cards = catAr?.length < 0 ? [] : catArr.map((color: any) => makeColorCard(color));
```

instead of pushing the value `makeColorCard(colorData)` onto a new variable that is used in the `cards` array, you can return it directly.

src/components/StyleGuide/Colors.mdx
```diff
@@ -46,6 +52,59 @@ transparency are tough to talk about when it comes to color combinations and
 color contrast. That being the case, we have used question marks for the
 contrast values and WCAG status.
 
+## Primitives
+
+The primitive colors and the related colors scales were reverse engineered using
+the legacy semantic color palette. Because of that, the primitive colors are not
+used extensively throughout the semantic color assignments.
```
EdwinGuzman: This answers my question about why they're not "root colors" or used in the base semantic tokens.

src/theme/foundations/colors.ts
```diff
@@ -21,6 +21,145 @@ import { hexToRGB } from "../../utils/utils";
  */
 
 /** Reusable variables: */
+// Primitives
+export const primitives = {
```
EdwinGuzman, line 25: I really like the names of the primitives, like `scienceBlue`. Too bad we won't use them elsewhere.

EdwinGuzman (commented):
Some minor updates

src/components/AccessibilityGuide/HighlightingText.mdx
bigfishdesign13: Yes I will include a link. However, I did speak with Clare about the `<mark>` element and she said it is not supported very well, as noted [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/mark#accessibility).

src/components/StyleGuide/ColorCard.tsx
bigfishdesign13: The text color assignment needs to be set based on the background color for the card, so there is no single color value that work for all instances of the `ColorCard`. What I can do is set a default color value and then only add a value when the default needs to be overridden.

bigfishdesign13: Ha! I already have the text defaulting to `ui.white`, so I will remove all of the value in all the instance where I am setting the text to white.

src/theme/foundations/colors.ts
bigfishdesign13: Yes, I did have some fun finding appropriate names for the primitives. But alas, the colors used in designs should be the semantic colors.

EdwinGuzman (approved):
Thanks for the updates!

src/components/AccessibilityGuide/HighlightingText.mdx
bigfishdesign13: UPDATE: I spoke to Clare again about the `<mark>` element prior to my final updates. She had done some research and found that support for the element has improved in recent years, so we updated our docs to reflect that and ultimately recommend that we do use it.

bigfishdesign13:
I asked the Design Team to review these updates and I will merge this PR once they have had a chance to review and comment.

PR #1775, opened 2025-04-03
DSD-1916: `Button` height sync with VDL

7emansell:
Fixes JIRA ticket [DSD-1916](https://newyorkpubliclibrary.atlassian.net/browse/DSD-1916)

- Assigns each `Button` variant a height value to align with VDL (except `pill` buttons which still fit content)
- Creates `sizes` design tokens which so far only are used for button height

### Accessibility Checklist

## Open Questions

The ticket mentions adding these as tokens– does that mean each of the 3 button height values should be added as a design token separate from the button sizes?

### Checklist:

### Front End Review:

[DSD-1916]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-1916?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Button/buttonChangelogData.ts
+++ b/src/components/Button/buttonChangelogData.ts
@@ -10,6 +10,15 @@ import { ChangelogData } from "../../utils/ComponentChangelogTable";
 
 export const changelogData: ChangelogData[] = [
   {
+    date: "Prerelease",
+    version: "Prelease",
+    type: "Update",
+    affects: ["Styles"],
+    notes: [
+      "Explicitly assigns the height value for each size option to align with VDL.",
+    ],
+  },
+  {
     date: "2024-07-25",
     version: "3.2.0",
     type: "Update",
--- a/src/components/DevelopmentGuide/DesignTokens.mdx
+++ b/src/components/DevelopmentGuide/DesignTokens.mdx
@@ -432,7 +432,7 @@ The `radii` design tokens can be used for the `borderRadius` (js) and
 
 ### Sizes
 
-The `sizes` design tokens can be used for the `height` attributes.
+The `sizes` design tokens can be used for the `height` (css/js) attribute.
 
 |            | JS Token              | CSS Token                          | Value |
 | ---------- | :-------------------- | :--------------------------------- | :---- |
--- a/src/components/DevelopmentGuide/DesignTokens.mdx
+++ b/src/components/DevelopmentGuide/DesignTokens.mdx
@@ -209,6 +209,7 @@ The DS design tokens are broken into the following categories:
 - {<Link href="#breakpoints" target="_self">Breakpoints</Link>}
 - {<Link href="#color" target="_self">Color</Link>}
 - {<Link href="#radii" target="_self">Radii</Link>}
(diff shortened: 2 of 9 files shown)


src/components/Button/buttonChangelogData.ts
```diff
@@ -9,6 +9,15 @@
 import { ChangelogData } from "../../utils/ComponentChangelogTable";
 
 export const changelogData: ChangelogData[] = [
+  {
+    date: "Prerelease",
+    version: "Prelease",
+    type: "Update",
+    affects: ["Styles"],
+    notes: [
+      "Explicitly assigns the height value for each size option to align with VDL.",
```
bigfishdesign13, line 18: Nit: Update the changelog entry to align with this wording.

src/theme/components/button.ts
```diff
@@ -69,14 +69,13 @@ const generalSizeValues = (size = "medium", isPill = false) => {
   const sizes = {
     small: {
       fontSize: "desktop.button.small",
-      height: "fit-content",
-      minHeight: "auto",
+      height: isPill ? "fit-content" : "26px",
```
bigfishdesign13: After looking at this in the PR, I would like to establish `sizes` tokens to use here. We currently don't have any `sizes` tokens, so adding them would be somewhat of a foundational change. 

@EdwinGuzman What would you suggest as the best way to handle this? Do you think that update should be a separate ticket? If so, we could branch off of this feature branch and eventually merge back in here. Essentially, we would need a new `sizes.ts` file in the `theme/foundations` directory with this code:

```
const sizes = {
  button: {
    default: {
      height: "40px",
    },
    small: {
      height: "26px",
    },
    medium: {
      height: "40px",
    },
    large: {
      height: "56px",
    },
  },
};
export default sizes;
```

bigfishdesign13 (commented):
The update works well. I know we talked about this a bit during standup, but I would like to consider adding some new design tokens.

src/theme/components/button.ts
EdwinGuzman: I think it's small enough to do in this PR if you're ok with that @7emansell . I'd make this a separate ticket if there's also a need to update references in other components, but it seems like it's only for `Button`, for now, right?

7emansell: Sounds good, yes I can do that in this PR

bigfishdesign13: Correct. For now, it would only impact the styles for `Button`. But going forward, the `sizes` theme object will be another token group we can utilize for `width` and `height` values.

src/components/DevelopmentGuide/DesignTokens.mdx
```diff
@@ -429,6 +430,17 @@ The `radii` design tokens can be used for the `borderRadius` (js) and
 |                   |                |                             |        |
 | **Round**         | round          | --nypl-radii-round          | 100%   |
 
+### Sizes
```
bigfishdesign13, line 433: Thank you for adding this!

bigfishdesign13 (approved):
Looks great!

PR #1792, opened 2025-04-15
DSD-2032: heading font weight

bigfishdesign13:
Fixes JIRA ticket [DSD-2032](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2032)

- Updates the `Heading` component to increase the font weight and to set the letter spacing to `0`  for the `size` styles.
- Updates the `Typography` and `Design Tokens` docs to reflect the new font weight styles for the `Heading` component.

## How has this been tested?

- local Storybook

- While this update is not specifically related to a11y, the font weight updates increase readability.

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[DSD-2032]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-2032?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


oliviawongnyc (approved):
This looks good to me, Marty!

src/theme/foundations/typography.ts
```diff
@@ -238,15 +243,15 @@ const typography: Typography = {
...
-      display1: fontWeightValues["thin"],
-      heading1: fontWeightValues["thin"],
-      heading2: fontWeightValues["thin"],
-      heading3: fontWeightValues["light"],
-      heading4: fontWeightValues["regular"],
-      heading5: fontWeightValues["medium"],
-      heading6: fontWeightValues["medium"],
-      heading7: fontWeightValues["medium"],
-      heading8: fontWeightValues["medium"],
+      display1: fontWeightValues["lightImportant"],
```
EdwinGuzman: Do you _have_ to create these new `important` variables? If you set it as `light: "300 !important"`, will it override other styles where `light` is used? If not, I think it's fine to use `!important` in the existing variables.

EdwinGuzman (commented):
One question but looks good otherwise.

src/theme/foundations/typography.ts
bigfishdesign13: The only reason we are doing this is because of the annoying overrides coming from the Header app using DS v2. And logically, we do want the styles from v3 to be _THE_ styles and not be overridden by what is coming out of the Header and Footer. Having said that, yes, I also now agree that we could put the `!important` right in the original variables. I will do that an push the updates for another review.

EdwinGuzman (approved):

PR #1794, opened 2025-04-16
DSD-1964: Tabs border color

bigfishdesign13:
Fixes JIRA ticket [DSD-1964](https://newyorkpubliclibrary.atlassian.net/browse/DSD-1964)

- Updates the `Tabs` component to sync the border color with the VDL.

## How has this been tested?

- local Storybook

- n/a

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[DSD-1964]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-1964?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/theme/components/tabs.ts
+++ b/src/theme/components/tabs.ts
@@ -110,7 +110,7 @@ const tablistWrapper = {
   alignItems: "center",
   borderBottom: {
     base: "0",
-    md: "1px solid black",
+    md: "1px solid var(--nypl-colors-ui-gray-xx-dark)",
   },
   height: { base: "58px", md: "auto" },
   margin: "0",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 - Updates `Tooltip` text color.
 - Updates the `SearchBar` component to remove the text input right border.
 - Updates the `SkeletonLoader` component to make the button placeholder look more like a button.
+- Updates the `Tabs` component to sync the border color with the VDL.
 
 ## 3.6.0 (April 10, 2025)
 
--- a/src/components/Tabs/Tabs.mdx
+++ b/src/components/Tabs/Tabs.mdx
@@ -9,10 +9,10 @@ import { changelogData } from "./tabsChangelogData";
 
 # Tabs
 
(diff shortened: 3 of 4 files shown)


oliviawongnyc:
Hi Marty! The component audit doc says all borders should be black, but the ticket says the border should be ui.gray.xx-dark. Can you explain this? TY!

Screenshot 2025-04-17 at 1 28 47 PM

src/theme/components/tabs.ts
```diff
@@ -110,7 +110,7 @@ const tablistWrapper = {
   alignItems: "center",
   borderBottom: {
     base: "0",
-    md: "1px solid black",
+    md: "1px solid var(--nypl-colors-ui-gray-xx-dark)",
```
oliviawongnyc, line 113: It seems like this is only impacting the end of the border? This is what happens when I change the color to red:

Screenshot 2025-04-17 at 3 31 54 PM

Is this intentional?

bigfishdesign13: There is a lot of backstory with this that I did not relay in the PR, so let me try...

After reconsidering the original ask that you noted from the audit doc (👀), I recalled that the original designs for the `Tabs` component were put together before we made the move away from using pure black for text and UI elements. With that in mind, I opted to flip the request and instead update the VDL component to use gray.

Further, the theme for the `Tabs` styles has some underlying colors that are being set by Chakra. Those colors are not easily changed in the theme file, so I decided to stick with the gray color that is already being used for the borders. Most of the borders in the `Tabs` component are being set by Chakra and with that the only thing that needed to be tweaked to get a uniform border color was the one line that you flagged. That border is on an element that we included when we created our `Tabs` component.

oliviawongnyc: I see. So you're saying Chakra uses the same gray color we are using for the borders?

bigfishdesign13: One of our darker gray colors is being used (curious where that is initially being set), which is better for this than our default border color.

EdwinGuzman (approved):

PR #1797, opened 2025-04-23
DSD-2006: update props with html equiv pt 1

oliviawongnyc:
Fixes JIRA ticket [DSD-2006](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2006)

- Removes `aria` props from components in favor of native alternatives and updates documentation:
     - Removes `ariaLabel` from `Accordion`
     - Removes `ariaLabel` from `Banner`
     - Removes `ariaAtomic` and `ariaLive` from `HelperErrorText`
     - Removes `ariaLabel` from `Notification`
- This is pt 1, the `aria` props. There will be another PR for the remaining props.

## How has this been tested?

- Locally in Storybook

- None.

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[DSD-2006]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-2006?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Accordion/Accordion.tsx
```diff
@@ -205,16 +202,20 @@ const getElementsFromData = (
  * Accordion component that shows content on toggle. Can be used to display
  * multiple accordion items together.
  */
+
+type ExtendedAccordionProps = AccordionProps &
```
EdwinGuzman: I personally prefer interfaces over types (more flexible), but don't change it. My actual comment is that this might be hard to scale if new props are needed in the future. That seems like a low risk for `Accordion`, but something to consider for other components.

EdwinGuzman: Can you also export this type and then export it in the main `index.tsx` file?

src/components/HelperErrorText/HelperErrorText.tsx
```diff
@@ -39,17 +27,21 @@ export interface HelperErrorTextProps {
  * This pattern guarantees accessibility guidelines are met if the text content
  * is dynamically updated by the app or component that implements it.
  */
+
+type ExtendedHelperErrorTextProps = HelperErrorTextProps &
```
EdwinGuzman: Three comments:
- you can `extends` the interface with the second clause (interfaces are not preferable over types, but we've been using them throughout the repo, so just for consistency)
- export the prop so consuming apps can use it if needed,
- I'd rename `HelperErrorTextProps` to something more generic like `BaseHelperErrorTextProps` and then call the new one `HelperErrorTextProps`. When this is exported, it'll make more sense for a consumer to use `HelperErrorTextProps` rather than `ExtendedHelperErrorTextProps`.

EdwinGuzman (changes requested):
Looks good just some minor comments.

src/components/Accordion/Accordion.tsx
oliviawongnyc: What about this instead?

```
type AccordionButtonAttributes = Pick<
  ComponentProps<typeof AccordionButton>,
  "aria-label"
>;

export interface AccordionProps extends AccordionButtonAttributes {
...
}
```

EdwinGuzman: That looks good!

EdwinGuzman (approved):
Looks good!

src/components/Accordion/Accordion.tsx
```diff
@@ -21,12 +21,14 @@ export interface AccordionDataProps {
   label: string | JSX.Element;
   panel: string | React.ReactNode;
 }
-export interface AccordionProps {
+
+type AccordionButtonAttributes = Pick<
+  ComponentProps<typeof AccordionButton>,
+  "aria-label"
+>;
```
bigfishdesign13: Fort my own clarity, is this extracting the `aria-label` attribute from the props that were added to an instance of the `<Accordion>` component?

src/components/HelperErrorText/HelperErrorText.mdx
```diff
@@ -50,40 +50,40 @@ use case of the `HelperErrorText` component is to dynamically render error text
...
-### ariaAtomic
+### ARIA Atomic
 
-The `ariaAtomic` prop sets the `aria-atomic` HTML attribute. This allows the
-entire DOM element to be presented to assistive technologies. When it is set to
-`false`, only additions or removals will be read by assistive technologies.
+By default, `aria-atomic` is set to `true` for the `HelperErrorText` component.
+This allows the entire DOM element to be presented to assistive technologies. If
+desired, you can pass `aria-atomic={false}` instead so that only additions or
+removals will be read by assistive technologies.
```
bigfishdesign13: I know this is how it was written previously, but can you clearly define what `additions and removals` means?

bigfishdesign13 (approved):
Looks good! One minor suggestion.

src/components/HelperErrorText/HelperErrorText.mdx
```diff
@@ -50,42 +50,42 @@ use case of the `HelperErrorText` component is to dynamically render error text
...
-By default, this prop is set to `"polite"`.
-
 <Canvas of={HelperErrorTextStories.AriaControls} />
 
+### ARIA Atomic
+
+When `aria-atomic` is set to `true` within an ARIA live region, the entire DOM
+element is presented to assistive technologies. By default, `aria-atomic` is
+set to `true` for the `HelperErrorText` component. If desired, you can pass
+`aria-atomic={false}` instead so that only the changed node is presented.
```
EdwinGuzman: I saw that @bigfishdesign13 made a comment about "additions and removals". "only the changed node" is helpful but that update might not always be a node. It's possible that the only change inside an element with aria-atomic is a string, such as
```
<div aria-atomic="false">
 This is some string that doesn't change but ${variableThatUpdatesConstantly} does change.
</div>
```

In this case, with `aria-atomic={false}`, only the string value from `variableThatUpdatesConstantly` will be read by screen readers. So that's the "additions and removals" from before.

dgcohen (approved):
Looks great!

PR #1802, opened 2025-04-25
DSD-2056 & DSD-1903: Template + 12-columns

bigfishdesign13:
Fixes JIRA ticket [DSD-2056](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2056) & [DSD-1903](https://newyorkpubliclibrary.atlassian.net/browse/DSD-1903)

- Updated the new `Template` component to use a 12-column grid to underlie the whole layout.
- Updated the `narrow` variant of the new `Template` component to use the grid for sizing, rather than relying on `max-width`.

**NOTE:** This could have been done a few ways, but I chose to continue working with the template areas idea. Also, I did not add anything to the CHANGELOG because this is just an extension of the new `Template` work that is already noted.

## How has this been tested?

- local Storybook

- n/a

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[DSD-2056]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-2056?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DSD-1903]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-1903?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/theme/components/template.ts
+++ b/src/theme/components/template.ts
@@ -14,7 +14,12 @@ const Template = defineStyleConfig({
       minWidth: "288px",
       m: "0 auto",
       p: responsiveMargin,
-      gridTemplateAreas: `"breakout" "top" "main" "bottom"`,
+      gridTemplateAreas: `
+        "breakout"
+        "top"
+        "main"
+        "bottom"
+      `,
       gridTemplateColumns: "100%",
       gridTemplateRows: "auto",
       columnGap: responsiveGap,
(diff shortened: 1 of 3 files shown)


src/theme/components/template.ts
```diff
@@ -24,27 +29,64 @@ const Template = defineStyleConfig({
...
-        xl: "minmax(100px, 1fr) minmax(300px, 3fr)",
+        base: `
+          "breakout"
+          "top"
+          "sidebar"
+          "main"
+          "bottom"
+        `,
+        sm: `
+          "breakout breakout breakout breakout breakout breakout breakout breakout breakout breakout breakout breakout"
```
EdwinGuzman, line 40: Is there a different way to write this? It _looks_ brittle, and I imagine if I need to update this and lose count, I can introduce a layout issue.

bigfishdesign13: Yes, when the wrong number of names are included, the layout blows up and it can be hard to figure out where the issue is coming from. There are other methods that will accomplish the same thing (i.e. named grid lines and spans). That other method would still require accuracy to calculate the correct starting point and span values, but maybe the whole page won't get out of whack if an incorrect value is used. I will do some testing with that method and see what happens.

oliviawongnyc: Would creating a utility function that takes the area name and the number of times it needs to be repeated and returns the proper string help?

```
  md: [
    `"${repeatArea("breakout", 12)}"`,
    `"${repeatArea("top", 12)}"`,
    `"${repeatArea("sidebar", 4)} ${repeatArea("main", 8)}"`,
    `"${repeatArea("bottom", 12)}"`,
  ].join("\n"),
```

src/theme/components/template.ts
```diff
@@ -24,27 +29,64 @@ const Template = defineStyleConfig({
...
       gridTemplateAreas: {
-        base: `"breakout" "top" "sidebar" "main" "bottom"`,
-        md: `"breakout breakout" "top top" "sidebar main" "bottom bottom"`,
-      },
-      gridTemplateColumns: {
-        base: "100%",
-        md: "repeat(2, minmax(100px, 1fr))",
-        lg: "minmax(100px, 1fr) minmax(200px, 2fr)",
-        xl: "minmax(100px, 1fr) minmax(300px, 3fr)",
+        base: `
```
oliviawongnyc, line 32: For the base size (0 - 480px) you just want one column?

src/theme/components/template.ts
```diff
@@ -61,9 +103,20 @@ const TemplateBreakout = defineStyleConfig({
...
+      display: "grid",
+      gridTemplateAreas: {
+        base: `"mainNarrow"`,
+        md: `". mainNarrow mainNarrow mainNarrow mainNarrow mainNarrow mainNarrow mainNarrow mainNarrow mainNarrow mainNarrow ."`,
+        lg: `". . mainNarrow mainNarrow mainNarrow mainNarrow mainNarrow mainNarrow mainNarrow mainNarrow . ."`,
+      },
+      gridTemplateColumns: { base: "100%", md: "repeat(12, 1fr)" },
+      // maxWidth: "720px",
+      // m: "0 auto",
+    };
```
oliviawongnyc, line 119: What about instead of having a `TemplateMainNarrow` component at all, we create a `isNarrow` variant and then add these styles to the `Template` styles? Something like this:

``` 
isNarrow: {
      gridTemplateAreas: {
        base: `
          "breakout"
          "top"
          "main"
          "bottom"
        `,
        md: `
          "breakout breakout breakout breakout breakout breakout breakout breakout breakout breakout breakout breakout" 
          "top top top top top top top top top top top top" 
          ". main main main main main main main main main main ." 
          "bottom bottom bottom bottom bottom bottom bottom bottom bottom bottom bottom bottom"
        `,
        lg: `
          "breakout breakout breakout breakout breakout breakout breakout breakout breakout breakout breakout breakout" 
          "top top top top top top top top top top top top" 
          ". . main main main main main main main main . ." 
          "bottom bottom bottom bottom bottom bottom bottom bottom bottom bottom bottom bottom"
        `,
      },
    },

```

src/theme/components/template.ts
bigfishdesign13: Yes, the small mobile breakpoint range would be one always-stacked column. The template jumps to 2-columns for the next breakpoint (large mobile).

src/theme/components/template.ts
bigfishdesign13: Consuming apps would have to add that as a `className`?
```
<TemplateMain className="isNarrow">...</TemplateMain>
```
Or as a boolean prop?
```
<TemplateMain isNarrow>...</TemplateMain>
```

oliviawongnyc: I am suggesting it would be a prop on `Template` just as `sidebar` is now.

Then it would be passed here:

```
// If `isNarrow` is true, there cannot be a sidebar. If `isNarrow` is false (which should be the default), whatever is passed as `sidebar` should be the variant used ("none" by default).

const templateVariant = isNarrow ? "narrow" : sidebar
const styles = useStyleConfig("Template", { variant: templateVariant  });
```

Could also potentially combine into one prop that specifies sidebar left, sidebar right, narrow, or default (no sidebar, not narrow).

bigfishdesign13: > Could also potentially combine into one prop that specifies sidebar left, sidebar right, narrow, or default (no sidebar, not narrow).

That would eliminate the potential problem with combining sidebar and the narrow layout, which would be nice.

bigfishdesign13:
Moving forward with the `span` syntax in [this PR](https://github.com/NYPL/nypl-design-system/pull/1823).

PR #1806, opened 2025-04-28
DSD-1900: ARIA landmarks guide

7emansell:
Fixes JIRA ticket [DSD-1900](https://newyorkpubliclibrary.atlassian.net/browse/DSD-1900)

- Adds guide to ARIA landmarks to Storybook documentation

### Accessibility Checklist

### Checklist:

### Front End Review:

[DSD-1900]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-1900?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/SearchBar/SearchBar.stories.tsx
```diff
@@ -246,6 +248,28 @@ export const DisabledState: Story = {
   ),
 };
 
+export const WithSearchElement: Story = {
+  render: () => (
+    <search>
+      <SearchBar
```
EdwinGuzman: `SearchBar` still contains `role="search"` so this is technically not the appropriate pattern (and it won't be until v4 where the DOM is updated). I think you should replace `SearchBar` with separate heading, input, button, and checkbox components.

src/components/AccessibilityGuide/AriaLandmarks.mdx
```diff
@@ -0,0 +1,119 @@
...
+- {<Link href="#navigation" target="_self">navigation</Link>}
+- {<Link href="#region" target="_self">region</Link>}
+- {<Link href="#search" target="_self">search</Link>}
+- {<Link href="#resources" target="_self">Resources</Link>}
+
+## Landmarks overview
+
+Landmarks are a useful tool for people who use screen readers to orient themselves to important areas of a web page, and help quickly navigate to these areas within a website or application.
+
+ARIA’s 8 landmark roles are specifically meant to describe page organization and structure, providing keyboard navigation to each larger section. Each of these roles has a corresponding native HTML element:
```
EdwinGuzman: nit: spell out `8` (any number under 10)

src/components/AccessibilityGuide/AriaLandmarks.mdx
```diff
@@ -0,0 +1,119 @@
...
+
+| Landmark role | HTML element        |
+| ------------- | ------------------- |
+| banner        | document `<header>` |
```
EdwinGuzman: what do you mean by "document" here?

src/components/AccessibilityGuide/AriaLandmarks.mdx
```diff
@@ -0,0 +1,119 @@
...
+| contentinfo   | document `<footer>` |
+| form          | `<form>`            |
+| main          | `<main>`            |
+| navigation    | `<nav>`             |
+| region        | `<section>`         |
+| search        | `<search>`          |
+
+Landmarks can be indicated with `role="landmark"` or by using the landmark's HTML element.
+
+**Some Reservoir components already define their own landmark roles, and some landmarks must be defined by the consuming app since they could wrap many components.**
```
EdwinGuzman: This makes it seem like the components are defining roles beyond the eight native ones.

src/components/AccessibilityGuide/AriaLandmarks.mdx
```diff
@@ -0,0 +1,119 @@
...
+
+### banner
+
+The `Header` is Reservoir's only component that falls under the `banner` category: an area that has global, rather than page-specific, information or actionable elements.
+For example, a website's logo and primary navigation would be expected child elements of a `banner` landmark. There should only be a single `banner` landmark per interface.
+
+Our `Header` contains other landmarks: `complementary` for temporary sitewide alerts, `navigation` for the site navigation. (The DS `Banner` is a `complementary` landmark).
```
EdwinGuzman: The DS `Banner` sentence can be its own sentence and perhaps highlight that the name matches a landmark role but only in name and not in semantic meaning.

EdwinGuzman (commented):
This is great, thanks! Just some minor comments.

src/components/SearchBar/SearchBar.stories.tsx
7emansell: Oh true. Then maybe this should go into `v4` since the original intention of this guide was to clarify how the search landmark should be used, and I'd like to keep this specific example with `Searchbar`

src/components/AccessibilityGuide/AriaLandmarks.mdx
7emansell: Does this update make it clearer

src/components/SearchBar/SearchBar.stories.tsx
7emansell: Discussed in standup– this is not gonna go into `v4` so replacing the searchbar with a `TextInput` for now so the pattern is clear

src/components/SearchBar/SearchBar.stories.tsx
```diff
@@ -246,6 +248,19 @@ export const DisabledState: Story = {
   ),
 };
 
+export const WithSearchElement: Story = {
+  render: () => (
+    <search>
+      {/* TODO: Replace with Searchbar component. */}
+      <TextInput id={"search-input"} labelText={"Search"} />
```
EdwinGuzman: nit: you don't need braces around string prop values.

EdwinGuzman: This could also use a simple button component.

EdwinGuzman (approved):
Minor comments but not blockers.

src/components/AccessibilityGuide/AriaLandmarks.mdx
```diff
@@ -0,0 +1,119 @@
...
+| banner        | `<header>`   |
+| complementary | `<aside>`    |
+| contentinfo   | `<footer>`   |
+| form          | `<form>`     |
+| main          | `<main>`     |
+| navigation    | `<nav>`      |
+| region        | `<section>`  |
+| search        | `<search>`   |
+
+Landmarks can be indicated with `role="landmark"` or by using the landmark's HTML element.
```
oliviawongnyc: Is one of these preferred or does it not matter?

src/components/AccessibilityGuide/AriaLandmarks.mdx
```diff
@@ -0,0 +1,119 @@
...
+
+**Some Reservoir components already internally set a landmark role, but some landmarks must be set by the consuming app since they could wrap many components.**
+
+### banner
```
oliviawongnyc, line 47: For the other landmarks, you've structured the description as general info, and then how it applies to Reservoir. Would recommend the same structure here! Would also put more emphasis on the call out of the DS `Banner` piece as this is potentially confusing. So:

"The `banner` landmark is an area that has global, rather than page-specific, information or actionable elements.
For example, a website's logo and primary navigation would be expected child elements of a `banner` landmark. There should only be a single `banner` landmark per interface.

The `Header` is Reservoir's only component that falls under the `banner` category. Our `Header` also contains other landmarks: `complementary` for temporary sitewide alerts and `navigation` for the site navigation. 

Note: The DS `Banner` is a `complementary` landmark, despite its title matching `banner`."

src/components/AccessibilityGuide/AriaLandmarks.mdx
```diff
@@ -0,0 +1,119 @@
...
+
+Our `Header` contains other landmarks: `complementary` for temporary sitewide alerts, `navigation` for the site navigation. The DS `Banner` is a `complementary` landmark, despite its title matching `banner`.
+
+### complementary
+
+The `complementary` landmark indicates an area that contains additional information that "complements" the main content. Though complementary, the content it represents should also be understandable on its own.
+
+The `Banner`, `StatusBadge`, `Notification` , `Hero`, and `FeaturedContent` components could all be `complementary` landmarks.
+
+### contentinfo
```
oliviawongnyc, line 62: Can there only be one of these on the page, like `banner`?

src/components/AccessibilityGuide/AriaLandmarks.mdx
oliviawongnyc: Maybe add something about how landmarks may be nested within one another and there are examples of that below (ie, within the `Header`)?

src/components/AccessibilityGuide/AriaLandmarks.mdx
```diff
@@ -0,0 +1,119 @@
...
+The `main` landmark is meant to designate the primary content of a page. In most browsers, it's natively conveyed by use of the `<main>` HTML element.
+
+There should only be a single `main` landmark per interface. Since `main` represents the primary content of a page, it is the most important destination for [skip links](../?path=/docs/accessibility-guide-skip-navigation--docs), bypassing recurring elements like the header or navigation.
+
+The `Template` [component](../?path=/docs/components-page-layout-template--docs) renders the `main` landmark (and the `header` and `footer`).
+
+### navigation
+
+The `navigation` landmark identifies groups (e.g. lists) of links that are intended to be used for website or page content navigation.
+There can be multiple `navigation`s on a page, but if a `navigation` landmark has the same set of links as another `navigation` landmark on the page (think mobile versus desktop), use the same label for both.
```
oliviawongnyc: label meaning HTML `label` attribute?

src/components/AccessibilityGuide/AriaLandmarks.mdx
```diff
@@ -0,0 +1,119 @@
...
+
+Elements under [Navigation](../?path=/docs/components-navigation) should be wrapped by the `navigation` landmark.
+
+### region
```
oliviawongnyc, line 102: I wonder how to differentiate between complementary and region. Could you include something about how to think about this?

oliviawongnyc (commented):
I learned a lot from this! Some small suggestions included.

src/components/AccessibilityGuide/AriaLandmarks.mdx
7emansell: Added a note below

EdwinGuzman (approved):
Approving latest update, thanks for swapping the searchbar out.

src/components/AccessibilityGuide/AriaLandmarks.mdx
oliviawongnyc: Great! Small typo: "content should supports or modifies" - remove extra s's.

oliviawongnyc (approved):

src/components/AccessibilityGuide/AriaLandmarks.mdx
```diff
@@ -0,0 +1,125 @@
...
+| banner        | `<header>`   |
+| complementary | `<aside>`    |
+| contentinfo   | `<footer>`   |
+| form          | `<form>`     |
+| main          | `<main>`     |
+| navigation    | `<nav>`      |
+| region        | `<section>`  |
+| search        | `<search>`   |
+
+Landmarks can be indicated with `role="landmark"` or by using the landmark's HTML element– both methods mean the same thing in the document.
```
bigfishdesign13: If you can get away with it, replace the dash with `—` (em dash). If that does not render properly, use `&mdash;`. Also, remove the space after the em dash character.

src/components/AccessibilityGuide/AriaLandmarks.mdx
```diff
@@ -0,0 +1,125 @@
...
+
+Some landmarks can be nested inside each other (see `<header>`, for instance) or appear multiple times on a page (distinguished by a unique id), but it depends on the specific landmark.
+
+**Some Reservoir components already internally set a landmark role, but some landmarks must be set by the consuming app since they could wrap many components.**
+
+### banner
+
+The banner landmark is an area that has global, rather than page-specific, information or actionable elements.
+For example, a website's logo and primary navigation would be expected child elements of a `banner` landmark. There should only be a single `banner` landmark per interface.
```
bigfishdesign13: It was my understanding, and it is also stated on [w3schools.com](https://www.w3schools.com/tags/tag_header.asp), that it is ok to have multiple `<header>` elements on one page. For example, each significant section on a page may have a `<header>` element.

src/components/AccessibilityGuide/AriaLandmarks.mdx
```diff
@@ -0,0 +1,125 @@
...
+
+The Header is Reservoir's only component that falls under the banner category. Our `Header` contains other landmarks: `complementary` for temporary sitewide alerts, `navigation` for the site navigation.
+
+Note: The DS `Banner` is a `complementary` landmark, despite its title matching `banner`.
+
+### complementary
+
+The `complementary` landmark indicates an area that contains additional information that "complements" the main content. Though complementary, the content it represents should also be understandable on its own.
+
+The `Banner`, `StatusBadge`, `Notification` , `Hero`, and `FeaturedContent` components could all be `complementary` landmarks.
```
bigfishdesign13: Out of the box, `Banner` and `Notification` are `<aside>` elements, so they are inherently `complementary` landmarks. Would consuming apps need to explicitly add the role to the other components that are mentioned?

src/components/AccessibilityGuide/AriaLandmarks.mdx
```diff
@@ -0,0 +1,125 @@
...
+
+### contentinfo
+
+`contentinfo` is another globally repeating landmark, like `banner`, with information about the parent document. If `banner` is usually the site's header, `contentinfo` is usually the footer. There should only be a single `contentinfo` landmark per interface.
```
bigfishdesign13: Same thing here...it is generally accepted that more than one `<footer>` element can be added to a page.

src/components/AccessibilityGuide/AriaLandmarks.mdx
```diff
@@ -0,0 +1,125 @@
...
+
+<Banner
+  content={
+    <>
+      <strong>IMPORTANT:</strong> Use the search landmark instead of the form
+      landmark when the form is used for search functionality.
+    </>
+  }
+  type="warning"
+/>
```
bigfishdesign13, line 81: Nice!

src/components/AccessibilityGuide/AriaLandmarks.mdx
```diff
@@ -0,0 +1,125 @@
...
+
+Elements under [Form Elements](../?path=/docs/components-form-elements) should be wrapped by the `form` landmark.
+
+### main
+
+The `main` landmark is meant to designate the primary content of a page. In most browsers, it's natively conveyed by use of the `<main>` HTML element.
+
+There should only be a single `main` landmark per interface. Since `main` represents the primary content of a page, it is the most important destination for [skip links](../?path=/docs/accessibility-guide-skip-navigation--docs), bypassing recurring elements like the header or navigation.
+
+The `Template` [component](../?path=/docs/components-page-layout-template--docs) renders the `main` landmark (and the `header` and `footer`).
```
bigfishdesign13, line 93: This will need to be updated for v4, as the header and footer regions are not part of the updated `Template` component.

src/components/AccessibilityGuide/AriaLandmarks.mdx
```diff
@@ -0,0 +1,125 @@
...
+
+### navigation
+
+The `navigation` landmark identifies groups (e.g. lists) of links that are intended to be used for website or page content navigation.
+There can be multiple `navigation`s on a page, but if a `navigation` landmark has the same set of links as another `navigation` landmark on the page (think mobile versus desktop), use the same HTML `label` attribute for both.
```
bigfishdesign13: Nit: change "multiple `navigation`s" to "multiple `navigation` elements"

src/components/AccessibilityGuide/AriaLandmarks.mdx
```diff
@@ -0,0 +1,125 @@
...
+
+Elements under [Navigation](../?path=/docs/components-navigation) should be wrapped by the `navigation` landmark.
```
bigfishdesign13: Only two components in our `Navigation` category do not have a `<nav>` element baked in. In fact, one of the two that doe snot have it should be updated to have it added. So that would leave `Link` as the only component that needs to potentially be wrapped in a `navigation` landmark, and even that would depend on how it was being used.

This is a long winded way of saying this comment should be reworded.

src/components/AccessibilityGuide/AriaLandmarks.mdx
bigfishdesign13: Further, should `region` landmarks be nested inside the `main` landmark or do they typically sit outside `main`?

src/components/AccessibilityGuide/AriaLandmarks.mdx
```diff
@@ -0,0 +1,125 @@
...
+
+`region` should always be applied by the consuming app, since it is specific to the content being presented.
+
+Note: There's not always a clear line between `complementary` content and content that deserves its own `region`, but a good rule of a thumb is that `complementary` content should support or modify the main content of the page, while `region` should represent its own separate information.
+
+### search
+
+The `search` landmark contains the page's search and filter functionality, which will likely include form controls (such as an `input type="search"`).
+
+Inspect below to see `<search>` in the DOM– notice that it wraps the `Heading`, the `TextInput`, the `Button`, and the `Checkbox`, not just the searchbar:
```
bigfishdesign13: Replace the dash with an em dash and remove the trailing space.

If we are going to ask users to look for the DS components, let's tell them to click on the `Show code` button in the lower right corner. Conversely, if we are going to ask them to inspect the DOM, then let's ask them to look for the native HTML elements.

bigfishdesign13 (changes requested):
Great information. I just have some suggestions.

src/components/AccessibilityGuide/AriaLandmarks.mdx
EdwinGuzman: I'd follow [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/banner_role)'s doc and not w3schools (they have a bad reputation). That said, you can have more than one header/banner role on a page but if so, it should have a unique label. So that should be updated here (and for contentinfo)

src/components/AccessibilityGuide/AriaLandmarks.mdx
7emansell: Gotcha– since this isn't going into `v4` I'm gonna leave it for now and make a note when we do update `Template`

bigfishdesign13 (approved):
💯 !

PR #1808, opened 2025-04-30
DSD-2056: 12-column grid using named grid lines syntax

bigfishdesign13:
Fixes JIRA ticket DSD-2056

- Use named grid lines instead of the grid template areas repeating syntax.

## How has this been tested?

- local Storybook

- n/a

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

### DIFF
--- a/src/theme/components/template.ts
+++ b/src/theme/components/template.ts
@@ -14,79 +14,41 @@ const Template = defineStyleConfig({
       minWidth: "288px",
       m: "0 auto",
       p: responsiveMargin,
-      gridTemplateAreas: `
-        "breakout"
-        "top"
-        "main"
-        "bottom"
-      `,
-      gridTemplateColumns: "100%",
-      gridTemplateRows: "auto",
+      // Using named grid lines to implicitly define grid template areas
+      gridTemplateColumns:
+        "[breakout-start top-start sidebar-start main-start bottom-start] 1fr [breakout-end top-end sidebar-end main-end bottom-end]",
+      gridTemplateRows:
+        "[breakout-start] auto [breakout-end top-start] auto [top-end main-start] auto [main-end sidebar-start] auto [sidebar-end bottom-start] auto [botom-end]",
       columnGap: responsiveGap,
       "& > *:not(:last-child)": { mb: responsiveGap },
     };
   }),
   variants: {
     left: {
-      gridTemplateAreas: {
-        base: `
-          "breakout"
-          "top"
-          "sidebar"
-          "main"
-          "bottom"
-        `,
-        sm: `
-          "breakout breakout breakout breakout breakout breakout breakout breakout breakout breakout breakout breakout" 
-          "top top top top top top top top top top top top" 
-          "sidebar sidebar sidebar sidebar sidebar sidebar main main main main main main" 
-          "bottom bottom bottom bottom bottom bottom bottom bottom bottom bottom bottom bottom"
-        `,
-        md: `
-          "breakout breakout breakout breakout breakout breakout breakout breakout breakout breakout breakout breakout" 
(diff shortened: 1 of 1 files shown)


oliviawongnyc (commented):
This should point to `reservoir-v4`, right?

bigfishdesign13:
> This should point to `reservoir-v4`, right?

@oliviawongnyc It should actually point to the initial 12-column branch. I fixed it.

src/theme/components/template.ts
```diff
@@ -14,79 +14,41 @@ const Template = defineStyleConfig({
...
-        "breakout"
-        "top"
-        "main"
-        "bottom"
-      `,
-      gridTemplateColumns: "100%",
-      gridTemplateRows: "auto",
+      // Using named grid lines to implicitly define grid template areas
+      gridTemplateColumns:
+        "[breakout-start top-start sidebar-start main-start bottom-start] 1fr [breakout-end top-end sidebar-end main-end bottom-end]",
```
EdwinGuzman, line 19: I'm not opposed to this but from the syntax info here https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-columns#syntax, it can get very complicated. I think this is minor, but we should have more documentation for this if this is the approach going forward.

bigfishdesign13: Right, there are A LOT of viable formatting options for the `grid-template-columns` attribute. But which method is easier to understand? I am so deep in this that it is hard for me to make a judgment. I have decent working knowledge of both methods at this point.

For clarity, this is the page that I used as a reference...
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_grid_layout/Grid_layout_using_named_grid_lines

@oliviawongnyc, @7emansell, and @dgcohen - It would be great to get your thoughts on this. The other method for applying the grid areas is in the [original PR](https://github.com/NYPL/nypl-design-system/pull/1802/files/1288275245adb2df8e71dd2f177c4e62b5cda9dc#diff-76de9d322d4aa66b4eb366987b6631316d030d890f8dfae8a644015f19f3c101). See the `variants` section starting on line 29.

EdwinGuzman: I'll leave this up to everyone else to decide. What's confusing to me is the `-start` and `-end` suffixes. Doesn't that mean that the currently named `breakout`, `top`, `sidebar`, etc need to have those suffixes as well?

bigfishdesign13: This update removed the explicitly named grid areas and in their place uses an implicit method for defining the grid areas. It is the `-start` and `-end` suffixes that execute that implicit method. For example, `breakout-start` and `breakout-end` define the `breakout` grid area.

7emansell: I'm finding this method more readable than the original PR but second the request for a lot of documentation

7emansell (approved):
lgtm as a template approach

oliviawongnyc:
What I was wondering is if you could use `gridTemplateColumns: "repeat(12, 1fr)",` on the `Template` and then `gridColumn` on each of the children to specify which columns they should start and end at. Is that what you tried already?

bigfishdesign13:
> What I was wondering is if you could use `gridTemplateColumns: "repeat(12, 1fr)",` on the `Template` and then `gridColumn` on each of the children to specify which columns they should start and end at. Is that what you tried already?

@oliviawongnyc Using the span notation (start / stop) was how I put together my original POC many months back. I then ran with the method you used in the initial dev for the updated `Template` component. And now we have the method that is being used here. Maybe it's overkill, but I'd like to put together one more POC using the span method to see what that feels like now that we are privy to these other two methods.

dgcohen (approved):

PR #1810, opened 2025-05-05
DSD-2036: Removes useNYPLBreakpoints hook where possible

oliviawongnyc:
Fixes JIRA ticket [DSD-2036](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2036)

- Removes use of `useNYPLBreakpoints` hook where possible without causing breaking change.
- Does NOT remove `useNYPLBreakpoints` in `MultiSelectGroup` and `FilterbarInline` where breaking changes would occur (will make these changes in v4)

## How has this been tested?

- Locally in Storybook

- I did not remove the use of `useNYPLBreakpoints` in one instance in `Searchbar` where it is being used to dynamically add or remove an `aria-label` (`aria-label={isLargerThanMobile ? "" : "Search"}`) because I think this is an ok use-case for it.

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[DSD-2036]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-2036?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -13,7 +13,6 @@ import {
 import React, { forwardRef } from "react";
 import Button from "../Button/Button";
 import ButtonGroup from "../ButtonGroup/ButtonGroup";
-import useNYPLBreakpoints from "../../hooks/useNYPLBreakpoints";
 import useDSHeading from "../../hooks/useDSHeading";
 
 export interface BaseProps {
@@ -99,11 +98,7 @@ export const BaseModal: ChakraComponent<
     isOpen,
     ...rest
   } = props;
-  const xlarge = "xl";
-  const fullSize = "full";
-  const { isLargerThanMobile } = useNYPLBreakpoints();
-  // For larger screens, set the size to xl, otherwise set it to full.
-  const size = isLargerThanMobile ? xlarge : fullSize;
+
   const finalTitle = useDSHeading({
     title: headingText,
     id,
@@ -127,7 +122,7 @@ export const BaseModal: ChakraComponent<
           isOpen={isOpen}
           onClose={onClose}
           scrollBehavior="inside"
-          size={size}
+          size={{ base: "full", md: "xl" }}
           {...rest}
         >
           <ModalOverlay />
@@ -150,7 +145,7 @@ export const BaseModal: ChakraComponent<
           isOpen={isOpen}
           onClose={onCancel}
           scrollBehavior="inside"
-          size={size}
+          size={{ base: "full", md: "xl" }}
           {...rest}
         >
           <ModalOverlay />
--- a/src/components/MultiSelectGroup/MultiSelectGroup.tsx
+++ b/src/components/MultiSelectGroup/MultiSelectGroup.tsx
@@ -69,7 +69,6 @@ export const MultiSelectGroup: ChakraComponent<
           className={className}
           columnGap="xs"
           data-testid={`multiselect-group-${id}`}
-          direction={finalLayout}
           id={id}
           ref={ref}
           rowGap="xs"
@@ -80,6 +79,7 @@ export const MultiSelectGroup: ChakraComponent<
               : null
           }
           sx={{
+            flexDirection: { base: "column", md: layout },
             "> div": {
               _notFirst: {
                 mx: "0",
--- a/src/components/RadioGroup/RadioGroup.tsx
+++ b/src/components/RadioGroup/RadioGroup.tsx
@@ -160,7 +160,11 @@ export const RadioGroup: ChakraComponent<
             {showRequiredLabel && isRequired && <span> (required)</span>}
           </Box>
           <ChakraRadioGroup {...radioGroupProps}>
-            <Stack direction={[layout]} spacing={spacingProp}>
+            <Stack
+              className="radioGroupStack"
+              direction={[layout]}
+              spacing={spacingProp}
+            >
               {newChildren}
             </Stack>
           </ChakraRadioGroup>
--- a/src/theme/components/feedbackBox.ts
+++ b/src/theme/components/feedbackBox.ts
@@ -8,6 +8,7 @@ const { defineMultiStyleConfig, definePartsStyle } =
     "drawerContent",
     "drawerHeader",
     "openButton",
+    "radioGroup",
   ]);
 
 const FeedbackBox = defineMultiStyleConfig({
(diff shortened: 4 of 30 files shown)


src/components/Modal/Modal.tsx
```diff
@@ -127,7 +122,7 @@ export const BaseModal: ChakraComponent<
           isOpen={isOpen}
           onClose={onClose}
           scrollBehavior="inside"
-          size={size}
+          size={{ base: "full", md: "xl" }}
```
bigfishdesign13, line 125: Have we ever used the Charka responsive object syntax for anything other than styles? This seems new to me. And it works?

src/components/MultiSelectGroup/MultiSelectGroup.tsx
```diff
@@ -69,7 +69,6 @@ export const MultiSelectGroup: ChakraComponent<
           className={className}
           columnGap="xs"
           data-testid={`multiselect-group-${id}`}
-          direction={finalLayout}
```
bigfishdesign13, line 72: This is removed, but the `useNYPLBreakpoints` hook and associated conditions are still on lines 56–59. Should that code also be removed?

src/theme/components/searchBar.ts
```diff
@@ -21,6 +22,14 @@ const SearchBar = defineMultiStyleConfig({
           " > span": {
             display: "none",
           },
+          "> svg": {
+            ...size["medium"],
+          },
```
bigfishdesign13: Interesting. This really is just a cleanup and better execution of the container query styles. Nice.

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -160,7 +160,11 @@ export const RadioGroup: ChakraComponent<
             {showRequiredLabel && isRequired && <span> (required)</span>}
           </Box>
           <ChakraRadioGroup {...radioGroupProps}>
-            <Stack direction={[layout]} spacing={spacingProp}>
+            <Stack
+              className="radioGroupStack"
```
bigfishdesign13, line 164: I know we've done things like this before where a generic custom class is added to an element which then allows us to target the element in a situation where it is being used to compose another component. (Phew, longwinded!) Is this class name following the same naming convention that we have used for other class names like this?

bigfishdesign13 (commented):
Some comments and questions.

src/components/Modal/Modal.tsx
oliviawongnyc: It works! I'm not sure if we have used it elsewhere or not, but `size` is a Chakra prop whose value can either be a string or an object of responsive styles: https://github.com/chakra-ui/chakra-ui/blob/v2/packages/styled-system/src/theming-props.ts.

src/components/MultiSelectGroup/MultiSelectGroup.tsx
oliviawongnyc: Yes, we are still using the hook elsewhere in this component. To remove it completely, I will need to implement a breaking change and will do so in v4.

src/theme/components/searchBar.ts
oliviawongnyc: Yes! I noticed we were doing some styling with container queries, some with media queries so the radius was changing at a slightly different point than everything else.

src/components/RadioGroup/RadioGroup.tsx
oliviawongnyc: These are the classNames I could find that we added for styling purposes:

`Subnav` uses a `className` of `selectedItem`
`StructuredContent` uses a `className` of `structuredcontent-body`
`SocialMediaLinks` uses a `className` of `platLink`
`SearchBar` uses a `className` of `textInput`

Doesn't seem like we have too much of a convention but happy to adjust if something better comes to mind.

bigfishdesign13:
The refactored responsive styles are working as expected. However, the `MultiSelectGroup` component is still flashing the mobile styles before rendering the desktop styles. It is a complex component with a lot of javascript under the hood and it looks like even the Chakra responsive style object is getting hung up similar to the `useNYPLBreakpoints` hook, with the page needing to slog through a bunch of javascript before showing the final render of the styles. It's still `css-in-js`.

Cc @EdwinGuzman

src/components/RadioGroup/RadioGroup.tsx
bigfishdesign13: This discussion is related to [DSD-2038](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2038). That ticket says, "Initial thought is to remove prefixes from classnames/ids that use them," which is interesting because my feeling here was to add some type of prefix to all of the classes in order to ensure that they were not too general and would not accidentally pick up unrelated styles from elsewhere.

What are you thinking?

src/theme/components/searchBar.ts
EdwinGuzman: Yes, this is great. My only suggestion is to move the `size` object to a more generic file so that `searchBar` doesn't import from `icon`.

src/components/RadioGroup/RadioGroup.tsx
EdwinGuzman: Yeah, that is my "initial thought" and it definitely doesn't mean it's correct! It's to start the discussion and if `radioGroupStack` works, or `radioGroup-stack`, or some other variant, then let's go for it. The current name works and refining this name and others can be done in DSD-2038.

src/theme/components/feedbackBox.ts
```diff
@@ -79,6 +80,11 @@ const FeedbackBox = defineMultiStyleConfig({
         zIndex: "10000 !important",
       },
     },
+    radioGroup: {
```
EdwinGuzman, line 83: This is ok but I think there's a side effect where the gap is reduced between radio options:
Screenshot 2025-05-08 at 3 22 01 PM

EdwinGuzman: PR on the left, prod on the right.

EdwinGuzman (commented):
Thanks! This is what I was hoping for and the flash and jump from components using the appropriate layout as the page loads is gone (mostly, I know more updates need to be made).

src/theme/components/feedbackBox.ts
bigfishdesign13: Good catch. The `layout` prop for the `RadioGroup` component not only adjusts the `flexDirection`. It also adjusts the flex `gap` value (`"s"` vs `"l"`). That will need to also be set here.

oliviawongnyc: Yes good catch!

bigfishdesign13 (approved):
This looks good!

EdwinGuzman (approved):
Thanks! Big update!

PR #1811, opened 2025-05-09
DSD-2036/remove children and classname props

oliviawongnyc:
Fixes JIRA ticket [DSD-2036](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2037)

- Removes explicit `className` and `children` props in favor of expanded prop type definitions
- Attempted to expand most prop definitions by Chakra's `BoxProps` as it's a good baseline set of props - includes `className` and Chakra styling props.
- This should not be a breaking change, but we are reserving for v4 because it is a wide-scale change that affects every component.

## How has this been tested?

- Locally, by removing explicit className and children props and attempting to pass them still in the component stories

- None.

### Accessibility Checklist

## Open Questions

- Do we update the documentation so consumers know they can still pass `className`s even though they are not explicitly listed in the props list?

### Checklist:

### Front End Review:

[DSD-2036]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-2036?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
   - `StatusBadge`'s `level` prop
   - `StyledList`'s "tag", "mini" values in `textSizesArray`
 - Removes `role="search"` from `Searchbar` wrapper.
+- Removes explicit `className` and `children` props in favor of expanded prop type definitions
```
bigfishdesign13: I feel that this should include something along the lines of `in all components`.

src/components/AudioPlayer/audioPlayerChangelogData.ts
```diff
@@ -9,6 +9,13 @@
 import { ChangelogData } from "../../utils/ComponentChangelogTable";
 
 export const changelogData: ChangelogData[] = [
+  {
+    date: "Prerelease",
+    version: "Prerelease",
+    type: "Update",
+    affects: ["Functionality"],
+    notes: ["Removed explicit `classname` prop."],
```
bigfishdesign13: Should be `className`.

src/components/Banner/bannerChangelogData.ts
```diff
@@ -9,6 +9,15 @@
 import { ChangelogData } from "../../utils/ComponentChangelogTable";
 
 export const changelogData: ChangelogData[] = [
+  {
+    date: "Prerelease",
+    version: "Prerelease",
+    type: "Update",
+    affects: ["Functionality"],
+    notes: [
+      "Removes explicit `className` prop as interface can be extended to include Chakra prop or HTML attribute types.",
```
bigfishdesign13, line 18: Why is this description so verbose as compared to the description added for `AudioPlayer`?

src/components/Breadcrumbs/Breadcrumbs.mdx
```diff
@@ -47,6 +47,8 @@ const breadcrumbsData = [
 
 ## Component Props
 
+You may pass any Chakra `Box` prop, as well as the props below.
```
bigfishdesign13, line 50: This note is formatted different from the others.

src/components/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -15,8 +15,6 @@ import { LayoutTypes } from "../../helpers/types";
 import { spacing } from "../../theme/foundations/spacing";
 
 export interface CheckboxGroupProps {
```
bigfishdesign13: Why does this one extend `BoxProps`?

src/components/DatePicker/DatePicker.tsx
```diff
@@ -51,9 +52,9 @@ interface DateRangeRowProps {
 
 // Interface used by the `div` or `fieldset` parent wrapper element.
 // Internal use only.
-interface DatePickerWrapperProps extends DateRangeRowProps {
-  /** Additional className. */
-  className?: string;
+interface DatePickerWrapperProps
+  extends Omit<BoxProps, "id" | "onChange">,
```
bigfishdesign13, line 56: Just curious...does TypeScript flag an error when these props are not omitted? I am wandering how you identify the props to include in the `omit` statements?

src/components/Grid/SimpleGrid.mdx
```diff
@@ -39,6 +39,8 @@ row) will be triggered.
 
 <Canvas of={SimpleGridStories.WithControls} />
 
+You may pass any Chakra `Box` prop, as well as the props below.
```
bigfishdesign13, line 42: This message is shorter than the message used in other components.

src/components/Heading/Heading.mdx
```diff
@@ -41,6 +41,8 @@ semantic elements can be overwritten using the `size` prop.
 
 ## Component Props
 
+You may pass any Chakra `Box` prop, as well as the props below.
```
bigfishdesign13, line 44: This message is shorter than the message used in other components.

src/components/Icons/Icon.mdx
```diff
@@ -34,6 +34,8 @@ import Link from "../Link/Link";
 
 <Canvas of={IconStories.WithControls} />
 
+You may pass any Chakra style props, as well as the props below.
```
bigfishdesign13, line 37: This message is shorter than the message used in other components.

src/components/Icons/Icon.tsx
```diff
@@ -24,11 +26,10 @@ export type IconRotations = typeof iconRotationsArray[number];
 export type IconSizes = typeof iconSizesArray[number];
 export type IconTypes = typeof iconTypesArray[number];
 
-export interface IconProps {
+export interface IconProps
+  extends Pick<BoxProps, "className" | keyof ChakraProps> {
```
bigfishdesign13, line 30: Why was this necessary?

src/components/Link/Link.mdx
```diff
@@ -12,10 +12,10 @@ import * as LinkStories from "./Link.stories";
...
 # Link
 
-| Component Version | DS Version |
-| ----------------- | ---------- |
-| Added             | `0.0.4`    |
-| Latest            | `3.4.2`    |
+| Component Version | DS Version   |
+| ----------------- | ------------ |
+| Added             | `0.0.4`      |
+| Latest            | `Prerelease` |
```
bigfishdesign13, line 18: The props message was not added.

src/components/Link/Link.tsx
```diff
@@ -27,9 +28,9 @@ export const linkTypesArray = [
 ] as const;
 export type LinkTypes = typeof linkTypesArray[number];
 
-export interface LinkProps extends ChakraLinkProps {
-  /** Additional class name to render in the `Link` component. */
-  className?: string;
+export interface LinkProps
+  extends Pick<BoxProps, "as" | keyof ChakraProps>,
```
bigfishdesign13: Why did you need to single out the `as` prop? Is that the only prop you wanted from `BoxProps` because this component is better off extending `AnchorHTMLAttributes`? Additionally, seeing this here makes me question whether we should allow the `as` prop in other components. It doesn't make sense for most of them.

src/components/List/List.mdx
```diff
@@ -9,10 +9,10 @@ import { changelogData } from "./listChangelogData";
...
 # List
 
-| Component Version | DS Version |
-| ----------------- | ---------- |
-| Added             | `0.7.0`    |
-| Latest            | `3.6.0`    |
+| Component Version | DS Version   |
+| ----------------- | ------------ |
+| Added             | `0.7.0`      |
+| Latest            | `Prerelease` |
```
bigfishdesign13, line 15: The blurb about the `Box` props is missing.

src/components/Menu/Menu.mdx
```diff
@@ -36,6 +36,8 @@ The component allows users to perform multiple actions from an element with a sm
 
 <Canvas of={MenuStories.WithControls} />
 
+The `Menu composes a Chakra `Box`component, so you may pass any Chakra`Box` prop, as well as the props below.
```
bigfishdesign13: Missing the closing tic mark for `Menu`. It also looks like that may have affected the auto-formatting throughout the sentence.

src/components/Menu/Menu.tsx
```diff
@@ -297,7 +296,7 @@ export const Menu: ChakraComponent<
 
       const initialRef = useRef();
       return (
-        <Box ref={ref}>
+        <Box className={className} ref={ref}>
```
bigfishdesign13: Why was `className` added here? It will also be added as part of `...rest` below.

src/components/RadioGroup/RadioGroup.mdx
```diff
@@ -9,10 +9,10 @@ import { changelogData } from "./radioGroupChangelogData";
...
 # RadioGroup
 
-| Component Version | DS Version |
-| ----------------- | ---------- |
-| Added             | `0.25.0`   |
-| Latest            | `3.5.2`    |
+| Component Version | DS Version   |
+| ----------------- | ------------ |
+| Added             | `0.25.0`     |
+| Latest            | `Prerelease` |
```
bigfishdesign13, line 15: The blurb about `Box` props is missing.

src/components/SkeletonLoader/SkeletonLoader.mdx
```diff
@@ -33,11 +33,15 @@ be customized.
 
 Please note that the initial rendering below has the `width` prop set to `300px`
 to better view the example. The default value is `100%`.
+zxggk
```
bigfishdesign13: Inadvertent addition?

src/components/SocialMediaLinks/SocialMediaLinks.tsx
```diff
@@ -1,4 +1,9 @@
-import { ChakraComponent, chakra, useStyleConfig } from "@chakra-ui/react";
+import {
+  ChakraComponent,
+  chakra,
+  useStyleConfig,
+  BoxProps,
```
bigfishdesign13: Nit: move this into its appropriate alphabetical location in the list

src/components/StyledList/StyledList.tsx
```diff
@@ -35,44 +37,45 @@ export const StyledList: ChakraComponent<
...
-        variant: style,
-      });
-      let finalChildren;
+  forwardRef<
+    HTMLDivElement & HTMLUListElement,
+    React.PropsWithChildren<StyledListProps>
+  >((props, ref?) => {
+    const {
+      children,
+      className,
```
bigfishdesign13: Why is this still included? Where is `...rest`?

src/components/StyledList/StyledList.tsx
```diff
@@ -35,44 +37,45 @@ export const StyledList: ChakraComponent<
...
 
-      return (
-        <Box as="ul" className={className} id={id} ref={ref} __css={styles}>
-          {finalChildren}
-        </Box>
-      );
-    }
-  )
+    return (
+      <Box as="ul" className={className} id={id} ref={ref} __css={styles}>
```
bigfishdesign13: `className` is still here and there is no `...rest` prop.

bigfishdesign13 (changes requested):
Some suggestions and a lot of questions. Also, the message above the prop tables are not all consistent. Some are simplified when compared to the others and I was wondering if that was intentional.

src/components/Banner/bannerChangelogData.ts
oliviawongnyc: I wanted all of them to say this, but missed `AudioPlayer`. `AudioPlayer` is now updated.

src/components/Breadcrumbs/Breadcrumbs.mdx
oliviawongnyc: I intentionally used a different note because I didn't feel that the language "composes `Box`" was appropriate as `ChakraBreadcrumb` does not use a `Box`: https://github.com/chakra-ui/chakra-ui/blob/v2/packages/components/src/breadcrumb/breadcrumb.tsx

src/components/CheckboxGroup/CheckboxGroup.tsx
oliviawongnyc: Originally, I did not extend this by `BoxProps` because I wasn't removing a `className` prop that would necessitate extending, but I have updated this now for consistency.

src/components/DatePicker/DatePicker.tsx
oliviawongnyc: The props I am omitting have a name that matches one of the props we are passing explicitly, but have a different type. As an example, if BoxProps' interface contains a `value` prop that could be "string" or "number" and we have a `value` prop that needs to be "string" and perhaps we cannot allow a number to be passed because the underlying Chakra component won't accept any other type. In these scenarios, we are saying omit the prop of this name from BoxProps' interface and use the one we are providing explicitly.

src/components/Grid/SimpleGrid.mdx
oliviawongnyc: See comment above. Happy to adjust, but I don't want to misrepresent the component by saying there is an underlying `Box` if there isn't.

src/components/Icons/Icon.tsx
oliviawongnyc: In this case, `BoxProps` conflicts with many of the props that Chakra's `Icon` expects. For example, `onCopy` exists in both interfaces, but the types are incompatible. Another example from Typescript: `Type 'ClipboardEventHandler<HTMLDivElement>' is not assignable to type 'ClipboardEventHandler<SVGElement>'.` Therefore, instead of adding `BoxProps` and omitting a ton of props, I picked out the props I wanted from `Box`, which are just the `className` and the style props. There are other ways to do this. For example, I could extend by Chakra's `IconProps` or the HTML svg attributes, but I did this for consistency with the other components.

src/components/Link/Link.tsx
oliviawongnyc: I think I added the anchor attributes here because ChakraLinkProps was being extended previously and I wanted to replace that with the HTML equivalent, which is what we discussed doing (extend by `Box`, then extend by the HTML attributes if helpful), but really for this PR I don't need the HTML attributes so I removed.

As for the `as` prop – it looks like all Chakra components accept `as`. You can see here that their `HTMLChakraProps` includes `AsProps` and I believe `HTMLChakraProps` are part of every component: https://github.com/chakra-ui/chakra-ui/blob/87215e02b6f5e1eefada6ef2f1b1e23fa124d7a5/packages/components/src/system/system.ts#L123

src/components/SkeletonLoader/SkeletonLoader.mdx
oliviawongnyc: Oops

src/components/StyledList/StyledList.tsx
oliviawongnyc: Removed `className`, added `...rest`.

src/components/DatePicker/DatePicker.tsx
oliviawongnyc: And yes, Typescript throws an error.

src/components/Menu/Menu.tsx
oliviawongnyc: Hmm not sure! I removed. Good catch.

src/components/Breadcrumbs/Breadcrumbs.mdx
bigfishdesign13: Ok, understood. I was caught up on the fact that we are extending `BoxProps`, but not thinking about actual composition.

src/components/Link/Link.tsx
bigfishdesign13: If a DS component is extending `BoxProps`, but not composing `Box`, what happens? And if it is composing `Box`?

oliviawongnyc: In either scenario, consumers can pass any `Box` prop and it'll be passed wherever `...rest` is. I think this is ok because every component that does not "compose a `Box`" either boils down to a `div` underneath it all or boils down to another element (`nav`, `p`, etc.) that can still take any `div` attribute.

bigfishdesign13: I am most curious about what happens if the `as` props is passed.

oliviawongnyc: Oh I see. You're asking about what happens if a consumer passes the "as" prop in other components. I tried passing the `as` prop to a few different components (`Breadcrumbs` and `Text`, which do not compose `Box` and `Label` which does compose `Box`). Basically, it does in fact change the underlying element to whatever is passed. Sometimes this completely breaks the component, sometimes the component still renders, but does not behave as we would expect. I'm not too worried about this as I don't see people trying to use the `as` prop more broadly. What do you think?

bigfishdesign13: Interesting. I say leave it as is. If devs do try to use it, I'd be curious to see if they find any useful applications.

bigfishdesign13 (approved):
Nice work.

src/components/Banner/Banner.tsx
```diff
@@ -70,12 +71,10 @@ export const bannerHighlightColorsArray = [
 ] as const;
 export type BannerHighlightColors = typeof bannerHighlightColorsArray[number];
 
-export interface BannerProps {
+export interface BannerProps extends Omit<BoxProps, "content"> {
```
EdwinGuzman, line 74: This is interesting, `content` is usually used for the `meta` tag, so why is it explicitly removed here?

src/components/CheckboxGroup/CheckboxGroup.mdx
```diff
@@ -37,6 +37,9 @@ import Link from "../Link/Link";
 
 <Canvas of={CheckboxGroupStories.WithControls} />
 
+The `CheckboxGroup` composes a Chakra `Box` component, so you may pass
+any Chakra `Box` prop, as well as the props below.
```
EdwinGuzman: Not sure if this is needed, but perhaps a message that says that the `onChange` prop has a different shape might be helpful. I'm reading this as, "I can use any Box prop so I can pass what I would normally expect `onChange` to be", but that's not exactly the case.

src/components/Image/Image.tsx
```diff
@@ -84,8 +84,9 @@ interface ImageWrapperProps {
 }
 
 export interface ImageProps
-  extends ImageWrapperProps,
-    ImgHTMLAttributes<HTMLImageElement> {
+  extends Pick<BoxProps, keyof ChakraProps>,
+    ImageWrapperProps,
+    Omit<ImgHTMLAttributes<HTMLImageElement>, "color" | "height" | "width"> {
```
EdwinGuzman, line 89: Why are these three omitted?

src/components/Link/Link.tsx
```diff
@@ -27,9 +28,8 @@ export const linkTypesArray = [
 ] as const;
 export type LinkTypes = typeof linkTypesArray[number];
 
-export interface LinkProps extends ChakraLinkProps {
-  /** Additional class name to render in the `Link` component. */
-  className?: string;
+export interface LinkProps
+  extends Pick<BoxProps, "as" | "className" | keyof ChakraProps> {
```
EdwinGuzman, line 32: `ChakraLinkProps` seems more inlined with what this component needs, so was this update to `BoxProps` necessary? because of `classname`?

src/components/Menu/Menu.mdx
EdwinGuzman: You're missing a closing backtick for `Menu`

src/components/Select/Select.mdx
```diff
@@ -40,6 +40,9 @@ is left blank, a value will be generated for you.
 
 <Canvas of={SelectStories.WithControls} />
 
+`Select` composes an HTML `select` element, so you may pass any HTML `select` attribute, as well
```
EdwinGuzman, line 43: nit: missing second "as" in  "as well as any".

EdwinGuzman (commented):
Thanks! This is a massive PR and the different language for each type of component update is helpful and makes sense. Just have some comments/questions.

src/components/Banner/Banner.tsx
oliviawongnyc: There is apparently a `content` that is included in `BoxProps` that conflicts with the `content` in our props list. This is the error I get when I don't omit it:

```
Interface 'BannerProps' incorrectly extends interface 'BoxProps'.
  Type 'BannerProps' is not assignable to type 'Omit<PropsOf<"div">, "ref" | keyof StyleProps>'.
    Types of property 'content' are incompatible.
      Type 'string | Element' is not assignable to type 'string'.
        Type 'Element' is not assignable to type 'string'.ts(2430)
```

src/components/Image/Image.tsx
oliviawongnyc: "color" | "height" | "width" conflict with the props of the same names in `ChakraProps`

src/components/Banner/Banner.tsx
EdwinGuzman: Hmm interesting. Ok to keep then.

src/components/CheckboxGroup/CheckboxGroup.mdx
oliviawongnyc: How about this?

`The `CheckboxGroup` composes a Chakra `Box` component, so you may pass
most Chakra `Box` props, as well as the props below. Note, if using the
`onChange` prop, it must be of the type listed below.`

src/components/Link/Link.tsx
oliviawongnyc: I think there are a few components where, yes, it would be simpler/look nicer to extend by the underlying component's Chakra props. Definitely open to discussion about this, but there are two reasons I didn't do this: 1) consistency - now all components extend `Box` to some degree 2) we didn't want to use the Chakra component props beyond `Box` in case they included props that would break the component or not do what was expected if passed. This is not the case for `ChakraLinkProps`, but for example [`Breadcrumb`](https://v2.chakra-ui.com/docs/components/breadcrumb/props). Let me know your thoughts.

src/components/Menu/Menu.mdx
oliviawongnyc: I thought I fixed this! Thank you.

src/components/CheckboxGroup/CheckboxGroup.mdx
EdwinGuzman: Yes, that makes sense.

src/components/Link/Link.tsx
EdwinGuzman: That's helpful, thanks. Let's keep it for consistency.

EdwinGuzman (approved):

PR #1814, opened 2025-05-20
DSD-2061: Removes `useNYPLBreakpoints` in `FilterBarInline` and `MultiSelectGroup`

oliviawongnyc:
Fixes JIRA ticket [DSD-2061](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2061)

- Removes `useNYPLBreakpoint` hook from `MultiSelectGroup` and `FilterBarInline`.
- This should reduce/eliminate the style flash for these two components.
- This should not be a breaking change.

## How has this been tested?

- Locally, in Storybook.

- None.

### Accessibility Checklist

## Open Questions

- In the docs for `MultiSelectGroup` we recommend NOT passing `row` for layout with `full` for width. However, we show an example of this configuration in the docs and in fact, it is Marty's preferred combination. This PR makes no changes wrt to this, but we should decide our stance.
- `FilterBarInline` enforces the above by saying if layout is `row`, `multiSelectWidth` must be `fitContent` and if layout is `column`, `multiSelectWidth` must be `full`. This PR does not change this, but we may want to re-examine if this is what we want.

### Checklist:

### Front End Review:

[DSD-2061]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-2061?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/FilterBarInline/FilterBarInline.tsx
```diff
@@ -114,7 +106,7 @@ export const FilterBarInline: ChakraComponent<
               isBlockElement: layout === "column",
               closeOnBlur: layout === "row",
               layout: layout,
-              width: finalWidth,
+              width: layout === "row" ? "fitContent" : "full",
```
oliviawongnyc, line 109: This enforces `fitContent` with `row` layout and `full` with `column` as the component does currently, though we may change this in future?

If `fitContent` is passed here, but the viewport is mobile, then `MultiSelect` knows to set `width` to "100%" (in `MultiSelect`'s theme file). This was already the case so the use of `useNYPLBreakpoints` in this component was unecessary.

src/components/MultiSelectGroup/MultiSelectGroup.tsx
```diff
@@ -87,7 +82,10 @@ export const MultiSelectGroup: ChakraComponent<
             },
           }}
         >
-          {renderMultiSelect({ isBlockElement, multiSelectWidth: finalWidth })}
+          {renderMultiSelect({
+            isBlockElement: layout === "column",
+            multiSelectWidth,
```
oliviawongnyc, line 88: We can simply pass through `multiSelectWidth` here because (as stated in the comment above) if `fitContent` is passed, `MultiSelect` knows to render at a width of "100%" if the viewport is mobile. This means we probably do not need to have it be a prop on `MultiSelectGroup`, but I am not changing that in this PR.

bigfishdesign13 (approved):
The code looks good and the responsive component styles are functioning as expected.

dgcohen (approved):

PR #1815, opened 2025-05-20
DSD-2048: Removes `isDarkMode`

oliviawongnyc:
Fixes JIRA ticket [DSD-2048](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2048)

- Removes `isDarkMode` function from `Slider` and `Accordion`.
- Adds `currentColor` option to `iconColorArray`.
- Should not be a breaking change.

## How has this been tested?

- Locally, in Storybook.

- None.

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[DSD-2048]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-2048?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bigfishdesign13 (commented):
Can the color styles for `Accordion`  be moved from the component file into the theme file? I really wish I remembered why we jumped through all these hoops to use the `isDarkMode` props.

oliviawongnyc:
> Can the color styles for `Accordion` be moved from the component file into the theme file? I really wish I remembered why we jumped through all these hoops to use the `isDarkMode` props.

@bigfishdesign13, curious about this too. from what I can tell we are not even using the theme file! I can try to change this, but probably should be another ticket.

bigfishdesign13 (approved):
With the idea of moving the styles for the `Accordion` component into the theme file now flagged in another ticket ([DSD-2064](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2064)), this PR looks good.

src/components/Accordion/Accordion.tsx
```diff
@@ -166,29 +164,29 @@ const getElementsFromData = (
                 }
                 bg={
                   !content.accordionType
```
dgcohen: This functionality is working well, but i believe it could be simplified for clarity and legibility.

For example, it's a bit confusing initially that there are accordion type-to-background color hashmaps, yet there's an additional ternary checking the accordion type in the `bg` prop itself.

Would it be possible to fall back to the defaults in `colorMapLight` and `colorMapDark` without the additional ternary here?

eg
```
bg={colorMapLight[content?.accordionType] || colorMapLight['default']}
```

oliviawongnyc: Tried to simplify, lmk what you think!

dgcohen (approved):

src/components/Accordion/Accordion.tsx
dgcohen: Looks great! Thanks

PR #1821, opened 2025-05-28
DSD-2044: Removes the need for consuming apps to import mediaMatchMock

oliviawongnyc:
Fixes JIRA ticket [DSD-2044](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2044)

- Removes `useNYPLBreakpoints` from final component, `SearchBar` so that consuming apps are no longer required to import the MatchMedia API mock to test. 
- Now, instead of an `aria-label` being applied or removed based on whether the button is big enough to contain the word "Search", the word "Search" is always present and simply hidden with screenreader-only styles on mobile.

## How has this been tested?

- Using the unit tests in the repo.

- None, although it should be noted that the device used to make the `SearchBar` button accessible to screen readers has changed - instead of an `aria-label`, the word "Search" is always present and simply made visible or hidden depending on the the screen size.

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[DSD-2044]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-2044?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/hooks/useNYPLBreakpoints.mdx
+++ b/src/hooks/useNYPLBreakpoints.mdx
@@ -83,8 +83,8 @@ When development needs more granular control when targetting viewports:
 
 ## Testing
 
-Because `useNYPLBreakpoints` uses the `MediaMatch` API, you will need to mock the
-API when unit testing. There is a mock you can import from Reservoir if you'd like.
+Because `useNYPLBreakpoints` uses the `MediaMatch` API, you will need to mock the API
+when unit testing. There is a mock you may import from Reservoir, or you can use your own.
 
 <Source
   code={`
--- a/src/hooks/useNYPLBreakpoints.mdx
+++ b/src/hooks/useNYPLBreakpoints.mdx
@@ -81,6 +81,20 @@ When development needs more granular control when targetting viewports:
 - use `isLargerThanSmallTablet` to conditionally target the `large tablet` viewport
 - use `isLargerThanLargeTablet` to conditionally target the `desktop` viewport
 
+## Testing
+
+Because `useNYPLBreakpoints` uses the `MediaMatch` API, you will need to mock the
+API when testing. There is a mock you can import from Reservoir if you'd like.
+
+<Source
+  code={`
+import { MatchMedia } from "@nypl/design-system-react-components";
+
+new MatchMedia();
+`}
+language="jsx"
+/>
+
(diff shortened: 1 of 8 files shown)


bigfishdesign13 (approved):
This looks good. We should also ask Clare to review it due to the changes that impact a11y.

7emansell (approved):

src/hooks/useNYPLBreakpoints.mdx
```diff
@@ -81,6 +81,20 @@ When development needs more granular control when targetting viewports:
 - use `isLargerThanSmallTablet` to conditionally target the `large tablet` viewport
 - use `isLargerThanLargeTablet` to conditionally target the `desktop` viewport
 
+## Testing
+
+Because `useNYPLBreakpoints` uses the `MediaMatch` API, you will need to mock the
+API when testing. There is a mock you can import from Reservoir if you'd like.
```
EdwinGuzman: Nit: update the language to reflect testing in unit tests because it can be confusing if you're testing in storybook or in a consuming app. Also, the mock is necessary to import in consuming app's unit tests, right? If so, then it's required and not if they'd like to import it.

EdwinGuzman (commented):
Thanks! This looks great, but just a minor documentation comment.

src/hooks/useNYPLBreakpoints.mdx
oliviawongnyc: Well couldn't they just add their own `MediaMatch` mock if they wanted? They don't need to use ours.

oliviawongnyc: I don't even think this mock exists in the Chakra v2 repo anymore. I see this instead:
https://github.com/chakra-ui/chakra-ui/blob/cbe51b994104602c914568d2bcb60a0e67d09d14/vitest.setup.ts#L14

EdwinGuzman: Good point, so yes, they can add their own. Perhaps say import from Reservoir or from elsewhere?

EdwinGuzman (approved):
Nice!

PR #1823, opened 2025-05-28
DSD-2056 & DSD-1903: 12-column grid using span syntax

bigfishdesign13:
Fixes JIRA ticket xxx

- Another option for handling the grid layout. This one uses the span syntax. I opted to use class names instead of trying to pass variables values down through the DOM tree.
- Removed the `TemplateTop` and `TemplateBottom` child components and added the `TemplateFull` component.
- Removed the `sidebar` prop and replaced with `variant` prop.

## How has this been tested?

- local Storybook

- n/a

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:


src/theme/components/template.ts
```diff
@@ -14,79 +14,68 @@ const Template = defineStyleConfig({
...
+       * width of the content area from a mobile-first viewpoint. Using -1 for
+       * the "last column" value ensures that these elements span all columns in
+       * the grid. */
+      "& .reservoir-template-full": {
+        gridColumn: "1 / -1",
+      },
+      "& .reservoir-template-main": {
+        gridColumn: "1 / -1",
+      },
+      "& .reservoir-template-sidebar": {
```
oliviawongnyc: I don't think we need this because if a sidebar is used, we will skip this style and use the one in the variant section.

oliviawongnyc (commented):
Want to look at this further but some initial thoughts:

1) With this approach, could we get rid of `TemplateBottom`/`TemplateTop` and just replace with one component called `TemplateFull` that can be used more than once if desired?
2) What if there were 3 variants instead of 2: `sidebarLeft`, `sidebarRight`, and `narrow`? Then we could get rid of the `TemplateMainNarrow` component altogether and not have a grid within a grid.

```
    narrow: {
      "& .reservoir-template-main": {
        gridColumn: {
          base: "1 / -1",
          md: "2 / 12",
          lg: "3 / 11",
        },
      },
    },
```

src/theme/components/template.ts
bigfishdesign13: Good point. I will remove.

src/components/Template/Template.tsx
```diff
@@ -1,15 +1,20 @@
...
 
-export const sidebarPlacementArray = ["none", "left", "right"] as const;
-export type SidebarPlacement = typeof sidebarPlacementArray[number];
+export const variantArray = [
+  "full",
+  "narrow",
+  "sidebarLeft",
+  "sidebarRight",
+] as const;
+export type Variant = typeof variantArray[number];
```
oliviawongnyc: Should this be `TemplateVariant` and `templateVariantArray`? Just because we may end up using the word `variant` in other components over `type`.

src/components/Template/Template.tsx
```diff
@@ -29,8 +34,8 @@ const Template: ChakraComponent<
   React.PropsWithChildren<TemplateProps>
 > = chakra(
   forwardRef<HTMLDivElement, React.PropsWithChildren<TemplateProps>>(
-    ({ children, sidebar = "none" }, ref?) => {
-      const styles = useStyleConfig("Template", { variant: sidebar });
+    ({ children, variant = "full" }, ref?) => {
+      const styles = useStyleConfig("Template", { variant: variant });
```
oliviawongnyc: Nit: could just have `{ variant }`

src/theme/components/template.ts
```diff
@@ -14,79 +14,74 @@ const Template = defineStyleConfig({
...
+      /** Set the full and main content areas to span the full width of the
+       * content area from a mobile-first viewpoint. Using -1 for the "last
+       * column" value ensures that these elements span all columns in the grid.
+       * */
+      "& .reservoir-template-full": {
+        gridColumn: "1 / -1",
+      },
+      "& .reservoir-template-main": {
+        gridColumn: "1 / -1",
+      },
```
oliviawongnyc: If you change this to:

```
 "& .reservoir-template-main": {
        gridColumn: { base: "1 / -1" },
      },
```

then you don't have to include `base: "1 / -1"` in the variants below. For example, sidebarLeft would just have:

```
   "& .reservoir-template-main": {
        gridColumn: {
          sm: "7 / -1",
          md: "5 / -1",
          lg: "4 / -1",
        },
      },
```

oliviawongnyc (commented):
Looking good! Are `TemplateTop`, `TemplateBottom`, and `TemplateMainNarrow` still here because you haven't updated the docs yet? What are all the pictures for?

src/components/Template/Template.tsx
bigfishdesign13: I was leaning toward simplified and common names across components, but maybe that idea should only be applicable to the prop names and types should be unique. What do you think?

Cc @EdwinGuzman

src/components/Template/Template.mdx
```diff
@@ -93,21 +92,21 @@ import {
...
+and are proven to make it harder for people to focus, while short lines of text
+with fewer than 30 characters per line force the eye to travel back too often,
+breaking a reader's rhythm.
 
-This means you may use `TemplateMain` or `TemplateMainNarrow` depending on your
-needs, but the two should not be used in conjunction.
+The `TemplateMain` component has a max width of `1280px`, which is much wider
+than the ideal line length for standard text. When page content is predominantly
+large blocks of text, the `Template` component should be set to use the `narrow`
+variant to restrict the width of the `TemplateMain` component.
```
EdwinGuzman: How do you feel about creating a subsection for the text width topic and another subsection for the `main` element and structure?

src/components/Template/Template.mdx
```diff
@@ -307,100 +302,114 @@ offered:
...
-  ...
-</Template>
-`}
-  language="jsx"
+<DocsWithImage
+  docs={
+    <List type="ul">
+      <li>Ideal for displaying content that is primarily text</li>
+      <li>
+        Max width of <code>821px</code>
```
EdwinGuzman, line 454: Not 820px?

src/components/Template/Template.test.tsx
```diff
@@ -59,53 +50,46 @@ describe("Template components accessibility", () => {
...
   it("renders a #mainContent id when using TemplateMain", () => {
-    const { container } = render(templateComponents("left"));
-
-    expect(container.querySelector("#mainContent")).toBeInTheDocument();
-    expect(screen.getByRole("main")).toHaveAttribute("id", "mainContent");
-  });
-
-  it("renders a #mainContent id when using TemplateMainNarrow", () => {
-    const { container } = render(templateComponents("none", true));
+    const { container } = render(templateComponents("sidebarLeft"));
```
EdwinGuzman, line 104: This should be the default or "narrow".

src/components/Template/Template.tsx
EdwinGuzman: I'm slightly torn. I think `TemplateVariant` makes sense to tie values to the component internally. This is also exporting `Variant` which might make things confusing. This _could_ conflict with `Variant` in other components if the same name is used (although I know this is exported so it's used in other template-related files). The values in this generic `Variant` type is not really usable in other components. That said, I think `TemplateVariant` is the way to go.

src/components/Template/Template.tsx
```diff
@@ -49,20 +54,19 @@ const Template: ChakraComponent<
...
 
 /**
  * This optional component renders content at a max width of 1280px and
  * will render below `TemplateBreakout` (if being used) and above the
  * main content and sidebar (if one exists).
  */
+const TemplateFull: React.FC<React.PropsWithChildren> = ({ children }) => (
+  <Box className="reservoir-template-full">{children}</Box>
+);
 const TemplateTop: React.FC<React.PropsWithChildren> = ({ children }) => (
```
EdwinGuzman: Just for consistency, can you give component js doc style comments?

src/utils/utils.ts
```diff
@@ -1,5 +1,5 @@
 import { HelperErrorTextType } from "../components/HelperErrorText/HelperErrorText";
-import { SidebarPlacement } from "../components/Template/Template";
+import { Variant } from "../components/Template/Template";
```
EdwinGuzman: Similar to the comment above, the name makes it seem generic, but if you read where it's imported from, it's tied to the `Template`. Maybe it's just me but `TemplateVariant` for the name makes sense.

EdwinGuzman (commented):
Left some comments but I had two which I went over with @oliviawongnyc : I got confused about `TemplateMainNarrow` and the two Top and Bottom components. The plan is to remove them and I agree, it simplifies this component.

EdwinGuzman:
BTW, have not (and will not?) review the other PRs. I think this approach works well.

src/components/Template/Template.mdx
bigfishdesign13: Yes, that's a good idea.

src/components/Template/Template.mdx
bigfishdesign13: That would be nice, right? However, the width of the narrow variant is based on the responsive grid system, which does not adhere to nice round numbers. The actual max width is `821.33px`, so I rounded.

src/components/Template/Template.test.tsx
bigfishdesign13: @EdwinGuzman Not necessarily. This could test _any_ instance of `TemplateMain` to verify the `id` value. Maybe I should add an explicit render for each variant option?

src/components/Template/Template.tsx
bigfishdesign13: I am not entirely sure what you meant, but I updated the code comments. I missed them the first time around.

src/components/Template/Template.mdx
EdwinGuzman: :\

src/components/Template/Template.test.tsx
EdwinGuzman: If there are only four, yes.

src/components/Template/Template.tsx
EdwinGuzman: Might not be relevant anymore after the lastest commit.

src/components/Template/Template.stories.tsx
```diff
@@ -247,14 +254,14 @@ export const TemplateFullExample: Story = {
             subHeaderText={otherSubHeaderText}
           />
         </TemplateBreakout>
-        <TemplateTop>
+        <TemplateFull>
           <Banner
             content="This is the top content area!"
             heading="Content Top"
             type="informative"
```
oliviawongnyc: Do we still want to use this `content` or `heading`? Feels a little like relics of the old component now. We could swap it for the values you used in the "Complex Example":

```
<TemplateFull>
          <Banner
            content="This is a full content area!"
            heading="Full content"
            type="informative"
          />
        </TemplateFull>
```
If you decide to update, then same goes for the "Content Bottom" version and those in other stories.

src/components/Template/Template.mdx
```diff
@@ -93,21 +87,25 @@ import {
...
-text, you should use the `TemplateMainNarrow` component instead. Note that
-`TemplateMainNarrow` is not meant to be used in conjunction with a sidebar.
+### Content Width
 
-This means you may use `TemplateMain` or `TemplateMainNarrow` depending on your
-needs, but the two should not be used in conjunction.
+Wide lines of text with more than 80 characters per line are difficult to read
+and are proven to make it harder for people to focus, while short lines of text
+with fewer than 30 characters per line force the eye to travel back too often,
+breaking a reader's rhythm.
```
oliviawongnyc: My vote would be to remove this paragraph and add a `Resources` link instead that talks about this. The second paragraph stands well on its own and contains the most important info.

src/components/Template/Template.mdx
```diff
@@ -245,9 +240,8 @@ offered:
...
       <li>Should be used for the main content of a page</li>
       <li>
-        Positioned below <code>TemplateTop</code> and to either the left or
-        right of
-        <code>TemplateSidebar</code>, when it is rendered
+        Positioned below <code>TemplateFull</code> and to either the left or
+        right of <code>TemplateSidebar</code>, when it is rendered
       </li>
       <li>
         May be used in conjunction with <code>TemplateSidebar</code>
```
oliviawongnyc: Typo in "percentage" below

src/components/Template/Template.stories.tsx
```diff
@@ -10,29 +10,31 @@ import Form, { FormField, FormRow } from "../Form/Form";
...
   title: "Components/Page Layout/Template",
   component: Template,
   argTypes: {
     id: { control: false },
-    sidebar: {
+    variant: {
       control: { type: "radio" },
-      options: sidebarPlacementArray,
-      table: { defaultValue: { summary: "none" } },
+      options: templateVariantArray,
```
oliviawongnyc, line 41: I'm not sure why, but the `variant` type is showing up as `string` on the Storybook args table.

oliviawongnyc (commented):
Love the images in the Storybook page!!

Github won't let me make comments on these, but what do you think about adding "Often" to the bullets talking about where each child component should live?

For `TemplateBreakout`:
- Often positioned directly below the NYPL universal header and above any other `Template` child components

For `TemplateFull`:
- Often positioned directly below `TemplateBreakout` or directly below `TemplateSidebar` and `TemplateMain`

For `TemplateSidebar`:
- Positioned either the left or right of `TemplateMain` and often below `TemplateFull`

For `TemplateMain`:
- Often positioned below `TemplateFull` and to either the left or right of `TemplateSidebar`  (if being used)

There's a typo in this sentence:
If an app needs to render the SkipNavigation component, it should be [REMOVE: to be] explicitly rendered immediately before the Template component

src/components/Template/Template.stories.tsx
bigfishdesign13: I noticed this a while ago when we first started reworking the `Template` component. The description is missing as well. We talked about it a little bit, but we never got to the bottom of the problem.

EdwinGuzman: Is it because `variant` is an underlying chakra prop?

bigfishdesign13: I don't think so. This was also the case when the prop name was still `sidebar`.

EdwinGuzman (commented):
This looks good to me. I'll approve once Olivia's comments have been addresssed.

src/components/Template/Template.stories.tsx
```diff
@@ -441,45 +484,380 @@ export const TemplateFullExampleWithFeaturedContent = {
...
+                p="2rem"
+                size="display1"
+                text="Overlay heading"
+                textAlign="center"
+                textShadow="0px 0px 8px #000"
+                width="100%"
+              />
+            </Box>
+            <SubNav
+              id="asdasd"
```
EdwinGuzman: nit: give this a better id

src/components/Template/Template.tsx
```diff
@@ -42,87 +48,126 @@ const Template: ChakraComponent<
...
- * It must be paired with the `TemplateMain` component. An optional `sidebar`
- * prop value of "left" or "right" must be passed to the `Template` wrapper
- * to render the correct CSS styles.
+ * This optional component will span the full width of the content area and will
+ * render content at a max width of 1280px.
  */
-const TemplateSidebar: React.FC<React.PropsWithChildren> = ({ children }) => (
-  <Box gridArea="sidebar">{children}</Box>
-);
+// const TemplateFull: React.FC<React.PropsWithChildren<TemplateChildProps>> = ({
```
EdwinGuzman: I think you can remove the comments.

src/theme/components/template.ts
```diff
@@ -13,115 +13,141 @@ const Template = defineStyleConfig({
...
+      gridColumn: "1 / -1",
       ml: "calc(-50vw + 50%)",
       px: responsiveMargin,
+      width: "100vw",
     };
   }),
 });
 
-const TemplateMainNarrow = defineStyleConfig({
+const TemplateFull = defineStyleConfig({
```
EdwinGuzman, line 122: Suggestion: keep `TemplateFull`, remove `TemplateHeader` and `TemplateFooter`, and use `const styles = useStyleConfig("TemplateFull", {});` for the header and footer components.

src/components/Template/Template.mdx
```diff
@@ -119,30 +117,135 @@ Resources
...
+      </li>
+      <li>
+        Ideal for including custom header elements like main navigation, hero,
+        or page title in apps that do not use the NYPL universal header
+      </li>
+      <li>
+        Positioned above any other <code>Template</code> child components
+      </li>
+      <li>
+        It is recommended to use the <code>TemplateBreakout</code> child
```
EdwinGuzman: Why?

src/components/Template/Template.mdx
```diff
@@ -564,10 +668,10 @@ import {
   <SkipNavigation />
```
EdwinGuzman: Minor note outside the scope of this pr: if the global header is added here as an example, then `SkipNavigation` is not needed.

src/components/Template/Template.mdx
```diff
@@ -564,10 +668,10 @@ import {
...
   <Template sidebar="...">
     <TemplateBreakout>...</TemplateBreakout>
-    <TemplateTop>...</TemplateTop>
+    <TemplateFull>...</TemplateFull>
     <TemplateSidebar>...</TemplateSidebar>
     <TemplateMain>...</TemplateMain>
```
EdwinGuzman: Add `id="mainContent"` to show the recommendation for the skip nav target from the global header.

EdwinGuzman (changes requested):
Some minor updates, but I think this is good to go after the updates.

src/components/Template/Template.mdx
bigfishdesign13: Good catch. I neglected to update the DOM here based on the changes related to the `TemplateMain` component and the addition of the `TemplateContent` component.

EdwinGuzman (approved):
Thanks for the updates. Let's get more reviews on this and then merge! @oliviawongnyc @7emansell

src/theme/components/template.ts
oliviawongnyc: I would take this a step further and remove `TemlateHeader` and `TemplateFooter` components all together in favor of one `TemplateFull` component, but perhaps you are keeping all of them so you can have the `as` prop on the `Box`? What about letting the consumer add the `as`?

oliviawongnyc: As discussed, we will keep `TemplateHeader` and `TemplateFooter`.

oliviawongnyc (approved):

dgcohen (approved):

PR #1824, opened 2025-06-03
DSD-2047: Combine shared types

7emansell:
Fixes JIRA ticket [DSD-2047](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2047)

 Consolidates types that are shared across a couple components:

 - `bannerTypesArray` and `statusBadgeArray` have similar values --> `messageTypeArray`
 - `bannerBgColorsArray and actionBackgroundColorsArray` have similar values --> `bgColorArray`
 - the set of highlight color values are the same in `Banner`, `SubNav`, and `NewsletterSignup` --> `highlightColorsArray`
 -  size types in `ButtonSizesArray` and what’s in `SocialMediaList` --> `sizesArray`
        

### Accessibility Checklist

### Checklist:

### Front End Review:

[DSD-2047]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-2047?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


7emansell:
Will update changelogs accordingly once we feel confident this is how we want to do it– particularly wondering if these arrays should just be imported as their types rather than letting each component name their custom type? Can't decide which way is more readable cc @oliviawongnyc @bigfishdesign13 @EdwinGuzman

src/theme/sharedTypes.ts
```diff
@@ -0,0 +1,57 @@
...
+  "dark.section.connect.primary",
+  "dark.section.education.primary",
+  "dark.section.locations.primary",
+  "dark.section.research.secondary",
+  "dark.section.research-library-lpa.primary",
+  "dark.section.research-library-schomburg.primary",
+  "dark.section.research-library-schwarzman.primary",
+  "dark.section.whats-on.primary",
+];
+export const messageTypeArray = [
```
oliviawongnyc: We're going to be moving away from the word `type` for props that control a component's variant and actually just using the word `variant`, so maybe we can change this to `messageVariantsArray`.

oliviawongnyc (commented):
I'm not sure how you decided which types were similar enough to combine, but a couple others to consider: 

- `logoSizesArray` and `imageSizesArray`, 
- `textSizesArray` in `StyledList` and `statusBadgeFontSizeArray` and `tableBodyTextSizesArray` and `textSizesArray` in `Text`? 
- Also maybe `iconColorsArray` could go with the other colors? 
- Maybe `iconAlignArray` and `Menu`'s `listAlignment` type? 

Maybe none of these make sense to combine, but maybe worth looking into.

oliviawongnyc:
> I'm not sure how you decided which types were similar enough to combine, but a couple others to consider:
> 
> * `logoSizesArray` and `imageSizesArray`,
> * `textSizesArray` in `StyledList` and `statusBadgeFontSizeArray` and `tableBodyTextSizesArray` and `textSizesArray` in `Text`?
> * Also maybe `iconColorsArray` could go with the other colors?
> * Maybe `iconAlignArray` and `Menu`'s `listAlignment` type?
> 
> Maybe none of these make sense to combine, but maybe worth looking into.

Per our standup discussion, most of these shouldn't be combined. I do think you could make the argument for the last one tho?

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -59,21 +60,7 @@ export interface NewsletterSignupProps {
...
-  "section.education.primary",
-  "section.locations.primary",
-  "section.research.primary",
-  "section.research-library.lpa",
-  "section.research-library.schomburg",
-  "section.research-library.schwartzman",
-  "section.whats-on.primary",
-] as const;
-export type HighlightColorTypes = typeof highlightColorTypesArray[number];
+export type HighlightColorTypes = typeof highlightColorsArray[number];
```
EdwinGuzman: Using `highlightColorsArray` adds _more_ highlight colors, specifically the dark variants. That, imo, is ok. What seems to be missing is `ui.gray.medium` though. It's the default value on line 94 so it's still needed. I think you may need to just* combine this single value with `highlightColorsArray` into a new array.

EdwinGuzman:
> Per our standup discussion, most of these shouldn't be combined. I do think you could make the argument for the last one tho?

I think it's worth looking into but doesn't seem, from first glance, that it's straightforward. "none", for example, might not make much sense for `Menu`.

EdwinGuzman:
> particularly wondering if these arrays should just be imported as their types rather than letting each component name their custom type? Can't decide which way is more readable

@7emansell great discussion topic. I'm in favor of having a single variable for all to use, such as `export type SizeType = typeof sizesArray[number];`

Then, `SocialMediaLinks` and `Button` can import this type rather than create their own in the repo. This reduces the types and might make things simpler, but what I like about the custom types is how they're tied down semantically to the component when you import them, such as `import { ButtonSizes }, Button from ....`

7emansell:
> > particularly wondering if these arrays should just be imported as their types rather than letting each component name their custom type? Can't decide which way is more readable
> 
> @7emansell great discussion topic. I'm in favor of having a single variable for all to use, such as `export type SizeType = typeof sizesArray[number];`
> 
> Then, `SocialMediaLinks` and `Button` can import this type rather than create their own in the repo. This reduces the types and might make things simpler, but what I like about the custom types is how they're tied down semantically to the component when you import them, such as `import { ButtonSizes }, Button from ....`

@bigfishdesign13 thoughts on this? otherwise i think i will leave it as custom types for now

src/theme/sharedTypes.ts
```diff
@@ -0,0 +1,57 @@
...
+  "dark.section.connect.primary",
+  "dark.section.education.primary",
+  "dark.section.locations.primary",
+  "dark.section.research.secondary",
+  "dark.section.research-library-lpa.primary",
+  "dark.section.research-library-schomburg.primary",
+  "dark.section.research-library-schwarzman.primary",
+  "dark.section.whats-on.primary",
+];
+export const messageVariantsArray = [
```
bigfishdesign13, line 49: The `messageVariants` prefix initially didn't sit well with me, but I am good with it now. At first glance, I wanted it to be something like `meaningVariants` or `intentVariants`. Or even something very literal like `semanticVariants`.

bigfishdesign13:
> > Then, `SocialMediaLinks` and `Button` can import this type rather than create their own in the repo. This reduces the types and might make things simpler, but what I like about the custom types is how they're tied down semantically to the component when you import them, such as `import { ButtonSizes }, Button from ....`
> 
> @bigfishdesign13 thoughts on this? otherwise i think i will leave it as custom types for now

We use varying size values throughout the DS components. Yes, the values for `SocialMediaLinks` and `Button` overlap, but that overlap is arbitrary. And the values used for `Heading`, `Icon`, `Image`, and `Logo` (and others) are either completely different or they are a more extensive set. And those types are being allowed to remain as custom types. 

That said, I don't think the size values for `SocialMediaLinks` and `Button` should be changed to use a common value set. For consistency, I feel the should remain as custom types.

Cc @EdwinGuzman @oliviawongnyc

src/components/NewsletterSignup/NewsletterSignup.tsx
bigfishdesign13: The approach described here (and what was finally executed in the code) aligns with how I envisioned the color definitions: predefined sets of colors that can be combined with individual color values to create a final master set within a component.

bigfishdesign13 (commented):
Just a few comments.

src/theme/sharedTypes.ts
7emansell: glad you have come around because naming a variable `semanticVariants` would haunt me 😅

7emansell:
@oliviawongnyc @EdwinGuzman @bigfishdesign13 leaving it as custom types for now but something to keep in mind!

EdwinGuzman (approved):

PR #1831, opened 2025-06-11
DSD-2038: Consistent id and classname update

EdwinGuzman:

- This is the "main" feature branch for updating all component's `id`s and `classname`s to a consistent pattern which will be `ds-[componentName][-additionalUserPassedId]` for ids. For class names, we won't be combining the passed classname value with the component's because they can be separate for various styling purposes.

## How has this been tested?

- Unit tests.

- Will make a separate PR to address recommendation for the set of components that should have unique ids.

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

### DIFF
--- a/src/components/AlphabetFilter/AlphabetFilter.stories.tsx
+++ b/src/components/AlphabetFilter/AlphabetFilter.stories.tsx
@@ -11,7 +11,6 @@ const meta: Meta<typeof AlphabetFilter> = {
   argTypes: {
     activeLetters: { control: false },
     currentLetter: { control: false },
-    id: { control: false },
     isDisabled: argsBooleanType(),
     onClick: { control: false },
   },
@@ -30,7 +29,6 @@ export const WithControls: Story = {
     currentLetter: undefined,
     descriptionText: "This is description text.",
     headingText: "AlphabetFilter",
-    id: "alphabet-filter-id",
     isDisabled: false,
     onClick: undefined,
   },
@@ -68,6 +66,7 @@ export const SetActiveLetters: Story = {
 export const WithCustomHeading: Story = {
   render: () => (
     <AlphabetFilter
+      id="custom-heading"
       onClick={undefined}
       headingText={<Heading level="h4">Custom H4 Heading</Heading>}
     />
--- a/src/components/AlphabetFilter/AlphabetFilter.tsx
+++ b/src/components/AlphabetFilter/AlphabetFilter.tsx
@@ -1,4 +1,4 @@
-import React, { forwardRef, useRef, useState } from "react";
+import { forwardRef, useEffect, useRef, useState } from "react";
 import {
   Box,
   BoxProps,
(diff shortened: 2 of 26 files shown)


oliviawongnyc (commented):
Makes sense to me! Eventually will want component changelogs updated.

src/utils/utils.ts
```diff
@@ -220,3 +220,7 @@ export const getTextFromElement = (
   }
   return getTextFromElement(children);
 };
+
+export const generateComponentId = (name: string, id: string | undefined) => {
+  return `ds-${name}${id ? `-${id}` : ""}`;
```
bigfishdesign13, line 225: I also asked this in the Google Doc: why not use `nypl` for the prefix?

src/components/AlphabetFilter/AlphabetFilter.stories.tsx
```diff
@@ -68,6 +66,7 @@ export const SetActiveLetters: Story = {
 export const WithCustomHeading: Story = {
   render: () => (
     <AlphabetFilter
+      id="custom-heading"
```
bigfishdesign13, line 69: Why was the new naming pattern not used for this `id`?

bigfishdesign13: Sorry, ignore that question. This is the story file.

src/components/AlphabetFilter/AlphabetFilter.tsx
```diff
@@ -51,7 +51,7 @@ export const AlphabetFilter: ChakraComponent<
     } = props;
 
     const styles = useMultiStyleConfig("AlphabetFilter", {});
-
+    const mainId = generateComponentId("alphabetFilter", id);
```
bigfishdesign13, line 54: The prefix we've used for similar conditional values is `final`. So that would be `finalId`. How do you feel about that?

src/utils/utils.ts
EdwinGuzman: `ds` for reservoir but anyone, if they wanted to, could use `nypl` in their own app. It's not specific enough.

src/components/AlphabetFilter/AlphabetFilter.tsx
EdwinGuzman: I was debating that. Some components use the top-level `id` for internal components so "final" didn't feel right, whereas "main", to me, signifies the parent or top-level id. Maybe "parent" is better?

EdwinGuzman: But then not all DS components are complex and have multiple internal components, so another argument for "main".

src/utils/utils.ts
bigfishdesign13: Ok, I see what you are saying. Then yes, either `ds` or `reservoir` would work.

src/components/AlphabetFilter/AlphabetFilter.tsx
bigfishdesign13: Understood. With that in mind, `main`, `root`, or `base` would all work.

src/utils/utils.ts
```diff
@@ -220,3 +220,7 @@ export const getTextFromElement = (
   }
   return getTextFromElement(children);
 };
+
+export const generateComponentId = (name: string, id: string | undefined) => {
```
7emansell, line 224: nice

7emansell (approved):

EdwinGuzman:
Closing, don't like this approach after working through the updates...

PR #1832, opened 2025-06-12
DSD-2053 & DSD-2054: Add JavaScript disabled fallback warning to AudioPlayer and VideoPlayer

dgcohen:
Jira tickets:
[DSD-2053](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2053)
[DSD-2054](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2054)

- Renders a `noscript` tag with a JavaScript disabled fallback warning for AudioPlayer and VideoPlayer.

- This will only work when rendered via Server-Side Generation or Server-Side Rendering.

## How has this been tested?

- I verified by publishing the DS package with `yalc` and consuming in the `research-catalog` app to test the functionality with JS disabled. Here's a screenshot:

Screenshot 2025-06-13 at 12 05 50 PM

## Open Questions

- Should we get formal approval/input on the copy? If so from whom?

### Checklist:

[DSD-2053]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-2053?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DSD-2054]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-2054?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/AudioPlayer/AudioPlayer.tsx
+++ b/src/components/AudioPlayer/AudioPlayer.tsx
@@ -155,7 +155,9 @@ export const AudioPlayer: ChakraComponent<
           ) : (
             <>{embedElement}</>
           )}
-          {/* Displays a fallback message to users when JavaScript is disabled. */}
+          {/* Displays a fallback message to users when JavaScript is disabled.
+              NOTE: This relies on the component being Server-Side Rendered (SSR) or Statically Generated (SSG).
+          */}
           <noscript>
             <Box __css={styles.invalid}>
               The audio player requires JavaScript to function. Please enable
--- a/src/components/AudioPlayer/AudioPlayer.tsx
+++ b/src/components/AudioPlayer/AudioPlayer.tsx
@@ -157,11 +157,9 @@ export const AudioPlayer: ChakraComponent<
           )}
           {/* Displays a fallback message to users when JavaScript is disabled. */}
           <noscript>
-            <Box 
-              __css={styles.invalid} 
-            >
-              The audio player requires JavaScript to function. 
-              Please enable JavaScript in your browser settings and refresh the page.
+            <Box __css={styles.invalid}>
+              The audio player requires JavaScript to function. Please enable
+              JavaScript in your browser settings and refresh the page.
             </Box>
           </noscript>
         </ComponentWrapper>
--- a/src/components/AudioPlayer/AudioPlayer.tsx
+++ b/src/components/AudioPlayer/AudioPlayer.tsx
@@ -159,10 +159,17 @@ export const AudioPlayer: ChakraComponent<
               NOTE: This relies on the component being Server-Side Rendered (SSR) or Statically Generated (SSG).
           */}
           <noscript>
-            <Box __css={styles.invalid}>
+            <div
+              style={{
+                margin: "1rem 0",
+                backgroundColor: "var(--nypl-colors-ui-bg-default)",
+                border: "1px solid var(--nypl-colors-ui-border-default)",
+                padding: "var(--nypl-space-s)",
+              }}
+            >
               The audio player requires JavaScript to function. Please enable
               JavaScript in your browser settings and refresh the page.
-            </Box>
+            </div>
           </noscript>
         </ComponentWrapper>
       );
--- a/src/components/AudioPlayer/AudioPlayer.tsx
+++ b/src/components/AudioPlayer/AudioPlayer.tsx
@@ -155,6 +155,15 @@ export const AudioPlayer: ChakraComponent<
           ) : (
             <>{embedElement}</>
           )}
+          {/* Displays a fallback message to users when JavaScript is disabled. */}
+          <noscript>
+            <Box 
+              __css={styles.invalid} 
+            >
+              The audio player requires JavaScript to function. 
+              Please enable JavaScript in your browser settings and refresh the page.
+            </Box>
+          </noscript>
         </ComponentWrapper>
       );
     }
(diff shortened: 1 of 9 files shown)


oliviawongnyc (commented):
Looks good! A couple thoughts:
1. I know you mentioned making this a reusable component at some point. Are we holding off bc it's only two components for now?
2. Are there any apps not using SSR/SSG that would need this?

EdwinGuzman (commented):
The outcome is correct but I don't agree on using chakra components or having the feature rely on the component being rendered SSR or SSG. The fallback should work _anywhere_, and I think this could work if `Box` is replaced with a `div`. That way, it doesn't rely on Chakra to be enabled and it could work in any situation.

EdwinGuzman (changes requested):
Sorry, should've been clear that I'm requesting changing the implementation.

dgcohen:
> The outcome is correct but I don't agree on using chakra components or having the feature rely on the component being rendered SSR or SSG. The fallback should work _anywhere_, and I think this could work if `Box` is replaced with a `div`. That way, it doesn't rely on Chakra to be enabled and it could work in any situation.

Updated the implementation to use divs and prop-based inline styles per your request. I wasn't able to use generic inlines style due to a type mismatch, not sure how important it is however since the message is rendered within a react component that couldn't render without JS (on the server of client-side) anyway.

dgcohen:
> Looks good! A couple thoughts:
> 
> 1. I know you mentioned making this a reusable component at some point. Are we holding off bc it's only two components for now?
> 2. Are there any apps not using SSR/SSG that would need this?

Both good questions!

1.  It would be possible to make it a reusable component, though it may be out of scope for this ticket (let me know if you disagree). I imagine it could be useful for (server-rendered) consuming apps that need to output a JS-disabled warning with base NYPL styles. 

2. I would imagine so, but as far as I understand they would need to output their own noscript message near the top of their app, since a client-only React app wouldn't render at all with JS-disabled.

EdwinGuzman (approved):

7emansell (approved):

7emansell:
It would be nice to have this as a reusable component if only so that the copy was consistent across all JS-dependent components but agreed that it's out of scope here

src/components/AudioPlayer/AudioPlayer.tsx
```diff
@@ -155,6 +155,22 @@ export const AudioPlayer: ChakraComponent<
...
+            <div
+              style={{
+                margin: "1rem 0",
+                backgroundColor: "var(--nypl-colors-ui-bg-default)",
+                border: "1px solid var(--nypl-colors-ui-border-default)",
+                padding: "var(--nypl-space-s)",
+              }}
+            >
+              The audio player requires JavaScript to function. Please enable
+              JavaScript in your browser settings and refresh the page.
```
bigfishdesign13, line 171: This message works fine, but if we are going to eventually move to a shared component with a more universal and generic message in the future, let's go ahead and update the messages in the PR to reflect where we are headed. Maybe something like this:
```
<h2 style={{ fontSize: "16px" }}>JavaScript is disabled in your browser.</h2>
<p>
  Some features of this website may not function correctly without 
  JavaScript enabled. To experience the full functionality, please enable 
  JavaScript and refresh the page.
</p>
```

src/components/VideoPlayer/VideoPlayer.tsx
```diff
@@ -199,6 +199,23 @@ export const VideoPlayer: ChakraComponent<
...
+              NOTE: This relies on the component being Server-Side Rendered (SSR) or Statically Generated (SSG).
+          */}
+          <noscript>
+            <div
+              style={{
+                margin: "1rem 0",
+                backgroundColor: "var(--nypl-colors-ui-bg-default)",
+                border: "1px solid var(--nypl-colors-ui-border-default)",
+                padding: "var(--nypl-space-s)",
+                height: "auto",
```
bigfishdesign13, line 212: Why is `height: "auto"` added here but not on the other one?

bigfishdesign13 (commented):
A thought and a question.

src/components/AudioPlayer/AudioPlayer.tsx
EdwinGuzman: The generic message is a part of this guide https://nypl.github.io/nypl-design-system/reservoir/v3/?path=/docs/accessibility-guide-progressive-enhancement--docs

I was thinking of separating the global-app message from the "inlined" component-specific message. It will most likely be the case, or rather ideally, that _both_ messages display; the global app-wide message that js is needed to view/interact with some features, and component-level message for the audioplayer. I'm imagining the scenario where it's possible for 50-80% of the page is still interactive without js, but there's one feature, such as audio player, that doesn't work.

If this is true, or the approach we want, can we keep this more specific message?

src/components/VideoPlayer/VideoPlayer.tsx
dgcohen: I added it to replicate the videoPlayer-specific invalid styles found here: https://github.com/NYPL/nypl-design-system/blob/528bc51e6cd95f225a318737568acf3cff80dd22/src/theme/components/videoPlayer.ts#L25

src/components/AudioPlayer/AudioPlayer.tsx
bigfishdesign13: Understood. Yes, keep this message as is.

src/components/VideoPlayer/VideoPlayer.tsx
bigfishdesign13: Ah, yes, that makes sense.

PR #1835, opened 2025-06-13
DSD-2071: Change `type` props to `variant`

7emansell:
Fixes JIRA ticket [DSD-2071](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2071)

- Renames component props that use "type" to indicate a group of styles to "variant" (or in a couple cases, to something else that's more accurate)
- Apologies in advance for how many files this changes, this is a very very breaking change

   - Accordion: `accordionType` --> `variant`
   - Banner: `type` --> `variant`
   - Breadcrumbs: `breadcrumbsType` --> `variant`
   - Button: `buttonType` --> `variant`
   - Hero: `heroType` --> `variant`
   - Icon: `iconType` --> `variant`
   - Link: `type` --> `variant`
   - List: `type` --> `variant`**
   - Modal: `type` --> `variant`
   - Notification: `notificationType` --> `variant`
   - Select: `selectType` --> `variant`
   - StatusBadge: `type` --> `variant`
   - TagSet: `type` --> `variant`
   - TextInput: `textInputType` --> `textInputVariant` (`type` stays the same here)**

** These ones were weird, definitely would like feedback on name choices

### Accessibility Checklist

## Open Questions

- Are any of these overkill/am I missing ones you would expect? 

### Checklist:

### Front End Review:

[DSD-2071]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-2071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


EdwinGuzman (commented):
I'll completely look into this next week, but I'm, at first glance, surprised that this doesn't involve more changes in respective theme files. I'm guessing it's mostly a name update but not a chakra-based approach for the variant in the theme file?

7emansell:
> I'll completely look into this next week, but I'm, at first glance, surprised that this doesn't involve more changes in respective theme files. I'm guessing it's mostly a name update but not a chakra-based approach for the variant in the theme file?

right and i think we're already using variant for a lot of the theme files that are using the "type" prop to define styles

EdwinGuzman:
> List: type --> listContainer **
> TextInput: textInputType --> variant, type --> inputType **

For `listContainer`, I can see why this is weird. I think `variant` is better, but what do you think of `as`? I _think_ this is weird or might be weird because it's not updating a variant as much as the parent element. The Chakra convention is `as` which I think works here.

For `textInputType`, this is only for internal purposes, so that makes me worried that if it's `variant`, it could be more widely used. I think it's ok to keep this name. I'd keep `type` over `inputType` since `type` is an attribute for input elements and is commonly used.

src/components/Banner/Banner.tsx
```diff
@@ -145,27 +145,27 @@ export const Banner: ChakraComponent<
...
       id,
       isDismissible = false,
-      type = "neutral",
+      variant = "neutral",
       ...rest
     } = props;
     const [isOpen, setIsOpen] = useState(true);
     const handleClose = () => setIsOpen(false);
-    const overrideType = !!(backgroundColor && highlightColor);
+    const overrideGivenVariant = !!(backgroundColor && highlightColor);
```
oliviawongnyc: Nit: `given` isn't always going to be accurate since the consumer isn't required to pass in a variant (just like they weren't with type)

src/components/Button/Button.tsx
```diff
@@ -94,11 +94,11 @@ export const Button: ChakraComponent<
       );
 
       if (childCount === 1 && hasIcon) {
-        variant = "iconOnly";
+        finalVariant = "iconOnly";
       }
 
       styles = useMultiStyleConfig("ReservoirButton", {
-        variant,
+        finalVariant,
```
oliviawongnyc: Does this work or does it need to be `variant: finalVariant`?

src/components/Card/__snapshots__/Card.test.tsx.snap
```diff
@@ -114,6 +114,7 @@ exports[`Card Renders the UI snapshot correctly 2`] = `
           id="link-online"
           rel={null}
           target={null}
+          type="buttonPrimary"
```
oliviawongnyc: Why is this `type` here now?

src/components/Accordion/Accordion.tsx
```diff
@@ -12,9 +12,9 @@ import React, { forwardRef, useEffect, useState } from "react";
 
 import Icon from "../Icons/Icon";
 
-export type AccordionTypes = "default" | "warning" | "error";
+export type AccordionVariants = "default" | "warning" | "error";
```
oliviawongnyc: Nit: seems like we mostly use an array for these types, can you update this for consistency?

example:
```
export const heroVariantsArray = ["primary", "tertiary", "campaign"] as const;
export type HeroVariants = typeof heroVariantsArray[number];
```

src/components/Link/Link.tsx
```diff
@@ -188,46 +190,46 @@ export const Link: ChakraComponent<
...
-      variant = type;
+      finalVariant = "moreLink";
+    } else if (variant.includes("button")) {
+      finalVariant = variant;
     }
     const styles = useMultiStyleConfig("Link", {
       finalIsUnderlined,
       hasVisitedState,
-      variant,
+      finalVariant,
```
oliviawongnyc: Same as with `Button`, think this needs to be `variant: finalVariant`

src/components/NewsletterSignup/__snapshots__/NewsletterSignup.test.tsx.snap
```diff
@@ -15,24 +15,11 @@ exports[`NewsletterSignup Snapshots Renders the bad email UI snapshot correctly
     <a
       className="chakra-link css-1xdhyk6"
       href="https://www.nypl.org/help/about-nypl/legal-notices/privacy-policy"
-      rel="nofollow noopener noreferrer"
-      target="_blank"
+      rel={null}
+      target={null}
+      type="external"
```
oliviawongnyc: Why did all this change and why is `type` here?

src/components/Select/Select.tsx
```diff
@@ -16,9 +16,9 @@ import Icon from "../Icons/Icon";
 import Label from "../Label/Label";
 import { getAriaAttrs } from "../../utils/utils";
 
-export const selectTypesArray = ["default", "searchbar"];
+export const selectVariantsArray = ["default", "searchbar"];
 export const labelPositionsArray = ["default", "inline"];
```
oliviawongnyc: I know this wasn't part of this work, but these arrays should end with `as const`

oliviawongnyc:
> > List: type --> listContainer **
> > TextInput: textInputType --> variant, type --> inputType **
> 
> For `listContainer`, I can see why this is weird. I think `variant` is better, but what do you think of `as`? I _think_ this is weird or might be weird because it's not updating a variant as much as the parent element. The Chakra convention is `as` which I think works here.
> 
> For `textInputType`, this is only for internal purposes, so that makes me worried that if it's `variant`, it could be more widely used. I think it's ok to keep this name. I'd keep `type` over `inputType` since `type` is an attribute for input elements and is commonly used.

For `List`, I think `variant` makes the most sense because we actually are passing it to the theme file as a variant.
Agree with Edwin on the `TextInput`, except maybe we could call the first `textInputVariant` to differentiate from `type`.

src/components/NewsletterSignup/__snapshots__/NewsletterSignup.test.tsx.snap
oliviawongnyc: Ah, I think it's bc you need to change `type` in the `Link` within the `NewsletterSignup` component.

src/components/Card/__snapshots__/Card.test.tsx.snap
oliviawongnyc: This is because you need to change the `type`s to `variant`s in the `Card.test.tsx` file.

src/components/Pagination/__snapshots__/Pagination.test.tsx.snap
```diff
@@ -129,6 +130,7 @@ exports[`Pagination Rendering renders the UI snapshot correctly 1`] = `
         id="firstPage-Next"
         rel={null}
         target={null}
+        type="action"
```
oliviawongnyc: Here, you need to change `type` in the `Pagination` component's `Link`.

oliviawongnyc (changes requested):
I think this is a great improvement! Couple little things to fix.

src/components/Banner/Banner.stories.tsx
```diff
@@ -64,7 +64,7 @@ export const WithControls: Story = {
   parameters: {
     design: {
       type: "figma",
-      url: "https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Main?type=design&node-id=86601-97661&mode=design&t=wZy1nqVOOZ4Dzuu2-11",
+      url: "https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Main?variant=design&node-id=86601-97661&mode=design&t=wZy1nqVOOZ4Dzuu2-11",
```
bigfishdesign13: The query string for the Figma URL should not have been changed.

src/components/Banner/Banner.tsx
bigfishdesign13: I also don't feel the addition of `Given` into the variable name is necessary, but I am curious about the reasoning for adding it. Can you share your thinking?

src/components/Banner/Banner.tsx
```diff
@@ -228,7 +228,7 @@ export const Banner: ChakraComponent<
     return (
       <Box
         as="aside"
-        data-type={type}
+        data-variant={variant}
```
bigfishdesign13, line 180: Why are we including this? @EdwinGuzman, do you recall?

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -34,8 +34,8 @@ export interface BreadcrumbsDataProps {
 export interface BreadcrumbProps extends BoxProps {
   /** Breadcrumb links as an array */
   breadcrumbsData: BreadcrumbsDataProps[];
-  /** Used to control how the `Hero` component will be rendered. */
-  breadcrumbsType?: BreadcrumbsTypes;
+  /** Used to control how the `Breadcrumbs` component will be rendered. */
```
bigfishdesign13, line 37: Ha! Funny copy-and-paste typo.

src/components/Button/Button.mdx
```diff
@@ -30,7 +30,7 @@ import Link from "../Link/Link";
 
 <Description of={ButtonStories} />
 
-Update the `buttonType` in the Controls section to see the types of buttons
+Update the `variant` in the Controls section to see the types of buttons
 that can be rendered.
```
bigfishdesign13: I find it interesting that we felt we needed to include instructions like this, but I don't think this line is necessary anymore.

src/components/Button/Button.mdx
```diff
@@ -105,16 +105,16 @@ is an `aria-label` passed to the `Button` component `aria-label="Previous"`.
...
-a call out. The "Back to Top" button must be of `buttonType="secondary"` and
+a call out. The "Back to Top" button must be of `variant="secondary"` and
 must include an up arrow icon on the right side.
 
 <Canvas of={ButtonStories.Patterns} />
 
 ## Other Examples
 
-The variations modified by the `buttonType` prop:
+The variations modified by the `variant` prop:
```
bigfishdesign13, line 112: I've seen this a few times now. We use the word "variation" in a sentence with `variant`. And it feels a little odd to me. I don't think it's a big deal, but I'm wondering if anyone else feels the same way.

src/components/Button/Button.tsx
```diff
@@ -94,11 +94,11 @@ export const Button: ChakraComponent<
       );
 
       if (childCount === 1 && hasIcon) {
-        variant = "iconOnly";
+        finalVariant = "iconOnly";
       }
```
bigfishdesign13, line 99: Ugh, I forgot that we force a specific style of `iconOnly` buttons.

src/components/List/List.tsx
```diff
@@ -39,7 +39,7 @@ export interface ListProps extends Omit<BoxProps, "title"> {
    * to Description Lists and will render above the list. */
   title?: string | JSX.Element;
   /** The type of list: "ol", "ul", or "dl". "ul" by default. */
-  type: ListTypes;
+  listContainer: ListContainers;
```
bigfishdesign13: We are talking about list types, so I think it is appropriate to use a prop name that includes `"type"`. And `listType` might make the most sense. I don't think `variant` is correct for this because we are talking about the root element. And `type` is also not correct because of all the reasoning behind this ticket.

src/components/StatusBadge/StatusBadge.tsx
```diff
@@ -43,10 +43,10 @@ export const StatusBadge: ChakraComponent<
 > = chakra(
   forwardRef<HTMLDivElement, React.PropsWithChildren<StatusBadgeProps>>(
     (props, ref?) => {
-      const { children, id, type, ...rest } = props;
+      const { children, id, variant, ...rest } = props;
```
bigfishdesign13, line 38: Not sure if this would be out of scope, but the default value for `variant` can be set to `"neutral"`, then the configuration for `useStyleConfig` can be simplified.

bigfishdesign13 (changes requested):
Overall, this looks very good. I left a few comments to consider. Also, and this could be an ask that eats up a bunch of time, the renamed props should be alphabetically repositioned in the interfaces and when components are implemented in tests, stories, etc.

src/components/Banner/Banner.tsx
7emansell: Just wanted to differentiate but i see how it's unnecessary and sometimes wrong– removed

src/components/Banner/Banner.tsx
7emansell: Is this for QA to access this element more easily?

7emansell:
Sorry for deserting this PR for a minute– as it stands now:
-  `List`'s  `type` /`listContainer` is now `variant` since 2/3 felt that made more sense... I'm neutral since I don't think any of the names so far have really been exactly right. @bigfishdesign13 what do you think– is this a change you want to discuss more?
- `TextInput`'s `textInputType`/`variant` is now `textInputVariant`
- `TextInput`'s `type`/`inputType` is now  back to `type`

and @oliviawongnyc thank you for double checking the snapshots 🙏
cc @EdwinGuzman

src/components/Banner/Banner.tsx
bigfishdesign13: Yes, it is for QA. All good here.

bigfishdesign13:
> * `List`'s  `type` /`listContainer` is now `variant` since 2/3 felt that made more sense... I'm neutral since I don't think any of the names so far have really been exactly right. @bigfishdesign13 what do you think– is this a change you want to discuss more?

No need to discuss further. I am ok with using `variant` for the prop name.

bigfishdesign13 (approved):
The positioning of the prop names should eventually be updated, but that is not a showstopper.

src/components/Breadcrumbs/Breadcrumbs.tsx
EdwinGuzman: 🙈

src/components/StatusBadge/StatusBadge.tsx
EdwinGuzman: True.... but out of scope!

src/components/StatusBadge/StatusBadge.mdx
```diff
@@ -56,14 +56,15 @@ variant. In this way, the user will be able to understand the priority of the
 
 The component offers two props for setting the visual style:
 
-- `type`
+- `variant`
+- `fontSize`
```
EdwinGuzman, line 60: I see the description says "two props" but did this always only have `type`? `fontSize` is more of the typical style prop that can be used for any component, so now sure we need to necessarily highlight it here.

Okay to keep, but just curious about this.

EdwinGuzman (approved):
Wow! Thank you so much for this. I hope the general `[componentName]Type` to `variant` update makes it cognitively easier for devs when coding. The consistency is nice to see.

PR #1836, opened 2025-06-13
DSD-2038: id and classname update

EdwinGuzman:

- I'm upset with the previous implementation of #1831 where I'm trying to set a default of `ds-[componentName][-${userPassedId}`. It's just not scalable in some components that have nested DS components. It was troublesome to see in the dom and in tests id values such as `ds-notification-ds-heading-heading1` or `ds-card-ds-link-ds-icon`, etc.

I'm scrapping those PRs but I want a closer review on this before I update the rest of the components.

## New Approach

1) All `id` props will be optional.
2) There are some components where we associate the id to different elements to meet accessibility requirements. For example, `input` and `label` _need_ a proper id value so they are "linked" together. In those situations, if no `id` is passed, we'll use the [`React.useId()`](https://react.dev/reference/react/useId) hook to generate an id randomly. If the user passes an `id`, that one is used instead.

This follows similar patterns in Material UI, Chakra, and Radix.

Aside: there might be some concerns with `useId` but it actually works well with Chakra. Chakra seems to export the same function as well but I rather stick with the React version. Since we're not supporting RSC, it's not a problem to use `useId` the way it's set up.

## How has this been tested?

- Locally and in Storybook. The `Accordion` tests were not modified and they continue to pass.

- None, but we should still ask Clare for our more complicated components.

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:


src/components/Accordion/Accordion.mdx
```diff
@@ -111,6 +112,19 @@ You may pass any Chakra `Box` props, as well as the props below.
 
 ## Accessibility
 
+<Banner
+  content={
+    <>
+      <strong>NOTE:</strong> It's recommended to pass an `id` value for
```
EdwinGuzman, line 120: This should be the message for every component that internally needs an `id`.

src/utils/utils.ts
```diff
@@ -160,6 +160,11 @@ export const sanitizeStringForAttribute = (str: string) => {
   return sanitizedStr;
 };
 
+/** Sanitize any string */
+export const sanitizeString = (str: string) => {
```
7emansell, line 164: can this just replace the stricter `sanitizeStringForAttribute` above it (i think that one is only used for link ids so likely would be removed as this work continues

src/hooks/useSafeId.tsx
```diff
@@ -0,0 +1,13 @@
+import { useId } from "react";
+
+import { sanitizeString } from "../utils/utils";
+
+/**
+ * This function is used to generate a unique ID for the component.
+ * It uses the `useId` hook from React to generate a unique ID and
+ * sanitizes it using the `sanitizeString` function.
+ */
+export function useSafeId(userId?: string) {
```
7emansell, line 10: do we care if an unsafe id is passed?

7emansell (approved):
Approach looks good just a couple questions about the overall implementation

oliviawongnyc (commented):
LGTM! Tho I think Emma has some good q's.

src/components/Accordion/Accordion.mdx
```diff
@@ -111,6 +112,19 @@ You may pass any Chakra `Box` props, as well as the props below.
...
+      accessibility requirements. If no `id` is passed, a random `id` will be
+      generated.
```
bigfishdesign13, line 122: This was a pattern that we used when the DS was very young and we ran into problems with randomly generated values being out of sync in certain testing situations. Will that not be a problem here because of `React.useId()`?

bigfishdesign13 (commented):
Question.

src/components/Accordion/Accordion.mdx
EdwinGuzman: Will add this as its own separate PR for all components at the end.

src/utils/utils.ts
EdwinGuzman: Yes, I was going to update this when I got to the Link component. But now that I actually got to it, I think we may need it and should be removed in a separate pr. The implementation for the link and why this is needed is tricky and maybe out of scope for this set of changes.

src/hooks/useSafeId.tsx
EdwinGuzman: 😨 is it our (DS) fault or the consumer's fault? You _could_ pass an unsafe id value to any HTML element right now and nothing will really stop you...

So I'm following that approach, but you are correct, we should have _some_ sanitization. I'll update this function, but in a later PR.

src/components/Accordion/Accordion.mdx
EdwinGuzman: Right. This was a problem before React 18 introduced `useId`. Hopefully, this new function helps us out. The docs are very helpful for this function.

oliviawongnyc (approved):

dgcohen (approved):

oliviawongnyc:
@bigfishdesign13 and I noticed there were some components that you added `extends BoxProps` to. I am assuming these are components I missed in my initial update? If so, the docs probably need to be updated with our line about "you may pass any Box props as well as...".

src/components/Banner/__snapshots__/Banner.test.tsx.snap
```diff
@@ -3,13 +3,15 @@
...
+  data-testid="ds-banner"
   data-variant="informative"
+  id="-r0-"
 >
   <svg
     aria-hidden={true}
     className="chakra-icon banner-icon css-1976tye"
     data-file-name="SvgErrorOutline"
-    data-testid="undefined-banner-icon"
+    data-testid="-r0--banner-icon"
```
bigfishdesign13: This seems a little odd: `-r0-`.

src/components/FeaturedContent/FeaturedContent.tsx
```diff
@@ -90,26 +92,27 @@ export const FeaturedContent: ChakraComponent<
...
+          id={mainId}
           __css={styles.base}
           ref={ref}
           {...rest}
         >
           <Box __css={styles.wrapper} data-wrapper>
             <Box
-              data-testid="featuredcontent-bg-image"
+              data-testid="ds-featuredContent-bgImage"
+              data-imagewrapper
```
bigfishdesign13, line 102: I don't recall seeing this done in other components. Should this be here and on the other main sections in this component?

src/components/FeedbackBox/__snapshots__/FeedbackBox.test.tsx.snap
```diff
@@ -3,11 +3,13 @@
...
   className="css-1xdhyk6"
-  id="feedbackbox"
+  data-testid="ds-feedbackBox"
+  id="-r0-"
 >
   <button
     className="chakra-button css-1xdhyk6"
-    id="open"
+    data-testid="ds-button"
+    id="-r0--feedback-open"
```
bigfishdesign13: This again (`-r0-`).

src/components/Accordion/Accordion.tsx
```diff
@@ -309,20 +309,21 @@ export const Accordion: ChakraComponent<
     return (
       <ChakraAccordion
         allowMultiple
+        data-testid="ds-accordion"
         index={expandedPanels}
         onChange={(expandedIdxs: number[]) => setExpandedPanels(expandedIdxs)}
         onKeyDown={handleKeyDown}
-        id={id}
+        id={mainId}
```
bigfishdesign13, line 316: Nit: alphabetize prop name

src/components/AudioPlayer/AudioPlayer.tsx
```diff
@@ -140,8 +132,8 @@ export const AudioPlayer: ChakraComponent<
           headingText={headingText}
           descriptionText={descriptionText}
           helperText={helperText}
-          id={`${id}-componentWrapper`}
-          data-testid="audio-player-component"
+          id={id}
+          data-testid="ds-audioPlayer"
```
bigfishdesign13: Nit: alphabetize prop name

bigfishdesign13 (commented):
Looks great. I just have one question about something that seems out of place.

src/components/Banner/__snapshots__/Banner.test.tsx.snap
EdwinGuzman: It's a random value from `React.useId`. They actually return values in the form of `:r4:` but I prefer `-` over `:` (`:` is completely valid in ids, but if the value is ever used for styling or querying, it's easier to use `-`).

src/components/FeaturedContent/FeaturedContent.tsx
EdwinGuzman: This is for the container query.

src/components/Banner/__snapshots__/Banner.test.tsx.snap
bigfishdesign13: That's what I though. I just found it odd that it was at the beginning of the `id` and that there was no prefix.

bigfishdesign13 (approved):

PR #1837, opened 2025-06-16
DSD-2038: id classname update pt2

EdwinGuzman:

- Continues the work started in #1836 .

## How has this been tested?

- Locally and with tests passing.

- N/A

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:


src/components/AudioPlayer/AudioPlayer.mdx
```diff
@@ -28,8 +28,8 @@ import { changelogData } from "./audioPlayerChangelogData";
 
 The `AudioPlayer` component is used to embed a streaming audio player from 3rd
 party services such as `Libsyn`, `SoundCloud`, and `Spotify` on a page. The
-component requires the `embedCode` prop or `filePath` prop. The component will
-not function properly if you try to set both of them.
+component requires the `embedCode` prop. The component will not function
+properly if you try to set both of them.
```
oliviawongnyc: This sentence doesnt make sense now  ("both of them").

src/components/Icons/Icon.tsx
```diff
@@ -37,8 +37,6 @@ export interface IconProps
   decorative?: boolean;
   /** Rotates the icon clockwise in increments of 90deg */
   iconRotation?: IconRotations;
-  /** ID that other components can cross reference for accessibility purposes */
```
oliviawongnyc, line 40: Still showing up on args table bc of the Story I believe

oliviawongnyc:
Is it true that you will always need to add an `id` to snapshot tests now so that `useSafeId` doesn't return a randomly generated id and break the test?

oliviawongnyc (changes requested):
Nice to see the lack of `undefined`s! Will there be a part 3 for this?

Could you add a few tests for what happens when ids are not passed? I know the id will be random, but maybe there is a test to show that it is not `undefined`.

src/components/AudioPlayer/AudioPlayer.mdx
EdwinGuzman: Removed.

EdwinGuzman:
> Is it true that you will always need to add an id to snapshot tests now so that useSafeId doesn't return a randomly generated id and break the test?

Good question... it doesn't seem like it's needed for all snapshot tests. It's "random" but not really random and somehow returns the same id for the component that is rendered. I imagine it's a react thing (similar to keys in loops). I  tested this locally and it's ok to not have an `id` value set for snapshot tests, they'll pass on every run.

EdwinGuzman:
> Will there be a part 3 for this?

Yes, and more.

> Could you add a few tests for what happens when ids are not passed? I know the id will be random, but maybe there is a test to show that it is not undefined.

That's a good point. To not make this PR even bigger, I'm going to update all tests in a later PR to group changes together.

oliviawongnyc (approved):

src/components/AlphabetFilter/AlphabetFilter.tsx
```diff
@@ -161,7 +159,13 @@ export const AlphabetFilter: ChakraComponent<
     };
 
     return (
-      <Box as="nav" ref={ref} aria-label="Filter by letter">
+      <Box
+        as="nav"
+        aria-label="Filter by letter"
+        data-testid="alphabetFilter"
+        id={id}
```
dgcohen, line 166: Do we still want this ID if it was removed from the component's props type?

src/components/AudioPlayer/AudioPlayer.tsx
```diff
@@ -140,8 +132,8 @@ export const AudioPlayer: ChakraComponent<
           headingText={headingText}
           descriptionText={descriptionText}
           helperText={helperText}
-          id={`${id}-componentWrapper`}
-          data-testid="audio-player-component"
+          id={id}
```
dgcohen, line 135: Same question as above

src/components/AlphabetFilter/AlphabetFilter.tsx
EdwinGuzman: Yes, it's an implicit prop now like various other HTML attributes.

src/components/AudioPlayer/AudioPlayer.tsx
EdwinGuzman: Yes, I'm removing them from the list of props but they're still available, just not listed. This is similar to other html attributes which can generally be used on any element.

dgcohen (approved):

dgcohen:
Thanks for the clarifications, looks good aside from the merge conflicts.

PR #1842, opened 2025-06-18
DSD-2084: enabled fontSize attribute for Table component

bigfishdesign13:
Fixes JIRA ticket [DSD-2084](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2084)

- Removes the `tableTextSize` prop from the `Table` component in favor of using the native `fontSize` CSS prop.

## How has this been tested?

- local Storybook

- n/a

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[DSD-2084]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-2084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Table/Table.tsx
```diff
@@ -20,10 +20,9 @@ interface CustomColors {
   color?: string;
 }
 
-export const tableBodyTextSizesArray = ["body1", "body2", "caption"] as const;
-export type TableBodyTextSizes = typeof tableBodyTextSizesArray[number];
-
 export interface TableProps extends BoxProps {
+  /** Additional class name for the `Table` component. */
+  className?: string;
```
oliviawongnyc: We shouldn't need to add this explicitly bc it's part of `BoxProps`. Was something not working?

src/components/Table/Table.stories.tsx
```diff
@@ -173,6 +168,7 @@ export const WithControls: Story = {
       columnHeaders={args.useColumnStyles ? columnHeadersAlt : columnHeaders}
       columnStyles={args.useColumnStyles ? columnStylesComplex : undefined}
       tableData={args.useColumnStyles ? tableDataAlt : tableData}
+      fontSize="8px"
```
oliviawongnyc: Were you just testing here with `8px`?

oliviawongnyc (changes requested):
Nice! I'm glad you figured it out, Marty! 

One thing I noticed is that the column headers do change to the `fontSize` when the layout switches to the mobile view. Is this intentional?

src/components/Table/Table.stories.tsx
bigfishdesign13: Oops. Yes. A silly use case, but doable.

src/components/Table/Table.tsx
bigfishdesign13: Good point.

bigfishdesign13:
> One thing I noticed is that the column headers do change to the `fontSize` when the layout switches to the mobile view. Is this intentional?

@oliviawongnyc Yes. In that use case, aligning the font styles between the headers and the content was intentional.

src/components/Table/Table.mdx
```diff
@@ -114,6 +115,31 @@ instead of rendering the visisble header.
 
 <Canvas of={TableStories.HideTitleText} />
 
+## Font Size
+
+The font size of the table content defaults to `1rem` (16px), but a custom size
+can be set using the `fontSize` CSS attribute. The font size of the column
+headers will always be `.75rem` (12px).
```
oliviawongnyc: Maybe add "...except for in mobile viewports?"

oliviawongnyc (approved):
One last little non blocking suggestion!

dgcohen (approved):

src/components/Table/Table.mdx
```diff
@@ -114,6 +115,31 @@ instead of rendering the visisble header.
...
+
+Recommended font sizes:
```
EdwinGuzman: Perhaps add after "sizes" " (design tokens):" to be explicit that these are ds-specific design tokens?

EdwinGuzman (approved):
Thanks for simplifying this. Made a minor comment but not a blocker.

PR #1845, opened 2025-06-24
DSD-2038: id classname update pt3

EdwinGuzman:

This does NOT update tests or changelogs, but will update them in a later PR. This PR is too big as it is and more components need to be updated.

- Continues the work in #1837 .
- General updates:
  - Adds `data-testid="ds-[componentName]"` on the top-level DOM element.
  - Uses `useSafeId` for component that need ids for accessibility requirements.
  - Leaves some components' ids alone if they go into a `Box`, since they don't render as "undefined".
  - Some more complicated components have additional suffixes for internal button or helper text components.

## How has this been tested?

- Storybook and locally.

- N/A

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:


src/components/ButtonGroup/__snapshots__/ButtonGroup.test.tsx.snap
```diff
@@ -3,16 +3,19 @@
 exports[`ButtonGroup Snapshot renders the UI snapshot correctly 1`] = `
 <div
   className="chakra-stack css-11okbqf"
+  data-testid="ds-buttonGroup"
```
dgcohen, line 6: Should this be `ds-button-group` to preserve pure kebab-case? Or are we ok mixing kebab and SnakeCase since the second part refers a Component?

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -80,7 +78,7 @@ const tooltipWrapperOrText = ({
           name="arrow"
           size="small"
           iconRotation="rotate90"
-          id={`${breadcrumbsID}__backarrow`}
+          id={id ? `${id}-breadcrumbs-backarrow` : undefined}
```
dgcohen, line 81: Is this meant to be a regular `id` or `data-testid`? I would imagine we'd get the same id attribute for each breadcrumb, since "id" seems to be passed down from the `BreadCumbs` container component.

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -50,15 +48,15 @@ const breadcrumbTextLength = 40;
...
   customLinkComponent,
   renderIcon = false,
+  id,
   isCurrentPage = false,
 }: {
   breadcrumbsData: BreadcrumbsDataProps;
-  breadcrumbsID: string;
   customLinkComponent: React.ElementType;
   renderIcon?: boolean;
+  id: string;
```
oliviawongnyc: Should this prop be optional? Seems like we are checking whether it exists below.

src/components/ButtonGroup/__snapshots__/ButtonGroup.test.tsx.snap
EdwinGuzman: I was thinking about it, but I want to do camel case for component names so that we can use kebab-case for additional internal components. Such as the following as an example:

```
  <Box data-testid="ds-buttonGroup">
    <button data-testid="ds-buttonGroup-close-btn">
```

src/components/Breadcrumbs/Breadcrumbs.tsx
EdwinGuzman: Hmm good point. I'll update this component.

oliviawongnyc: I'm not sure how the accessibility test works, but I am a bit surprised this doesn't make it fail.

src/components/DatePicker/DatePicker.tsx
```diff
@@ -487,25 +482,28 @@ export const DatePicker: ChakraComponent<
...
         isDateRange={isDateRange}
         showLabel={showLabel}
         labelText={labelText}
         isRequired={isRequired}
         showRequiredLabel={showRequiredLabel}
         {...rest}
       >
-        <DateRangeRow id={id} isDateRange={isDateRange}>
-          <FormField id={`${id}-start-form`}>
+        <DateRangeRow isDateRange={isDateRange}>
```
oliviawongnyc: `DateRangeRow` is no longer getting an id passed to it so it will have an `id` of `undefined-formRow`. I'm guessing it should be passed `mainId` like the others?

src/components/FilterBar/FilterBar.tsx
```diff
@@ -166,7 +165,13 @@ export const FilterBar: React.FC<React.PropsWithChildren<any>> = chakra(
         }
       );
       return (
-        <Box id={`filter-bar-${id}`} __css={styles} ref={ref} {...rest}>
+        <Box
+          id={`${mainId}`}
```
oliviawongnyc: Nit: simplify to id={mainId}

oliviawongnyc (changes requested):
Mostly just the `DateRangeRow` thing that needs to be adjusted. Then just a couple small questions.

src/components/Breadcrumbs/Breadcrumbs.tsx
EdwinGuzman: Maybe because this only renders on "mobile" view but tests don't have a "mobile" view....

src/components/Breadcrumbs/Breadcrumbs.tsx
EdwinGuzman: Yes, will update.

src/components/DatePicker/DatePicker.tsx
EdwinGuzman: Ah, will update.

EdwinGuzman:
Updated, will fix conflicts when the base branch is updated/merged

oliviawongnyc (approved):

dgcohen (approved):
Thanks for the updates!

PR #1848, opened 2025-06-25
DSD-2038: Updating id usage in List to Select components

EdwinGuzman:
Fixes JIRA ticket [DSD-2038](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2038)

## This PR does the following:

- Same as #1845 
- Updates components, alphabetically, from List to Select.

## How has this been tested?

- tests and storybook

## Accessibility concerns or updates

- Components that required ids now generated it if it's not passed to keep elements "connected".

### Accessibility Checklist

- [ ] Checked Storybook's "Accessibility" tab for color contrast and other issues.
- [ ] The feature works with keyboard inputs including tabbing back and forward and pressing space, enter, arrow, and esc keys.
- [ ] For hidden text or when `aria-live` is used, a screenreader was used to verify the text is read.
- [ ] For features that involve UI updates and focusing on DOM `ref`s, focus management was reviewed.
- [ ] The feature works when the page is zoomed in to 200% and 400%.

## Open Questions

-

### Checklist:

- [ ] I have updated the Storybook documentation and changelog accordingly.
- [ ] I have added relevant accessibility documentation for this pull request.
- [ ] All new and existing tests passed.

### Front End Review:

- [ ] Review the Vercel preview deployment once it is ready.

[DSD-2038]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-2038?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1850, opened 2025-06-25
DSD-2038: last PR to update the id usage for all components

EdwinGuzman:
Fixes JIRA ticket DSD-2038

- This is a continuation of #1848 
- This updates the rest of the components in the repo to have more consistent usage of the `id` prop and the data-testid prop for their parent wrapper.

## How has this been tested?

- tests and storybook.

- Components that have internal elements that require ids for accessibility reqs, randomly generate the id value if it's not provided.

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:


src/components/Tabs/Tabs.tsx
```diff
@@ -240,6 +239,7 @@ export const Tabs: ChakraComponent<
 
       return (
         <ChakraTabs
+          data-testid="ds-tabs"
           defaultIndex={defaultIndex}
           id={id}
```
oliviawongnyc: Should this be `id={mainId}`?

src/components/TagSet/TagSetExplore.tsx
```diff
@@ -13,7 +13,7 @@ export interface TagSetExploreDataProps {
   /** The content to display; should be a link-type component. */
```
oliviawongnyc, line 13: Should the `id` above be optional? It is not in `TagSetFilterDataProps` and maybe why there are so many `undefined`s in `src/components/TagSet/__snapshots__/TagSet.test.tsx.snap`?

oliviawongnyc (changes requested):
Same issue with the `id`s in the props tables and two little questions!

src/components/Tabs/Tabs.tsx
EdwinGuzman: hmm yes

src/components/TagSet/TagSetExplore.tsx
EdwinGuzman: Instead of optional, I updated the id value in the component.

EdwinGuzman: It's optional for "explore" but required for "filter"

oliviawongnyc (approved):

src/components/StructuredContent/StructuredContent.tsx
```diff
@@ -119,7 +117,11 @@ export const StructuredContent: ChakraComponent<
       });
       const calloutTextUpdate = calloutText ? (
         typeof calloutText === "string" ? (
-          <Heading id={`${id}-callout`} level="h3" size="heading5">
+          <Heading
+            id={id ? `${id}-callout` : undefined}
```
bigfishdesign13, line 121: Do you recall why there is an `id` on this element? I can understand for the title in the main heading, but the need for this `id` is unclear.

bigfishdesign13 (approved):
Looks good! I have one question, but it's not a blocker.

src/components/StructuredContent/StructuredContent.tsx
EdwinGuzman: No... I don't recall, I can also just remove it.

bigfishdesign13: I'm fine either way. Just something I noticed.

PR #1853, opened 2025-06-27
DSD-2038: updated tests for id prop update

EdwinGuzman:

- Wraps up the `id` prop update with accessibilty and updated tests. Sorry it's so big!
- The general update consists of two new additional tests for each component:
1) an accessibility test that does not include the `id` prop. There are some components with accessibility tests that don't include the `id` prop. I left those alone.
2) a test that does not include the id prop. For most components, it is simply verifying that there is no id attribute for the parent element. For the components that previously required an `id` prop, the test verifies that a "randomly generated" (through jest mock) id value is set.

## How has this been tested?

- all test?

- n/a

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:


src/components/Banner/Banner.test.tsx
```diff
@@ -7,7 +7,18 @@ import Heading from "../Heading/Heading";
 import Icon from "../Icons/Icon";
 import Banner from "./Banner";
 
+jest.mock("../../hooks/useSafeId", () => ({
+  ...jest.requireActual("../../hooks/useSafeId"),
+  useSafeId: jest.fn((id) => id || "test-id"),
+}));
```
oliviawongnyc, line 13: Is this mock worth pulling out to a shared file?

src/components/DatePicker/DatePicker.test.tsx
```diff
@@ -35,6 +40,13 @@ describe("DatePicker Accessibility", () => {
     expect(await axe(container)).toHaveNoViolations();
   });
 
+  it("passes axe accessibility for a single date input wiht no id", async () => {
```
oliviawongnyc: Nit: typo in "with"

src/components/Fieldset/Fieldset.test.tsx
```diff
@@ -15,6 +15,15 @@ describe("Fieldset Accessibility", () => {
     expect(await axe(container)).toHaveNoViolations();
   });
 
+  it("passes axe accessibility with id", async () => {
```
oliviawongnyc: Should this description switch with the one above that has an id being passed?

src/components/NewsletterSignup/NewsletterSignup.test.tsx
```diff
@@ -18,6 +23,25 @@ describe("NewsletterSignup Accessibility", () => {
   const onSubmit = jest.fn();
   const onChange = jest.fn();
   const valueEmail = "";
+
+  it("passes axe accessibility test with no id", async () => {
+    const { container } = render(
+      <NewsletterSignup
+        className="my-class"
+        id="my-id"
```
oliviawongnyc: This `id` should be removed.

src/components/RadioGroup/RadioGroup.test.tsx
```diff
@@ -75,9 +75,36 @@ describe("Radio Accessibility", () => {
     );
     expect(await axe(container)).toHaveNoViolations();
   });
+  it("passes axe accessibility with no id", async () => {
+    const { container } = render(
```
oliviawongnyc, line 79: I think generally you've only added this test in the components that need an id (and generate a random one if none is passed).

oliviawongnyc (changes requested):
Whew! Lot of work 🎉. Couple tiny things.

Not sure how much this really matters, but this test description – "should randomly generate an id if no id was passed" – the test doesn't really check this. It really only asserts that `useSafeId` runs and that an `id` is added to the component, so maybe it's "adds id to the component even if none is passed?"

Nit: this is _slightly_ nicer sounding imo but up to you
"should not render an id if none is passed" rather than "should not render an id if it's not passed"

src/components/Banner/Banner.test.tsx
EdwinGuzman: I was debating that, but it's not used everywhere so I don't want to set a global mock.

src/components/Fieldset/Fieldset.test.tsx
EdwinGuzman: I missed the word "no", so "with no id".

src/components/RadioGroup/RadioGroup.test.tsx
EdwinGuzman: Yes, generally. There are some components that did not have an id for the accessibility test, but I did not add an additional test _with_ the `id` value passed.

EdwinGuzman:
> Not sure how much this really matters, but this test description – "should randomly generate an id if no id was passed" – the test doesn't really check this. It really only asserts that useSafeId runs and that an id is added to the component, so maybe it's "adds id to the component even if none is passed?"

Yes, this is nicer and a bit more accurate. I am mocking the `useSafeId` function so it returns a specific value `test-id` and not test react's `useId`. So that distinction makes sense.

oliviawongnyc (approved):

7emansell (approved):

PR #1855, opened 2025-06-27
DSD-2041/remove children map

oliviawongnyc:
Fixes JIRA ticket [DSD-2041](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2041)

- Removes `Children.map` logic from `Button`, `ButtonGroup`, `Card`, `CheckboxGroup`, `Form`, `List`, `RadioGroup`, and `Tooltip` to reduce complexity and improve performance.
- Some more detail on how each component was handled can be found here: https://docs.google.com/spreadsheets/d/1fZ-C557yWZuPYsbDHnsnJ35gtGceKOckhy5E33tgZbc/edit?usp=sharing

## How has this been tested?

- Locally, with unit tests and in Storybook.

- None.

### Accessibility Checklist

## Open Questions

- Approach for any of the components is up for discussion, especially `Card` where I made the least change.

### Checklist:

### Front End Review:

[DSD-2041]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-2041?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Button/Button.mdx
```diff
@@ -38,13 +38,11 @@ import { changelogData } from "./buttonChangelogData";
 - `"secondary"` is used for actions that move the user back, such as cancellations.
 - `"text"` is used for equally weighted actions in a text based list.
 - `"callout"` is used for call to action text such as "Donate".
+- `"iconOnly"` is used to apply proper styling when the button contains an icon, but no text.
```
EdwinGuzman, line 41: Can you add this new variant to the "Other Examples" section?

Screenshot 2025-06-30 at 9 14 53 AM

EdwinGuzman: And Sizes (if that makes sense)

src/components/Button/buttonChangelogData.ts
```diff
@@ -13,14 +13,15 @@ export const changelogData: ChangelogData[] = [
...
-    affects: ["Functionality", "Styles"],
+    affects: ["Documentation", "Functionality", "Styles"],
     notes: [
       "Removes deprecated `link` buttonType.",
       "Changes theme name from `CustomButton` to `ReservoirButton` for consistency.",
       "Removes explicit `className` prop as interface can be extended to include Chakra prop or HTML attribute types.",
       "Removes `type` and `onClick` props (as HTML equivalents may be used) and updates docs.",
       "Renames `buttonType` to `variant`.",
       "Replaces `buttonSizesArray` with shared sizes array.",
+      "Removes `Children.map` logic and asks consuming apps to pass the `iconOnly` variant if the button only contains an icon. Also adds a hook to consume `ButtonGroup` context and moves `gap` style to theme file.",
```
EdwinGuzman: How about instead of "asks", it's rephrased such as "consuming apps must now use the iconOnly variant for icon-only buttons."

src/components/Card/Card.tsx
```diff
@@ -261,56 +262,52 @@ export const Card: ChakraComponent<
...
-            child.type === CardContent ||
-            child.props.mdxType === "CardContent"
-          ) {
-            const elem = React.cloneElement(child, { key });
-            cardContents.push(elem);
-          } else if (isCardActions) {
-            const elem = React.cloneElement(child, { key, isCentered, layout });
+      childrenArray.forEach((child, key) => {
+        const isCardActions =
+          child.type === CardActions || child.props.mdxType === "CardActions";
```
EdwinGuzman, line 267: I forgot why we needed to do `props.mdxType` and at this point, I think Storybook has evolved so much that we may not need it. Something to update (remove?) for another PR.

src/components/List/List.tsx
EdwinGuzman, line 1: 💯

CHANGELOG.md
```diff
@@ -51,6 +51,7 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 - Removes explicit `className` and `children` props in favor of expanded prop type definitions in all components.
 - Removes the use of the `window`'s `scrollIntoView` dependency in the `Tabs` component.
 - Removes the `tableTextSize` prop from the `Table` component in favor of using the native `fontSize` CSS prop.
+- Removes `Children.map` logic from `Button`, `ButtonGroup`, `Card`, `CheckboxGroup`, `Form`, `List`, `RadioGroup`, and `Tooltip` to reduce complexity and improve performance.
```
EdwinGuzman: I think `Form` is still using `Children.map` (so does `FilterBar` but that component needs to be completely removed from the repo so ok to ignore for now).

EdwinGuzman (commented):
This is great! Thanks. Just a few minor comments. I'm glad the context works and it seems lightweight.

Tbh, I need to work out how it's able to keep context separate when there are many of the same components on the same page. Something for me to further understand but not a blocker. I verified this in the stories by adding different props to the same component in a group. For example, `checkboxGroup`'s Layout story has two `CheckboxGroup` components side-by-side. If one is disabled and the other invalid, the context works correctly.

EdwinGuzman:
@charmingduchess this PR fixes https://newyorkpubliclibrary.atlassian.net/browse/DSD-1672 just an fyi. If you have time to review, please feel free to do so.

src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -49,33 +47,13 @@ export const ButtonGroup: ChakraComponent<
...
-                return;
-              }
-            }
-            const disabledProps = isDisabled ? { isDisabled } : {};
-            newChildren.push(
-              React.cloneElement(child, { key, ...disabledProps })
-            );
-          }
-        }
+      const buttonGroupContextValue = useMemo(
```
charmingduchess: I think this `useMemo` might be unnecessary. You could use `isDisabled` as the single value passed in to the button group context, though I understand the pattern you're going for here.

charmingduchess (commented):
Great solution for the CheckBox group! I'm excited to see the impact this has on RC performance.

src/components/ButtonGroup/ButtonGroup.tsx
oliviawongnyc: Adjusted!

src/components/Card/Card.tsx
oliviawongnyc: I did some minimal testing, and the components seem to work the same in storybook without the props.mdxType check. Would you like me to remove?

EdwinGuzman: Yes, but different ticket.

oliviawongnyc: Kay, will create

oliviawongnyc:
> Great solution for the CheckBox group! I'm excited to see the impact this has on RC performance.

Thank you for flagging and for taking a look at the PR! Hope it helps! @charmingduchess

package.json
```diff
@@ -53,16 +53,16 @@
     "types-ramda": "0.29.4"
   },
   "dependencies": {
-    "@chakra-ui/react": "2.8.2",
-    "@chakra-ui/system": "2.6.2",
+    "@chakra-ui/react": ">=2.8.2",
+    "@chakra-ui/system": ">=2.6.2",
```
EdwinGuzman: These are set for the peer dependencies but why are they set for the normal set of dependencies?

src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -89,7 +62,9 @@ export const ButtonGroup: ChakraComponent<
           sx={styles}
           {...rest}
         >
-          {newChildren}
+          <ButtonGroupContext.Provider value={isDisabled}>
```
EdwinGuzman, line 65: ~I like the `useButtonGroup` hook pattern more than this approach, but nbd.~
Nvm, got the file mixed up, that hook is used in `Button.tsx`.

EdwinGuzman (approved):
Just a comment about package.json.

package.json
oliviawongnyc: This is fixed now, but I am not sure why it happened automatically. It took a weird workaround for me to be able to install rollup without having these update.

PR #1856, opened 2025-06-27
DSD-2088: remove default spacing

bigfishdesign13:
Fixes JIRA ticket [DSD-2088](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2088)

- Removed `noSpace` prop from `Heading` component.
- Removed `noSpace` prop from `Text` component.
- Removed default spacing values set on the `DatePicker`, `Heading`, `HelperErrorText`, `HorizontalRule`, `Label`, `List`, `StyledList`, and `Text` components.

## How has this been tested?

- local Storybook

- n/a

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[DSD-2088]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-2088?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/FeaturedContent/FeaturedContent.stories.tsx
```diff
@@ -71,14 +71,14 @@ export const WithControls: Story = {
       isFullWidth={args["isFullWidth"]}
       textContent={
         <div>
-          <Heading level="h2" overline="Featured">
+          <Heading level="h2" mb="s" overline="Featured">
             Spotlight on the Public Domain
           </Heading>
-          <p>
+          <Text mb="s">
```
EdwinGuzman, line 77: Would it make sense to add `mb="s"` to the `p` tag for a global style rule?

src/components/FeedbackBox/FeedbackBox.tsx
```diff
@@ -454,14 +454,14 @@ export const FeedbackBox: ChakraComponent<
                           Thank you for submitting your feedback.
                         </Text>
                         {showEmailField && (
-                          <Text>
+                          <Text mt="s">
```
EdwinGuzman, line 457: The others had mb but this has mt, why?

src/components/Hero/Hero.stories.tsx
```diff
@@ -110,12 +111,15 @@ export const WithControls: Story = {
...
         backgroundImageSrc={getPlaceholderImage()}
         heading={
-          <Heading
-            level="h1"
-            id="1"
-            subtitle="Example Subtitle"
-            text="Hero Primary"
-          />
+          <Box mb="s">
+            <Heading
```
EdwinGuzman: why?

src/components/Hero/Hero.stories.tsx
```diff
@@ -157,14 +161,17 @@ export const Primary: Story = {
...
-          id="primary-hero"
-          level="h1"
-          noSpace
-          overline="Hero Example"
-          subtitle="Integer posuere erat a ante venenatis dapibus posuere velit aliquet. Donec sed odio dui. Vestibulum id ligula porta felis euismod semper. Aenean eu leo quam."
-          text="Hero Primary"
-        />
+        <div>
+          <Box>
+            <Heading
```
EdwinGuzman: Why!

src/components/Logo/Logo.stories.tsx
```diff
@@ -84,8 +84,8 @@ const logoRow = (logo, opts: any = {}) => {
   let key = logo;
 
   if (logo.indexOf("White") !== -1 || logo.indexOf("Negative") !== -1) {
-    styles.backgroundColor = "var(--nypl-colors-ui-gray-xx-dark)";
-    styles.color = "var(--nypl-colors-ui-white)";
+    styles.backgroundColor = "var(--nypl-colors-dark-ui-bg-default)";
+    styles.color = "var(--nypl-colors-ui-white) !important";
```
EdwinGuzman: Any way you can remove the `!important`?

src/components/StructuredContent/StructuredContent.stories.tsx
```diff
@@ -122,17 +122,17 @@ export const WithHTMLStringTextContent: Story = {
         "eiusmod tempor incididunt ut labore et dolore magna aliqua. " +
         "Pellentesque massa placerat duis ultricies lacus sed turpis tincidunt. " +
         "<a href='#'>Porttitor eget dolor</a> morbi non arcu risus quis varius. " +
-        "Faucibus in ornare quam viverra orci sagittis.</p><p>Nisl vel pretium " +
+        "Faucibus in ornare quam viverra orci sagittis.</p><p style='margin-top: 1rem'>Nisl vel pretium " +
```
EdwinGuzman, line 125: Can we target the `p` element when it's within the `StructureContent` component?

src/components/StyleGuide/ColorCard.tsx
```diff
@@ -1434,34 +1434,24 @@ export const ColorCard = (props: PropsWithChildren<ColorCardProps>) => {
...
             {backgroundColor}
           </Heading>
           <Text
-            fontWeight="medium"
-            lineHeight="20px"
-            noSpace
-            size="tag"
-            m="0"
             color={textColor}
+            fontSize="var(--nypl-fontSizes-desktop-caption) !important"
```
EdwinGuzman, line 1442: This should work with the design token, no?

src/theme/foundations/global.ts
```diff
@@ -1,5 +1,5 @@
 import { activeFocus } from "../components/global";
-import { textMargin } from "../components/global";
+// import { textMargin } from "../components/global";
```
EdwinGuzman: remove the comments

src/theme/foundations/spacing.ts
```diff
@@ -106,6 +106,12 @@ const reservoirSpacingTokens = {
       py: reservoirSpacingValues["s"],
     },
   },
+  heading: {
```
EdwinGuzman, line 109: I was wondering whether this update should go into the spacing style guide but I realized we don't really show any semantic tokens there. Should we?

EdwinGuzman (commented):
Looks good but have a few questions.

src/components/FilterBarPopup/FilterBarPopup.tsx
```diff
@@ -84,8 +84,8 @@ export const FilterBarPopup: ChakraComponent<
       });
 
       const generalHeadingProps = {
+        mb: "heading.default",
```
oliviawongnyc: This was `noSpace` before, so are we updating spacing here?

src/components/FeaturedContent/FeaturedContent.stories.tsx
bigfishdesign13: That is essentially what was removed as part of this update with the expectation that consuming apps could add something like that on the app level.

src/components/FeedbackBox/FeedbackBox.tsx
bigfishdesign13: I used `mt` here because of the way the conditions are set up. The first `Text` element will always be there, but we only want spacing below it when there are subsequent elements. Therefore, it made the most sense to add the spacing to the conditional elements.

src/components/Hero/Hero.stories.tsx
bigfishdesign13: Oops. This is leftover from when I first discovered the props problem within the `Heading` component. The extra wrapping elements will be removed.

src/components/Hero/Hero.stories.tsx
bigfishdesign13: Like above, this is leftover from when I first discovered the props problem within the `Heading` component. The extra wrapping elements will be removed.

src/components/StyleGuide/ColorCard.tsx
bigfishdesign13: If this did not need the `!important`, yes, I could use the standard Chakra token convention. But let me test again to see why I needed `!important` in the first place.

src/components/FeaturedContent/FeaturedContent.stories.tsx
EdwinGuzman: Sorry, I meant "global" in the sense of global for the component but still restricted to the `FeaturedContent` component.

src/components/StructuredContent/StructuredContent.stories.tsx
bigfishdesign13: No one is using this component, so I don't think we should try to get too fancy.

src/theme/foundations/spacing.ts
bigfishdesign13: Ah, yes, these additions should be added to the Spacing Style Guide. We do include semantic tokens there. 

image

I also need to update the Design Tokens Development Guide.

src/theme/components/global.ts
```diff
@@ -126,6 +125,10 @@ const labelLegendText = {
   },
 };
 // Used in form inputs that require nuanced spacing.
+const labelLegendTextDefaultSpacing = {
```
oliviawongnyc: Where is this expected to be used? Is the comment meant to be repeated below?

src/components/Logo/Logo.stories.tsx
bigfishdesign13: Apparently I can remove it. Not sure why I needed it when I did the initial development. :)

src/components/FeaturedContent/FeaturedContent.stories.tsx
bigfishdesign13: The content area for the `FeaturedContent` component is built up entirely by a consuming app, so I don't think we should add any global styles. If we went that route, where would we draw the line for which components get default styles and which do no?

src/theme/components/helperErrorText.ts
```diff
@@ -1,31 +1,17 @@
...
-const baseStyle = definePartsStyle(
-  ({ isInvalid }: HelperErrorTextBaseStyle) => ({
+const HelperErrorText = defineStyleConfig({
+  baseStyle: defineStyle(({ isInvalid }: HelperErrorTextBaseStyle) => ({
     fontSize: "desktop.helper.default",
     color: isInvalid ? "ui.error.primary" : null,
     _dark: {
       color: isInvalid ? "dark.ui.error.primary" : null,
     },
-    innerChild: innerChildStyle,
```
oliviawongnyc, line 23: Looks like `styles.innerChild` can be removed from the component file.

src/theme/components/list.ts
```diff
@@ -116,9 +113,7 @@ const List = defineMultiStyleConfig({
     ul: definePartsStyle((props) => ({
       base: unorderedStyles(props),
     })),
-    ol: definePartsStyle({
-      base: textMargin,
-    }),
+    ol: definePartsStyle({}),
```
oliviawongnyc: I don't think this is necessary, but maybe it's clearer?

oliviawongnyc (commented):
Wow, looks like this required a lot of focus! So many little changes. Edwin suggested we do a review of every storybook page as part of the v4 qa so will be keeping this update in mind then in case anything was missed. 

I wonder how much work this will put on consuming apps. It could be easy to miss if it's only a visual breaking change, but could make a lot of things look off. Will need to think about this and include in our migration guide and estimate.

EdwinGuzman:
I'll re-review this after the `reservoir-v4` branch is merged in.

src/components/FilterBarPopup/FilterBarPopup.tsx
bigfishdesign13: Good catch. This should be removed.

src/theme/components/global.ts
bigfishdesign13: These can actually both be replaced with design tokens. I will be updating the spacing values to reflect this change.

src/theme/components/list.ts
bigfishdesign13: Yes, that is why I left it. But I'll remove it and leave a comment instead.

bigfishdesign13:
> I'll re-review this after the `reservoir-v4` branch is merged in.

@EdwinGuzman Done.

oliviawongnyc (approved):

EdwinGuzman (approved):

PR #1857, opened 2025-07-02
DSD-2097: props bug in Heading

bigfishdesign13:
Fixes JIRA ticket [DSD-2097](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2097)

- Updates how `...rest` is applied in `Heading` component.

## How has this been tested?

- local Storybook

- n/a

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[DSD-2097]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-2097?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Heading/Heading.stories.tsx
+++ b/src/components/Heading/Heading.stories.tsx
@@ -67,9 +67,14 @@ export const WithControls: Story = {
       <Heading
         {...args}
         bgColor="brand.primary"
-        color="ui.typography.inverse.heading"
+        color="ui.white"
         px="s"
         py="xs"
+        __css={{
+          h2: {
+            color: "ui.white",
+          },
+        }}
       />
     </VStack>
   ),
--- a/src/components/Heading/Heading.stories.tsx
+++ b/src/components/Heading/Heading.stories.tsx
@@ -70,11 +70,6 @@ export const WithControls: Story = {
         color="ui.white"
         px="s"
         py="xs"
-        __css={{
-          h2: {
-            color: "ui.white",
-          },
-        }}
       />
     </VStack>
   ),
--- a/src/components/Heading/Heading.tsx
+++ b/src/components/Heading/Heading.tsx
@@ -195,10 +195,8 @@ export const Heading: ChakraComponent<
               ...styles.base,
               ...wrapperStyles,
             };
-      /** The `...rest` prop should only be applied to the <h> element when
-       * neither overline nor subtitle are passed. If either of those two props
-       * are passed, `...rest` will be applied to the <hgroup> element as
-       * stipulated in the final return statement. */
+      /** If there is an `overline` or a `subtitle`, `...rest` will be passed to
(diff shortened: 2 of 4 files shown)


src/components/Heading/Heading.tsx
```diff
@@ -195,6 +195,11 @@ export const Heading: ChakraComponent<
               ...styles.base,
               ...wrapperStyles,
             };
+      /** When overline or subtitle is passed, the ...rest props should not be
+       * applied to the <h> element and should only be applied to the <hgroup>
+       * element. The conditional statement below is written such that when
+       * neither of those props are passed, ...rest is a applied. */
+      const headingRest = !overline && !subtitle && { ...rest };
```
EdwinGuzman, line 200: If it's either overline or subtitle, then change the first `&&` to `||`

bigfishdesign13: This should not be an `or` statement. `...rest` should only be applied here when both props are not passed.

EdwinGuzman: Gotcha, the beginning of the comment threw me off, "When overline or subtitle is passed, ..."

src/components/Heading/Heading.stories.tsx
```diff
@@ -67,9 +67,14 @@ export const WithControls: Story = {
...
         bgColor="brand.primary"
-        color="ui.typography.inverse.heading"
+        color="ui.white"
         px="s"
         py="xs"
+        __css={{
+          h2: {
+            color: "ui.white",
+          },
+        }}
```
oliviawongnyc: Why is this necessary?

src/components/Heading/Heading.tsx
```diff
@@ -195,6 +195,11 @@ export const Heading: ChakraComponent<
               ...styles.base,
               ...wrapperStyles,
             };
+      /** The `...rest` prop should only be applied to the <h> element when
+       * neither overline nor subtitle are passed. If either of those two props
+       * are passed, `...rest` will be applied to the <hgroup> element as
+       * stipulated in the final return statement. */
```
oliviawongnyc: What about simplifying this? 

"If there's an `overline` or a `subtitle,` `...rest` will be passed to the `<hgroup>`, otherwise, it will be passed directly to the `<h>`."

oliviawongnyc (approved):
Makes sense. Small suggestion and question!

EdwinGuzman (approved):

src/components/Heading/Heading.stories.tsx
bigfishdesign13: Hmm, I am not sure. I recall that the `color` value was not being passed properly to the `<h>` element, but that seems to be working fine now. I'll get rid of this.

PR #1870, opened 2025-07-14
DSD-1818: Storybook show code fix

EdwinGuzman:
Fixes JIRA ticket [DSD-1818](https://newyorkpubliclibrary.atlassian.net/browse/DSD-1818)

- Resolves the "show code" issue for some components where the rendered name was not displaying. This was fixed by adding `.displayName = "..."` and the name of the component wrapper in a component's story.
- Also adds Storybook 9's "Code panel" which is a replacement for the storysource addon: https://storybook.js.org/docs/writing-docs/code-panel
  - This doesn't change much for us but I can simply remove it if it's too much noise. Let me know what you think @oliviawongnyc @7emansell @bigfishdesign13 
  - I think Storybook wants to push everyone to visit the "Controls" over a docs page (it is meant to be an interactive docs framework).
Screenshot 2025-07-14 at 4 46 03 PM

## How has this been tested?

- Storybook.

NOTE: How we run storybook locally in dev mode is _not_ the same as the static build in Vercel or gh-pages. The component names in the "show code" blocks rendered properly locally in dev mode. I had to test by building and reviewing the static site. Something to keep in mind for related issues in the future.

- N/A

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[DSD-1818]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-1818?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.storybook/main.ts
```diff
@@ -23,11 +23,6 @@ const config: StorybookConfig = {
     "@storybook/addon-jest",
   ],
 
-  // This gets added automatically through react-vite,
```
EdwinGuzman, line 26: Cleaned up this file a bit.

.storybook/main.ts
```diff
@@ -49,7 +44,6 @@ const config: StorybookConfig = {
     // Type-check stories during Storybook build.
     check: true,
     // Display the compiled value options.
-    reactDocgen: "react-docgen-typescript",
```
EdwinGuzman, line 52: The default is "react-docgen" in S9 and is recommended. We don't lose any of the props in the args tables.

EdwinGuzman:
There are more updates we can make in Storybook 9, but they're not blockers for v3 or v4.

jackiequach (approved):
This looks good! I think that the "Show code" buttons in the components on doc pages should be disabled if possible since they don't add any value as there are usually code examples provided in the docs.

src/components/FilterBarInline/FilterBarInline.stories.tsx
```diff
@@ -193,6 +193,7 @@ const FilterBarStory = (args) => {
     />
   );
 };
+FilterBarStory.displayName = "FilterBarStory";
```
oliviawongnyc, line 193: This is better than before, but still a little confusing to me bc it looks like `FilterBarStory` is the name of the component.

Image

EdwinGuzman: That's true. It's because we're rendering the component wrapper and not a direct `FilterBarInline` component. Perhaps it's better to just not display this at all.

oliviawongnyc: Yeah I guess ideally we would just write our own code snippets (assuming we can't get them to make sense automatically), but that would be a lot of work.

EdwinGuzman: So remove for now?

bigfishdesign13: We've been adding more and more of our own code examples, so I think it would be best to remove this functionality. If the problem persists (here and in the new `Code` panel!), then this really is not useful. The new pattern for the story header could include a code example below the rendered component.

EdwinGuzman:
@oliviawongnyc @bigfishdesign13 removed the "show code" button from `Story` and `Canvas` blocks.

oliviawongnyc (approved):

PR #1881, opened 2025-07-22
V4 QA feedback updates

EdwinGuzman:
Updates based on QA feedback reported in:

- https://newyorkpubliclibrary.atlassian.net/browse/DSD-2119
- https://newyorkpubliclibrary.atlassian.net/browse/DSD-2120
- https://newyorkpubliclibrary.atlassian.net/browse/DSD-2121


src/components/AlphabetFilter/AlphabetFilter.stories.tsx
```diff
@@ -27,8 +27,8 @@ export const WithControls: Story = {
   args: {
     activeLetters: undefined,
     currentLetter: undefined,
-    descriptionText: "",
-    headingText: "",
+    descriptionText: "This is description text.",
+    headingText: "AlphabetFilter",
```
bigfishdesign13: I removed these when I updated the story header. I opted to show the component in it's default state without the extra heading and description text. Did that cause problems somewhere?

src/components/Slider/Slider.tsx
```diff
@@ -263,7 +263,7 @@ export const Slider: ChakraComponent<
             // Specific margins for each text input to
             // push the elements inside.
             marginRight: type === "start" ? "s" : null,
-            marginLeft: type === "end" ? "s" : null,
+            marginLeft: type === "end" ? "xs" : null,
```
bigfishdesign13: I looked at the comment from Alkim and I don't think this is addressing the issue. The problem seems to be with the actual text input element. They are too wide for both instances (left and right).

src/theme/components/multiSelectItemsCountButton.ts
```diff
@@ -13,7 +13,7 @@ const MultiSelectItemsCountButton = defineStyleConfig({
     fontSize: "desktop.caption",
     justifyContent: "flex-end",
     left: "15px",
-    marginRight: "xs",
+    marginRight: "m",
```
bigfishdesign13, line 16: Any thoughts on what would have caused this shift in the spacing?

bigfishdesign13 (commented):
Just some questions.

src/components/AlphabetFilter/AlphabetFilter.stories.tsx
EdwinGuzman: Not problems, but more so that Alkim pointed out the difference from prod as an issue. I can remove it and make a note of it.

src/components/Slider/Slider.tsx
EdwinGuzman: I kind of thought so as well but wasn't sure. I'll take another stab at this.

src/theme/components/multiSelectItemsCountButton.ts
EdwinGuzman: Perhaps a merge conflict?

oliviawongnyc: I didn't see this issue in Alkim's notes. Did I miss it or is this from somewhere else?

EdwinGuzman: It is `SS_4.png` in https://newyorkpubliclibrary.atlassian.net/browse/DSD-2119

oliviawongnyc: Ah I see now. Thank you!

src/components/DatePicker/DatePicker.stories.tsx
```diff
@@ -81,7 +81,7 @@ export const WithControls: Story = {
     isInvalid: false,
     isRequired: false,
     labelText: "Select the date you want to visit NYPL",
-    maxDate: "1/1/2025",
+    maxDate: "1/1/2027",
```
oliviawongnyc, line 84: Do you think we should future proof this?

oliviawongnyc (commented):
You said this should be fixed now in the ticket: "NewsletterSignup > Submit button flexed under email box on QA vs next to the box on Prod.  SS_5", but I don't see what update fixes this. Also, is this just a discrepancy bc of the new breakpoints ?

src/components/DatePicker/DatePicker.stories.tsx
EdwinGuzman: Kind of?... We're going to run into this issue all the time :\

EdwinGuzman:
> Also, is this just a discrepancy bc of the new breakpoints ?

@oliviawongnyc yes. It's hard to notice the fix because it's an issue we created in v4, so we're not comparing it against prod

bigfishdesign13:
> You said this should be fixed now in the ticket: "NewsletterSignup > Submit button flexed under email box on QA vs next to the box on Prod. SS_5", but I don't see what update fixes this. Also, is this just a discrepancy bc of the new breakpoints ?

@oliviawongnyc The `NewsletterSignup` button bug was an issue that I handled in a separate ticket, so there should be no work in this PR related to that.

oliviawongnyc:
> > You said this should be fixed now in the ticket: "NewsletterSignup > Submit button flexed under email box on QA vs next to the box on Prod. SS_5", but I don't see what update fixes this. Also, is this just a discrepancy bc of the new breakpoints ?
> 
> @oliviawongnyc The `NewsletterSignup` button bug was an issue that I handled in a separate ticket, so there should be no work in this PR related to that.

Ah, got it. Thanks @bigfishdesign13.

src/components/Slider/Slider.tsx
bigfishdesign13: Were any additional updates made for the styles in `Slider`?

oliviawongnyc: @bigfishdesign13, does your `Searchbar` fix fix this? I know you adjusted the `TextInput` styles and it looks good in the vercel preview:

https://nypl-design-system-git-dsd-2066-searchbar-ui-issues-nypl.vercel.app/?path=/story/components-form-elements-slider--range-slider-with-controls

EdwinGuzman: Yes, it seems like it does. I will revert this and rely on the Searchbar fix.

EdwinGuzman:
The Slider margin issue was fixed in the SearchBar UI updates PR. I reverted the update here. @bigfishdesign13 @oliviawongnyc

src/components/Slider/Slider.tsx
bigfishdesign13: The `Slider` looks good now. Yay for shared elements! 🎉

bigfishdesign13 (approved):
The updates in this PR look good.

PR #1888, opened 2025-07-28
DSD-2124: Heading component text color bug

bigfishdesign13:
Fixes JIRA ticket [DSD-2124](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2124)

- Adjusted `Heading` styles to support custom colors.

## How has this been tested?

- local Storybook and new examples in Storybook

- n/a

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[DSD-2124]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-2124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Heading/Heading.tsx
+++ b/src/components/Heading/Heading.tsx
@@ -164,9 +164,11 @@ export const Heading: ChakraComponent<
         ? "subtitle1"
         : "subtitle2";
 
-      /** The styles that should be applied to the outer-most wrapper of the
-       * Heading component. */
-      const wrapperStyles = styles.headingWrapper;
+      /** To circumvent how Chakra controls style props, the text color value
+       * was moved into its own theme "part" and then applied to the UI
+       * appropriately. This method allows custom colors to override the default
+       * color value. */
+      const colorStyle = styles.colorStyle;
 
       /** The styles for the actual native heading element. If the native
        * element is going to sit by itself, without the overline or subtitle
@@ -179,7 +181,7 @@ export const Heading: ChakraComponent<
             }
           : {
               ...styles.base,
-              ...wrapperStyles,
+              ...colorStyle,
             };
       /** If there is an `overline` or a `subtitle`, `...rest` will be passed to
        * the `<hgroup>`, otherwise, it will be passed directly to the `<h>`. */
@@ -232,7 +234,7 @@ export const Heading: ChakraComponent<
           aria-roledescription="Heading group"
           data-testid="ds-heading-group"
           role="group"
-          sx={{ ...wrapperStyles }}
+          sx={{ ...colorStyle }}
           {...rest}
         >
           {finalContent}
--- a/src/components/Heading/Heading.tsx
+++ b/src/components/Heading/Heading.tsx
@@ -164,28 +164,21 @@ export const Heading: ChakraComponent<
         ? "subtitle1"
         : "subtitle2";
 
-      /** To circumvent how Chakra controls style props, the text color value
-       * was moved into its own theme "part" and then applied to the UI
-       * appropriately. This method allows custom colors to override the default
-       * color value. */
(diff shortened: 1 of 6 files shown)


src/theme/components/heading.ts
```diff
@@ -162,6 +161,9 @@ const Heading = defineMultiStyleConfig({
           : isLowercase
           ? "lowercase"
           : null,
+      },
+      colorStyle: {
```
oliviawongnyc: What do you think of calling this `defaultColorStyle`?

src/theme/components/heading.ts
```diff
@@ -5,7 +5,7 @@ import { baseLinkStyles } from "./link";
 // This function creates a set of function that helps us
 // create multipart component styles.
 const { defineMultiStyleConfig, definePartsStyle } =
-  createMultiStyleConfigHelpers(["base", "headingWrapper"]);
```
oliviawongnyc, line 8: Were we using `headingWrapper` in this theme file? I don't see it below.

src/components/Heading/Heading.tsx
```diff
@@ -164,9 +164,11 @@ export const Heading: ChakraComponent<
         ? "subtitle1"
         : "subtitle2";
 
-      /** The styles that should be applied to the outer-most wrapper of the
-       * Heading component. */
-      const wrapperStyles = styles.headingWrapper;
+      /** To circumvent how Chakra controls style props, the text color value
+       * was moved into its own theme "part" and then applied to the UI
+       * appropriately. This method allows custom colors to override the default
+       * color value. */
```
oliviawongnyc: Do you think this note will make sense to our future selves :)? I think this is captured in the other note about things going into hgroup vs h, and the precedence piece is important.

oliviawongnyc (commented):
Good thinking on creating the extra part!! I think it would have taken me awhile to get there.

I liked your idea of combining the two props because they both fall in the same situation. They need to be applied to the heading if there is no overline/subtitle and applied to the hgroup if there are. 

What do you think about the following for comments and syntax?

```
   /**
       * `colorStyle` defines the default color for the heading.
       * Any custom color passed as a prop will override this default,
       * following standard CSS precedence rules.
       */
      const colorStyle = styles.colorStyle;

      /** If there is an `overline` or a `subtitle`, `...rest` and `...colorStyle`
       * will be passed to the `<hgroup>`, otherwise, they will be passed directly
       * to the `<h>`. */
      const headingProps =
        overline && subtitle
          ? { sx: { ...styles.base } }
          : {
              ...rest,
              sx: { ...styles.base, ...colorStyle },
            };
```

Then, you would pass `headingProps` like this:

```
 <ChakraHeading
            as={asHeading}
            data-testid="ds-heading"
            id={id}
            ref={ref}
            {...headingProps}
          >
            {content}
          </ChakraHeading>
```

src/theme/components/heading.ts
bigfishdesign13: I like that. Interestingly, I struggled with the name for this variable.

src/theme/components/heading.ts
bigfishdesign13: No. It was declared, but then never implemented.

src/components/Heading/Heading.tsx
bigfishdesign13: I will alter the wording. This certainly is not a definitive method for overcoming the Chakra native style prop issue. Rather, it is a solution that worked for us here.

oliviawongnyc (approved):
🎉

EdwinGuzman (approved):

PR #1893, opened 2025-07-30
DSD-2135: Fixing Select helper text alignment

EdwinGuzman:
Fixes JIRA ticket [DSD-2135](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2135)

- Pr branch says "slider" but the ticket's title was initially wrong. The issue is in the `Select` component. The helper text does not align properly when `labelPosition` is "inline" and `showLabel` is false. 

Screenshot 2025-07-30 at 5 22 30 PM

## How has this been tested?

- Storybook.

- N/A

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[DSD-2135]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-2135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Select/selectChangelogData.ts
+++ b/src/components/Select/selectChangelogData.ts
@@ -21,6 +21,7 @@ export const changelogData: ChangelogData[] = [
       "Rename `selectType` to `variant`.",
       "Updates the `id` prop to be optional and generates a random id if not provided.",
       "Adds `data-testid` value of `ds-select` to the parent element.",
+      "Fixes helper error text alignment when label is not shown and it is inlined.",
     ],
   },
   {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -560,6 +560,7 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 - Removes console warnings from the `CheckboxGroup` component when a non-`Checkbox` component is passed as a child.
 - Fixes the responsive styles related to the image in the `FeaturedContent` component.
+- Fixes helper error text alignment in the `Select` component when label is not shown and it is inlined.
 
 ### Breaking Changes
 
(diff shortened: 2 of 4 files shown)


oliviawongnyc:
Interesting. I see the screenshot but I can't reproduce the issue so far.

bigfishdesign13 (commented):
I didn't dig into the changes in the theme file, but I am seeing that the bug still exists if you do the following with the controls:
- set `showLabel` to `false`
- set `labelPosition` to `"inline"`
- set `showLabel` to `true`

That said, this issue only seems to happen when changing the prop values with the controls. If the prop values are set initially, the position of the label is fine. I am not too concerned about this `"bug"`.

EdwinGuzman:
Took a closer look. I think it's a Storybook "bug". The `Select` component dynamically updates the left margin for the input and helper text based on the width of the label. It generally works, but not in storybook where I think it doesn't pick up the control value updates. I'll make a note of it when Alkim reviews it but I agree it's not something to be overly concerned about.

src/components/Select/selectChangelogData.ts
```diff
@@ -21,6 +21,7 @@ export const changelogData: ChangelogData[] = [
       "Rename `selectType` to `variant`.",
       "Updates the `id` prop to be optional and generates a random id if not provided.",
       "Adds `data-testid` value of `ds-select` to the parent element.",
+      "Fixes helper error text alignment when label is not shown and it is inlined.",
```
bigfishdesign13, line 24: Was this an bug prior to `v4`? If not, it does not need a changelog entry.

bigfishdesign13 (approved):
Works well. Just one minor question.

src/components/Select/selectChangelogData.ts
EdwinGuzman: Yes, it's in prod so it's one of the few fixes we have.

PR #1902, opened 2025-08-27
DSD-2050: Menu ARIA label fix

7emansell:
Fixes JIRA ticket [DSD-2050](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2050)

- Adds `labelAsAriaLabel` prop to the `Menu`
- Adds a story and updates docs

- Should resolve a11y issue with "labelText: selectedItem : selectedItem" ARIA label on the `Menu` button by allowing devs to override

### Accessibility Checklist

## Open Questions

- See [this](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2050?focusedCommentId=137868) for explanation of approach– trying to avoid breaking changes

### Checklist:

### Front End Review:

[DSD-2050]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-2050?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Menu/Menu.mdx
+++ b/src/components/Menu/Menu.mdx
@@ -13,7 +13,7 @@ import { changelogData } from "./menuChangelogData";
   componentName="Menu"
   summary="List of interactive elements rendered in a dropdown format and used for executing actions or navigating"
   versionAdded="3.0.0"
-  versionLatest="4.0.0"
+  versionLatest="Prerelease"
 />
 
 <Canvas of={MenuStories.WithControls} />
--- a/src/components/Menu/Menu.mdx
+++ b/src/components/Menu/Menu.mdx
@@ -77,10 +77,11 @@ type when there is limited space such as a table or on mobile.
 
 ## Menu label
 
-The default label of the menu will be `labelText`. With `showSelectionAsLabel`
+The default label of the menu will be `labelText`.
+
+With `showSelectionAsLabel`
 enabled, the selected item will then display as the menu's label. This is useful
-when the choice
-should be displayed after selection, like choosing sort/filter criteria, or if
+when the choice should be displayed after selection, like choosing sort/filter criteria, or if
 an option needs to be selected and displayed when the page first loads.
 
 `labelText` is still required even if `showSelectionAsLabel` is true, because it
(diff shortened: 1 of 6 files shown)


oliviawongnyc (commented):
If someone uses your new prop, is the `aria-label` only accessible if they are also doing the thing you described, where they are dynamically adding selected values to the label?

7emansell:
> If someone uses your new prop, is the `aria-label` only accessible if they are also doing the thing you described, where they are dynamically adding selected values to the label?

yes the label has to include the selected value @oliviawongnyc

oliviawongnyc:
> > If someone uses your new prop, is the `aria-label` only accessible if they are also doing the thing you described, where they are dynamically adding selected values to the label?
> 
> yes the label has to include the selected value @oliviawongnyc

@7emansell got it, thank you. This looks good to me!

oliviawongnyc (approved):

src/components/Menu/Menu.mdx
oliviawongnyc, line 1: Ooh just need to update the last updated version

src/components/Menu/Menu.tsx
```diff
@@ -39,6 +39,9 @@ export interface MenuProps extends BoxProps {
   /** Optional boolean value used to toggle the visibility of the label text for the button element.
    * If false, this value will be used to set an aria-label attribute on the button element.  */
   showLabel?: boolean;
+  /** Optional boolean value used to pass the labelText as the button element's aria-label.
+   * If false, the aria-label will be built from the labelText and the selected item. */
+  labelAsAriaLabel?: boolean;
```
EdwinGuzman: nit: move this above `labelText`

src/components/Menu/Menu.stories.tsx
```diff
@@ -33,6 +34,10 @@ const meta: Meta<typeof Menu> = {
       control: { type: "boolean" },
       defaultValue: { summary: "true" },
     },
+    labelAsAriaLabel: {
```
EdwinGuzman: nit: move this above `labelText` just to order the props

EdwinGuzman (commented):
This looks good. Two things:

- can you write a few tests?
- should we have a deprecation note for the other pattern in this PR or a separate pr?

7emansell:
where would the deprecation note go? @EdwinGuzman

EdwinGuzman (approved):
I don't think we need the deprecation note just yet (or at all in case that is one pattern a dev wants to use). I'd add the note in the prop's description so it shows up in the args table and then in the `Menu label` section in storybook. To be clear, I was wondering if we needed one but I don't think it's a blocker for this PR.

PR #1903, opened 2025-09-03
DSD-2145: Add padding for TextInput clearable X

7emansell:
Fixes JIRA ticket [DSD-2145](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2145)

- Adds padding to `input` when clearable button is visible, to prevent text from overlapping with the button

## How has this been tested?

- Enter a bunch of text into `TextInput` with `isClearable`

### Accessibility Checklist

### Checklist:

### Front End Review:

[DSD-2145]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-2145?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/TextInput/textInputChangelogData.ts
```diff
@@ -9,6 +9,15 @@
 import { ChangelogData } from "../../utils/ComponentChangelogTable";
 
 export const changelogData: ChangelogData[] = [
+  {
+    date: "Prerelease",
+    version: "Prerelease",
+    type: "Update",
+    affects: ["Styles"],
+    notes: [
+      "Adds padding so clearable X button doesn't ovverlap with text input.",
```
EdwinGuzman: nit: typo `ovverlap`.

EdwinGuzman (approved):
Works for me!

oliviawongnyc (commented):
Looks good on `TextInput`, but don't think it's working on `SearchBar` yet?

Image

oliviawongnyc:
> Looks good on `TextInput`, but don't think it's working on `SearchBar` yet?
> 
> Image

Nevermind! Tested again and looked good. Must have been looking at wrong tab.

oliviawongnyc (approved):

src/components/TextInput/TextInput.tsx
```diff
@@ -321,7 +321,12 @@ export const TextInput: ChakraComponent<
               {labelText}
             </Label>
           )}
-          <Box position="relative">
+          <Box
+            position="relative"
+            sx={{
+              input: { pr: clearButtonOutput ? "inset.wide" : "inset.default" },
+            }}
```
bigfishdesign13: There is an explicit attribute for this in the theme file. To follow common DS patterns, it would be preferred to handle this logic in the theme file rather than in the component. See line 23 in `textInput.ts`. It will require that the `isClearable` prop value is passed into `useMultiStyleConfig`, but we are already doing that with two other props.

bigfishdesign13: Also, I was mistaken about the design token. It should be `inset.extrawide`, which is equal to `3rem` (48px).

bigfishdesign13 (changes requested):
Two comments.

src/components/TextInput/TextInput.tsx
7emansell: `inset.extrawide` seems imbalanced to me but not sure what design we're working off. Also, `isClearable` is not the only condition being checked to display the clearable button, so either I need to pass several other values to the theme file (`isDisabled`, `isHidden`, `isTextArea`) or we're okay with applying this padding regardless of those conditions (shouldn't really be visible or cause any issues as far as I can see, but just wanted to flag).

bigfishdesign13: The `extrawide` value was used to make sure there was decent space based on the hover and focus states of the clear button.

Good point about the other conditions. `isDisabled` is the only one I would worry about. Do you foresee anything that would make it difficult to pass that value into the theme file to use in the condition statement?

src/theme/components/textInput.ts
```diff
@@ -93,11 +90,11 @@ const TextInput = defineMultiStyleConfig({
         },
       },
     },
-    input,
+    input: getInputStyles(isClearable),
     label: { marginBottom: "label.default" },
     textarea: {
-      ...input,
-      lineheight: "1.5",
```
EdwinGuzman, line 100: 😓

EdwinGuzman (approved):
Approving the updates in the latest commit.

bigfishdesign13 (approved):
Looks good and works well.

PR #1914, opened 2025-09-22
DSD-2156: Hero background color

bigfishdesign13:
Fixes JIRA ticket [DSD-2156](https://newyorkpubliclibrary.atlassian.net/browse/DSD-2156)

- Updates the `Hero` component to add the `textBackgroundColor` prop to be used in lieu of the `backgroundColor` prop.

**NOTE:** Ultimately, extracting the `backgroundColor` prop from `rest` was not feasible. Instead, a new textBackgroundColor` props was added and used where backgroundColor was used previously to set the background color behind the text. 

**QUESTION:** For backward compatibility, should `backgroundColor` still be supported? Should conditions be added to use that prop if it is passed?

## How has this been tested?

- local Storybook

- n/a

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[DSD-2156]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-2156?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Hero/Hero.mdx
+++ b/src/components/Hero/Hero.mdx
@@ -100,7 +100,7 @@ Each `Hero` variation can be rendered through the `variant` prop.
 ### Primary
 
 The `"primary"` hero variant should be used with a `backgroundImageSrc` and a
-`heading`. `backgroundColor`, `foregroundColor`, and `subHeaderText` props may
+`heading`. `subHeaderText`, `textBackgroundColor`, and `textColor` props may
 also be used.
 
 <Story of={HeroStories.Primary} />
@@ -127,7 +127,7 @@ also be used.
 ### Tertiary
 
 The `"tertiary"` hero variant should be used with a `heading` prop.
-`backgroundColor`, `foregroundColor`, and `subHeaderText` props may also be used.
+`subHeaderText`, `textBackgroundColor`, and `textColor` props may also be used.
 
 <Story of={HeroStories.Tertiary} />
 
@@ -201,16 +201,16 @@ The `"tertiary"` hero variant should be used with a `heading` prop.
 
 ### Campaign
 
-The `"campaign"` hero variant should be used with `backgroundImageSrc`, `heading`,
-and `imageProps` props. `backdropBackgroundColor`, `backgroundColor`, `foregroundColor`,
-and `subHeaderText` props may also be used.
+The `"campaign"` hero variant should be used with `backgroundImageSrc`,
+`heading`, and `imageProps` props. `backdropBackgroundColor`, `subHeaderText`,
+`textBackgroundColor`, and `textColor` props may also be used.
 
 The `"campaign"` variant is special in that it has both a background and foreground.
 We understand that this naming can be confusing, but it helps to think of this
 variant having two layers.
 
-To modify the foreground, the `backgroundColor`, `foregroundColor`, `imageProps`
-props should be used.
+To modify the foreground, the `textBackgroundColor`, `textColor`, and
+`imageProps` props should be used.
 
 <Story of={HeroStories.Campaign} />
 
@@ -387,7 +387,6 @@ content.
   code={`
   <Hero
     backgroundImageSrc="https://iiif.nypl.org/iiif/2/5164274/full/!900,900/0/default.jpg"
-    foregroundColor="ui.white"
     heading={
       <Heading
         id="primary-hero"
@@ -397,6 +396,7 @@ content.
         text="Hero Primary"
       />
     }
+    textColor="ui.white"
     textBackgroundColor="brand.primary"
     variant="primary"
   />
@@ -407,7 +407,6 @@ content.
 <Source
   code={`
   <Hero
-    foregroundColor="ui.white"
     heading={
       <Heading
         level="h1"
@@ -418,6 +417,7 @@ content.
       />
     }
     subHeaderText={otherSubHeaderText}
+    textColor="ui.white"
     textBackgroundColor="brand.primary"
     variant="tertiary"
   />
@@ -429,7 +429,6 @@ content.
   code={`
   <Hero
     backgroundImageSrc={getPlaceholderImage()}
-    foregroundColor="ui.white"
     heading={
       <Heading
         level="h1"
@@ -443,6 +442,7 @@ content.
       src: "https://iiif.nypl.org/iiif/2/5164274/full/!900,900/0/default.jpg",
     }}
     subHeaderText={otherSubHeaderText}
+    textColor="ui.white"
     textBackgroundColor="brand.primary"
     variant="campaign"
   />
@@ -454,18 +454,18 @@ content.
 
 By default, the `Hero` component renders a dark background color with light
 text. This color treatment can be overridden using a combination of the
-`foregroundColor`, `textBackgroundColor`, and `isDarkText` props.
+`textBackgroundColor`, `textColor`, and `isDarkText` props.
 
-- `foregroundColor`: custom color value for the text
 - `textBackgroundColor`: custom color value for the text background
+- `textColor`: custom color value for the text
 - `isDarkText`: used to render default dark color value for the text
 
 <Banner
   content={
     <>
-      <strong>IMPORTANT:</strong> The <code>foregroundColor</code> prop will
-      override the <code>isDarkText</code> prop, so they should not be used at
-      the same time.
+      <strong>IMPORTANT:</strong> The <code>textColor</code> prop will override
+      the <code>isDarkText</code> prop, so they should not be used at the same
+      time.
     </>
   }
   mt="s"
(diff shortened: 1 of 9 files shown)


src/components/Hero/Hero.tsx
```diff
@@ -56,6 +56,9 @@ export interface HeroProps extends BoxProps {
   /** Optional string used for the subheader that displays
    * underneath the heading element. */
   subHeaderText?: string | JSX.Element;
+  /** Optional hex color value used to override the default background color for
```
EdwinGuzman: This says hex color value, but the storybook examples use ds token values.

src/components/Hero/Hero.tsx
```diff
@@ -243,11 +247,17 @@ export const Hero: ChakraComponent<
...
+      /** The "tertiary" variant doesn't have an apparent content box, so that
+       * element will default to a transparent background in the "teriary"
+       * variant. */
       contentBoxStyling = {
         ...(foregroundColor && { color: foregroundColor }),
-        ...(backgroundColor
-          ? { backgroundColor }
-          : { bgColor: defaultBackgroundColor }),
+        ...(textBackgroundColor
+          ? { backgroundColor: textBackgroundColor }
```
EdwinGuzman: What's the difference between `backgroundColor` and `bgColor`? I always think they're the same.

src/components/Hero/Hero.mdx
```diff
@@ -370,59 +370,114 @@ the `backgroundImageSrc` will take precedence.
...
+## Text color
 
 By default, the `Hero` component renders a dark background color with light
 text. This color treatment can be overridden using a combination of the
-`backgroundColor`, `foregroundColor` and `isDarkText` props.
+`foregroundColor`, `textBackgroundColor`, and `isDarkText` props.
 
-**NOTE:** The `foregroundColor` prop will override the `isDarkText` prop, so they
-should not be used at the same time.
+The examples below only show the `"campaign"` variant, but this same treatment
```
EdwinGuzman: Should these examples render images? None of them currently do in the vercel preview.

src/components/Hero/Hero.mdx
```diff
@@ -370,59 +370,114 @@ the `backgroundImageSrc` will take precedence.
...
+applies to all variants.
+
+<Banner
+  content={
+    <>
+      <strong>IMPORTANT:</strong> The <code>foregroundColor</code> prop will
```
EdwinGuzman: This seems overly complicated to implement. I have no good solution though.

src/components/Hero/Hero.stories.tsx
```diff
@@ -296,52 +295,78 @@ export const CampaignDarkBackgroundImage: Story = {
...
+      }
+      imageProps={imageProps}
+      subHeaderText={otherSubHeaderTextLong}
+    />
+  ),
+};
+export const TextBackgroundColor: Story = {
+  render: () => (
+    <Stack spacing="l">
+      <div>
```
EdwinGuzman, line 321: Can the divs be removed? I don't think they're needed unless they're not stacking correctly.

src/components/Hero/Hero.tsx
```diff
@@ -56,6 +56,9 @@ export interface HeroProps extends BoxProps {
   /** Optional string used for the subheader that displays
    * underneath the heading element. */
   subHeaderText?: string | JSX.Element;
+  /** Optional hex color value used to override the default background color for
+   * a text content area in a given `Hero` variant. */
+  textBackgroundColor?: string;
```
EdwinGuzman, line 67: This seems very similar to `backdropBackgroundColor`, at least conceptually. Or maybe it's just me being confused?

src/components/Hero/Hero.mdx
```diff
@@ -31,8 +31,9 @@ import { changelogData } from "./heroChangelogData";
 - [Overview](#overview)
 - [Component props](#component-props)
 - [Accessibility](#accessibility)
-- [Active variants](#active-variants)
-- [Text colors](#text-colors)
+- [Variants](#variants)
+- [Text background color](#text-background-color)
+- [Text color](#text-color)
```
oliviawongnyc, line 36: I'm sure there is a reason for this, but I really want this prop to be `textColor` rather than `foregroundColor`, especially now that we are changing the name of `backgroundColor` to `textBackgroundColor`.

src/components/Hero/Hero.mdx
oliviawongnyc: What seems overly complicated? I think this is actually one of the easier-to-understand parts of the Hero 😅.

src/components/Hero/Hero.mdx
oliviawongnyc: They seem to be working now.

src/components/Hero/Hero.tsx
oliviawongnyc: One is behind the text and one is behind the background image right?

oliviawongnyc: Should we add the note "Note: not all `Hero` variants utilize this prop."

oliviawongnyc (commented):
I think we should change all instances of `backgroundColor` to `textBackgroundColor`. From what it looks like we still have some variants using `backgroundColor` even though it does the same thing as `textBackgroundColor`. Am I right about this? With that being said, I do believe it should be backwards compatible as much as possible.

src/components/Hero/Hero.tsx
bigfishdesign13: Agreed. We don't need to be that specific. That was the language used for the other color prop, so I just copied it. I updated to just say `color value`. To me, that covers hex values as well as design tokens.

src/components/Hero/Hero.tsx
bigfishdesign13: You are correct. In Chakra land, they are the same. Originally, we were just passing along the `backgroundColor` prop, but now we need the value of the `textBackgroundColor` prop. I just kept the long-form prop name from the original code, but I can update to use `bgColor` for consistency.

src/components/Hero/Hero.mdx
bigfishdesign13: Yes, there should be images in those example. They seem fine to me.

src/components/Hero/Hero.mdx
bigfishdesign13: I added some additional text to help clarify.

src/components/Hero/Hero.stories.tsx
bigfishdesign13: I'd prefer to keep the divs to help with spacing. The space between the headings and the heroes is 16px, but the spacing between each example is 32px.

src/components/Hero/Hero.tsx
bigfishdesign13: Olivia has it right. `textBackgroundColor` sets the background color for the text area. `backdropBackgroundColor` set the background color for the element behind the image and text element in the `campaign` variant.

`textBackgroundColor` can be used by all variants.

src/components/Hero/Hero.mdx
bigfishdesign13: Yes, `textColor` would make a lot more sense. This was the prop name used in the original component from day one and we never changed it. I can add a new `textColor` prop and update the conditions to look for `textColor` or foregroundColor`. 

@EdwinGuzman What do you think?

bigfishdesign13:
> I think we should change all instances of `backgroundColor` to `textBackgroundColor`. From what it looks like we still have some variants using `backgroundColor` even though it does the same thing as `textBackgroundColor`. Am I right about this? With that being said, I do believe it should be backwards compatible as much as possible.

@oliviawongnyc I was just looking at the places where `backgroundColor` is still used and considering whether those could or should be replaced with `textBackgroundColor`. I think it is just the `tertiary` variant where it is used and `backgroundColor` is actually setting the background color for the whole component.

I will take another look and see if it makes sense to switch.

src/components/Hero/Hero.mdx
EdwinGuzman: That makes sense. Let's go with `textColor` but keep `foregroundColor` for backwards compatibility. Perhaps deprecate it? DC needs this update soon, so we can think long term about it later on.

bigfishdesign13: Ok, I will make that change today. I will add `textColor`, but keep `foregroundColor` for backward compatibility. And yes, I will add information noting that `foregroundColor` is being deprecated.

EdwinGuzman (approved):
I think the three variants are quite distinct and we should consider breaking this component apart (but for later on). All the logic for rendering the proper variant is getting confusing.

bigfishdesign13:
> I think the three variants are quite distinct and we should consider breaking this component apart (but for later on). All the logic for rendering the proper variant is getting confusing.

@EdwinGuzman I totally agree. It's been confusing for some time now. Do you feel it would be worth it to do as part of v4? Or wait for the next major version?

PR #1918, opened 2025-10-06
ISW-5048: Add `as` prop to `SubNavLink`

7emansell:
Fixes JIRA ticket [ISW-5048](https://newyorkpubliclibrary.atlassian.net/browse/ISW-5048)

- Adds the `as` prop to the `SubNavLink`
- Doesn't add to docs since I think this just allows what would already be expected given the `Link` behavior

## How has this been tested?

- Hard to really test until it's in another app but follows the pattern of `Link`

### Accessibility Checklist

### Checklist:

### Front End Review:

[ISW-5048]: https://newyorkpubliclibrary.atlassian.net/browse/ISW-5048?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/SubNav/SubNav.tsx
+++ b/src/components/SubNav/SubNav.tsx
@@ -7,7 +7,6 @@ import {
   Flex,
   BoxProps,
 } from "@chakra-ui/react";
-import type { As } from "@chakra-ui/react";
 import Button from "../Button/Button";
 import Link from "../Link/Link";
 import List from "../List/List";
@@ -49,9 +48,8 @@ interface SubNavItemProps {
   screenreaderOnlyText?: string;
 }
 
-interface SubNavLinkProps extends SubNavItemProps {
+interface SubNavLinkProps extends SubNavItemProps, Pick<BoxProps, "as"> {
   href: string;
-  as?: As;
 }
 
 interface SubNavButtonProps extends SubNavItemProps {
--- a/src/components/SubNav/SubNav.tsx
+++ b/src/components/SubNav/SubNav.tsx
@@ -6,8 +6,8 @@ import {
   useMultiStyleConfig,
   Flex,
   BoxProps,
-  As,
 } from "@chakra-ui/react";
+import type { As } from "@chakra-ui/react";
 import Button from "../Button/Button";
 import Link from "../Link/Link";
 import List from "../List/List";
--- a/src/components/SubNav/SubNav.tsx
+++ b/src/components/SubNav/SubNav.tsx
@@ -6,6 +6,7 @@ import {
   useMultiStyleConfig,
   Flex,
   BoxProps,
+  As,
 } from "@chakra-ui/react";
 import Button from "../Button/Button";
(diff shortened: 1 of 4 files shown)


src/components/SubNav/SubNav.tsx
```diff
@@ -7,6 +7,7 @@ import {
   Flex,
   BoxProps,
 } from "@chakra-ui/react";
+import type { As } from "@chakra-ui/react";
```
7emansell: same type that `Link` is getting from `BoxProps`/`HTMLChakraProps` but maybe there's a better way to define this ?

EdwinGuzman: What do you mean by this?

EdwinGuzman (approved):

src/components/SubNav/SubNav.tsx
7emansell: I guess i thought it would follow the `Link` more closely to `Pick` it off the `BoxProps` instead of just directly import but i don't know why that would be better or worse 🤷‍♀️

EdwinGuzman: Ah. Good point. Can you update this to use the same pattern of picking from `BoxProps` as `Link`?

EdwinGuzman (changes requested):
I approved too soon and the question you raised does mean a slight modification.

EdwinGuzman (approved):

src/components/SubNav/subNavChangelogData.ts
```diff
@@ -9,6 +9,13 @@
 import { ChangelogData } from "../../utils/ComponentChangelogTable";
 
 export const changelogData: ChangelogData[] = [
+  {
+    date: "Prerelease",
+    version: "Prerelease",
+    type: "Update",
+    affects: ["Functionality"],
+    notes: ["Adds `as` prop to `SubNavLink`."],
```
bigfishdesign13, line 17: Do you feel that using the `as` prop with `SubNavLink` is a pattern that would be applicable to any NYPL app that implements the `SubNav` component? If so, I think an example in the docs would be useful.

oliviawongnyc (approved):

src/components/SubNav/subNavChangelogData.ts
7emansell: I think only ones using a routing library– I'm not sure there's anything to add here besides what's already in the `Link` docs: https://nypl.github.io/nypl-design-system/reservoir/v4/?path=/docs/components-navigation-link--docs#link-with-routers. Should I add a note directing to that?

bigfishdesign13: Yes, I think that would be helpful. Thanks.

bigfishdesign13 (approved):
💯

PR #1927, opened 2025-11-05
ISW-4353: Fix for map bug on /locations

isastettler:
Fixes [JIRA ticket](https://newyorkpubliclibrary.atlassian.net/browse/ISW-4353),  [JIRA ticket](https://newyorkpubliclibrary.atlassian.net/browse/ISW-4590) and  [JIRA ticket](https://newyorkpubliclibrary.atlassian.net/browse/ISW-4620)

- removes `iframe` and `[tabindex]` selectors from the globally set `:focus` style in theme/foundations/global

## Context

This fixes an issue on nypl.org/locations where after the first panning of the map the map is no longer interactive. The issue is the `zIndex` set in the `onFocus` function. 

### Front End Review:

I am not sure if this needs a Front End review on the Design System side. For a review in Scout:
- Visit the currently live page [Locations Finder](https://www.nypl.org/locations)
  - To see the issue, try to move the map twice in a row
- For testing the updated Design System:
  - Replace `"@nypl/design-system-react-components": "4.0.0",` with `"@nypl/design-system-react-components": "https://github.com/NYPL/nypl-design-system.git#ISW-4353/map-bug",` on your local scout instance
  - Delete node modules folder, .next folder and package-lock.json
  - run `npm install`
  - run `npm run build && npm start`
  - visit [Local Location Finder](http://localhost:3000/locations)
  - try moving the map twice


EdwinGuzman:
Add an entry into the changelog, please!

src/theme/foundations/global.ts
```diff
@@ -38,7 +38,7 @@ const global = {
       _focus: activeFocus(),
     },
   },
-  "iframe, [href], [tabindex], [contentEditable='true']": {
```
oliviawongnyc, line 41: This is probably a question for @EdwinGuzman and @bigfishdesign13, but how do we know what the impact of this change is outside of the bug that Isa's resolving?

oliviawongnyc: And, out of curiosity, is this issue resolvable on the Scout side rather than in DS? Not saying it shouldn't be changed in the DS, but if we are unclear on the overall impact, is it an option?

EdwinGuzman: I am slightly wary of this. It would technically be a breaking change, but I see it as a good breaking of unneeded styles. I don't know who currently uses iframes, and this problem hasn't come up before, so I think the risk is low. We can ask in our slack channel if that'll help?

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 ### Updates
 
+- Removes `iframe` and `[tabIndex]` selector from the gobal set focus styles
```
EdwinGuzman: typo: `gobal`

src/theme/foundations/global.ts
oliviawongnyc: Cool, if low risk, this works for me. I think people may know if their app uses an `iframe`, but not necessarily all the elements with `tabindex` so not sure if asking would be helpful.

EdwinGuzman (approved):

EdwinGuzman:
@oliviawongnyc and I were testing and talking through this update. It _should_ be ok but if we find a regression, we'll have to revert. I think if we revert, we'll add back all the styles for `tabIndex` specifically but remove the zindex value.

src/theme/foundations/global.ts
isastettler: I did some more reading about this issue and I think we should go ahead and remove the iframe and the tabindex selector. 
Two reasons:
1. I tried one more time to apply a fix on Scout side but was not successful even now that I understand the issue and was more explicit and targeted in my approaches. 
2. As I understand from my quick afternoon research , using attribute selectors is not best practice when adding z-index properties globally through a design system. 
Given both of those findings plus the fact that we have 3 open tickets on ISW that flag this issue, I would move ahead with this approach. 
I like the idea of making an announcement on slack to make sure engineers that use Reservoir are aware. 
How does this sound?

oliviawongnyc (approved):

src/theme/foundations/global.ts
bigfishdesign13: With the removal of the `[tabindex]` selector, what will happen when focus is placed on a text element? For example, after using pagination on a list of search results, we often scroll to the top of the list and move focus to either a text element that describes the updated results list or focus is placed on the first item in the list.

EdwinGuzman: @oliviawongnyc and I were chatting a bit about this the other day. I do agree that using an attribute selector is not best practice. If we really need to, we should instead consider creating a new rule block just for iframe and tabIndex that uses a similar set of focus styles, but doesn't include the z-index rule that is giving us trouble.

I also tested DS components and programmatically focusing on other elements still works. This is true for `NewsletterSignup` and `FeedbackBox`, although `FeedbackBox` was not always consistent.

> what will happen when focus is placed on a text element? 

This should continue to work, but let's further test in turbine.

EdwinGuzman:
@bigfishdesign13 can we add this as part of the 4.1.0 RC to test?

bigfishdesign13:
> @bigfishdesign13 can we add this as part of the 4.1.0 RC to test?

@EdwinGuzman Do you want to keep it as a separate branch until testing is completed? Or can we go ahead and merge it into `development` and create the new `rc` from there?

Cc @isastettler

isastettler:
Hey all (@EdwinGuzman, @oliviawongnyc , @bigfishdesign13 ), I was thinking we could search in the repo for components that use tabindex and add a class to them that could then be the selector for the focus styles instead of the tabindex atttibute. That way the current setup would stay in place for ds components but would be scoped better for consuming apps?

oliviawongnyc:
> Hey all (@EdwinGuzman, @oliviawongnyc , @bigfishdesign13 ), I was thinking we could search in the repo for components that use tabindex and add a class to them that could then be the selector for the focus styles instead of the tabindex atttibute. That way the current setup would stay in place for ds components but would be scoped better for consuming apps?

@isastettler, this sounds like a good idea to me!

EdwinGuzman:
Yes, good idea, but we're not using classes or targeting classes in apps from the DS. We could try it with this single update, but it's an outlier from our current approach.

Perhaps you can make a new `activeFocus` function that contains the same rules _except_ the zindex rule?

PR #1932, opened 2025-11-17
ISW-5151: styles for full-click Card hover state

bigfishdesign13:
Fixes JIRA ticket [ISW-5151](https://newyorkpubliclibrary.atlassian.net/browse/ISW-5151)

- Adds hover state styles for "full-click" functionality in the `Card` component.
- Adds `bg` color palette options that are defined using alpha transparency (rgba).

## How has this been tested?

- local Storybook

- n/a

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[ISW-5151]: https://newyorkpubliclibrary.atlassian.net/browse/ISW-5151?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/theme/components/card.ts
+++ b/src/theme/components/card.ts
@@ -129,10 +129,55 @@ const ReservoirCard = defineMultiStyleConfig({
           bodyMargin = "0";
         }
       }
+      // Hover state styles to be used with full-click functionality (mainActionLink)
+      const hoverState = () => {
+        if (mainActionLink) {
+          return {
+            outline: "0 solid",
+            outlineColor: "ui.bg.defaultAsAlpha",
+            transition: [
+              "background-color ease-out",
+              "outline ease-out",
+              "transform ease-out",
+            ].join(", "),
+            transitionDuration: "normal",
+            ".chakra-heading a": {
+              textDecorationColor: "transparent !important",
+              transition: "text-decoration-color ease-out",
+              transitionDuration: "normal",
+            },
+            img: {
+              transition: "opacity ease-out, filter ease-out",
+              transitionDuration: "normal",
+            },
+            _hover: {
+              backgroundColor: "ui.bg.defaultAsAlpha",
+              outline: "8px solid",
+              outlineColor: "ui.bg.defaultAsAlpha",
+              ".chakra-heading a": {
+                textDecorationColor:
+                  "var(--nypl-color-ui-link-secondary) !important",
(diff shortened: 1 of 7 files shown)


src/theme/components/card.ts
```diff
@@ -129,10 +129,55 @@ const ReservoirCard = defineMultiStyleConfig({
...
+      const hoverState = () => {
+        if (mainActionLink) {
+          return {
+            outline: "0 solid",
+            outlineColor: "ui.bg.defaultAsAlpha",
+            transition: [
+              "background-color ease-out",
+              "outline ease-out",
+              "transform ease-out",
+            ].join(", "),
```
EdwinGuzman: Can you make this a string rather than a joined array? Unless if there was a specific reason for doing this? It's not done below online 146.

src/theme/components/card.ts
```diff
@@ -129,10 +129,55 @@ const ReservoirCard = defineMultiStyleConfig({
           bodyMargin = "0";
         }
       }
+      // Hover state styles to be used with full-click functionality (mainActionLink)
```
EdwinGuzman: Nit: I don't think `(mainActionLink)` is helpful here, especially because it's in the code below that it should be true for the styles to be on.

It's minor but something that we consider when we write code is whether we need verbose comments when the code is enough.

EdwinGuzman (approved):

src/theme/components/card.ts
bigfishdesign13: Seeing the attributes stack was easier to read than a long sting. But yes, I can convert back to a basic string.

bigfishdesign13:
@EdwinGuzman I just updated some of the language for the `Cards as Navigation Links` docs. Forgot to do that earlier. Letting you know in case you want to take one last look before I merge.

src/theme/components/card.ts
EdwinGuzman: Do you still need the commas for each rule that is being transformed?

EdwinGuzman (approved):

src/theme/components/card.ts
bigfishdesign13: Oops. Yes.

PR #1949, opened 2026-02-03
ISW-5336: TagSet semantic variants

bigfishdesign13:
Fixes JIRA ticket [ISW-5336](https://newyorkpubliclibrary.atlassian.net/browse/ISW-5336)

- Updates the `TagSet` component to add semantic color variants.
- Fixes the hex color value for `dark.ui.warning.secondary`.

## How has this been tested?

- local Storybook

- The updated color styles all meet WCAG 2.1 contrast requirements.

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[ISW-5336]: https://newyorkpubliclibrary.atlassian.net/browse/ISW-5336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/TagSet/TagSet.mdx
```diff
@@ -213,6 +216,26 @@ Resources:
...
+
+Semantic variants align with specific uses.
+
+<Unstyled>
+  <Banner
+    content={
+      <>
+        <strong>RECOMMENDATION:</strong> Avoid using the "informative" variant
+        for "filter" tags as the colors are similar to the default colors of
+        "explore" tags and imply that tags are clickable.
```
oliviawongnyc: But they are clickable if dismissable, so is the recommendation only if not dismissable?

src/theme/components/tagSet.ts
```diff
@@ -12,62 +13,651 @@ const { defineMultiStyleConfig, definePartsStyle } =
...
-  })),
+
+      // Set the bg, borderColor, colo, & fill attributes (light and dark)
+      ...getSemanticColorStyles({ semanticVariant }),
+    };
+  }),
 });
 
-const TagSetExplore = defineStyleConfig({
+const _TagSetExplore = defineStyleConfig({
```
oliviawongnyc: Is this set of styles needed?

src/components/TagSet/TagSet.mdx
```diff
@@ -213,6 +216,26 @@ Resources:
...
 
+## Semantic variants
+
+Semantic variants align with specific uses.
+
+<Unstyled>
+  <Banner
+    content={
+      <>
+        <strong>RECOMMENDATION:</strong> Avoid using the "informative" variant
```
oliviawongnyc, line 227: But filter tags are clickable if `isDismissable` is true, so does this recommendation still stand in that scenario?

oliviawongnyc (commented):
Looks great!

Somewhat unrelated to your PR, but I noticed that in the scenario where there is a TagSet of variant `filter` with an `onClick` prop, but no `isDismissable` prop, there are no hover styles. Is this intentional? If not, perhaps we switch the checks in the theme file from whether `isDismissable` is true to whether `onClick` has a value? It seems like if `isDismissable` is true, there should always be an `onClick`, but not necessarily the other way around. Not positive that is right either. The documentation is a bit confusing:

```
The isDismissible and onClick props are optional for this variant. However, when isDismissible is true, the onClick prop is required. The onClick prop will only fire when isDismissible is true and the user clicks a tag. The function will return the entire tag object that was clicked to remove.
```

src/components/TagSet/TagSet.mdx
bigfishdesign13: I can write this better. The issue is that the `informative` colors mirror the link colors.

src/theme/components/tagSet.ts
bigfishdesign13: Nope. That was an experiment that should have been removed.

src/components/TagSet/TagSet.mdx
EdwinGuzman: I would think it still stands. A dismissiable filter tag that has the "informative" variant will look even more like an interactive element. This rec could clarify to also not use `isDismissable` as well. Would that help?

src/components/TagSet/TagSet.stories.tsx
```diff
@@ -152,119 +271,255 @@ export const FormattingExamples: Story = {
...
+            { id: "yellow", label: "Yellow" },
+            { id: "green", label: "Green" },
+            { id: "blue", label: "Blue" },
+            { id: "indigo", label: "Indigo" },
+            { id: "violet", label: "Violet" },
+          ]}
+          variant="filter"
+        />
+        <TagSet
+          isDismissible
```
EdwinGuzman, line 359: The `TagSet` below also includes the "x" close button, but not the "Clear filters" link. It doesn't include `isDismissible`.

Okay, I just checked the component and `isDismissible` is mostly used for the "Clear filters" button. I just forgot how it worked, but slightly confusing looking at this and both have the "x"

src/components/TagSet/TagSetFilter.tsx
```diff
@@ -9,6 +9,9 @@ import Button from "../Button/Button";
 import Icon from "../Icons/Icon";
 import { IconNames } from "../Icons/Icon";
 import { TooltipWrapper } from "./TooltipWrapper";
+import { messageVariantsArray } from "../../theme/sharedTypes";
+
+export type TagSemanticVariants = typeof messageVariantsArray[number];
```
EdwinGuzman, line 14: Can you set this in `sharedTypes.ts` so it's already exported as a variable? That way, it's not declared in both `TagSetExplore.tsx` and `TagSetFilter.tsx`.

src/theme/components/tagSet.ts
```diff
@@ -12,134 +13,647 @@ const { defineMultiStyleConfig, definePartsStyle } =
...
+  ),
+});
+
+const TagSetExplore = defineStyleConfig({
+  baseStyle: defineStyle(({ semanticVariant }: TagSetExploreProps) => {
+    return {
+      display: "flex",
+      alignItems: "center",
+      justifyContent: "center",
+      // bg: "ui.link.primary-05",
```
EdwinGuzman: It's ok to remove these if they're not needed.

src/theme/components/tagSet.ts
```diff
@@ -12,134 +13,647 @@ const { defineMultiStyleConfig, definePartsStyle } =
...
-    base: {
-      alignItems: "center",
+const getSemanticColorStyles = ({
+  isDismissible = false,
+  semanticVariant,
+  variant = "explore",
+}) => {
+  const semanticColorStyles = {
+    default: {
+      bg: variant === "explore" ? "ui.link.primary-05" : "ui.bg.default",
```
EdwinGuzman: There are many `variant === "explore"` checks here. It's not wrong but I think it could be refactored for readability. Perhaps something like
```
const exploreVariant = {
  bg: "ui.link.primary-05",
  borderColor: "ui.link.primary",
 ...
};
const filterVariant = {
  bg: "ui.bg.default",
  borderColor: "ui.border.default",
  ...
};
```

Then
```
const semanticColorStyles = {
  default: variant === "explore" ? exploreVariant : filterVariant,
  ...
}
```

There might be a better way to do this but it's just an idea to reduce all the inlined checks.

EdwinGuzman (commented):
I did not check every style line in `tagSet.ts` but the update looks good. Left some comments.

src/components/TagSet/TagSet.mdx
bigfishdesign13: `isDismissible` is only applicable to the `filter` variant, so simply saying to avoid using the `"informative"` variant for any and all `filter` tags covers the use of `isDismissible`.

src/components/TagSet/TagSet.stories.tsx
bigfishdesign13: The visibility of the `Clear filters` button is connected to the `onClick` prop. That must not have been the case when this story was originally configured. I removed `onClick` and it looks good now.

src/components/TagSet/TagSetFilter.tsx
bigfishdesign13: Currently, `sharedTypes.ts` only contains and exports arrays. What are you suggesting be moved into `sharedTypes.ts`?

src/theme/components/tagSet.ts
bigfishdesign13: I refactored using your suggestions. Take a look and let me know how it feels.

src/components/TagSet/TagSetFilter.tsx
EdwinGuzman: I was suggesting `export type TagSemanticVariants = typeof messageVariantsArray[number];` be moved to the `sharedTypes` file since both this file and `TagSetExplore` import the array and create this new variable. If we're using the file just for arrays, then this is ok. This is not a big deal.

src/theme/components/tagSet.ts
```diff
@@ -12,134 +13,1331 @@ const { defineMultiStyleConfig, definePartsStyle } =
...
+          svg: {
+            fill: "dark.ui.warning.secondary",
+          },
+        },
+      },
+    },
+  };
+  const semanticColorStyles = {
+    default:
+      variant === "explore"
```
EdwinGuzman: Minor: you could have an `isExplore` variable for `variant === "explore"` but not a blocker.

src/theme/components/tagSet.ts
EdwinGuzman: It might be slightly longer but I think more readable/maintainable.

src/theme/components/tagSet.ts
```diff
@@ -12,134 +13,1331 @@ const { defineMultiStyleConfig, definePartsStyle } =
...
+    recommendation:
+      variant === "explore"
+        ? exploreColorStyles.recommendation
+        : filterColorStyles.recommendation,
+    warning:
+      variant === "explore"
+        ? exploreColorStyles.warning
+        : filterColorStyles.warning,
+  };
+  const _semanticColorStyles = {
```
EdwinGuzman: This is no longer needed.

src/theme/components/tagSet.ts
```diff
@@ -12,134 +13,1331 @@ const { defineMultiStyleConfig, definePartsStyle } =
...
+        paddingRight: isDismissible ? { base: "8px", md: "6px" } : "s",
+        ...transitionStyles,
+        "> span": {
+          display: "inline-block",
+          overflow: "hidden",
+          textOverflow: "ellipsis",
+          whiteSpace: "nowrap",
+        },
+
+        // Set the bg, borderColor, colo, & fill attributes (light and dark)
```
EdwinGuzman: minor typo: "colo".

EdwinGuzman (commented):
Looks good! I'll let @oliviawongnyc make the final approval.

src/components/TagSet/TagSet.mdx
oliviawongnyc: I like the updated text. Thank you.

oliviawongnyc (approved):

PR #1963, opened 2026-03-24
ISW-5522: Always render the helper text DOM in the ComponentWrapper

EdwinGuzman:
Fixes JIRA ticket [ISW-5522](https://newyorkpubliclibrary.atlassian.net/browse/ISW-5522)

- There is an error when an input element does not render helper text but it renders invalid text. The aria-live does not trigger because the div container is not in the DOM. It could be prevented if there is helper text along with invalid text, but it's also the case that an input will only have invalid text.
- ~Note: this adds 4px of extra margin so it will affect form layouts @bigfishdesign13~
- ~If this approach looks good, I'll update the rest of the changelogs.~

## How has this been tested?

- Storybook.

- Fixes an accessibility issue raised by Clare in the Events Registration form.

### Accessibility Checklist

### Checklist:

### Front End Review:

[ISW-5522]: https://newyorkpubliclibrary.atlassian.net/browse/ISW-5522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/ComponentWrapper/ComponentWrapper.tsx
+++ b/src/components/ComponentWrapper/ComponentWrapper.tsx
@@ -69,15 +69,13 @@ export const ComponentWrapper: React.FC<React.PropsWithChildren<any>> = chakra(
             <Text mb="paragraph.default">{descriptionText}</Text>
           )}
           {children}
-          {footnote && (
-            <HelperErrorText
-              id={helperErrorTextId}
-              isInvalid={isInvalid}
-              isRenderedText={showHelperInvalidText}
-              text={footnote}
-              __css={{ ...styles.helperErrorText, ...helperTextStyles }}
-            />
-          )}
+          <HelperErrorText
+            id={helperErrorTextId}
+            isInvalid={isInvalid}
+            isRenderedText={showHelperInvalidText}
+            text={footnote}
+            __css={{ ...styles.helperErrorText, ...helperTextStyles }}
+          />
         </Box>
       );
     }
--- a/src/components/ComponentWrapper/ComponentWrapper.tsx
+++ b/src/components/ComponentWrapper/ComponentWrapper.tsx
@@ -45,9 +45,8 @@ export const ComponentWrapper: React.FC<React.PropsWithChildren<any>> = chakra(
       // it is used as a wrapper component.
       const mainId = id ? `${id}-componentWrapper` : undefined;
       const helperErrorTextId = id ? `${id}-helperErrorText` : undefined;
-      const hasChildren = !!children;
-      const styles = useMultiStyleConfig("ComponentWrapper", { hasChildren });
       const footnote = isInvalid ? invalidText : helperText;
+      const styles = useMultiStyleConfig("ComponentWrapper", { hasHelperText: !!footnote });
       const finalHeadingText = useDSHeading({
         id,
         title: headingText,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Currently, this repo is in Prerelease. When it is released, this project will ad
 
 ## Prerelease
 
+### Updates
+
+- Updates the `ComponentWrapper` component so it always renders the `HelperErrorText` component. This resolves aria-live announcement issues when there is invalid text but no helper text.
+
 ## 4.1.5 (March 5, 2026)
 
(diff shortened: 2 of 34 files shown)


src/components/ComponentWrapper/ComponentWrapper.tsx
```diff
@@ -69,15 +69,13 @@ export const ComponentWrapper: React.FC<React.PropsWithChildren<any>> = chakra(
             <Text mb="paragraph.default">{descriptionText}</Text>
           )}
           {children}
-          {footnote && (
```
EdwinGuzman, line 72: This is the real fix.

oliviawongnyc (commented):
2 questions:

- Can you provide an example of how form layouts are affected?
- Why is the only component changelog update to the video player?

EdwinGuzman:
> Can you provide an example of how form layouts are affected?

Yeah, I can add another example in storybook without the helper text if that's helpful.

> Why is the only component changelog update to the video player?

I wrote in the description that if the approach works, then I'll update the rest of the changelogs. But I also wanted to use it as an exercise for the AI training.

oliviawongnyc:
> > Can you provide an example of how form layouts are affected?
> 
> Yeah, I can add another example in storybook without the helper text if that's helpful.
> 
> > Why is the only component changelog update to the video player?
> 
> I wrote in the description that if the approach works, then I'll update the rest of the changelogs. But I also wanted to use it as an exercise for the AI training.

Oops sorry about the changelog comment @EdwinGuzman

EdwinGuzman:
@oliviawongnyc no problem! All good.

EdwinGuzman:
@oliviawongnyc @bigfishdesign13 added an example of the input elements without helper text but with invalid text on https://nypl-design-system-git-isw-5522-aria-live-helpertex-c14674-nypl.vercel.app/?path=/docs/components-form-elements-form--docs#example-without-helper-text

There does seem to be more space between rows.

src/components/ComponentWrapper/ComponentWrapper.tsx
Toxiapo: This matches my understanding, thanks!

Toxiapo:
> Note: this adds 4px of extra margin so it will affect form layouts

@EdwinGuzman @bigfishdesign13 Do we need to apply any styles to the container when there’s no text? Would it make more sense to only apply styles when text is present. e.g. `__css: text ? styles : {}`

src/components/ComponentWrapper/ComponentWrapper.tsx
```diff
@@ -69,15 +69,13 @@ export const ComponentWrapper: React.FC<React.PropsWithChildren<any>> = chakra(
...
-            <HelperErrorText
-              id={helperErrorTextId}
-              isInvalid={isInvalid}
-              isRenderedText={showHelperInvalidText}
-              text={footnote}
-              __css={{ ...styles.helperErrorText, ...helperTextStyles }}
-            />
-          )}
+          <HelperErrorText
```
bigfishdesign13, line 71: Regarding the spacing issue...

The `ComponentWrapper` component has a conditional style that may have been added to address this issue. However, if that was the intention, that condition is not evaluating the correct value. This is the condition:
```
marginTop: hasChildren ? "helper.default" : 0,
```
In this statement, `hasChildren` is referring to the `ComponentWrapper` component, so the margin will always be applied. For the condition to execute as intended, the child evaluation needs to be shifted to look at the children in the `HelperErrorText` component.

bigfishdesign13: @Toxiapo Your comment/question is right on the money. What I am suggesting here will address just that.

Toxiapo: Great—thanks a lot for sharing this context, Marty

EdwinGuzman:
@bigfishdesign13 updated. I also updated the name of the arg since `children` is not as accurate.

Toxiapo (approved):

bigfishdesign13 (approved):
Very nice. The conditional styles work great. Are you planning to ask Clare to validate this change?

PR #1965, opened 2026-04-02
Release v4.1.6

EdwinGuzman:
### Updates

- Updates the `ComponentWrapper` component so it always renders the `HelperErrorText` component. This resolves aria-live announcement issues when there is invalid text but no helper text.


bigfishdesign13 (approved):
💯

oliviawongnyc (approved):

PR #1978, opened 2026-05-27
ISW-57546: remove aria-live from HelperErrorText

bigfishdesign13:
Fixes JIRA ticket [ISW-5756](https://newyorkpubliclibrary.atlassian.net/browse/ISW-5756)

- Updates the `HelperErrorText` component to remove the default `aria-atomic` and `aria-live` attributes.
- Removed all `HelperErrorText` docs referring to `aria-atomic` and `aria-live`.
- updated docs in `Accessibility Guide` to reflect new validation patterns.

## How has this been tested?

- local build of Storybook

- The helper text will no longer be announced to screen readers when the help text content is updated.

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[ISW-5756]: https://newyorkpubliclibrary.atlassian.net/browse/ISW-5756?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/HelperErrorText/HelperErrorText.tsx
oliviawongnyc: Can this comment be removed?

src/components/HelperErrorText/HelperErrorText.mdx
oliviawongnyc, line 1: Seems like we expect some people to still want/need to use aria-atomic and aria-live. Do we want to include anything here about that?

src/components/HelperErrorText/HelperErrorText.mdx
```diff
@@ -54,54 +54,16 @@ When used with form components, the text is associated with a related input
 field by using the `aria-describedby` attribute which ensures that the
 `HelperErrorText` is made available to screenreaders.
 
-This component will always render a `<div>` element with an `aria-live="polite"`
-attribute even if no text content is passed. This is to ensure that the
-accessibility guideline for dynamic content is met.
-
 The example below uses the `HelperErrorText` component to render the red-colored
```
oliviawongnyc: What do you think about updating this to include what the screenreader will hear?

From your doc:
When focus moves to the Email field, because the helper text is programmatically tied to the Email field using aria-describedby, the user hears "Email, invalid entry, Please include an '@' symbol in your email address."

oliviawongnyc (approved):
Couple small q's!

src/components/HelperErrorText/HelperErrorText.tsx
bigfishdesign13: Yes. I missed that one. 👀

src/components/HelperErrorText/HelperErrorText.mdx
bigfishdesign13: With the change in the recommendations, we don't expect apps to want or need to use those aria attributes with the `HelperErrorText` component, so I don't think we need to include anything in the docs here. There is more detailed info under the `Dynamic Content` Accessibility Guide.

src/components/HelperErrorText/HelperErrorText.mdx
bigfishdesign13: Yes, we can do that.

src/components/HelperErrorText/HelperErrorText.test.tsx
EdwinGuzman, line 1: This might be a bit of a hot take, but I think it's OK to remove all the tests associated with the `aria-live` or `aria-atomic` attributes. Now that we're treating these props like any other props and having special conditionals, the snapshot tests with `withOtherProps` pretty much covers this case. Maybe remove all these tests and add `aria-live="polite"` in that snapshot test.

EdwinGuzman (approved):
The comment in `HelperErrorText.test.tsx` is not a blocker.

Note: did you talk to Millie or Jiayong about this update? Can you have them review this because they proposed a different approach to pretty much the same problem. So if this covers their use case, they should be aware so they're not attempting to do as well.

PR #1986, opened 2026-06-06
ISW-5784: updating Chakra to 2.9 and React to 18.3

EdwinGuzman:
Fixes JIRA ticket [ISW-5783](https://newyorkpubliclibrary.atlassian.net/browse/ISW-5784)

- Updates react and react-dom to version `18.0.3`.
- Updates Chakra to `2.9.5`. This is NOT the latest 2.x.x version. The latest 2.10.x version has a few incompatibilities.
  - [Chakra changelog](https://github.com/chakra-ui/chakra-ui/blob/v2/.changelog/v2.mdx)
- Other minor updates:
  - `statusBadge.ts` - updated return type for the theme file.
  - Removed the `As` type import in `List`. This is no longer exported from Chakra.
  - Adds `key` prop directly as a value in `AccordionPanel`. The `key` prop cannot be part of an object that gets spread. This comes up often during local Scout development.

## How has this been tested?

- There is a test version [4.3.0-rc](https://www.npmjs.com/package/@nypl/design-system-react-components/v/4.3.0-rc) that can be used in apps for testing.
- There is also a 4.2.0-rc version. They're both the same but `4.3.0-rc` has recent and unrelated updates.
- This was tested by @sarangj in DC using `4.2.0-rc`. I don't thinl we need to retest in DC again. The PR seems to be working https://github.com/NYPL/digital-collections/pull/571 
- @7emansell will test in RC
- Scout update: https://github.com/NYPL/dxp-react-search/pull/1169

### Checklist:

### Front End Review:

[ISW-5783]: https://newyorkpubliclibrary.atlassian.net/browse/ISW-5783?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/theme/components/statusBadge.ts
```diff
@@ -78,7 +78,7 @@ export const recommendation = defineStyle({
   },
 });
 
-const StatusBadge = defineStyleConfig({
+const StatusBadge: ReturnType<typeof defineStyleConfig> = defineStyleConfig({
```
bigfishdesign13, line 81: Why is this needed only in this theme file, but not in others?

bigfishdesign13 (approved):
The updates look good. I just had one non-blocking question.

src/theme/components/statusBadge.ts
EdwinGuzman: It was unclear why. I thought it could be because the name matches a Chakra component, but then this issue would come up elsewhere.

oliviawongnyc (approved):

PR #1989, opened 2026-06-10
DSD-1756: Banner onClose

KaseyW31:
Fixes JIRA ticket [DSD-1756](https://newyorkpubliclibrary.atlassian.net/browse/DSD-1756)

- Adds the onClose prop to the Banner component

## How has this been tested?

- Local build of storybook

### Accessibility Checklist

## Open Questions

- I added in the warning for non-dismissible banners, should that be taken out?

### Checklist:

### Front End Review:

[DSD-1756]: https://newyorkpubliclibrary.atlassian.net/browse/DSD-1756?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Banner/Banner.mdx
+++ b/src/components/Banner/Banner.mdx
@@ -21,7 +21,7 @@ import { changelogData } from "./bannerChangelogData";
   componentName="Banner"
   summary="Provides contextual messages about page content and user actions"
   versionAdded="3.1.0"
-  versionLatest="4.0.0"
+  versionLatest="Prerelease"
 />
 
 <Canvas of={BannerStories.WithControls} />
--- a/src/components/Banner/Banner.mdx
+++ b/src/components/Banner/Banner.mdx
@@ -265,6 +265,8 @@ order to override the predefined variants.
 A dismissible `Banner` component can be created by setting the `isDismissible`
 prop to `true`. Once the "X" close icon on the upper right is clicked, the
 `Banner` will be removed from the DOM. Therefore, it only renders once.
+If the `onClose` prop is set, it will be called as well. `onClose` will be
+ignored if the `isDismissible` prop is `false`.
 
 <Canvas of={BannerStories.Dismissible} />
 
--- a/src/components/Banner/Banner.mdx
+++ b/src/components/Banner/Banner.mdx
@@ -24,7 +24,7 @@ import { changelogData } from "./bannerChangelogData";
   versionLatest="Prerelease"
 />
 
-<Canvas of={BannerStories.WithControls} />
+<Canvas of={BannerStories.WithControls} story={{ autoplay: true }} />
 
 ## Table of contents
 
--- a/src/components/Banner/Banner.stories.tsx
+++ b/src/components/Banner/Banner.stories.tsx
@@ -1,5 +1,6 @@
 import { VStack } from "@chakra-ui/react";
 import type { Meta, StoryObj } from "@storybook/react-vite";
+import { fn, expect, userEvent, screen } from "storybook/test";
 
 import Banner from "./Banner";
 
@@ -59,10 +60,8 @@ export const WithControls: Story = {
     heading: "Heading text",
     highlightColor: undefined,
     icon: undefined,
-    isDismissible: false,
-    onClose: () => {
-      console.log("custom close");
-    },
+    isDismissible: true,
+    onClose: fn(),
     variant: "neutral",
   },
   parameters: {
@@ -71,6 +70,14 @@ export const WithControls: Story = {
       url: "https://www.figma.com/file/qShodlfNCJHb8n03IFyApM/Main?type=design&node-id=86601-97661&mode=design&t=wZy1nqVOOZ4Dzuu2-11",
     },
     jest: ["Banner.test.tsx"],
+  },
+  play: async ({ args }) => {
+    const heading = screen.getByText("Heading text");
+    expect(heading).toBeInTheDocument();
+    const button = screen.getByRole("button");
+    await userEvent.click(button);
+    expect(args.onClose).toBeCalled();
+    expect(heading).not.toBeInTheDocument();
   },
   render: (args) => <Banner {...args} />,
 };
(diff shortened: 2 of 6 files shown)


src/components/Banner/bannerChangelogData.ts
```diff
@@ -9,6 +9,13 @@
 import { ChangelogData } from "../../utils/ComponentChangelogTable";
 
 export const changelogData: ChangelogData[] = [
+  {
+    date: "Prerelease",
+    version: "Prerelease",
+    type: "Update",
+    affects: ["Functionality"],
+    notes: ["Adds `onClose` prop."],
```
EdwinGuzman: Very minor: can this also include that it's only used when `isDismissible` is set to true?

EdwinGuzman (commented):
This is great, thank you!

Apologies that we didn't fill out the AC in the ticket before you started. The DS team has only recently been talking about this, but we want to include more automated tests besides unit tests. The unit tests you wrote in this PR are already good.

Can you also include storybook interaction tests? These would be added in `Banner.stories.tsx` as a `play` function within the `WithControls` object.

Here's an existing example: https://github.com/NYPL/nypl-design-system/blob/af1fd8eee2ab6bb727be8e5d9a73d68dc24ab04a/src/components/FeedbackBox/FeedbackBox.stories.tsx#L121
Here's storybook documentation https://storybook.js.org/docs/writing-tests#interaction-tests (we are using an older very of Storybook so I would use their docs as a guide but follow the existing `play` examples within our repo more closely).

The unit tests correctly test the low-level functionality. The interaction test should more closely test how a patron would interact with the component. So they would click on the X icon and expect the banner to not be visible. Aim to use the accessible roles as locators rather than ids or `data-testid`s.

The interaction tests will be visible in the "Interactions" tab under the "With Controls" story for the Banner component: https://nypl-design-system-git-dsd-1756-banner-onclose-nypl.vercel.app/?path=/story/components-feedback-banner--with-controls

src/components/Banner/Banner.stories.tsx
```diff
@@ -68,6 +71,14 @@ export const WithControls: Story = {
     },
     jest: ["Banner.test.tsx"],
   },
+  play: async ({ args }) => {
+    const heading = screen.getByText("Heading text");
+    expect(heading).toBeInTheDocument();
+    const button = screen.getByRole("button");
+    await userEvent.click(button);
+    expect(args.onClose).toBeCalled();
+    expect(heading).not.toBeInTheDocument();
```
EdwinGuzman, line 83: Great, thanks! One last thing, can the component rerender after this last expectation? It then becomes hard to see the component after the test runs.

KaseyW31: I was looking at documentation and was having some trouble finding a standardized way to do it - it seems like Storybook wants to keep the end state of stories. Other examples in DS don't fully close their components in the interaction tests so rerendering isn't necessary. Do you have any suggestions for this?

src/components/Banner/Banner.tsx
```diff
@@ -39,6 +39,8 @@ export interface BannerProps extends Omit<BoxProps, "content"> {
   /** Optional prop to control whether a `Banner` can be dismissed
    * (closed) by a user. */
   isDismissible?: boolean;
+  /** Function to call when the modal is closed. */
```
7emansell: nit: Banner, not modal

src/components/Banner/bannerChangelogData.ts
7emansell, line 1: 👍👍

src/components/Banner/Banner.stories.tsx
7emansell: I'm also not finding any solutions for this, besides literally adding a button into the story that you can click to remount it

7emansell (approved):
Great work!!!

src/components/Banner/Banner.stories.tsx
EdwinGuzman: One possible solution is to move the mount within the `play` function and remove the `render` property:

```
play: async ({ args, mount }) => {
    // Mount the component initially
    await mount(<Banner {...args} />);
    
    const heading = screen.getByText("Heading text");
    expect(heading).toBeInTheDocument();
    const button = screen.getByRole("button");
    await userEvent.click(button);
    expect(args.onClose).toBeCalled();
    expect(heading).not.toBeInTheDocument();

    // Re-render the banner after tests complete for manual testing
    await new Promise(resolve => setTimeout(resolve, 100));
    await mount(<Banner {...args} onClose={fn()} />);
  },
```

However, this breaks the storybook documentation page. You'll see this warning 

Image

And it links to https://storybook.js.org/docs/api/doc-blocks/doc-block-story#autoplay
I can't get the above to work.

So I'm thinking, in order to prioritize automated testing:
- keep the latest update made in this PR
- add a new story with a basic `Banner` component and use it in the `.mdx` file.
  - This `<Canvas of={BannerStories.WithControls} />` would become `<Canvas of={BannerStories.NameOfNewBannerExample} />`

Let me know what you think, but we can definitely meet later today or early next week to go over it in person.

src/components/Banner/Banner.mdx
bigfishdesign13: `versionLatest` should be set to `Prerelease`.

src/components/Banner/Banner.stories.tsx
KaseyW31: Sounds good! I was wondering if it makes sense to keep `WithControls` as it was originally, and move the interactions test into a new story (since it looks like other components that have `WithControls` use that in `<Canvas of>`), but I guess it wouldn't be standard to have `play()` in a separate story either. Feel free to put time on my calendar!

KaseyW31: Actually I think I can get the mount suggestion to work by adding `story={{ autoplay: true }}` to `<Canvas of`

EdwinGuzman: Great, that worked. I think I tried adding autoplay in the wrong place.

The output of the interaction tests don't display, but it does say it passes. Not a big deal to me.

EdwinGuzman (approved):
Nice! Thank you!

src/components/Banner/Banner.stories.tsx
KaseyW31: Sounds good, thanks for your help!

PR #2004, opened 2026-07-08
Hotfix: revert chakra fix

EdwinGuzman:
Reverts the Chakra version to 2.8.2.

This was seemingly the issue causing Scout to load slowly and return 502s. It may be something within the updated Chakra version that we can investigate, but the first step is to revert and publish this.

I tested this in Scout on the evening of 7/7 by deploying a release version `4.4.0-rc` to production. This deployed successfully. I spot checked various Scout pages and did not see any issues and there were no significant CPU spikes according to AWS:

Screenshot 2026-07-07 at 9 32 39 PM

The second vertical banner on the right is the evening deployment. CPU usage stayed about the same after the release. This makes me confident that we can safely rollback only the Chakra version to 2.8.2 and keep React to 18.3.0.


oliviawongnyc (approved):
When upgrading Chakra, you had:
`- Removed the As type import in List. This is no longer exported from Chakra.`
Does this need to be reverted?

EdwinGuzman:
Good point. Will update.

EdwinGuzman:
@oliviawongnyc updated. For `List.tsx` and `statusbadge.ts`.

src/components/RadioGroup/__snapshots__/RadioGroup.test.tsx.snap
```diff
@@ -27,7 +27,6 @@ exports[`RadioGroup renders the UI snapshot correctly 1`] = `
       >
         <label
           className="chakra-radio css-rfs4w5"
-          htmlFor="radio-2"
```
bigfishdesign13, line 30: Why is `htmlFor` being removed? I noticed this as I was working on the `Icon` update yesterday.

bigfishdesign13 (approved):
The code looks good and spot checking the Vercel build also looks good.

src/components/RadioGroup/__snapshots__/RadioGroup.test.tsx.snap
EdwinGuzman: I think this was _added_ in the updated Chakra version. So now they're all removed.

PR #2002, opened 2026-07-06
ISW-5893: Routine npm package updates for the repo, not components

EdwinGuzman:
Fixes JIRA ticket [ISW-5893](https://newyorkpubliclibrary.atlassian.net/browse/ISW-5893)

## This PR does the following:

- Updates npm packages to fix security issues flagged by dependabot and `npm audit`.
- Only updates dev dependencies and not dependencies that would affect the DS components themselves.
- Updates unit tests to reflect updated package APIs.

## How has this been tested?

- Will create npm release candidate for testing.

## Accessibility concerns or updates

- N/A

### Checklist:

- [ ] I have updated the Storybook documentation and changelog accordingly.
- [ ] I have added relevant accessibility documentation for this pull request.
- [ ] All new and existing tests passed.

### Front End Review:

- [ ] Review the Vercel preview deployment once it is ready.

[ISW-5893]: https://newyorkpubliclibrary.atlassian.net/browse/ISW-5893?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2007, opened 2026-07-10
SCC-5375: Multiselect lazy loading

KaseyW31:
Fixes JIRA ticket [SCC-5375](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5375)

- Adds the `lazy-load` option to the `listOverflow` prop of the `MultiSelect` component

## How has this been tested?

- Local build of storybook - "Lazy loading items" story

### Accessibility Checklist

## Open Questions

-

### Checklist:

### Front End Review:

[SCC-5375]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-5375?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -287,10 +329,53 @@ export const MultiSelect: ChakraComponent<
         }, 1); // Ensure focus logic runs after state update
       };
 
+      const onItemsListScroll = () => {
```
7emansell: I think this chunk can be replaced by setting `rootMargin` on the `IntersectionObserver`– i'm new to that API but it seems like if it's set (to 200px or something) it would trigger the same anticipatory load?

src/components/MultiSelect/MultiSelect.test.tsx
```diff
@@ -470,6 +494,58 @@ describe("MultiSelect", () => {
     );
   });
 
+  it("should lazily load more list items while scrolling in lazy load mode", async () => {
```
7emansell: Clever test!! also wondering: did you try a more involved mock of the intersection observer and it got too messy ? It seems like it would be difficult to mock

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -24,7 +30,11 @@ export interface MultiSelectItem {
 }
 export const multiSelectWidthsArray = ["fitContent", "full"] as const;
 export type MultiSelectWidths = typeof multiSelectWidthsArray[number];
-export const multiSelectListOverflowArray = ["scroll", "expand"] as const;
+export const multiSelectListOverflowArray = [
+  "scroll",
```
7emansell: Interested in marty/edwin thoughts but maybe lazy-load should be a flag you can set if you've picked `scroll`, not a standalone type

7emansell (changes requested):
overall looks good to me, and it's working great in Storybook, just a couple notes/questions on the implementation.
I do have to ask my favorite question though... have you tried this in RC advanced search and if so is it performant

src/components/MultiSelect/MultiSelect.tsx
KaseyW31: This chunk allows for a more gradual and scaled loading - without it, scrolling down just a bit loads a lot of items (even with `rootMargin` at 0px). But that would work too, any thoughts on which is preferred?
I'm also quite new to the API so I could definitely be missing something else, and I can continue to explore it more

src/components/MultiSelect/MultiSelect.test.tsx
KaseyW31: I did not 🤔 I could try it though if that would be preferred

src/components/MultiSelect/MultiSelect.tsx
KaseyW31: Ah that does make sense

KaseyW31:
> have you tried this in RC advanced search and if so is it performant

I did, without any other changes to RC ~~it doesn't seem to make a much of a difference in prod~~, but it does in dev mode... I haven't tried it with my changes in [this PR](https://github.com/NYPL/research-catalog/pull/764) yet, but I imagine it would make less of a difference since it wouldn't be rerendered as often
Do you have any thoughts on whether this is worth trying?

Actually on second look there is a slight performance improvement in the language checkbox and input fields (if I squint). Previously I was still on DS 4.3.0 which might have slowed it down a bit

src/components/MultiSelect/MultiSelect.tsx
EdwinGuzman: Let me review before you make this update. Without having looked too closely at the PR, Emma's comment sounds reasonable. Perhaps it can be an independent flag that works for both `scroll` and `expand`?

src/components/MultiSelect/MultiSelect.tsx
7emansell: No i see what you're saying here! I think it's worth the workaround to have a better scrolling experience

src/components/MultiSelect/MultiSelect.test.tsx
7emansell: No worries just wondering

src/components/MultiSelect/MultiSelect.tsx
7emansell: this shouldn't be possible on `expand` right?

7emansell:
> > have you tried this in RC advanced search and if so is it performant
> 
> I did, without any other changes to RC ~it doesn't seem to make a much of a difference in prod~, but it does in dev mode... I haven't tried it with my changes in [this PR](https://github.com/NYPL/research-catalog/pull/764) yet, but I imagine it would make less of a difference since it wouldn't be rerendered as often Do you have any thoughts on whether this is worth trying?
> 
> Actually on second look there is a slight performance improvement in the language checkbox and input fields (if I squint). Previously I was still on DS 4.3.0 which might have slowed it down a bit

if it makes a difference in dev mode that's awesome (thinking along the lines of our unit test duration..) Let me know if you have a branch that has this + 4.2.0 + no other changes in advanced search, i'd be interested to see. I like the idea of adding this as a non-breaking change, totally optional, to the `Multiselect` on the DS level and then handling the `useMemo` on the consuming client level, that seems like a fair separation of concerns to me

KaseyW31:
> Let me know if you have a branch that has this + 4.2.0 + no other changes in advanced search, i'd be interested to see.

Yes I pushed my changes to the `no-ref/test-multiselect-lazy` branch! I do think the unit tests run slightly faster, now around 40-50s

src/components/MultiSelect/MultiSelect.stories.tsx
```diff
@@ -270,6 +270,403 @@ const withItemCountItems = [
   },
 ];
 
+const withLazyLoadItems = [
```
EdwinGuzman: Minor: can you move this to another file within this directory? Just to make it easier to read and store data elsewhere.

src/components/MultiSelect/MultiSelect.test.tsx
EdwinGuzman: Agreed on the good test and good mock.

src/components/MultiSelect/MultiSelect.test.tsx
```diff
@@ -11,6 +11,14 @@ jest.mock("../../hooks/useSafeId", () => ({
   useSafeId: jest.fn((id) => id || "test-id"),
 }));
 
+const intersectionObserverMock = () => ({
+  observe: () => null,
+  disconnect: () => null,
+});
+window.IntersectionObserver = jest
```
EdwinGuzman: Emma already asked if you tested the rc npm release in Research Catalog. Did you also run tests in RC?

I noticed that whenever we mock global APIs in the DS, unit tests in consuming apps need to be updated as well. It's not bad, but we would just have to let teams know about the dependency.

src/components/MultiSelect/MultiSelect.tsx
EdwinGuzman: Would it be wrong to _always_ have lazy load be enabled for the `scroll` option? It works well and it doesn't seem like it has a big overhead.

src/components/MultiSelect/MultiSelect.test.tsx
```diff
@@ -1,5 +1,5 @@
 import { axe } from "jest-axe";
-import { render, screen, waitFor } from "@testing-library/react";
+import { fireEvent, render, screen, waitFor } from "@testing-library/react";
```
EdwinGuzman: Minor: it looks like `userEvent` is now the preferred and conventional function to use over `fireEvent`. Can you verify and if that's right, update the tests?

EdwinGuzman: It might also be the recommended function in later versions of RTL.

EdwinGuzman (commented):
Some comments. Before I review the main file, we should decide on the approach and whether the lazy loading should always be enabled. If so, it might change the code.

src/components/MultiSelect/MultiSelect.test.tsx
EdwinGuzman: To be clear, you don't have to update the testing library version in this PR. If you can switch to `userEvent` safely now, then let's go with that. 

I updated a few npm packages and added async/await to `userEvent` calls: https://github.com/NYPL/nypl-design-system/pull/2002/changes#diff-734ff340ae6de4d2aec5eaff40652614d776ed3d91ed14a41fc022bf89f06490L175

I based my update in the PR on this https://testing-library.com/docs/user-event/intro/

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -287,10 +329,53 @@ export const MultiSelect: ChakraComponent<
...
       };
 
+      const onItemsListScroll = () => {
+        if (!isOverflowLazy || !itemsListRef.current) {
+          return;
+        }
+
+        const { scrollTop, clientHeight, scrollHeight } = itemsListRef.current;
+        const scrollThreshold = scrollHeight / 2; // Scales with further scrolling
+        const isAtBottom =
```
EdwinGuzman: This is the bottom of the visible item list, not the entire list, correct?

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -430,13 +518,21 @@ export const MultiSelect: ChakraComponent<
...
                       {getMultiSelectCheckboxItem(item)}
                     </React.Fragment>
                   ))}
                 </CheckboxGroup>
                 {isOverflowExpand && <ExpandToggleButton />}
+                {isOverflowLazy &&
+                  visibleItemsList.length < itemsList.length && (
+                    <Box
+                      ref={lazyLoadTargetRef}
+                      data-testid={`${mainId}-lazy-load-observer-target`}
```
EdwinGuzman: Minor: if the `data-testid` is not used, it's ok to not include it.

src/components/MultiSelect/MultiSelect.tsx
bigfishdesign13: Yes, agreed. `lazy-load` should not be an option for `listOverflow`. It should be a boolean flag.

Here's another idea. Although, I am not sure if the timing of this is possible. Can lazy load be automatically enabled if the total options is greater than `x`? Otherwise, scrolling functions as is always has?

Do we know the threshold when performance takes a hit?

src/components/MultiSelect/MultiSelect.tsx
KaseyW31: Yes visible items

src/components/MultiSelect/MultiSelect.tsx
bigfishdesign13: Yes, the scrolling-loading could use some polishing. I'd love to have a chat to learn more about the API and the config.

src/components/MultiSelect/MultiSelect.test.tsx
KaseyW31: I don't believe that `userEvent` currently supports scrolling. I tried with `user.keyboard('{PageDown}')` but it doesn't seem to work in the testing environment. Another option might be to do a more extensive mock of `IntersectionObserver`, but I think an important part of the test is the actual user interaction of scrolling instead of making assumptions of when an intersection is triggered.

Or maybe this could be tested in `play` in `WithControls`?

src/components/MultiSelect/MultiSelect.test.tsx
KaseyW31: Yes, the `IntersectionObserver` mock is needed when `lazy-load` is selected. Would this be a factor at all in deciding whether lazy loading should be always enabled for `scroll`?

src/components/MultiSelect/MultiSelect.stories.tsx
KaseyW31: I moved it to a file called ~~`MultiSelectStoriesData.ts`~~ `MultiSelectWithLazyLoadItems.ts`. I'm not sure if there's a convention for the naming, please let me know if it should be renamed!

src/components/MultiSelect/MultiSelect.tsx
oliviawongnyc: What do you think about `isHalfwayThroughLoadedContent` rather than `isAtBottom`? Then someone reading this can easily see that when `isHalfwayThroughLoadedContent`, `loadMoreLazyItems()` and I think it would act like a little bit of documentation.

oliviawongnyc: Scratch this. After looking at this longer, I don't think we should use two triggers for the loading, and think we should use the observer over this onScroll.

src/components/MultiSelect/MultiSelect.tsx
oliviawongnyc: Image

I put these console.logs right before the `loadMoreLazyItems()` call in the observer and right before `loadMoreLazyItems()` call in the `onItemsListScroll` function.

The scroll handler keeps moving the sentinel away before the observer can reach it. So the observer fires only once, while the scroll handler takes over and it's running a lot.

I tried removing `onItemsListScroll` and I personally did not observe much of a difference, so would love to see the difference you're seeing.

Also with regards to, "scrolling down just a bit loads a lot of items" -- isn't the number of items related to the setState in `loadMoreLazyItems`, rather than anything to do with the `rootMargin`?

```      
setLazyItemsVisible((previousVisibleItems) =>
          Math.min(
            previousVisibleItems +
              defaultItemsVisible +
              previousVisibleItems / defaultItemsVisible, // Scales with further scrolling
            itemsList.length
          )
        );
```

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -245,6 +272,21 @@ export const MultiSelect: ChakraComponent<
...
+      const loadMoreLazyItems = useCallback(() => {
+        if (!isOverflowLazy) {
+          return;
+        }
+
+        setLazyItemsVisible((previousVisibleItems) =>
+          Math.min(
+            previousVisibleItems +
+              defaultItemsVisible +
+              previousVisibleItems / defaultItemsVisible, // Scales with further scrolling
```
oliviawongnyc: What are the benefits to scaling the batch?

src/components/MultiSelect/MultiSelect.tsx
KaseyW31: @oliviawongnyc I didn't realize that, thanks for pointing that out! When I was testing in research catalog using just the observer, scrolling fast would cause stuttering. After scrolling all the way to the bottom, it took some time to load more items, and by then the user would be scrolling past the panel entirely. So I think I overcorrected for that and messed up the observer in the process without realizing it.

As for the difference, I misinterpreted the comment and thought the chunk was referring to the scaling part of it. Removing the scroll handler changes the behavior slightly but doesn't make the difference I mentioned, sorry for the confusion.

I'll take a look at reducing stuttering while only using the IntersectionObserver.

src/components/MultiSelect/MultiSelect.tsx
KaseyW31: When I was experimenting with scrolling quickly, the scrolling speed can increase as I scroll further down. The scaling was to prevent too many items from loading at the beginning, while preventing the stuttering that I mentioned above if the user is scrolling quickly to the bottom (which I saw more on research catalog than storybook). Maybe I could set a constant higher amount to load, though @bigfishdesign13 suggested continuing with the scaling, which I can try adapting to the IntersectionObserver.

src/components/MultiSelect/MultiSelect.tsx
KaseyW31: I removed the scroll handler and tweaked some of the scaling. I tried calculating the `rootMargin` using pixel heights in the scroll panel but that didn't work out great, so the numbers I have now are from trial and error. Any thoughts on that or the scrolling experience now @7emansell @oliviawongnyc ?

src/components/MultiSelect/MultiSelect.tsx
KaseyW31: I'm not sure about the threshold. I'm open to logging performance/render times compared to the number of items, but I imagine that performance could also vary based on how it's used in the consuming app?

Right now `20 + defaultItemsVisible` are loaded initially per @bigfishdesign13 's suggestion (although `defaultItemsVisible` is only used for `expand` since the default appears to be ~7 for scrolling, but that might be a separate concern). Maybe this could be the threshold for when lazy loading is enabled?

src/components/MultiSelect/MultiSelect.test.tsx
KaseyW31: Also it turns out that my initial test was working because of the scroll handler and not the observer. I changed it to mock the observer instead and trigger the intersections manually, so now it does not need `userEvent` (though it does use `act` instead now)

src/components/MultiSelect/MultiSelect.tsx
oliviawongnyc: Changes looking good to me!

As for the scaling, what happens with a fixed root margin? `rootMargin: "0px 0px 100px 0px"`, for example? I would think we just want the observer to trigger loading before someone reaches the bottom of the visible list and not need to scale, but lmk your thoughts.

Same question about the batching. Does the increment size need to scale?

src/components/MultiSelect/MultiSelect.tsx
EdwinGuzman: Yes, let's keep it at `20 + defaultItemVisible` as the threshold for now and we can gauge its performance once it's live.

src/components/MultiSelect/MultiSelect.mdx
```diff
@@ -211,7 +212,8 @@ counting the items to display.
   content={
     <>
       <strong>IMPORTANT:</strong> This prop can only be used with{" "}
```
EdwinGuzman, line 214: Minor: update "with" to "when"?

src/components/MultiSelect/MultiSelect.tsx
bigfishdesign13: The current configuration works well. No matter how fast I try to scroll, I never hit the bottom. I would prefer to load a constant number of options during the scrolling, but I also don't want users to hit the bottom. 

The data for all of the options is always available, right? No API call to get more data from the server. So any lag that happens as the options are added is due to rendering, which could be impacted by the power of a users computer or device. If there are concerns about the scaling, let's try a static number and try to find the sweet spot.

src/components/MultiSelect/MultiSelect.test.tsx
```diff
@@ -1,16 +1,56 @@
...
+  intersectionCallback = null;
+  observeMock.mockClear();
+  disconnectMock.mockClear();
+
+  window.IntersectionObserver = jest.fn().mockImplementation((callback) => {
+    intersectionCallback = callback;
+    return {
+      observe: observeMock,
+      disconnect: disconnectMock,
+      unobserve: jest.fn(),
```
EdwinGuzman: This doesn't also need to be cleared?

src/components/MultiSelect/MultiSelect.test.tsx
```diff
@@ -1,16 +1,56 @@
...
+    };
+  });
+});
+
+const triggerIntersection = () => {
```
EdwinGuzman, line 37: Nice update. How did you catch that the mock scroll behavior was not actually working as expected?

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -245,6 +276,19 @@ export const MultiSelect: ChakraComponent<
...
       };
 
+      const loadMoreLazyItems = useCallback(() => {
+        if (!isOverflowLazy) {
+          return;
+        }
+
+        setLazyItemsVisible((previousVisibleItems) =>
+          Math.min(
+            previousVisibleItems * 1.2 + lazyLoadIncrementNum,
```
EdwinGuzman: Can you create a `CONSTANT_VARIABLE`  for `1.2`? This is to reduce magic numbers and better understand when reading this why it's set to `1.2`.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -291,6 +335,33 @@ export const MultiSelect: ChakraComponent<
...
+        const observer = new IntersectionObserver(
+          (entries) => {
+            if (entries.some((entry) => entry.isIntersecting)) {
+              loadMoreLazyItems();
+            }
+          },
+          {
+            root: itemsListRef.current,
+            threshold: 0,
+            rootMargin: `${12 * lazyItemsVisible}px`,
```
EdwinGuzman: Can you also create another constant var for `12`?

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -430,13 +503,17 @@ export const MultiSelect: ChakraComponent<
...
                 >
-                  {itemsList.map((item: MultiSelectItem) => (
+                  {visibleItemsList.map((item: MultiSelectItem) => (
                     <React.Fragment key={item.id}>
                       {getMultiSelectCheckboxItem(item)}
                     </React.Fragment>
                   ))}
                 </CheckboxGroup>
                 {isOverflowExpand && <ExpandToggleButton />}
+                {isOverflowLazy &&
```
EdwinGuzman, line 508: It makes sense why we need this, but can you add a short comment describing why? When we revisit this component in the future, it'll be clear why this is needed in the DOM.

EdwinGuzman (commented):
This is looking great. Just have a few comments and questions.

I think what's left to do is to always enable lazy loading when scrolling is selected. @bigfishdesign13 just want to make sure we agree on this; you suggested an additional boolean flag.

src/components/MultiSelect/MultiSelect.test.tsx
EdwinGuzman: Not a big factor but just something to call out when devs import the release version. Did you have to add it in a global pre-test jest file in RC?

src/components/MultiSelect/MultiSelect.stories.tsx
EdwinGuzman: This looks good! Thank you.

bigfishdesign13:
> I think what's left to do is to always enable lazy loading when scrolling is selected. @bigfishdesign13 just want to make sure we agree on this; you suggested an additional boolean flag.

@EdwinGuzman Yes, always enable lazy loading when scrolling is selected. Consistency. There's no need to add an additional boolean to toggle this feature on and off.

src/components/MultiSelect/MultiSelect.tsx
KaseyW31: I realized that a lot of the slowness issues I was having without scaling was because I was testing in research catalog dev mode earlier, my mistake. I removed the scaling and set the rootMargin to 300px which seems to work well enough, let me know if there are any issues with that

src/components/MultiSelect/MultiSelect.test.tsx
KaseyW31: The test was failing because no more items were being loaded on scroll - seems like the scroll handler was taking over. I also didn't realize earlier that the IntersectionObserver doesn't work in the test environment because the environment doesn't have the browser layout engine.

src/components/MultiSelect/MultiSelect.test.tsx
KaseyW31: Seems like it's only needed if the number of items in the MultiSelect > `defaultItemsVisible + 20`, when lazy loading is triggered. Otherwise putting this block of code
```
const intersectionObserverMock = () => ({
  observe: () => null,
  disconnect: () => null,
  unobserve: () => null,
})
window.IntersectionObserver = jest
  .fn()
  .mockImplementation(intersectionObserverMock)
```
either in the test files with the MultiSelect, or the global pre-test file is needed.

7emansell (approved):
Eek sorry thought I had approved this last week! 
As discussed offline– we could probably refine the IntersectionObserver bounds forever, I think this looks pretty good and is working well. Since as far as we know, RC is the only consumer that will even be affected by this change, we can observe in prod and return to tweak it if it seems necessary.
Tldr this is great let's merge it in

EdwinGuzman (approved):
Thank you! Fix the conflicts before merging.

PR #2014, opened 2026-08-10
ISW-5842: Update tag action

oliviawongnyc:
Fixes JIRA ticket [ISW-5842](https://newyorkpubliclibrary.atlassian.net/browse/ISW-5842)

- Updates `release.yml` to:
- Use the latest CHANGELOG notes as the body of the Github tag
- Remove/upgrade outdated dependencies
- Update comments for clarity and accuracy

## How has this been tested?

- It has not been tested....I don't think it's risky to wait until the next release and see if it worked.

[ISW-5842]: https://newyorkpubliclibrary.atlassian.net/browse/ISW-5842?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,51 +1,12 @@
-# This workflow will tag, release, and publish to npm.
+# This workflow will publish to npm, then tag and create a GitHub release.
 name: Production Release
 
-# workflow will only run when a pull request to the release branch is closed.
+# Runs on any push to the release branch.
 on:
   push:
     branches: [release]
 
 jobs:
-  release_master:
-    name: Tag and Release
-    # only run if PR is merged (not just closed)
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-
-      # we extract the latest changelog entry to use as the release body.
-      - name: get changelog entry
-        run: |
-          body=$(awk '/^## /{i++} i==1{print}' CHANGELOG.md)
-          echo "commitmsg<<EOF" >> $GITHUB_ENV
-          echo "$body" >> $GITHUB_ENV
-          echo "EOF" >> $GITHUB_ENV
-
-      # the tag will match the package.json version (eg. v1.0.0)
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -1,47 +1,51 @@
...
 
-# workflow will only run when a pull request to the release branch is closed.
+# Runs on any push to the release branch.
 on:
   push:
     branches: [release]
 
 jobs:
-  release_master:
-    name: Tag and Release
```
EdwinGuzman, line 11: Can you verify if this update will also remove the job from the workflow? In the past, I found it difficult to remove workflows, and also remove jobs within workflows.

EdwinGuzman (commented):
Great. Looks like it'll work but can you verify the question about the new job and if old ones get removed?

.github/workflows/release.yml
oliviawongnyc: Are you asking if this will remove the job from the required status checks in Github? I looked at the rulesets for our branches and this job was not a required status check, so I believe this name change is purely cosmetic and won't impact anything outside of the file.

oliviawongnyc: https://github.com/NYPL/nypl-design-system/settings/branches

EdwinGuzman: No, I meant on the actions: https://github.com/NYPL/nypl-design-system/actions/runs/28961362460

It probably will rename it. This was more of a hassle in the past. I also am thinking of the main `name` attribute, so this might not be an issue at all.

EdwinGuzman (approved):
Let's try it out for the next release.

.github/workflows/release.yml
oliviawongnyc: Ok! We'll see tomorrow!

PR #2015, opened 2026-08-12
ISW-5977: exporting react externals for next15 compatibility

EdwinGuzman:
Fixes JIRA ticket [ISW-5977](https://newyorkpubliclibrary.atlassian.net/browse/ISW-5977)

- Tightens the scope of which react version to use in `peerDependencies`.
- Updates the vite config to export external packages so there are fewer conflicts when apps build the app.

The problem started when the DC team tried to upgrade to next15 and ran into various issues. Not _all_ were DS-related, but the various react versions prevented the app from building. Here's the PR https://github.com/NYPL/digital-collections/pull/583 and check the vercel build error to see one specific build issue.

## How has this been tested?

- Tested this in DC with my own PR against the existing next15 one. https://github.com/NYPL/digital-collections/pull/604
- Test this in Scout https://github.com/NYPL/dxp-react-search/pull/1235
- Both Vercel previews work so `npm run build` is successful, and the app generally runs as expected. Have not encountered any run-time issues.

### Checklist:

### Front End Review:

[ISW-5977]: https://newyorkpubliclibrary.atlassian.net/browse/ISW-5977?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vite.config.ts
```diff
@@ -6,6 +6,16 @@ import dts from "vite-plugin-dts";
...
+const externalPackages = [
+  /^react($|\/)/,
+  /^react-dom($|\/)/,
+  /^@chakra-ui\//,
+  /^@emotion\//,
+  /^framer-motion($|\/)/,
+  /^react-datepicker($|\/)/,
+  /^react-popper($|\/)/,
+  /^@popperjs\//,
+];
```
oliviawongnyc, line 18: Should this list match what is in `peerDependencies`?

package.json
```diff
@@ -77,8 +77,8 @@
     "@emotion/react": ">=11.11.1",
     "@emotion/styled": ">=11.11.0",
     "framer-motion": ">=10.16.4",
-    "react": ">=17",
-    "react-dom": ">=17",
+    "react": ">=17 <19",
+    "react-dom": ">=17 <19",
     "typescript": ">=4.9.5"
   },
```
oliviawongnyc, line 83: This is kind of a separate thought, but do you think we should explicitly tell people they need to install these? Especially chakra?

vite.config.ts
EdwinGuzman: Hmm probably. Good question, I'll check.

package.json
EdwinGuzman: I _think_ in the past we did. We made it so team wouldn't have to install chakra and other deps separately. However, I see in various projects that each team does still include chakra and other packages. We should've done this with react from the beginning.

For v5, we can be more explicit and say, "only install this and do not install chakra or other deps separately. the ds will include them".

vite.config.ts
EdwinGuzman: It's already very close to what's in `peerDependencies`. Typescript is not here and from what I gathered, that's intentional because typescript is used to compile the library and it's not needed at runtime. The other packages are needed in the app at run time.

We should've been doing this all along :( . I think we were previously bundling react/chakra within the DS bundle _and_ asking consuming apps to install the packages under `peerDependencies`.

For v5, we should look into:
- moving all packages from `peerDependencies` to `dependencies`,
- deleting this `externals` config,
- and asking teams to remove any `@chakra-ui/react` (and other packages) from their own package.json.

Teams should not have to install anything separately, and on our side we can achieve this as a "dependency".

EdwinGuzman: To answer the question directly, no, this list does not need to match what's in `peerDependencies`.

oliviawongnyc (approved):

sarangj:
FYI, this is working great on https://github.com/NYPL/digital-collections/pull/583 (I merged your PR and merged our qa branch in)

PR #2018, opened 2026-09-09
ISW-5995: Menu - Add isDisabled Prop

nprotheroNYPL:
Fixes JIRA ticket [ISW-5995](https://newyorkpubliclibrary.atlassian.net/browse/ISW-5995?focusedCommentId=173568&sourceType=mention)

- Implements `isDisabled` prop to action menu item.
- Implements menu item disabled state styling. 

## How has this been tested?

- Tested disabled state through storybook component testing. 
- Tested on text, icon, and image menu items. 
- Tested menu item focus navigation through keyboard controls.

- Disabled menu item is not focusable through `ArrowUp` and `ArrowDown` keys for accessibility sake.

### Accessibility Checklist

## Open Questions

- Disabled items have the 'not-allowed' cursor styling to follow Chakra UI's styles, do we want this?

### Checklist:

### Front End Review:

[ISW-5995]: https://newyorkpubliclibrary.atlassian.net/browse/ISW-5995?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Menu/Menu.mdx
+++ b/src/components/Menu/Menu.mdx
@@ -182,7 +182,7 @@ optional and will display to the left of the menu text, and will render either
 an [Icon](./?path=/docs/components-media-icons-icon--docs) or
 [Image](./?path=/docs/components-media-icons-image--docs). The `media` prop
 requires a `type` property (icon or image) and then either the name of the icon
-or the source URL and alt text of the image.
+or the source URL and alt text of the image. The optional `isDisabled` prop can be used to prevent items from being selected.
 
 <Source
   code={`
@@ -206,8 +206,19 @@ or the source URL and alt text of the image.
       console.log("Item Title 2 clicked");
     },
   }
+  // Example of disabled action item
+  {
+    type: "action",
+    id: "item-title-3",
+    label: "I am a disabled option",
+    isDisabled: true,
+    onClick: () => {
+      console.log("Item Title 3 clicked");
+    },
+
+}
 `}
-  language="tsx"
+language="tsx"
 />
 
 Below are examples of different content variations:
--- a/src/components/Menu/Menu.mdx
+++ b/src/components/Menu/Menu.mdx
@@ -13,7 +13,7 @@ import { changelogData } from "./menuChangelogData";
   componentName="Menu"
   summary="List of interactive elements rendered in a dropdown format and used for executing actions or navigating"
   versionAdded="3.0.0"
-  versionLatest="4.0.1"
+  versionLatest="Prerelease"
 />
 
 <Canvas of={MenuStories.WithControls} />
--- a/src/components/Menu/Menu.stories.tsx
+++ b/src/components/Menu/Menu.stories.tsx
@@ -373,6 +373,42 @@ const wideListItems: ListItemsData[] = [
   },
 ];
 
+const disabledItems: ListItemsData[] = [
+  {
+    type: "action",
+    id: "item-title-1",
+    isDisabled: true,
+    label: "I'm item 1",
+    media: {
+      type: "image",
+      src: getPlaceholderImage("smaller"),
+      alt: "",
+    },
+    onClick: () => {
+      console.log("Item Title 1 clicked");
+    },
+  },
+  {
+    type: "action",
+    id: "item-title-1",
+    isDisabled: true,
+    label: "I'm item 2",
+    media: { type: "icon", name: "search" },
+    onClick: () => {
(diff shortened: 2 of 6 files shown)


src/components/Menu/Menu.stories.tsx
```diff
@@ -373,6 +373,42 @@ const wideListItems: ListItemsData[] = [
   },
 ];
 
+const disabledItems: ListItemsData[] = [
```
oliviawongnyc, line 376: I'm not sure what use case you have in mind for this, but I would personally would change the example to have a mix of disabled and non-disabled options. Having them all disabled might give someone the impression its all or nothing if they're looking at it quickly.

src/components/Menu/Menu.tsx
```diff
@@ -276,9 +278,12 @@ export const Menu: ChakraComponent<
           const menuItem = (
             <MenuItem
               key={item.id}
-              isFocusable={true}
+              isFocusable={!item.isDisabled}
```
oliviawongnyc: I think `isFocusable` defaults to true if `isDisabled` is true so ok to remove.

src/components/Menu/Menu.tsx
```diff
@@ -276,9 +278,12 @@ export const Menu: ChakraComponent<
...
+              isDisabled={item.isDisabled}
               data-testid={isSelected ? "selected-item" : ""}
-              onClick={() => handleSelect(item.id, item.onClick)}
+              onClick={() =>
```
oliviawongnyc: Similar situation here. It doesn't seem like disabled items are clickable, so I don't think you need to pass/handle the `isDisabled` case here.

src/components/Menu/Menu.mdx
oliviawongnyc, line 1: Update the `Latest` value to say "Prerelease" so we know to update it to the correct version when we next deploy.

Image

src/theme/components/menu.ts
```diff
@@ -89,6 +104,21 @@ const Menu = defineMultiStyleConfig({
...
+            _disabled: {
+              textColor: "ui.gray.semi-dark",
+              fontStyle: "italic",
+              cursor: "not-allowed",
+              svg: {
+                fill: "ui.disabled.primary",
+              },
+              img: {
+                opacity: 0.4,
+              },
```
oliviawongnyc: Double check, but I don't think you need to repeat these styles from the first `disabled` block:

```
fontStyle: "italic",
cursor: "not-allowed",
svg: {
     fill: "ui.disabled.primary",
},
img: {
     opacity: 0.4,
},
```

oliviawongnyc (changes requested):
Nice! A few clean up suggestions, but looks really good.

oliviawongnyc (approved):
LGTM! But let Edwin/Marty take a look before merging.

src/theme/components/menu.ts
```diff
@@ -73,6 +73,21 @@ const Menu = defineMultiStyleConfig({
             bg: "ui.bg.hover",
             fontWeight: "medium",
           },
+          _disabled: {
+            textColor: "ui.gray.semi-dark",
+            fontStyle: "italic",
+            cursor: "not-allowed",
```
EdwinGuzman, line 79: I like how explicit the icon is on hover. It's not a pattern we've used before, but I think this is ok.

EdwinGuzman (approved):

src/theme/components/menu.ts
bigfishdesign13: We do use this in `AlphabetFilter`. I feel we use it somewhere else as well, but I can't find any code to support that claim.

bigfishdesign13: This is also used for the disabled state of `Radio`, `Select`, TextInput`, and `Toggle`. However, that is not something we added.

src/components/Menu/Menu.stories.tsx
```diff
@@ -373,6 +373,35 @@ const wideListItems: ListItemsData[] = [
...
```
bigfishdesign13, line 376: Update the options so there are two with just text, two with icons, and two with images. Each pair should have one enabled option and one disabled option.

bigfishdesign13 (changes requested):
Overall, looks great! It just needs the examples expanded.

bigfishdesign13 (approved):
Very nice!

